Issue #79, opened 2026-05-09
PRD: OpenCode plugin auto-install must refresh every existing layout


vfarcic:
## PRD: OpenCode plugin auto-install must refresh every existing layout

**Problem**: When both `~/.config/opencode/` and `~/.opencode/` exist, `detect_opencode_root` prefers XDG, so auto-install only refreshes the XDG plugin. If the user's `opencode.jsonc` lives at the legacy root (and explicitly references the legacy plugin path), the legacy plugin file is never refreshed — its `BINARY_PATH` can rot to a deleted binary, every hook silently fails, and dashboard cards stay stuck on "No agent" with `Tools: 0`.

**Solution**: Make auto-install resilient when both layouts coexist — write to every existing plugin layout (or detect the active root via `opencode.jsonc` rather than the `plugin/` subdir) so the plugin OpenCode actually loads is always kept in sync with the running binary.

**Detailed PRD**: See [prds/79-opencode-plugin-auto-install-all-layouts.md](https://github.com/vfarcic/dot-agent-deck/blob/main/prds/79-opencode-plugin-auto-install-all-layouts.md)

**Priority**: Medium

PR #188, opened 2026-06-22
fix: keep delegated worker prompt single-line

### LINKED ISSUE #187: Bug: delegated worker prompt can land as compacted multi-line input needing manual Enter
## Symptom

When the orchestrator delegates work to a worker role, Dot Agent Deck writes the real task into `.dot-agent-deck/worker-task-<role>.md`, then injects a prompt into the worker pane telling the agent to read that file.

The injected prompt currently includes a multi-line `## When done` footer and fenced `dot-agent-deck work-done` command. In Claude Code, that multi-line injected prompt can land in the worker input box as a compacted/braced pasted block and sit there until the user manually presses Enter.

Observed flow:

~~~text
Read .dot-agent-deck/worker-task-coder.md for your task.

## When done

Signal completion by running this command via Bash:
```bash
dot-agent-deck work-done --task "Brief summary of what you accomplished. Include file paths and outcomes."
```
~~~

## Impact

Delegation appears to have succeeded, but the worker does not actually start until the operator notices the compacted input and submits it manually. This breaks unattended orchestration and makes the operator monitor worker panes.

## Scope Tested

Confirmed by local testing with Claude Code workers. Codex and OpenCode behavior is not claimed here; Codex support may still be in progress under the broader multi-agent work.

## Likely Cause

The daemon-side delegation path still injects a multi-line prompt through the PTY:

- `handle_delegate`
- `dispatch_one_owned`
- `compose_delegate_prompt`
- `write_to_pane_and_submit`

Because the injected text contains embedded newlines, it takes the bracketed-paste path. The actual durable task is already file-backed, so the pane-injected prompt does not need to be multi-line.

## Proposed Fix

Keep the pane-injected delegate prompt to a single-line pointer:

```text
Read .dot-agent-deck/worker-task-<role>.md for your task.
```

Move the `## When done` / `dot-agent-deck work-done` completion instructions into the generated worker task file instead. This preserves the worker protocol while avoiding multi-line PTY injection for the normal file-backed delegation path.

## Acceptance Criteria

- Delegated worker panes receive only a single-line file pointer prompt.
- `.dot-agent-deck/worker-task-<role>.md` contains the role prompt/template, task, and work-done instructions.
- Claude Code workers start without a manual Enter after delegation.
- Regression coverage asserts the injected prompt has no newline and the worker task file still includes the completion footer.

s33dunda:
## Summary
- Keep delegated worker pane prompts to a single line by sending only the file-backed task pointer through the PTY.
- Move the `work-done` completion instructions into `.dot-agent-deck/worker-task-<role>.md` instead of appending them to the injected pane prompt.
- Normalize delegate prompt input whitespace so future multiline call sites still produce a single-line pane prompt.

Fixes #187

## Scope / Manual Validation
- The reported workflow was manually validated with Claude Code workers. Codex/OpenCode behavior is not claimed by this PR.
- The change is agent-agnostic at the prompt transport layer: the pane receives one line, while the worker task file keeps the full task and completion instructions.

## Tests
- `rtk cargo fmt --check`
- `rtk cargo clippy -- -D warnings`
- `rtk cargo test -p dot-agent-deck compose_delegate_prompt_is_single_line_file_pointer`
- `rtk cargo test -p dot-agent-deck compose_delegate_prompt_normalizes_multiline_input`
- `rtk cargo test -p dot-agent-deck compose_worker_task_file_appends_work_done_footer`
- `rtk cargo test-fast`

### DIFF
--- a/changelog.d/187.bugfix.md
+++ b/changelog.d/187.bugfix.md
@@ -1,0 +1,3 @@
+## Delegated Workers Start Without a Manual Enter
+
+Delegating to a worker role now injects a single-line prompt into the worker pane instead of a multi-line block. Previously the multi-line prompt (which carried the `## When done` completion instructions) could land in the worker's input as a compacted bracketed-paste that sat unsubmitted until an operator pressed Enter — stalling unattended orchestration. The completion instructions now live in the worker's task file (`.dot-agent-deck/worker-task-<role>.md`), so the worker still knows to signal back via `dot-agent-deck work-done` once the single-line pointer auto-submits.
--- a/src/state.rs
+++ b/src/state.rs
@@ -125,25 +125,25 @@ pub struct AppState {
 
 pub type SharedState = Arc<RwLock<AppState>>;
 
+const WORK_DONE_FOOTER: &str = "## When done\n\n\
+Signal completion by running this command via Bash:\n\
+```bash\n\
+dot-agent-deck work-done --task \"Brief summary of what you accomplished. Include file paths and outcomes.\"\n\
+```";
+
 /// Compose the prompt that the daemon writes into a worker pane on
-/// delegation: the caller-supplied `task_body` (typically a one-liner
-/// pointing at `.dot-agent-deck/worker-task-{role}.md`), a blank line,
-/// then the work-done footer that reminds the worker to signal
-/// completion via `dot-agent-deck work-done` once they finish.
+/// delegation. In the normal file-backed path this is intentionally only
+/// the one-line pointer to `.dot-agent-deck/worker-task-{role}.md`.
+/// Keeping every injected PTY prompt single-line avoids bracketed paste
+/// and lets the synthetic CR follow the same reliable path as ordinary
+/// typed prompts.
 ///
 /// The footer used to be appended per-role by the TUI's
 /// `OrchestrationConfig.roles[*].prompt_template` wrapping. PRD #93
-/// round-5 moved dispatch into the daemon but left
-/// `OrchestrationConfig` out of scope, so without this helper every
-/// delegated task lost the footer and workers stopped signaling back.
+/// round-5 moved dispatch into the daemon; the durable worker context now
+/// lives in the task file instead of the injected pane prompt.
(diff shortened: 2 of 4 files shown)


vfarcic:
Thanks for this fix, @s33dunda! 🙌 The single-line approach is exactly right. I've pushed one commit on top (`f21fc98`) adding regression coverage and a changelog fragment — no changes to your `src/state.rs` fix itself.

**What I added**

- **`tests/delegate_prompt_injection.rs`** (fast tier, runs in CI): drives the real `handle_delegate` dispatch against a stub worker pane and asserts the injected prompt is the single-line file pointer (and *not* the multi-line `## When done` footer), while the footer + task body land in `.dot-agent-deck/worker-task-<role>.md`. This guards the wiring `handle_delegate → compose_delegate_prompt → write_to_pane_and_submit` so a future change can't silently reintroduce the multi-line bracketed-paste.
- **`tests/e2e_delegate_work_done_chain.rs`** (L2, local-only behind the `e2e` feature): real-agent proof of the behaviour your PR fixes — that a worker **auto-submits** the injected single-line prompt with no manual Enter.
  - **Claude/Haiku** runs the full `delegate → work-done` loop end to end.
  - **OpenCode** confirms the auto-submit half via a permission-free conversational round-trip. (OpenCode's own tool-permission sandbox gates `.dot-agent-deck` reads, which is orthogonal to this fix, so the arm deliberately avoids it.)
- **`changelog.d/187.bugfix.md`** — release-notes fragment for the fix.

**What this verified:** your fix is now empirically confirmed on **both** Claude Code and OpenCode — auto-submit works on both. Along the way I ruled out (with evidence) that earlier OpenCode hiccups were the fix, an account/quota issue, the model, or the plugin-loader (PRD #79); they were just OpenCode's permission sandbox.

I also confirmed this is daemon-side prompt composition only — no TUI↔daemon wire change — so no `PROTOCOL_VERSION` bump or breaking fragment is needed.

CI + review are running now; once green I'll merge. Thanks again!

vfarcic:
Merged into `main` 🎉 This will ship in the next release. Thanks again for the fix, @s33dunda!

PR #198, opened 2026-06-27
feat(config): add mouse.enabled to opt out of TUI mouse capture

### LINKED ISSUE #197: Add `mouse.enabled` config to opt out of TUI mouse capture
## PRD: Opt out of TUI mouse capture for native terminal selection / copy

**Problem**: `run_tui` enables mouse capture unconditionally, which puts the
terminal into mouse-reporting mode. That blocks the terminal's own text
selection and copy — users who want to select/copy output with their mouse
can't, except via a per-terminal Shift-drag workaround that isn't universally
available. There is no way to turn it off.

**Solution**: Add a `mouse.enabled` config key (default `true`, so current
behavior is preserved), mirroring the existing `bell` / `idle_art` config. When
`false`, dot-agent-deck does **not** issue `EnableMouseCapture`, so the terminal
handles the mouse natively (native select / copy). The in-app mouse affordances
(clickable cards/buttons, in-pane click-drag selection + its OSC52 clipboard
copy) become inert while disabled — every action still has a keyboard shortcut
(`?`). Bracketed paste stays enabled; the teardown `DisableMouseCapture` stays
unconditional (a harmless no-op when capture was never enabled).

Set via: `dot-agent-deck config set mouse.enabled false`

**Scope**: `MouseConfig { enabled: bool = true }` wired into
`CONFIG_KEYS` / `get_field` / `set_field`; gate the `EnableMouseCapture` call in
`run_tui` on it; docs (`configuration.md` "Mouse" section); changelog fragment;
unit tests (defaults, `[mouse]` parse, get/set roundtrip).

**Priority**: Low

**Related**: the `bell` / `idle_art` opt-in config (the mirrored pattern).

bustapipes:
## Summary

Add a `mouse.enabled` config option (default `true`) so users can opt out of the
TUI's mouse capture and let their terminal handle native text selection / copy.

`run_tui` enabled mouse capture unconditionally, putting the terminal into
mouse-reporting mode and preventing native select/copy. This makes it a config
toggle, mirroring the existing `bell` / `idle_art` config.

## Changes

- **config** (`src/config.rs`): `MouseConfig { enabled: bool = true }` wired into
  `CONFIG_KEYS` / `get_field` / `set_field`, with unit tests (defaults, `[mouse]`
  section parse, get/set roundtrip).
- **ui** (`src/ui.rs`): gate `EnableMouseCapture` on `config.mouse.enabled`.
  Bracketed paste and the teardown `DisableMouseCapture` stay unconditional (the
  latter is a no-op when capture was never enabled).
- **docs** (`docs/configuration.md`): new "Mouse" section.
- **changelog**: feature fragment.

## Behavior / compatibility

- Default `true` → **no behavior change** for existing users; click/drag UI and
  OSC52 copy keep working.
- `mouse.enabled = false` → terminal owns the mouse (native select/copy); in-app
  mouse affordances are inert, but every action still has a keyboard shortcut.
- Set via: `dot-agent-deck config set mouse.enabled false`

## Testing

`cargo test` — adds 4 unit tests: `mouse_config_defaults`,
`dashboard_config_without_mouse_section`, `dashboard_config_with_mouse_section`,
`mouse_enabled_get_set_roundtrip`.

Closes #197

### DIFF
--- a/changelog.d/+mouse-enabled-config.feature.md
+++ b/changelog.d/+mouse-enabled-config.feature.md
@@ -1,0 +1,1 @@
+Add a `mouse.enabled` config option (default `true`). Set it to `false` to disable the TUI's mouse capture so the terminal handles native text selection and copy: `dot-agent-deck config set mouse.enabled false`.
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -17,6 +17,19 @@ dot-agent-deck config get default_command
 
 `default_command` is the agent command pre-filled in the **new-pane form**'s Command field and the value that seeds the **schedule-authoring** agent. Both the new-pane form and the Scheduled Tasks **Add/Edit** flow use the same form — you type the command directly into the **Command** field (it accepts `claude`, `opencode`, a path, or any command), pre-filled from `default_command`. If `default_command` is unset, the schedule-authoring agent falls back to `claude`.
 
+## Mouse
+
+```bash
+# Disable mouse capture so your terminal handles text selection / copy
+dot-agent-deck config set mouse.enabled false
+```
+
+| Key | Default | Description |
+|---|---|---|
+| `mouse.enabled` | `true` | Capture the mouse for in-TUI interaction (click cards/buttons, click-drag to select within a pane, OSC52 copy-to-clipboard). |
+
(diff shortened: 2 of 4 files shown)


bustapipes:
Basically useless PR.

Issue #201, opened 2026-07-10
PRD: Pi as a first-class agent — deterministic orchestrator + third agent type


vfarcic:
## PRD: Pi as a first-class agent — deterministic orchestrator + third agent type

**Problem**: dot-agent-deck is a control plane over black-box agent processes; the orchestrator is the weakest seam — delegation is prompt-and-pray CLI strings and status is scraped from installed hooks (`~/.claude/settings.json`), so we can only nudge orchestrator behavior, not control it.

**Solution**: Add Pi (`earendil-works/pi`) as a third, first-class, status-tracked agent type and use its TypeScript extension API to make the orchestrator deterministic — native `delegate`/`work-done` tools and event-bus-driven status feeding the existing `EventType`/`AgentEvent` protocol, with no hooks for a Pi pane. Pi is detected on PATH (not bundled as a runtime); only the extension ships in-binary. Opt-in behind the `experimental` flag.

**Detailed PRD**: See [prds/201-pi-agent-integration.md](https://github.com/vfarcic/dot-agent-deck/blob/main/prds/201-pi-agent-integration.md)

**Priority**: Medium

PR #210, opened 2026-07-15
fix: status reporting for wrapper-launched Pi and current-version OpenCode agents

vfarcic:
## Problem

In a mixed-agent orchestration, only **Claude** cards reported status in the dashboard; **Pi** and **OpenCode** cards stayed on **"No agent"** both after start and while working. Claude works because its hooks live in `~/.claude/settings.json` (first-party, installed globally at startup, read regardless of launcher). Investigation (live, on a dogfooding machine) found **two independent** root causes.

## Bug 1 — Pi: status setup coupled to the launch command

Pi's orchestrator extension was materialized at spawn time, gated on the command basename being literally `pi` (`AgentType::from_command == Some(Pi)`). The real role command `devbox run pi-big` → basename `devbox` → `None` → extension never written → Pi never reported → "No agent". Any wrapper (`devbox run`, a script, an absolute path) broke it.

**Fix (command-agnostic, at the right seam):** materialize the extension **once at daemon startup** (the `dot-agent-deck daemon serve` entry, covering both the lazy-spawned and headless daemon) — never per spawn, never keyed on a command. This is true parity with how Claude/OpenCode install their hooks/plugin at startup. It also honors Pi's own `PI_CODING_AGENT_DIR` (else `~/.pi/agent`), mirroring pi's `getAgentDir()`, so a user who relocates Pi's dir still gets a status-tracked pane — and tests isolate cleanly by pointing that var at a throwaway dir.

> Note: an earlier revision of this PR materialized unconditionally *at spawn time*. Review (Greptile P2) plus a discovered test-pollution bug — the fast `daemon_protocol` tests spawn in the real HOME, so per-spawn materialize wrote the developer's real `~/.pi` during `cargo test` — drove the move to the daemon-startup seam. Now nothing writes `~/.pi` outside a real daemon's own startup.

## Bug 2 — OpenCode: plugin installed where OpenCode never looks

The plugin was installed to a nested subdir `~/.config/opencode/plugin/dot-agent-deck/index.js`, but OpenCode discovers local plugins with a **one-level** glob (from its bundle): `{plugin,plugins}/*.{ts,js}` — no `**`. The nested file was never scanned → never loaded → no events → "No agent". Proven with probe plugins under real OpenCode 1.17.18: a flat `plugin/x.js` loaded, a nested `plugin/x/index.js` did **not**.

**Fix:** install a **flat** `plugin/dot-agent-deck.js` (what the glob matches), and delete the stale nested dir on install so existing users migrate on their next launch with **no manual step**. Ships to all users via the existing startup `opencode_manage::auto_install()`.

## Not a protocol change

Neither fix touches the TUI↔daemon wire, hook payloads, or delegate routing — the daemon/hook contract is unchanged. So **no `PROTOCOL_VERSION` bump** (bugfix → patch, per `docs/develop/versioning.md`).

## Tests

- `opencode_manage`: flat-file layout + stale-nested-dir migration + uninstall.
- `orchestrator_ext`: `expand_tilde`, the pure agent-dir resolver, and `PI_CODING_AGENT_DIR` precedence.
- Pi e2e: in-process-daemon tests (`chain-smoke/pi/00{1,2}`) stage the extension (they bypass the `daemon serve` entry); real-daemon tests (`pi/live/*`, `scheduler/pi/001`) exercise the startup seam.
- **Gates:** fast tier (1010) green with **no real `~/.pi` write**; `clippy -D warnings` clean (incl. `--features e2e`); Pi e2e subset **53/53 green**, real `~/.pi` clean afterward.

### Verified live
- Real CLI `hooks install --agent opencode` writes the flat file and removes a seeded stale nested dir; real OpenCode 1.17.18 loads the plugin at the flat path.
- Real `pi` (via OpenRouter) reports live status through the daemon-startup-materialized extension in `pi/live/*` and `scheduler/pi/001`.


vfarcic:
**Re: Greptile P1 "Daemon Startup Mutates Pi" (`main.rs`) — accepted, not addressed, with rationale.**

The finding is literally accurate: on a machine where `pi` is on the daemon's PATH, the daemon-startup materialize writes `~/.pi/agent/extensions/dot-agent-deck/` even for a Claude-only/shell-only session, and that extension is then discoverable by *any* `pi` session (including standalone `pi` outside the deck). We're shipping this deliberately:

- **We can't scope it to "an actual Pi launch" without reintroducing the bug.** The only spawn-time signal is the command, and this PR exists precisely because keying on the command basename (`AgentType::from_command`) silently skipped every wrapper launch (`devbox run pi-big`, scripts, absolute paths). Re-adding any command gate re-breaks those.
- **Gating on "Pi's agent dir already exists" is unsound.** Empirically, `pi` does not always create `~/.pi/agent` (e.g. `pi --version` / `pi list` don't), so a genuine Pi user could have no such dir — the guard would then skip materialization and regress them to "No agent". That's the exact failure we're fixing.
- **The residual is low-impact by design.** Status reporting and seed delivery are best-effort and swallow all failures when there's no pane env / daemon (`index.ts` `reportStatus` / `deliverPendingSeed`), so a standalone `pi` session sees no errors or breakage — only two clearly-labeled `delegate`/`work_done` tools that return a clean error if ever invoked. The write is idempotent and confined to dot-agent-deck's own two files.
- **It matches the codebase pattern** of installing a detected agent's integration at startup (Claude hooks, OpenCode plugin), and the write only ever lands in a real daemon's own startup — the fast-tier suite no longer touches `~/.pi` at all.

Truly isolating the deck's extension from the user's global `~/.pi` (e.g. a deck-managed `PI_CODING_AGENT_DIR`) interacts with Pi's auth/settings/session continuity and is a PRD #201 refinement rather than this bugfix — intentionally out of scope, and not worth a follow-up given the graceful degradation above.

Issue #140, opened 2026-06-08
Bug: concurrent orchestrations cross-deliver worker reports (shared per-user daemon not session-partitioned)


vfarcic:
## Symptom
Running more than one orchestration at the same time — **even in separate directories** — can deliver a worker's \`work-done\` report / notification to the *wrong* orchestrator's stream. Observed in practice; recovery required killing the app, which left in-flight worker tasks uncommitted.

## Evidence / root-cause hypothesis
- The daemon is **global per user**: its socket is keyed by UID (\`\$XDG_RUNTIME_DIR/dot-agent-deck.sock\`, falling back to \`/tmp/dot-agent-deck-{uid}.sock\`) — one daemon shared across **all** directories (see \`src/config.rs\`).
- Per-directory coordination files (\`.dot-agent-deck/work-done-<role>\`, \`worker-task-<role>\`) isolate *same-dir* state, but the shared daemon's delegate/work-done **notification routing does not appear to be partitioned by orchestration session / cwd**.
- Net: concurrent orchestrations route through one daemon with no session scoping, so reports cross-deliver regardless of directory.

## Impact
Concurrent orchestrations are currently unsafe. Documented workaround: run **one orchestration at a time per user/machine** (separate directories do NOT isolate them).

## Proposed fix
Partition orchestration sessions in the daemon (e.g. by cwd / a session id) so that \`delegate\` / \`work-done\` / notification delivery is scoped to the originating orchestration. Then concurrent orchestrations in different directories become safe.

## First step
Confirm the exact routing path: which daemon code delivers \`work-done\` notifications to attached TUIs, and where session scoping must be added.

🤖 Logged via Claude Code (dot-agent-deck orchestrator)

irizzant:
I can reproduce this with two instances of the **same orchestration** opened in separate tabs. Steps:

1. `dot-agent-deck`
2. `Ctrl+N` → select orchestration → start task A
3. `Ctrl+N` → select same orchestration → start task B

The two tabs interfere: delegate/work-done signals cross-deliver between them, making both workflows unreliable. The workaround (one orchestration tab at a time) works but severely limits parallel work.

vfarcic:
Thanks for the clean repro, @irizzant — that's exactly the case that's still open.

[...]

Issue #126, opened 2026-05-25
PRD: Agent-driven notifications with minimal deck-side fallback


vfarcic:
## PRD: Agent-driven notifications with minimal deck-side fallback

**Problem**: Long-running agents and scheduled tasks have no out-of-band way to reach an absent user. The current PRD #99 proposes the deck implement notification channels (Slack/SMTP/webhook/desktop) behind a pluggable trait — duplicating MCPs/CLIs that already exist, forcing the deck to hold third-party credentials, and only covering orchestrator-completion events.

**Solution**: Delegate notification delivery to agents via their own MCPs/CLIs (Slack MCP, ntfy, osascript, etc.). The deck only injects a prompt hint at spawn and operates an inactivity timer that fires *via prompt injection* — no new sending path. One minimal local desktop channel handles deck-side events (scheduler-side failures, agent crashes) where no agent exists to delegate to. Closes #99.

**Detailed PRD**: See [prds/126-agent-driven-notifications.md](https://github.com/vfarcic/dot-agent-deck/blob/main/prds/126-agent-driven-notifications.md)

**Priority**: Medium

vfarcic:
Shipped in #223, merged as 51ee52b.

**What landed:** a daemon-side idle-worker detector (the daemon reports a delegated worker that has gone silent past `worker_response_timeout_minutes`, default 120; `0` disables it — it never notifies anyone itself), an orchestrator-only Telegram notification recipe, published docs at `docs/idle-workers-and-notifications.md`, and retirement of the ntfy dogfood that preceded it.

Non-breaking: `PROTOCOL_VERSION` stays 6, confirmed by the rule-12 cross-version test in both directions (run twice — again after PRD #140 merged, since #140 landed after v0.34.1 and reworked the routing the matrix exercises).

Demo reel: https://youtu.be/arJLOwhiynE

Remaining work is tracked in the PRD rather than here: M5 validates the recipe across real orchestrated runs, including one long enough to compact (PRD #82 interaction).

PR #223, opened 2026-07-25
feat(prd-126): daemon idle-worker detector + orchestrator-only Telegram notifications

vfarcic:
## Summary

PRD #126 shipped as a **real, shipped feature** (rescoped 2026-07-27 from the original no-code dogfood — see the PRD's [Scope decision](https://github.com/vfarcic/dot-agent-deck/blob/prd-126-agent-driven-notifications-dogfood/prds/126-agent-driven-notifications.md#scope-decision-2026-07-27-rescope-to-a-real-feature)). The Phase 1 dogfood (config-only, ntfy over `curl`) ran first and did its job: it proved the channel and immediately surfaced the one thing config alone cannot do — detect a worker that was delegated to and then went silent. That finding is what this PR actually builds.

1. **A daemon-side idle-worker detector** (`src/state.rs`, `src/agent_pty.rs`). The daemon tracks every outstanding delegation and, if no `work-done` arrives within `worker_response_timeout_minutes` (default **120**), injects one self-describing prompt into the orchestrator over the existing prompt-injection path. The daemon holds **no notification logic and no credentials** — it only reports the condition; the orchestrator's own instructions decide what to do about it. `0` disables the detector entirely; 1–10080 minutes are accepted; an out-of-range value falls back to the default rather than being clamped. Delivery is identity-bound (not pane-id-bound), race-safe against a near-simultaneous `work-done`, and cancels cleanly on pane close.
2. **An orchestrator-only Telegram notification recipe** in this repo's own `.dot-agent-deck.toml`, firing short, fire-and-forget messages at the test-plan gate, an escalation, the merge gate, full-run-done, and the daemon's idle-worker event. `chat_id` is **mandatory** (not merely default) because the underlying bot has no allowed-chat check and falls back to the most recently active chat if `chat_id` is omitted — an interceptable default. Workers never notify; only the orchestrator does, via `work-done` escalation.
3. **Published user docs** — `docs/idle-workers-and-notifications.md` (in `site/sidebars.js`), split into the feature (what the daemon does and its config knob) and an example recipe (Telegram, placeholders only, framed as "an example that works for us").
4. **Retirement of the Phase 1 dogfood** — `scripts/notify.sh` and its ntfy `curl` glue are gone, every role's ntfy prompt instructions are removed from `.dot-agent-deck.toml`, and `docs/develop/notifications-dogfood.md` is rewritten as a short historical pointer so the `CONTRIBUTING.md` link stays live.

## Gates

- **Fast tier: 1252/1252.**
- **Full e2e: 2182 tests, 2178 passed, 4 failed, 0 skipped.** The 4 failures are a **pre-existing host-environment gap, not a regression**: all four are Codex real-agent tests failing because this host's Codex auth is an API key while `CODEX_TEST_MODEL_DEFAULT` is subscription-only. They pass under a reachable model override, and none of the helpers involved are touched by this branch. On `main` these same 4 print `SKIP:` and count as PASS — this branch's new `DOT_AGENT_DECK_REQUIRE_REAL_E2E` mode is what makes the gap visible instead of silently skipping it. Judged non-blocking.
- **Rule 12 cross-version manual test: non-breaking confirmed**, run twice — once before `main` was merged in, and again afterward, because PRD #140 (merged via `4e61423`) landed after `v0.34.1` and reworked the routing the matrix exercises (the `(name, cwd)` tuple → `OrchestrationIdentity`). Both directions (old daemon + new TUI, new daemon + old TUI) route delegates and deliver hooks correctly in both runs; the idle-worker prompt itself only ever travels over the existing prompt-injection path, so no new wire message type exists and `PROTOCOL_VERSION` stays **6** on both sides throughout.
- `cargo fmt --check`, `cargo clippy --all-targets -- -D warnings` (with and without `--features e2e`), and `cargo xtask linkage-check` are all clean.
- A reviewer and a security auditor both examined the implementation; every agreed finding was resolved, including two blockers — identity-bound delivery (so a stale timer can no longer auto-submit into a pane's new occupant after a close/reuse) and a race-safe pane-close/arm interaction.

## Compatibility

- `changelog.d/126.feature.md` already exists (feature, non-breaking) — no changelog change needed here.
- `changelog.d/140.breaking.md` is also present on this branch from the `main` merge; that break is PRD #140's own semantic routing-identity change, not introduced by this PR.
- No version bump in this PR.

## Test plan

- [x] `cargo test-fast` — 1252/1252
- [x] `cargo test-e2e` — 2178/2182 passing, 4 known-environment-gap failures explained above
- [x] Rule 12 cross-version manual test — non-breaking, run pre- and post-merge
- [x] `cargo fmt --check`, `cargo clippy --all-targets -- -D warnings` (both feature sets), `cargo xtask linkage-check`
- [x] Reviewer + security-auditor pass, blockers resolved

Demo reel for PRD #126: https://youtu.be/arJLOwhiynE. Watch before merging.


vfarcic:
Demo reel for PRD #126: https://youtu.be/arJLOwhiynE. Watch before merging.

PR #226, opened 2026-07-26
fix(pi): report Idle on turn-end, not "Needs Input" (#224)

### LINKED ISSUE #224: Pi agent shows "Needs Input" on every turn-end (should be Idle, like Claude/OpenCode/Codex)
## Problem

A Pi pane reads **"Needs Input"** whenever it ends a turn — e.g. an orchestrator that just delegated to a worker, or an auditor that finished its pass — even though it is functionally **idle** (waiting on a worker's `work-done`, which the daemon injects into its PTY to wake it). This is confusing: "Needs Input" reads as "the human must act", but nothing needs the human.

## Root cause

The status machine has no delegation/worker awareness — "Needs Input" is a pure lifecycle label. The Pi extension mapped **both** `session_start` and `agent_settled` (Pi's turn-end signal) to `waiting` → `WaitingForInput` → "Needs Input" (`pi-extension/src/orchestrator.ts`). So Pi effectively never showed "Idle" between turns; it only reached Idle on `session_shutdown`.

Every other backend maps turn-end to **Idle** and surfaces "Needs Input" only on a genuine user-blocking signal:
- Claude Code: `Stop` → Idle; `Notification`/`PermissionRequest` → Needs Input
- OpenCode: `session.idle` → Idle; `permission.asked` → Needs Input
- Codex: `Stop` → Idle; `PermissionRequest` → Needs Input (no `Notification` installed)

Pi was the only outlier.

## Fix

Map Pi's `session_start` and `agent_settled` to `finished` (Idle), for parity. Pi exposes no permission/attention lifecycle event today, so it simply never reports `waiting` — exactly like a Claude agent that never hits a permission prompt. The `waiting` CLI `--type` is kept so a future Pi user-blocking event can map to it with no wire change.

No protocol change (the `waiting`/`running`/`finished` wire vocabulary is unchanged; only which event emits which value moved).

vfarcic:
## What & why

A Pi pane read **"Needs Input"** whenever it finished a turn — most visibly an orchestrator that had just delegated to a worker, or an auditor that finished its pass — even though nothing needed the user: it was simply idle, waiting for the worker's result (the daemon wakes a finished orchestrator when its worker reports back). Pi was the only backend that behaved this way.

The Pi extension mapped **both** `session_start` and `agent_settled` (Pi's turn-end signal) to `waiting` → `WaitingForInput` → "Needs Input". Every other backend maps turn-end to **Idle** and reserves "Needs Input" for a genuine user-blocking signal:

| Backend | Turn-end | "Needs Input" from |
|---|---|---|
| Claude Code | `Stop` → Idle | `Notification` / `PermissionRequest` |
| OpenCode | `session.idle` → Idle | `permission.asked` |
| Codex | `Stop` → Idle | `PermissionRequest` |
| **Pi (before)** | `agent_settled` → **Needs Input** | every turn-end |

## Change

Map Pi's `session_start` and `agent_settled` to `finished` (Idle), for parity. Pi exposes no permission/attention lifecycle event today, so it never reports `waiting` — exactly like a Claude pane that never hits a permission prompt. The `waiting` CLI `--type` is kept so a future Pi user-blocking event can map to it with no wire change.

**No protocol change.** The `waiting`/`running`/`finished` wire vocabulary is unchanged; only which Pi event emits which value moved. Old daemon + new extension (and vice-versa) still interoperate → bugfix, patch bump. `changelog.d/224.bugfix.md` added.

## Tests / docs

- Unit tests (`pi-extension/test/orchestrator.test.ts`) updated — 28/28 pass.
- Live e2e `pi/live/001` re-anchored: its extension-only proof moves from the retired `Needs Input` to `Thinking` (`agent_start`→running), which the daemon never synthesizes for a hook-less Pi pane (a fresh pane defaults to Idle). Headless Pi e2e assertions already tolerate `Idle`/`Thinking`; stale comments updated.
- README, `docs/develop/pi-extension.md`, `tests/CATALOG.md` updated.

## Validation

`cargo fmt --check`, `cargo clippy --all-targets -- -D warnings` (default + `--features e2e`), `cargo nextest run` (1116 pass), and the e2e feature compiles.

⚠️ The real-`pi` e2e tier was **not** run locally (`pi` not on PATH here, so those tests runtime-skip; per rule 5 they don't run in CI). This change is covered by the mapping unit tests, the compile-checked/updated e2e, and the logic.

Closes #224


vfarcic:
Addressed the P2 in 26870ca: added a load-bearing post-settlement `Idle` assertion to `pi/live/001` that exercises the changed `agent_settled`→finished/Idle mapping (polled on the current grid, so it can only match the settled frame — a regression to "Needs Input" fails there), rather than only the unchanged `agent_start`→Thinking path.

Issue #161, opened 2026-06-14
Compatibility-based version-mismatch handling (revises #103)


vfarcic:
## PRD: Compatibility-based version-mismatch handling

**Problem**: Any TUI↔daemon build-id (commit-sha) difference forces a mandatory upgrade/recycle, even when the two builds are backwards-compatible — needlessly forcing users to upgrade both sides and, for the local daemon, killing all running agents (whose survival across detach is the daemon's whole purpose per #93).

**Solution**: Make the upgrade mandatory ONLY when the two builds are genuinely incompatible; otherwise warn-and-proceed ("suggest"). Use the version tag's semver as the compatibility signal (0.x: minor = incompatibility boundary; ≥1.0: major), keep `PROTOCOL_VERSION` as the structural fatal floor, and demote bare build-id skew to a non-blocking suggestion. Revises #103 from "detect-and-refuse" to "detect-classify-and-{suggest|refuse}".

**Detailed PRD**: See [prds/161-compatibility-based-version-mismatch-handling.md](https://github.com/vfarcic/dot-agent-deck/blob/main/prds/161-compatibility-based-version-mismatch-handling.md)

**Priority**: Medium

PR #237, opened 2026-07-28
fix(wrap): stop losing clear=true delegate prompts to Codex workers (#225)

### LINKED ISSUE #225: Codex delegate prompt is lost on clear=true respawn (wrapper readiness race + unstable launch shape)
## PRD: Reliable prompt delivery to Wrapper-strategy agents on respawn

**Problem**: A `clear = true` delegate to a Codex worker silently loses its prompt. The wrapper emits `SessionStart` immediately after fork, which satisfies the delegate's readiness gate seconds before Codex actually exists — so the prompt is written into a PTY nothing is reading. Compounding it, the pane's launch shape mutates across respawn (spawns unwrapped, respawns wrapped), so the failure only appears after the first delegate.

**Solution**: Make the readiness signal mean "the agent can accept input" rather than "the process was forked", and make the wrap decision stable across respawn. Codex becomes usable as an orchestration worker.

**Detailed PRD**: See [prds/225-wrapper-agent-readiness-and-stable-launch-shape.md](https://github.com/vfarcic/dot-agent-deck/blob/main/prds/225-wrapper-agent-readiness-and-stable-launch-shape.md)

**Priority**: High

**Depends on**: nothing — start this first. It is the fix for a live defect (Codex is unusable as an orchestration worker today).

**Related**: **PRD #234** (screen-state observation for hookless agents) closes the residual gap this PRD hands off. The fix here is *provenance*-based — ignore the wrapper's synthetic `SessionStart`, wait for the agent's real one — which structurally requires a real one to exist. A hookless wrapper agent (Gemini, #211) has none, so it keeps today's behaviour until #234 lands. Concrete coupling: the readiness discriminator chosen here should express "where does this agent's readiness signal come from" rather than "does it install hooks", so #234's observed-readiness source slots in as a third answer without re-migrating call sites. See #225 Open Question 2. **#234 does not block this PRD; this PRD should land first.**

vfarcic:
## Summary

Closes #225. A `clear = true` delegate to a Codex worker silently lost its prompt — two stacked defects:

1. **Readiness gate accepted the wrong signal.** `dot-agent-deck wrap` emits a `SessionStart` the instant it forks the child, purely so the dashboard card appears immediately for a slow-booting agent. The delegate readiness gate treated that fork-time event as proof the agent could accept input, so the prompt was written into a PTY nothing was reading yet — Codex's real interface didn't exist for several more seconds. The fork-time event is now marked card-surfacing only (`session_start_origin=wrapper_fork`); the gate waits for a genuine native `SessionStart` for agents that install hooks, and still fast-paths on the fork-time event for hookless wrapper agents (no regression there).
2. **Launch shape mutated across respawn.** A pane started from a command that doesn't name an agent (e.g. `devbox run codex-big`) launched unwrapped, then came back up *wrapped* after its first delegate once hook events revealed the real agent type — the same pane running a different process tree before and after. The hook-learned agent type now updates only the dashboard badge; the respawn wrap decision derives from the command actually being launched (`AgentType::from_command(command).or(spawn_agent_type)`), with the frozen creation-time identity only as a fallback for a command that implies nothing. An edited role command is honored — it relaunches as itself, not disguised as the pane's original identity.

Both changes are additive on the wire in both directions; no action required when upgrading.

## Gate results

- `cargo test-fast`: **1146/1146**, 0 skipped.
- `cargo test-e2e` (pre-PR tier, rule 5; post-merge and post-fix): `DOT_AGENT_DECK_CODEX_TEST_MODEL=gpt-5-nano DOT_AGENT_DECK_RECORD=1 cargo test-e2e` → **2058 run, 2058 passed, 0 skipped**, exit 0 (79.8s). The real-agent spec `orchestration/delegate/009` passes **in-suite** (42.4s) and in isolation (43.9s) — a real Codex worker receives the delegated prompt and acts on it (sentinel file content verified, not just existence).
- `cargo fmt --check`: clean. `cargo clippy --all-targets --features e2e -- -D warnings`: clean.

## Cross-version contract check (rule 12)

**Classification: PATCH.** No `PROTOCOL_VERSION` bump, no `changelog.d/225.breaking.md`.

The cross-version check ran twice — before and after the `#140` main merge — with a branch TUI/CLI against a real previous-release **v0.34.1** daemon (`PROTOCOL_VERSION` 6) on a fully isolated rig (separate socket/state/lock/session paths and `HOME`; the live dogfood daemon was never touched):

- **A delegate still routes.** Branch CLI → old daemon writes `worker-task-<role>.md` at ~+0.5s; a real worker acts on it and returns `work-done`. One round used the genuine operator path (not a test override): the branch TUI hit PRD #161's version-mismatch handshake against the old daemon, declined the restart, and delegated over that connection anyway.
- **Hooks still arrive.** The old daemon's own trace shows the full `clear = true` lifecycle: delegate signal, `SessionEnd`, new-session `SessionStart`, `Thinking`, 3× `ToolStart`/`ToolEnd`, `work-done`, `Idle`.
- **The reverse direction was staged, not argued.** Sniffed both binaries' payloads off a stand-in socket: identical JSON except an additive metadata key `session_start_origin=wrapper_fork` (`schema_version` unchanged, no field added/removed/retyped). A new wrapper's marker is **ignored, not rejected** by the old daemon. An old wrapper under the new daemon reproduces today's behavior exactly — measured **+0.15s** to the PTY, versus **+29.98s** for a marked wrapper taking the fallback. That's the reversion this PRD predicted under Risks, now quantified: no worse than current, and no error.

**Nuance:** `src/daemon_protocol.rs` is *not* byte-identical to the v0.34.1 tag (+59/−9) — `#140` added `orchestration_id` to `TabMembership`, already on `main` and already declared in `changelog.d/140.breaking.md`. The correct baseline for "did #225 break the contract?" is `origin/main`, not the release tag, and against `origin/main`, `daemon_protocol.rs` is identical (`PROTOCOL_VERSION` 6 at v0.34.1, `origin/main`, and `HEAD` alike). `#140`'s own fragment drives the release's minor bump independently — that bump is not attributable to #225.

## Known limitation (disclosed, not hidden)

For Codex, the readiness gate **never fast-paths**. codex-cli 0.145.0 posts its native `SessionStart` when the first turn starts — i.e. *after* a prompt is submitted — so every `clear = true` Codex delegate pays the full 30s `SESSION_START_WAIT_TIMEOUT` and then injects into an already-booted TUI. The prompt-loss defect is fixed and real-agent verified; the latency is a separate, deliberate scope decision (Option 1 in the PRD Work Log), documented in `docs/develop/agent-adapters.md`. A wrapper-side "TUI ready" signal that would remove this latency is filed as a follow-up rather than bolted on here.

## Also worth noting

Integrating post-`#140` main (`044bb17`) exposed that main's `49f93be` had silently turned the `codex/spawn/007` regression guard into a test that could no longer fail — the wrap rewrite now names the co-located build by absolute path, so a `$PATH` recorder stub never ran. Both spawn specs were migrated to main's `DOT_AGENT_DECK_WRAP_BIN` seam, and the repair was proven by mutation: reintroducing Defect 2 makes `spawn_007` fail with the seam and pass without it.

## Release hygiene

- `changelog.d/225.bugfix.md` — already present, user-facing framing.
- No `PROTOCOL_VERSION` bump, no `.breaking.md` (see rule-12 section above).

## Test plan

- [x] `cargo test-fast` green (1146/1146)
- [x] `cargo test-e2e` green (2058/2058, 0 skipped), real-agent Codex delegate spec passing in-suite and isolated
- [x] Cross-version manual test (rule 12) — old v0.34.1 daemon + new TUI/CLI (delegate + hooks round-trip), and new wrapper + old daemon (additive metadata ignored)
- [x] `cargo fmt --check` / `cargo clippy --all-targets --features e2e -- -D warnings` clean

---

The e2e casts recorded during the gate run are on disk under `.dot-agent-deck/recordings/` — demo reel is a separate, later step handled by another worker.

Demo reel for PRD #225: https://youtu.be/m_1o0nFReho. Watch before merging.


vfarcic:
Demo reel for PRD #225: https://youtu.be/m_1o0nFReho. Watch before merging.

Issue #199, opened 2026-07-04
delegate: worker pane restarts but prompt silently dropped on Claude Code >= 2.1.199


bustapipes:
### Description

When the orchestrator delegates work via the `delegate` command, the target worker pane respawns correctly but the task prompt never lands — no `Thinking` or `ToolStart` events appear, and the claude process sits idle. The worker appears to restart successfully (`SessionStart` fires) but the prompt bytes written to the PTY are silently consumed/dropped.

### Root Cause

Claude Code 2.1.199 changed its TUI startup timing such that `SessionStart` is emitted before the PTY input handler is ready to accept text. The daemon's `dispatch_one_owned` (in `src/state.rs`) writes the delegate prompt into the PTY **immediately** after `wait_for_session_start` returns `true`, racing against the claude TUI's own initialization. The prompt bytes arrive too early and are silently discarded.

The orchestrator's own prompt-injection path already works around exactly this problem: `src/ui.rs` injects a 500 ms `SPAWN_TIME_READINESS_BUFFER` after `SessionStart` before writing the orchestrator context prompt. The delegate injection path in `dispatch_one_owned` simply lacked this buffer.

### Reproduction

1. Configure an orchestration with at least one worker role (e.g. `researcher`, `coder`, etc.)
2. From the orchestrator pane, run: `delegate --task "some task" --to <role>`
3. Observe: worker pane restarts, but no `Thinking`/`ToolStart` events fire; the worker claude process remains idle with no transcript activity
4. Check the daemon log: `SessionStart` fires for the new session, the delegate prompt write succeeds (no error), but the worker never processes the task

### Versions Affected

- `dot-agent-deck`: any version (tested on v0.1.0)
- `claude`: 2.1.199, 2.1.200, 2.1.201 all fail
- `claude` pre-2.1.199 works (the original binary replaced by the version manager when 2.1.199 was installed)

### Fix

Ideally, a more durable solution can be found to replace the following hack.

Add a post-`SessionStart` readiness buffer in `dispatch_one_owned`, matching the existing pattern in the orchestrator prompt-injection path:

```rust
// In src/state.rs, after wait_for_session_start returns true:
const RESPAWN_READINESS_BUFFER: Duration = Duration::from_millis(1000);

// In the `observed == true` branch (the else of `if !observed`):
tokio::time::sleep(RESPAWN_READINESS_BUFFER).await;
```

This gives the claude TUI time to finish its input-handler setup before the prompt is written.

### References

[...]

PR #271, opened 2026-07-29
fix(delegate): gate task injection on a post-respawn readiness buffer (#249)

### LINKED ISSUE #249: PRD: Delegate prompt silently dropped on clear=true respawn (Claude/OpenCode workers)
## PRD: Delegate readiness gate for respawned workers

**Problem**: With `clear = true` on a worker role, `delegate` respawns the worker and writes the task prompt with **no readiness buffer** — while the structurally identical orchestrator spawn-time path has had a 500 ms one since v0.27.x. Depending on where the write lands on the agent's startup timeline, the prompt is either lost entirely (worker looks healthy, sits idle forever — #199) or lands unsubmitted until a human presses Enter (rarer, maintainer-reproduced).

**Solution**: Add a post-`SessionStart` readiness gate at the shared seam (`src/state.rs:1255`), mirroring `SPAWN_TIME_READINESS_BUFFER`, plus surface undelivered prompts instead of failing silently. Pi (native seed pull) and Codex (wrapper readiness) are already fixed individually; Claude and OpenCode share the unguarded path.

**Root cause is already proven in this repo**: `CHANGELOG.md:358-361` documents the identical mechanism on the spawn path — "Claude Code's `SessionStart` hook fires early in its boot sequence, before its TUI input is ready to interpret `\r` as submit" — with a toggle-verified regression test that fails at `BUFFER=0` and passes at `BUFFER=500`.

**Reported by**: @bustapipes (original + correct root cause), confirmed by @peterjcarroll, @AkosHosszu, @tomikonio; @xclydes narrowed it to `clear = true` and found the `clear = false` workaround.

**Consolidates**: #199

**Detailed PRD**: See [prds/249-delegate-respawn-readiness.md](https://github.com/vfarcic/dot-agent-deck/blob/main/prds/249-delegate-respawn-readiness.md)

**Priority**: High

vfarcic:
## Summary

Closes #249. Consolidates #199.

With `clear = true` on a worker role, `delegate` respawns the worker and wrote the task prompt with **no readiness buffer** — while the structurally identical orchestrator spawn-time path has had a 500 ms one since v0.27.x (`CHANGELOG.md:358-361`). Depending on where the write landed on the respawned agent's startup timeline, the prompt was either lost entirely (worker looks healthy, sits idle forever — #199's consistent report) or landed unsubmitted until a human pressed Enter (rarer, maintainer-reproduced). Three of four agents already had an individual fix (Pi native seed pull, Codex wrapper readiness in #237); Claude and OpenCode shared the unguarded legacy PTY injection path.

**Fix**: a post-`SessionStart` readiness gate at the shared seam (`dispatch_one_owned`, `src/state.rs`), applied on both the observed and timeout-fallback branches, mirroring `SPAWN_TIME_READINESS_BUFFER`. Default 1000 ms (the spawn path's warm-case 500 ms, doubled for a cold respawn start — a deliberate stopgap value, not a measured agent threshold), overridable via `DOT_AGENT_DECK_DELEGATE_READINESS_BUFFER_MS` (`0` disables it, values above 30 s are capped).

**Also landed, found during an independent review + security audit of the M1–M5 diff and fixed in this same PR** (see the PRD's "Review + audit round" section for full detail):
- The task-pointer write is now **identity-guarded** (`write_and_submit_guarded`) — the respawn wait window this PRD lengthens made a pre-existing cross-orchestration delivery race easier to hit; a close/respawn/re-home/teardown inside the wait could otherwise let a pointer land in a successor agent that inherited the pane id.
- A worker that receives its task pointer and then emits no turn-shaped agent event within a response window now produces a `warn!` plus a scrollback notice in the orchestrator's pane (`DOT_AGENT_DECK_DELEGATE_NO_EVENT_WINDOW_MS`), so "never got the task" is distinguishable from "still thinking." The watch is cancellable (work-done / supersede / pane close) and immune to a lagged event bus producing a false positive.
- `quote_untrusted_role`'s framing delimiter is no longer forgeable by a role name.
- Both new `_MS` env knobs parse into `u128` and clamp before conversion, so a value above `u64::MAX` is capped rather than silently falling back to a default that could leave the silence report unexpectedly disabled.

## Gate results

- `cargo test-fast`: **1328 passed / 0 failed / 0 skipped**
- `cargo test-e2e` (pre-PR tier, rule 5; `DOT_AGENT_DECK_RECORD=1`, this HEAD): **2544 passed / 0 failed / 0 skipped in 105.875s**
- `cargo fmt --check`: clean. `cargo clippy --all-targets -- -D warnings`: clean, zero warnings.
- `cargo xtask linkage-check`: pass (373 catalog ids, 295 annotations, 104 allowlisted, 7 rules)

## M2 — OpenCode exposure (empirical, negative result)

OpenCode shares the legacy injection path so was presumed affected, but no reporter confirmed it. A real interactive OpenCode worker respawned through a `clear = true` delegate, run 3× at `DOT_AGENT_DECK_DELEGATE_READINESS_BUFFER_MS=0` against the final shipped code: **all 3 passed** (48.203s / 44.328s / 43.578s), plus a shipped-default control at 54.754s. **Exposure was NOT reproducible in these attempts.** This is a negative result, not a positive one — it does not prove OpenCode was unaffected, only that this couldn't reproduce it here. Per the PRD's own risk table this is the harmless case (a short wait before a write that would have succeeded regardless), and `orchestration/delegate/015` covers the path either way. Full ledger: `.dot-agent-deck/prd-249-m2-verdict.md` (not part of this diff — a working-directory artifact, referenced for reviewer context).

## Cross-version contract check (rule 12)

**Classification: PATCH.** No `PROTOCOL_VERSION` bump, no `changelog.d/249.breaking.md` — daemon-internal timing before an existing PTY write plus one additional daemon→pane notice write; no wire shape, field, or field meaning moved.

Manual previous-release-daemon run against the **published v0.35.0 release asset** (not a rebuild) serving an isolated sandbox, with the branch TUI/CLI/hook binary on `PATH` against it (modelling a real upgrade: binary replaced on disk, old daemon still serving):

- **A delegate still routes.** The old daemon wrote the task pointer; the `clear = true` respawn's replacement agent (real interactive Claude on Haiku) **submitted** it and acted on it, producing the expected sentinel.
- **Hooks still arrive.** Status transitions rendered live off the old daemon's stream; `work-done` was delivered into the orchestrator pane.
- Honest limit: because the *serving* daemon was v0.35.0, the readiness gate itself wasn't in the exercised code path — this validates the contract, not the fix. The fix is pinned by the fast-tier toggle test and the real-agent e2e specs.

Full write-up: `.dot-agent-deck/prd-249-rule12.md` (working-directory artifact, not part of this diff).

## Notes for reviewers

- **`src/agent_pty.rs`**: this diff contains only PRD #249 content (the identity-guarded write plumbing and silence-watch cancellation). An unrelated endpoint-scrub fix that also touches this file is already merged to `main` (`ff5170d`) and correctly does not appear here — a duplicate of it was dropped during rebase.
- **M6 is deliberately unchecked** in the PRD. It closes the public bug report (#199) by referencing this shipped fix and asking reporters to confirm — a post-merge step, out of scope for this PR.
- No feature flag: this is a correctness fix on an existing delivery path with no new user-visible surface (CLAUDE.md rule 9 doesn't apply).

## Release hygiene

- `changelog.d/249.bugfix.md` — already present, reviewed for accuracy (the OpenCode section states the M2 result as negative/unproven, not positive).
- No `PROTOCOL_VERSION` bump, no `.breaking.md` (see rule 12 above).

## Test plan

- [x] `cargo test-fast` green (1328/1328)
- [x] `cargo test-e2e` green (2544/2544, 0 skipped)
- [x] Slow-readiness toggle test fails at buffer `0`, passes at the shipped default (`tests/delegate_prompt_injection.rs`)
- [x] Real-agent e2e: `tests/e2e_delegate_respawn_readiness.rs` (Claude + OpenCode respawn through `clear = true`, sentinel-verified)
- [x] Cross-version manual test (rule 12) — v0.35.0 released daemon + branch TUI/CLI (delegate routes, hooks arrive)
- [x] `cargo fmt --check` / `cargo clippy --all-targets -- -D warnings` clean
- [x] `cargo xtask linkage-check` clean

Demo reel for PRD #249: https://youtu.be/v7ONU4LTa3E. Watch before merging.


vfarcic:
Both Greptile findings from the prior review are addressed in 8c6adce:

- Finding 1 (src/state.rs:2646, unsequenced `cancel_silence_watch`): replaced with `retire_silence_watch`, which credits each work-done to the oldest unaccounted-for delegation via a per-record `superseded` counter, so a stale work-done can no longer cancel a newer delegation's silence watch. The old unsequenced path was deleted. Covered by a regression test pinning both directions (timely work-done still cancels; stale one does not).
- Finding 2 (src/state.rs:1883, unconditional readiness-buffer sleep): added a `pane_close_signal` + `tokio::select! { biased; ... }` so a pane closed mid-wait abandons the dispatch task promptly instead of riding out the 30s clamp. The normal path is still a real awaited delay.

vfarcic:
Demo reel for PRD #249: https://youtu.be/v7ONU4LTa3E. Watch before merging.

PR #295, opened 2026-07-30
ci: probe the action majors in #290/#291 before merging them

vfarcic:
Temporary probe. **Do not merge** — it exists to produce evidence for #290 and #291, both of which bump actions on paths no PR exercises.

| PR | Bump | Where it runs |
|---|---|---|
| #290 | `docker/setup-buildx-action` v3 → **v4** | `docs.yml` (path-filtered to `docs/**`, `site/**`) and `docs-publish.yml` (`workflow_call`) |
| #291 | `upload-artifact` v4 → **v7**, `download-artifact` v4 → **v8** | `release.yml` — **tag-only** |

So merging them blind means the first execution is a release.

## What the probe asserts

**The contract, not just that the action ran.** `release.yml` uploads one artifact per matrix target under a distinct name, then `finalize` downloads them all with `merge-multiple: true` so they land **flat** in `dist/` for `task checksums`. `download-artifact` jumps four majors. If that input changed, binaries land in per-artifact subdirectories, checksums finds nothing, and the release fails at `finalize` **with the tag already pushed**.

The probe therefore checks both files are directly under `dist/` and nothing is nested — and for buildx, that it bootstraps a builder and completes a real build, since the docs image build is what depends on it.

Once green, #290 and #291 can merge on evidence and this branch gets deleted.

### DIFF
--- a/.github/workflows/action-major-probe.yml
+++ b/.github/workflows/action-major-probe.yml
@@ -1,0 +1,80 @@
+# TEMPORARY PROBE — delete once #290 and #291 are decided.
+#
+# Both PRs bump actions that live ONLY on paths no PR exercises:
+#   #290  docker/setup-buildx-action v3 -> v4   (docs.yml is path-filtered to
+#         docs/** and site/**, so a workflow-only change does not run it;
+#         docs-publish.yml is workflow_call, i.e. release-time only)
+#   #291  actions/upload-artifact v4 -> v7 and download-artifact v4 -> v8,
+#         in release.yml, which runs ONLY on a tag
+#
+# So merging them blind means the first execution is a release. This probe runs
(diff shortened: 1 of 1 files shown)


vfarcic:
Probe served its purpose — closing without merging.

**Evidence produced:**
```
--- dist tree ---
dist/dot-agent-deck-linux-amd64
dist/dot-agent-deck-linux-arm64
OK: upload@v7 -> download@v8 with merge-multiple stays flat

buildx v0.35.0
OK: setup-buildx-action@v4 initialises and builds
```

`merge-multiple: true` still flattens across `upload-artifact` v4→v7 and `download-artifact` v4→v8, with nothing nested — so `release.yml`'s `task checksums` will find the binaries. #290 and #291 merged on that basis.

PR #314, opened 2026-08-01
fix(state): retire superseded cards correctly for Pi respawn and scheduler handoff (#284)

### LINKED ISSUE #284: Duplicate cards on one pane: Pi respawn vs scheduler handoff — fixing either causes the other
## Summary

Two panes-with-duplicate-cards bugs sit at the same seam in `AppState::apply_event`, and the obvious fix for either one **causes the other**. `78f92b6` fixed the Pi case and broke the scheduler case; it has been reverted (see below), so the Pi case is currently open again.

It also broke a third, unrelated-looking test — see "Collateral damage" — which is the strongest argument for treating this seam as high-blast-radius rather than a one-line condition.

Whoever picks this up: the trap is that both symptoms look identical on screen — two cards on one pane — so it is easy to "fix" it, watch the symptom disappear, and ship the same defect reached from the other direction. **A fix is only correct if it satisfies both cases simultaneously.**

## The two cases

| | Trigger | Agent type | First frame carrying the new `agent_id` | Guarding test |
|---|---|---|---|---|
| **A — Pi respawn** | `clear = true` delegate respawns a worker | Pi | `Thinking` / `Idle` | `tests/agent_event.rs` (in history at `78f92b6`) |
| **B — scheduler handoff** | synthetic placeholder superseded by the agent's real hook | ClaudeCode / OpenCode | `SessionStart` | `scheduler/live/004` in `tests/e2e_scheduler_live_surface.rs` |

**Case A.** A `clear = true` delegate respawn mints a fresh `agent_id` while the pane keeps the card it was spawned with. The retire in `apply_event` was gated on `event.event_type == EventType::SessionStart`. Pi never sends one — its extension reports status through `dot-agent-deck agent-event`, whose vocabulary is running/waiting/finished — so a respawned Pi worker's first frame is a `Thinking` carrying the NEW agent id. The reuse guard declines to fold it onto the old card (ids differ) and the retire block declines to retire (not a `SessionStart`). Result: two permanent cards on one pane, and because the orchestration deck derives its highlight by pane id, the second is unreachable.

**Case B.** A scheduled fire surfaces a synthetic placeholder card (`agent_id = None`, `display_name = <task name>`, badge "No agent"). The agent's real `SessionStart` then arrives carrying that pane's id and its registry agent id — a `Some`, distinct from the placeholder's `None` — and no `display_name` metadata. The placeholder must be retired and replaced by one live card. Under `78f92b6` it is not, so both survive:

```
┌ 1 ClaudeCode · 9f8e7d6c-5b ──── ● Idle ┐┌ 2 No agent · morning-digest ── ● No agent ┐
│Dir:  runbox                            ││Dir:  runbox                              │
│Last: 10s ago  Tools: 0                 ││Launch an agent to get started            │
```

Both cards claim the same pane — the injected hook carries the placeholder's own pane id.

## What `78f92b6` changed

`src/state.rs`, one condition plus a new guard:

```rust
- if event.event_type == EventType::SessionStart && event.agent_id.is_some()
+ if event.agent_id.is_some()
+     && event.timestamp >= session.last_activity   // new
```

The widening is sound in itself: a fresh `agent_id` is minted per spawn, so any event bearing one that differs already proves the pane changed hands. `SessionStart` was never what made the inference valid — it is just the event most hook-based agents happen to send first.

The monotonicity guard was added so that a DELAYED event from the **outgoing** agent cannot retire the card the **incoming** one just established.

## Bisect evidence

```
49d12b0  (parent)   scheduler/live/004  PASS   0.58s
78f92b6             scheduler/live/004  FAIL  11.44s     3/3 runs
```

Deterministic. Note `scheduler/live/004` **shipped green in v0.35.0** — it is not a long-standing RED test, despite its own doc comment saying `RED today:` (that described the state when it was written; the gap was closed afterwards and the comment was never updated). **That stale comment is itself a trap** — it caused an earlier investigation to classify the failure as intentional and stop looking.

## Collateral damage: `prompt/close-confirm/005`

`close_confirm_005_vanished_armed_session_closes_nothing` (`tests/e2e_pane_close.rs`) was **also** broken by `78f92b6`, and was **misattributed to `a005907` (PRD #241)** during investigation — because it was introduced by that PR, looked like a hole in its own feature, and nobody re-ran it against a tree without `78f92b6`.

It arms a close confirmation with `Ctrl+W`, replaces the armed session's identity on the same reused `pane_id`, then confirms. Expected footer: `Nothing closed`. Observed under `78f92b6`: `Closed pane 1` — the confirmation executed against the *replacement*.

That is coherent with the retire change: whether the armed session counts as "vanished" depends on exactly when a superseded session is retired, which is what `78f92b6` altered. Reverting makes it pass in **0.57s** instead of failing at **11.9s**.

Two lessons worth carrying:

- **PRD #241's safe-pane-closing work is sound.** The armed-close hole it appeared to have was not its own.
- **Blast radius here reaches beyond cards.** A change to the retire predicate can alter pane-close semantics, because close confirmation is keyed on session identity. Any future fix must run the **full** e2e tier, not just the card tests.

## Dead ends already ruled out

Recorded so nobody spends the time again. Both of these were tried against case B and **neither is the cause**:

1. **The `or_insert_with` inheritance path.** `display_name: inherited_display_name` is only applied when the session is *created*, so widening the retire to a later event whose session already exists would drop the inherited name. Plausible, and there is a real latent hole here — but it is not what `scheduler/live/004` exercises, and filling it does not turn the test green.
2. **The hardcoded timestamp versus the new monotonicity guard.** The test injects `"timestamp": "2026-06-08T12:00:00Z"`, roughly seven weeks in the past, which cannot satisfy `event.timestamp >= session.last_activity`. Replacing it with a current timestamp — which is what a real hook carries — **does not** fix the test either. (The hardcoded date is still worth changing on its own merits; it just is not the bug.)

The failure is at the **sync** step, `wait_for_absence("No agent")`, not at the title assertion. The friendly name is not lost — it is sitting on the leftover placeholder card. Any investigation that starts from "the title regressed" is starting from the wrong symptom.

## Suggested approach

1. Build a **fast-tier** reproduction at `apply_event` level asserting on `AppState` directly. Both existing guards are e2e at ~11s per run, which is too slow to iterate on and reports a rendered grid rather than the state that actually went wrong.
2. Restore the case-A tests from `78f92b6` (`tests/agent_event.rs`) **and** keep `scheduler/live/004` green. Either one alone is how this pair got into the tree.
3. Only then decide the predicate. Worth checking specifically: which of the retire conditions fails for case B — the pane match, the `session.agent_id != event.agent_id` comparison against a `None`, or the timestamp guard — because the widening alone should have made case B *more* likely to retire, not less.

## Current state

- `78f92b6` reverted, restoring v0.35.0 behaviour exactly. Neither symptom is a regression against what users run: case A returns to being a known bug, case B never ships.
- Case-A tests are recoverable from history at `78f92b6`.
- With the revert, the **full e2e tier is green: 2456/2456**. Both `scheduler/live/004` and `prompt/close-confirm/005` pass. Before it, those were the only two failures in the suite — and `78f92b6` was the sole cause of both.
- Neither case is in CI: `cargo test-e2e` is the pre-PR tier per CLAUDE.md rule 5, so both guards only run locally before a release. That is why a commit could break two e2e tests and still merge.

## Definition of done

- [ ] A fast-tier reproduction of case A that fails without a fix
- [ ] Case-A tests restored from `78f92b6`
- [ ] `scheduler/live/004` still green
- [ ] `prompt/close-confirm/005` still green
- [ ] Full `cargo test-e2e` green, not a filtered subset
- [ ] The stale `RED today:` comment on `scheduler/live/004` corrected, so it stops misleading the next reader

vfarcic:
## Summary

Two duplicate-card bugs sat at the same seam — the retire predicate in `AppState::apply_event` (`src/state.rs`) — where the obvious fix for either caused the other:

- **Case A — Pi respawn.** A `clear = true` delegate respawn mints a fresh `agent_id`, but Pi never sends `SessionStart` (its extension reports through `dot-agent-deck agent-event`, whose vocabulary is running/waiting/finished), so gating the retire on `SessionStart` left a respawned Pi worker's first frame (`Thinking`/`Idle`) stacking a second, permanent, unreachable card on the pane.
- **Case B — scheduler handoff.** A scheduled fire's synthetic placeholder (`agent_id = None`) must be retired by the agent's real `SessionStart`. A prior attempt (`78f92b6`) fixed case A by widening the retire condition off `SessionStart` entirely and adding a naive timestamp guard, which broke case B (`scheduler/live/004`) plus a third, unrelated-looking test (`prompt/close-confirm/005`), and was reverted (`8f579dd`).

The fix (`src/state.rs`) distinguishes what each kind of frame is evidence of, via a new `supersedes_generation` helper shared by both supersession sites:
- `SessionStart` is the incoming generation **announcing** itself — self-describing, so the takeover is asserted unconditionally, exactly as before #284.
- Any other frame only lets the takeover be **inferred** from a changed `agent_id`, which a delayed frame from the outgoing agent can also produce, so an inferred retire additionally requires the event to be no older than a high-water-mark `last_activity`.
- `SessionEnd` is excluded from claiming a generation at all (it's terminal, the opposite of a takeover) — an earlier version of this predicate let a delayed/forged `SessionEnd` retire a live sibling, leaving zero cards on a live pane.
- A new same-producer-key path (`status/supersede/005`) handles Pi's stable `{pane_id}-session` id: because only `agent_id` changes across repeated respawns, only the *first* respawn was retiring correctly by accident (different key); every respawn after that silently kept the previous generation's identity and history on the surviving entry. The card now changes hands (fresh generation state, same key) instead of only working once.

Net effect on the predicate: a pure widening of the pre-#284 `SessionStart`-only gate — `SessionStart` still retires unconditionally, and the additions (guarded non-start inference, `SessionEnd` exclusion) only ever narrow what else may retire.

Closes #284

## Known hazard shipped on purpose (not fixed here)

Close confirmation arms on the **session id alone** (`CloseTarget::Session`) and resolves by direct key lookup. Because Pi's producer key `{pane_id}-session` is stable across respawns, an armed close target stays resolvable across a generation change, so confirming can act on whichever generation currently occupies the pane.

This **predates #284** and is neither introduced nor worsened by it: before the fix the key resolved to a stale corpse entry, after it resolves to the live replacement — in both cases it maps to the pane's current card. The #284 identity refresh is a **prerequisite** for fixing this properly (arming on generation = session id + agent id), because only now is `agent_id` refreshed in place so a generation change is even detectable. Documented in `src/state.rs` and on the `status/supersede/005` catalog entry; a follow-up to fix the arm/resolve seam is planned separately.

Also: a test, `status/supersede/006`, was written during this work and then **removed** during review — it was logically unsatisfiable against `status/supersede/005` (it demanded the same HashMap key be simultaneously present and absent). Its gap is documented on `/005`'s catalog entry rather than left silent.

## Test plan

- **Fast tier:** 1410 / 1410 passed.
- **Full recorded e2e tier:** 2724 run, 2723 passed. The single failure, `pi_live_002_native_seeded_orchestration_delegates_live`, was investigated and ruled a **flake**: 5/5 pass isolated on this branch, 5/5 on `origin/main`, 1/1 under a 163-test contended slice, and the suspected mechanism (this PR's same-producer identity refresh) was proven structurally unreachable for that test's `clear=false` hook-based Claude worker. Not a #284 regression.
- **linkage-check:** ok (413 catalog ids, 338 annotations, 101 allowlisted, 7 rules).
- `cargo fmt --check` and `cargo clippy -- -D warnings`: clean.
- **CLAUDE.md rule 12 cross-version manual test:** PASS, observed directly — this branch's TUI against a daemon built from v0.35.3 (sha `9c9b905`), both at attach protocol 6; delegate routed, status hooks drove the card, work-done returned.

New/restored tests: `tests/agent_event.rs` (`status/agent-event/005`, `/006`), `tests/card_supersession.rs` (new `status/supersede/*` sub-area, `/001`–`/005`, `/007`), `tests/e2e_scheduler_live_surface.rs` comment fix (stale `RED today:` text corrected).

## Versioning

Patch / bugfix. No `PROTOCOL_VERSION` bump, no `changelog.d/284.breaking.md`. No wire shape or field meaning changed; old/new peers interoperate (confirmed by the cross-version test above). Changelog fragment already exists at `changelog.d/284.bugfix.md`.


vfarcic:
Filed four follow-up issues from the #284 review and audit, deliberately scoped **out** of this PR to keep its blast radius contained to the retire-predicate fix itself:

- #317 — Close confirmation arms on session id alone, so it can act on a replacement generation. The highest-value follow-up: #284's identity refresh is the prerequisite that finally makes this fixable (arm on generation = session id + agent id).
- #318 — Daemon does not bind hook-event provenance: any same-user process can drive another pane. Pre-existing, same-user-only today; carries a MUST-REASSESS trigger before any Kubernetes/network hook transport ships.
- #319 — Hook ingestion has no line-length or connection bound. Found during the same audit sweep; unrelated to the retire-predicate logic itself.
- #320 — Retire path needs a per-pane generation discriminator, not a timestamp. Tracks the residual late-frame-from-the-outgoing-agent case (both the pre-existing `SessionStart` shape and the new non-start sibling this PR admits) as a first-class issue rather than only a code comment.

None of these block this PR — they're documented as intentional, tracked follow-ups.

PR #332, opened 2026-08-01
fix: default generated delegation guidance to --task-file

### LINKED ISSUE #303: orchestrator context advertises --task, never --task-file, so shell-mangled delegations ship silently
## The defect

The orchestrator role context that dot-agent-deck generates tells the orchestrator to delegate with `--task "…"` and never mentions `--task-file`. That text is executed by the orchestrator's own shell, so any task containing backticks, `$VAR`, `$(…)`, `!`, or nested quotes is mangled by command substitution / expansion **before dot-agent-deck ever receives it**.

## Where it lives

The **Delegation protocol** block is hard-coded in Rust, not in a config asset: `src/ui.rs:2201-2221` builds it with `content.push_str` and shows five `--task "…"` invocations (four `delegate`, one `work-done`). `--task-file` appears nowhere in that block — confirmed by grep over both the builder and `assets/`.

Two things follow from that placement:

- `.dot-agent-deck/orchestrator-context.md` is **generated and gitignored** (`.gitignore:6`), so editing it locally fixes nothing — it is regenerated from the Rust source.
- Fixing only `assets/config_gen_prompt.md` would reach *newly generated* configs while leaving **every existing project** broken. (`--task-file` is currently absent from `assets/` too.)

The worker-side block at `src/state.rs:508` (`WORK_DONE_FOOTER`) has the same exposure for `work-done --task "…"`.

## The capability already exists

`--task-file <PATH>` is implemented and has been for a while:

- `src/main.rs:116` (`delegate`) and `src/main.rs:131` (`work-done`), both with `conflicts_with = "task_file"` on `--task`.
- Accepts `-` for stdin.
- Its own help text names *this exact hazard*: "The shell-safe way to pass a task containing backticks, quotes, `$VAR`, or newlines, which --task would otherwise let the caller's shell mangle."
- The mechanism is documented in comments at `src/main.rs:446` and `src/main.rs:1750`.
- Parse-time conflict tests exist at `src/main.rs:1820`, `:1862`, `:1909`.

So this is **not** a missing feature and **not** an unrecognised risk. The generator simply emits advice that predates its own fix.

## Why the CLI cannot defend itself

Command substitution happens in the shell *before* argv is built. dot-agent-deck receives already-corrupted text that is indistinguishable from text the caller intended to send. No validation, warning, or heuristic inside the CLI can detect it. That is precisely why the only real fix is an input path that bypasses the shell.

## Severity — the failure is usually SILENT

It is loud only when the substituted text happens not to be a runnable command.

Example: a task containing a backticked `reel` marker causes the shell to run `reel`, write "command not found" to stderr that dot-agent-deck never sees, and substitute **empty** — the marker is silently deleted from the task while the delegation reports success. A worker then acts on subtly wrong instructions and nobody knows.

## Two reproductions, 2026-07-30, during PRD #249 (PR #271)

1. A delegated task containing backticked catalog ids `status/agent-event/005` and `/006` arrived with those ids **stripped**: the delivered instruction read `catalog ID  has no …` — the ids removed from the very sentence that was about deleting them. This one was loud only because those strings are not commands.

2. **Knock-on effect.** The orchestrator re-issued the corrected task, so the worker had **two armed delegations**. It did one piece of work and sent one `work-done`. The completion was credited to the *older* delegation, leaving the newer one armed, and roughly two hours later PRD #126's idle-worker detector injected a **false STUCK notice** for work that was already finished and committed. Re-delegating after a corrupted delivery is a realistic path into that state, even though the protocol says never to send a new task to a worker that is still working.

## Proposed fix (small)

- **`src/ui.rs:2201-2221`** — present `--task-file <path>` as the **default** for any task containing backticks, quotes, `$`, `!`, or newlines; keep `--task "…"` for short one-liners. Keep this explicitly distinct from the existing advice about writing long context to `.dot-agent-deck/<task-slug>.md` and referencing that path — that addresses *context length*, not shell safety, and still routes the text through `--task`.
- **`src/state.rs:508`** — same treatment for the worker-side `work-done --task` example.
- **Test:** extend the existing generated-content test at `src/ui.rs:17795` (which already asserts the block contains `"Delegation protocol"`) to also assert it mentions `--task-file`. L1 content assertion; no PTY needed.
- **Changelog fragment**, since this changes generated text that users see.

vfarcic:
## Summary

Fixes #303. The orchestration guidance dot-agent-deck generates advertised only `delegate --to <role> --task "…"` and never `--task-file`. That text is executed by the orchestrator's own shell, so backticks, `$VAR`, `$(…)` and quotes were expanded before dot-agent-deck ever received argv — silently deleting parts of a task while the delegation still reported success.

- `src/ui.rs` — the generated Delegation protocol block now defaults to `--task-file`.
- `src/state.rs` — the worker-side `## When done` footer gets the same treatment, plus a collision-resistant role-specific report path outside the daemon's own `work-done-*` namespace, a positive allowlist replacing an unsound hazard-character blocklist, and a three-branch fallback for agents whose file-writing tool isn't authorized.
- `assets/config_gen_prompt.md` / `src/config_gen.rs` — the prompt that generates new project configs leads with `--task-file` too, so newly generated configs don't ship the old advice.
- `docs/orchestration.md` — documents the same guidance for readers of the docs site.
- `tests/CATALOG.md`, `tests/e2e_delegate_respawn_readiness.rs`, `tests/e2e_delegate_work_done_chain.rs` — updated/extended coverage for the new guidance.
- `changelog.d/303.bugfix.md` — release note.

## Contract check (CLAUDE.md rule 12)

This touches orchestration guidance, so the rule applies:

- **No TUI↔daemon contract change.** No wire shape, no field meaning, no hook payload, no delivery-semantics change. `src/main.rs`, `src/daemon.rs`, `src/daemon_protocol.rs`, and `src/init.rs` are untouched.
- **`PROTOCOL_VERSION` stays 6** — not bumped.
- **No `.breaking.md` fragment** — this is patch-level (bugfix).

## Gate results

| Gate | Result |
|---|---|
| `cargo clippy --all-targets --features e2e -- -D warnings` | pass, 0 warnings |
| Fast tier (`cargo nextest run -j 4 --build-jobs 2`) | 1,405 run, 1,405 passed, 0 failed, 0 skipped |
| E2E tier (`DOT_AGENT_DECK_RECORD=1 cargo nextest run --features e2e -j 4 --build-jobs 2`) | 2,719 run, 2,719 passed, 0 failed, 0 skipped, ~7m03s |

The e2e tier was run twice. The first run failed 1/2719 (`delegate_work_done_chain_claude`) because the new guidance required a `Write` tool the test harness had not authorized — that's actually the fallback-branch behavior working as intended. It was fixed in both the guidance (three-branch fallback, see above) and the harness (`--allowedTools Bash Read Write`); the rerun is fully green.

## Test plan

- [x] `cargo fmt --check`
- [x] `cargo clippy --all-targets --features e2e -- -D warnings`
- [x] Fast tier: `cargo nextest run -j 4 --build-jobs 2`
- [x] E2E tier: `cargo nextest run --features e2e -j 4 --build-jobs 2`

## Demo reel

Demo reel for PRD #303: https://youtu.be/7ZTG6zi4nCc. Watch before merging.


vfarcic:
Demo reel for PRD #303: https://youtu.be/7ZTG6zi4nCc. Watch before merging.

Issue #313, opened 2026-08-01
PRD: Zoom the focused agent pane


vfarcic:
## PRD: Zoom the focused agent pane

**Problem**: Even with the pane column given over entirely to the focused agent, roughly a third of the width is sidebar and several rows are chrome. On a laptop there is no way to say "I am working in this one agent now, nothing else matters".

**Solution**: A reversible zoom toggle that gives the focused agent effectively the whole terminal, and returns to the normal multi-agent view on a second press — the `tmux prefix+z` model.

**Detailed PRD**: See [prds/313-zoom-focused-agent-pane.md](https://github.com/vfarcic/dot-agent-deck/blob/main/prds/313-zoom-focused-agent-pane.md)

**Priority**: Medium

Issue #312, opened 2026-08-01
PRD: Retire the global stacked/tiled layout toggle


vfarcic:
## PRD: Retire the global stacked/tiled layout toggle

**Problem**: A single global `pane_layout` field drives three tab types with incompatible needs — after #311 removes collapsed frames, `Stacked` and `Tiled` are indistinguishable in orchestration tabs, while `Tiled` remains the only sensible arrangement for mode tabs whose side panes exist to be watched.

**Solution**: Delete the `Ctrl+T` toggle and the `toggle_layout` action, and make the arrangement a property of the tab type rather than a global mode the user has to manage.

**Detailed PRD**: See [prds/312-retire-global-layout-toggle.md](https://github.com/vfarcic/dot-agent-deck/blob/main/prds/312-retire-global-layout-toggle.md)

**Priority**: Medium

Issue #341, opened 2026-08-02
Make Command vs PaneInput mode unmistakable


vfarcic:
## PRD: Make Command vs PaneInput mode unmistakable

**Problem**: The release that colored the focused pane border by mode helped, but users still routinely believe they are in one mode while actually in the other. The border is a low-amplitude signal, and it is actively contradicted by the pane cursor — which renders identically in both modes, telling the user "type here" while the keyboard is driving the deck.

**Solution**: Make mode legible on several redundant channels at once — gate both pane cursors on `input_active`, add a persistent mode chip that names the *current* mode, dim the focused pane and overlay a decaying "COMMAND MODE" banner, make deck selection mode-aware, and enable agent-pane scroll in command mode so it becomes a genuine read-only inspect mode.

**Detailed PRD**: See [prds/341-make-ui-modes-unmistakable.md](https://github.com/vfarcic/dot-agent-deck/blob/main/prds/341-make-ui-modes-unmistakable.md)

**Priority**: High

PR #334, opened 2026-08-02
fix(prd-311): remove collapsed-pane frames from non-focused Stacked panes

prageethw:
## Summary

Implement PRD #311: Remove collapsed-pane-frames from non-focused Stacked panes.

This PR completes the removal of the visual "collapsed frame" rendering for non-focused Stacked panes, simplifying the layout geometry and PTY sizing. The previous behavior rendered non-focused Stacked panes as small "collapsed" frames; now they render at their calculated natural size, with their PTY windows sized as if the pane were focused.

**Closes #311**

## Changes Made

- **Rendering**: Dropped the collapsed-frame render arm from `render_terminal_panes`' Stacked branch in `src/ui.rs`
- **Geometry**: Fixed `pane_stack_rects` and `stacked_expanded_index` formulas to correctly compute pane positions and sizes without the collapsed-frame logic
- **PTY Sizing**: Non-focused Stacked panes' PTYs are now sized "as if focused" using the `pane_target_dims` function, ensuring the spawned agent runs at full capacity regardless of focus state
- **Tests**: 
  - RED tests added in `tests/e2e_orchestration_pane_column.rs` (PTY-attached L2 with real agent integration)
  - Extended regression guards in `tests/e2e_mode_tab_layout.rs`
  - L1 snapshot test in orchestration/layout/002
- **Docs**: Updated comments and test names to remove references to the removed collapsed-frame behavior
- **Shell Quoting**: Fixed shell-quote safety in new e2e fixture (auditor finding)

## Testing

All tests passing:
- `cargo test-fast`: 1421/1421 ✅
- `cargo test-e2e`: 2800/2803 passed (3 pre-existing failures in unrelated tests: dispatch_005, manager_010, restore_001 — tracked in separate PRDs #89, #**)
- 127 `.cast` recordings generated for demo-reel step

## Cross-Version Safety

Not applicable — TUI-only rendering/geometry change, no daemon protocol/hook/orchestration-routing changes. No breaking changes to the TUI↔daemon contract. Patch-level version bump.

## Files Changed

- `src/ui.rs` — geometry and rendering logic
- `tests/e2e_orchestration_pane_column.rs` (new) — L2 PTY test
- `tests/e2e_mode_tab_layout.rs` (new) — regression test
- `tests/fixtures/orch-focus-lifecycle` (new) — fixture for PTY test
- `tests/fixtures/mode-two-side-panes` (new) — fixture for regression test
- `tests/CATALOG.md` — recorded test entries
[...]

PR #342, opened 2026-08-02
feat(prd-336): toggle orchestration pane-column split ratio

### LINKED ISSUE #336: PRD: Toggle orchestration pane-column split ratio (1/3-2/3 <-> 1/4-3/4)
## PRD: Toggle orchestration pane-column split ratio

**Problem**: In the orchestration tab, the sidebar (left) and pane column (right) split at a fixed 34%/66% (`ORCHESTRATION_LEFT_PERCENT` / `ORCHESTRATION_PANES_PERCENT`, `src/ui.rs:1951-1952`). There's no quick way to reclaim more width for the working pane without a config edit.

**Solution**: A keybinding toggle (default `Ctrl+l`, remappable) that switches the orchestration tab's split between the default 34/66 and a narrower-sidebar 25/75, and back again on a second press.

**Detailed PRD**: See [prds/336-toggle-orchestration-pane-split-ratio.md](https://github.com/vfarcic/dot-agent-deck/blob/main/prds/336-toggle-orchestration-pane-split-ratio.md)

**Priority**: Medium

prageethw:
## Summary

Implement PRD #336: add a per-tab toggle for the orchestration pane-column split ratio.

`PaneLayout` orchestration tabs split the sidebar (role list, left) and the pane column (agent terminals, right) at a fixed `ORCHESTRATION_LEFT_PERCENT = 34` / `ORCHESTRATION_PANES_PERCENT = 66` ratio, with no quick way to reclaim width for the working pane. This PR adds `Ctrl+l` (remappable, `ToggleOrchestrationSplit` action) to toggle between the default 34/66 split and a narrower-sidebar 25/75 split, per orchestration tab, and back again on a second press.

**Closes #336**

## Changes Made

- **State**: New per-tab `orchestration_split_narrow: bool` field (defaults to `false` / 34-66) alongside existing per-tab layout state in `src/tab.rs`
- **Keybinding**: `ToggleOrchestrationSplit` action registered in `src/keybindings.rs`'s `ACTIONS` table, default chord `Ctrl+l`
- **Layout**: `orchestration_split_percents(narrow: bool) -> (u16, u16)` resolver added; the three `ORCHESTRATION_LEFT_PERCENT`/`ORCHESTRATION_PANES_PERCENT` call sites in `src/ui.rs` now resolve the active per-tab ratio instead of reading the fixed constants
- **Bug fix (found in review)**: `orchestration_role_pane_dims` was reading a stale thread-local across tabs, causing a cross-tab spawn-order staleness bug where a newly spawned orchestration tab could inherit another tab's split state. Fixed with an explicit `narrow: bool` parameter threaded through instead of the thread-local; regression test added (orchestration/layout/003) so new tabs always spawn at the default split regardless of another tab's toggled state
- **Tests**:
  - RED→GREEN TDD: `orchestration/layout/002` (L1 `insta` snapshot, both ratio states' geometry)
  - `tabs/orchestration/006` (L2 vt100 test, PTY-attached, drives the real chord and asserts the visible column-width change and round-trip)
  - `orchestration/layout/003` (regression test for the cross-tab spawn-order bug)
- **Docs**: `docs/keyboard-shortcuts.md` updated with the new binding; `changelog.d/336.feature.md` added

## Testing

- `cargo test-fast`: 1428/1428 passing
- `cargo test-e2e` (`DOT_AGENT_DECK_RECORD=1`, the required pre-PR gate): 2786/2788 passing. The 2 failures (`dispatch_005_respects_max_per_run`, `manager_010_blank_default_command_falls_back_to_claude`) are pre-existing and unrelated to this PRD's UI/keybinding changes — known flakes in this flaky-tolerant tier per CLAUDE.md rule 5
- `cargo fmt --check` / `cargo clippy -- -D warnings`: clean throughout all commits

## Cross-Version Safety

None — TUI-side rendering/keybinding state only, no daemon protocol, no hooks, no orchestration routing change. Patch-level version bump.

## Files Changed

- `src/tab.rs` — per-tab split-ratio state
- `src/keybindings.rs` — `ToggleOrchestrationSplit` action, default `Ctrl+l`
- `src/ui.rs` — resolver function, three call sites updated, cross-tab staleness fix
- `tests/e2e_orchestration_pane_column.rs` — L2 PTY test
- `tests/CATALOG.md` — recorded test entries
- `docs/keyboard-shortcuts.md` — new binding documented
- `changelog.d/336.feature.md` — changelog entry
- `prds/336-toggle-orchestration-pane-split-ratio.md` — PRD doc

## Review Focus

- Confirm the per-tab state is correctly scoped (toggling one orchestration tab doesn't affect another)
- Verify the cross-tab spawn-order fix (`narrow: bool` param replacing the thread-local) fully closes the staleness bug rather than narrowing its window
- Check that non-orchestration tabs (dashboard, mode tabs) are unaffected by the toggle
- Confirm the chord is cleanly remappable through the existing `keybindings.rs` mechanism

## Follow-Up

None — PRD #336 is complete. Out of scope by design: #312 (retiring the global stacked/tiled toggle), #313 (zoom), more than two ratio states, and persisting toggle state across restarts.


vfarcic:
@prageethw — heads-up: I've pushed to this branch rather than sending it back to you, because #334 landing made the rebase non-trivial and I wanted the feature unblocked. **Your branch was force-pushed twice**, so please `git fetch` and reset before touching it again. Nothing of yours was lost — both pushes used `--force-with-lease` against the exact commit I'd reviewed, and the feature commit is still authored to you.

Sorry for doing this much without checking in first. Below is everything that changed and why, and one open design question I'd like your view on.

## The rebase

The branch was 54 commits behind and conflicting. It also bundled all of PRD #311 as a squashed "CI shadow" commit, and #311 landed separately meanwhile via #334 — so I dropped that content and kept only #336. Because #334 claimed the spec ids first, the tests were renumbered:

| Was | Now |
|---|---|
| `orchestration/layout/002` (Ctrl+l L1) | `orchestration/layout/003` |
| `orchestration/layout/003` (spawn-order) | `orchestration/layout/004` → now per-tab isolation |
| `orchestration/layout/004` (#311) | dropped, on `main` as `layout/002` |
| `tabs/orchestration/006` (Ctrl+l L2) | `tabs/orchestration/007` |
| `tabs/orchestration/007` (PTY-forward) | `tabs/orchestration/008` |
| `tabs/orchestration/008`, `tabs/mode/001` (#311) | dropped, on `main` |

Worth knowing for next time: resolving those `src/ui.rs` and `CATALOG.md` conflicts by hand would have silently reverted #334's own review fixes, since this branch carried the pre-review version of #311. Dropping wholesale was the safe move.

## Five substantive changes

**1. The thread-local is gone.** `ACTIVE_ORCHESTRATION_SPLIT_NARROW` was replaced by threading `split_narrow` through `ActiveTabView::Orchestration` — already the per-tab render snapshot `compute_frame_layout` receives — plus an explicit `narrow` parameter on `orchestration_role_pane_dims`. It was *correct* as written; I checked the ordering. But its correctness rested on call ordering nothing enforced, and your own `layout/003` existed only to pin one hazard it created. Threading it as data removed both.

**2. `Ctrl+l` is now command-mode only — a behavior change, not in your PRD.** Greptile's P1 was right that the resolver claimed the chord everywhere, and your fix handled non-orchestration tabs. But it still swallowed `Ctrl+l` in **PaneInput on an orchestration tab** — which is where someone is most likely to want readline's clear-screen, in a role pane. So it now mirrors `close_pane` (PRD #241 M1), which is command-mode only for exactly this reason. Press `Ctrl+d` first, then `Ctrl+l`. The scoping lives in a pure `scope_orchestration_split(action, is_orchestration_tab, mode)` so it's unit-testable without a PTY, and the help overlay lists the binding under "Dashboard (command mode)" rather than "Global", as PRD #241 review F6 did for `close_pane`.

**3. `Ctrl+l` was missing from the in-app help overlay** (`?`), which the docs advertise as the complete shortcut list. Added. My first label rendered *truncated* in the snapshot, so it's now the shorter "Toggle orch tab split ratio".

**4. The PTY-forwarding test was rewritten.** It asserted that readline's `clear-screen` wiped a sentinel. That depends on the host's terminal setup — it failed on my machine while the forwarding was in fact working correctly, which I confirmed by reproducing the same non-clearing readline in a bare PTY with no deck involved. It now runs `cat -v` and asserts the pane echoes `^L`, observing the forwarded byte directly instead of a downstream side effect.

**5. Two smaller fixes.** The L2 edge-detection helper now accepts any box-corner glyph — PRD #341 renders the focused pane's border heavier in command mode, so pinning `┌` broke the test on a mode switch that was working. And the L1 chord assertion now checks the specific `Action` variant rather than `is_some()`, which would have passed even with the guard deleted.

## Greptile

It has re-raised the original P1 on the current head. **It's a false positive now.** It reads `global_action` (`src/ui.rs:6426`) in isolation without following the call site: the chord is un-resolved by `scope_orchestration_split` at `:8882`, well before `UiMode::PaneInput => handle_pane_input_key` at `:8961` forwards to the PTY, and `tabs/orchestration/008` proves it on a real PTY in exactly the scenario described. No action needed.

## Verification

Local: every gate green, e2e **3411/3411, 0 skipped**. CI green on all platforms — `build`, `build-macos`, `build-windows`, `security`, `windows-cross-check`, `devbox`, `docs-build`.

## The one open question: should the width be global?

Your PRD scopes the split **per orchestration tab** ("toggling one tab does not affect others"), and that's what's implemented and tested.

I've come to think global is the better design, and I'd like your view before changing it. My reasoning: I can't find a case where one person wants tab A narrow and tab B wide. Different *users* will want different widths, but a single user wanting them mixed seems unlikely — sidebar width reflects how someone reads, not which orchestration they opened. Meanwhile per-tab has a real cost: every new orchestration tab resets to 34/66, so anyone who prefers the narrow sidebar re-toggles it forever.

The nearest counter-case I can construct is transient — "right now I'm reading a wide diff in this tab" — but that's per-task rather than a per-tab preference, and #313 (zoom) is its proper home.

So I'm inclined to make it global, and to follow up separately with persistence so the choice survives a restart (that part is additive, so it doesn't need to block this).

**Did you have a reason for per-tab that I'm missing?** If you'd rather make the change yourself, say so and I'll leave it to you — it's your design and your PR. If you're happy with global, I can do it here. Either way I'll hold the merge until I hear from you.

## Also on this branch

The second commit (`chore(dev)`) is **unrelated tooling** — a `task run` / `task run-stop` pair for starting an isolated local deck, plus a CLAUDE.md rule and a `verify-pr` note. It's authored to me and kept separate deliberately so it can be dropped or split out before merge. It touches no shipped code.

Thanks for this one, and for #334 — sorry again for the heavy hand on the branch.

prageethw:
Agreed — global is the better design, and no, there's no per-tab reason you're missing. Your framing convinced me: sidebar width is a reading preference, not a property of which orchestration happens to be open, and the re-toggle-on-every-new-tab cost is real. The transient "wide diff right now" case does belong to #313 rather than here.

I'll make the change myself since it's my PRD — thanks for offering either way. Coming up: the global split, the PRD correction (it still specifies per-tab), and the test inversion. `orchestration/layout/004` currently pins per-tab *isolation*, so that assertion flips, and I'll add coverage for a newly opened orchestration tab adopting the current split, since that's the actual payoff of going global and would be easy to ship untested.

Everything else in the rebase looks right to me — dropping the #311 shadow wholesale rather than hand-resolving it, the renumbering, and replacing the thread-local by threading `split_narrow` through `ActiveTabView::Orchestration`. Agreed the thread-local was correct but rested on call ordering nothing enforced.

Scoping `Ctrl+l` to command mode is a behavior change beyond my PRD, and it's the right one. Swallowing it in PaneInput on an orchestration tab is precisely where someone wants readline's clear-screen, and mirroring `close_pane`'s scoping is more consistent than what I'd written. Same for the `cat -v` / `^L` rewrite of the PTY-forward assertion — observing the forwarded byte directly is much better than asserting on a downstream readline side effect that depends on the host's terminal setup.

No apology needed for the force-pushes. The branch was 54 commits behind and conflicting; you unblocked it.

prageethw:
Status update, since you said you'd hold the merge until you heard back — you have, and there's nothing outstanding from my side on this one.

The split is now **global**, as you proposed. Two commits since your review:

- **`daa6ccd`** — `TabManager::orchestration_split_narrow` is the single source of truth, flipped by `toggle_orchestration_split()`. That method rewrites the global and every open `Tab::Orchestration::split_narrow` mirror in one `&mut self` pass, so there's no window where tabs disagree and no call-ordering assumption anywhere — which was the specific thing wrong with the thread-local you removed. Both construction sites seed from the global instead of hardcoding `false`. PRD, `docs/keyboard-shortcuts.md` and the changelog fragment all updated to match, including the architecture rationale that previously argued *for* per-tab.
- **`1844451`** — a new L1 test, `orchestration/layout/006`, covering spawn-time PTY dims. That was an uncovered behaviour change riding along: opening a tab while the global is already narrow previously spawned the role PTY at 64 cols and only reflowed to 73 on the first frame. Verified the test actually fails without the fix (64 vs 73) rather than trusting that it passes.

One thing I did **not** cover, deliberately, and want to flag rather than leave you to find: the restore/hydrate call site got the same change, but it's currently unobservable. `apply_snapshot` runs once at cold start before the event loop, so `orchestration_split_narrow()` is always `false` at that point — byte-identical to the old hardcoded `false`. It becomes meaningful only once the split persists across launches, which is the follow-up you mentioned. It's noted in `orchestration/layout/006`'s "Does not assert" line rather than left silent.

CI is 10/10 green. Greptile's inline P1 on the `Ctrl+L` pane-input path is the same 2026-08-02 comment, unchanged — its `original_commit_id` is still the pre-rebase `79e29b5`, GitHub is just remapping the anchor forward, so it's the false positive you already assessed rather than a re-raise.

The `chore(dev)` tooling commit is still on the branch — that one's yours to drop, split or keep as you prefer; I've left it alone.

prageethw:
The fork has shipped the generalised version of this feature — a deck-global, command-mode-scoped `Ctrl+L` split toggle — released in v0.35.8 (see prageethw/dot-agent-deck#19).

This PR is deliberately not being re-scoped. The follow-up is queued as a *separate* upstream PR that will delete `split_narrow` in the same change, so upstream never briefly carries two generations of the mechanism. Reopening #342 to widen it would cost the two reviews it has already had and delay the `Ctrl+L` fix it already carries.

The follow-up is gated on this PR merging — that is the only reason it has not been opened.

For context: on an orchestration tab `Ctrl+L` was claimed **regardless of UI mode** in upstream, so a focused role pane's agent never received `0x0c` and Claude Code's own clear-screen never reached it. The fork fix addresses this. Your `scope_orchestration_split` is the seed the fork generalised.

Follow-up tracking: upstream issue #387.

prageethw:
No reason you're missing — go global. I don't have a case for per-tab; it was the narrower default rather than a considered choice, and your argument for global is the right one.

What's worth adding is that I've since reached the same conclusion twice, independently, on my fork:

- **#387** moves the split stage itself deck-global — the same setting, same reasoning.
- **#393** moves the command-entry lock deck-global, and the PRD there ended up phrasing it almost exactly as you did: the setting reflects *how someone is working right now, not which tab they happened to open*.

In both cases the deciding factor was the cost you named — a per-tab default means every new orchestration tab resets, so anyone with a preference re-toggles it forever. That's not a preference the tab should own. So this isn't me conceding the point; it's two more data points landing on your side.

**Please make the change here** — you're already in the branch and it'll be faster than a round trip. Persistence as a separate follow-up sounds right; agreed it shouldn't block this.

On the rest: no problem at all about pushing to the branch. Dropping the bundled #311 content wholesale was clearly the right call given it carried the pre-review version, and I'd rather you unblocked it than sent it back. Thanks for the `--force-with-lease` note — I'll fetch and reset before touching it.

Your five changes all read right to me. Two I'd call out specifically:

- **Threading `split_narrow` through `ActiveTabView::Orchestration` instead of the thread-local.** You're right that its correctness rested on call ordering nothing enforced — and that my `layout/003` existed only to pin a hazard the thread-local created. Removing the cause rather than testing the symptom is the better trade.
- **Scoping `Ctrl+l` to command mode.** Agreed, and it generalised: I hit the identical problem with `Ctrl+e` afterwards and followed the shape you used here — a pure `scope_*(action, is_orchestration_tab, mode)` resolved before the PaneInput branch, unit-testable without a PTY. That's #393/#404. Your `close_pane` precedent is doing a lot of work across several of these.

Also agreed on the Greptile re-raise being a false positive — reading `global_action` without following through to the un-resolution at the call site is exactly what it did.

Sorry for the slow reply on this one.

Issue #355, opened 2026-08-03
e2e: restore_001_no_flag_startup_restores_panes_from_snapshot fails on main — restored-beta pane never renders


prageethw:
## Bug

`cargo test-e2e restore_001_no_flag_startup_restores_panes_from_snapshot` (`tests/e2e_session_restore.rs:250`) fails consistently, not flakily — reproduced on plain `origin/main` (`4f188f9`) in an isolated rerun (11.2s), same panic both times:

> PRD #89 M2.1: launching with NO --continue and a 2-pane snapshot on disk must auto-restore BOTH saved panes (`restored-alpha`, `restored-beta`) as dashboard cards, but they never appeared. RED until the snapshot-load block in `run_tui` is made unconditional (today it is gated on `continue_session`).

Both `restored-alpha` and `restored-beta` appear as session-list cards, but only `restored-alpha`'s pane content actually renders on the right — `restored-beta`'s pane never draws.

Notably, **PRD #89 ("Auto-restore TUI state on attach; remove --continue flag") is already closed** — this test's own panic text ("RED until...") reads like an unfinished milestone (M2.1) that shipped in a red state, or a regression introduced after the PRD closed.

## Where found

Surfaced during the PRD #333 (orchestration tab label priority color) pre-PR e2e gate. That branch touches only tab-label-color rendering in `src/palette.rs`/`src/ui.rs` — zero overlap with `tests/e2e_session_restore.rs` or `run_tui`'s snapshot-restore path (confirmed via `git diff origin/main...HEAD --stat`). Independently reproduced on stock `origin/main` in a separate worktree, confirming this predates and is unrelated to PRD #333.

**Priority**: Medium — a real functional gap (multi-pane snapshot restore silently drops all but one pane), not test infra flake, and it affects a feature (#89) already marked closed/shipped.

prageethw:
This issue's repro is invalid, so I'm closing it.

**The repro claim was wrong.** This issue states the failure "reproduced on plain `origin/main` (`4f188f9`)". `4f188f9` is not plain `origin/main` — it is this branch's own PRD #336 commit (`fix(prd-336): toggle orchestration pane-column split ratio [CI shadow] (#2)`). So the repro actually ran against a tree that already contained the PRD #311 change that causes the failure, not against clean upstream `main`.

Credit to @vfarcic for catching this and measuring it properly:
- merge-base: passes 3/3
- `483fe3d1` (the PRD #311 commit): fails
[...]

Issue #351, opened 2026-08-03
cargo test-e2e: cluster of PTY-wait timeouts under full-parallel load (5-6 tests, different file each run)


prageethw:
## Bug

Full `cargo test-e2e` runs (2764 tests) intermittently drop a cluster of tests to PTY-wait timeouts or assertion mismatches — the same shape of flake CLAUDE.md rule 5 / issue #346 already root-caused for `tests/delegate_prompt_injection.rs`, but here spread across unrelated e2e files. Observed in one run:

- `e2e_mouse_picker::picker_001_click_confirm_opens_form` — PTY wait timeout, did not see "No active sessions" within 10s (`tests/common/mod.rs:801`).
- `e2e_mouse_picker::picker_001_double_click_enters_dir` — same timeout/needle.
- `e2e_new_pane_schedule_option::new_pane_010_issue_dispatch_option_flag_gated` — same timeout/needle.
- `e2e_scheduler_live_surface::live_003_scheduled_card_title_shows_friendly_name` — same timeout/needle.
- `e2e_render_contract::render_001_enlarge_fills_new_width` — PTY wait timeout, did not see "[Command Mode Ctrl+D]" within 10s (same helper, different needle).
- `e2e_issue_dispatch::dispatch_005_respects_max_per_run` — assertion `left==right` failed (max_per_run cap, left:1 right:2) in the full-suite run; passed clean in isolation (3.26s) on a separate rerun — confirms load-induced, not a real cap-logic bug.

Four of the six share the exact same shared PTY-wait timeout helper (`tests/common/mod.rs:801`) and needle text, strongly suggesting resource contention under full-parallel `nextest` e2e execution rather than independent bugs — the same category as #346, just in different files/needles.

## Where found

Surfaced during the issue #346 (`delegate_prompt_injection.rs` flakiness fix) pre-PR e2e gate. Unrelated to that change's diff (only touches `tests/delegate_prompt_injection.rs`).

**Priority**: medium — erodes trust in `cargo test-e2e` the same way #346 did for `cargo test-fast`; worth applying the same fix pattern (bounded wall-clock retry instead of a single fixed-timeout wait) to the shared PTY-wait helper at `tests/common/mod.rs:801`.

vfarcic:
Adding one more test to this cluster, found while verifying PR #334 (a rendering-only change that cannot plausibly cause it).

## `dispatch_013_orchestration_surfaces_and_delegates`

[...]

Issue #349, opened 2026-08-03
e2e: manager_010_blank_default_command_falls_back_to_claude panics — real regression, not flake


prageethw:
## Bug

`cargo test-e2e manager_010_blank_default_command_falls_back_to_claude` (`tests/e2e_scheduler_manager.rs:664`) fails consistently, not flakily — reproduced across two independent \`cargo test-e2e\` runs (full-suite and isolated reruns), both with the same panic:

> an unset/blank default_command must fall back to claude (not spawn a bare $SHELL): the authoring agent must run claude and deliver the base authoring seed — the claude recorder never received it

Isolated reruns (18.96s, 19.35s) reproduced it deterministically, ruling out timing/load flake.

## Where found

Surfaced twice independently while gating unrelated changes:
- During PRD #336 e2e validation.
- During the issue #346 (`delegate_prompt_injection.rs` flakiness fix) pre-PR e2e gate.

Neither change touches `tests/e2e_scheduler_manager.rs`, `src/ui.rs`, `keybindings.rs`, or `tab.rs` — this is a pre-existing regression in the scheduler-manager Add-flow, unrelated to both.

**Priority**: should be root-caused — it's a real functional gap (blank `default_command` not falling back to `claude`), not test infra flake.

PR #356, opened 2026-08-03
feat(prd-333): color orchestration tab labels by highest-priority pane status

### LINKED ISSUE #333: feature: multitabs show colors to show status ? working ,idle or need input?
## PRD: Orchestration tab label reflects highest-priority pane activity

**Problem**: Today you have to open an orchestration tab to see whether anything inside it needs attention — the tab itself carries no status signal, so a user with several orchestration tabs open has no at-a-glance way to know which one needs them.

**Solution**: Color the orchestration tab's label text using the same per-pane status palette already in `src/palette.rs`, reflecting the highest-priority state among that tab's panes: Error > Needs Input > Working > Thinking > Idle.

**Detailed PRD**: See [prds/333-orchestration-tab-status-color.md](https://github.com/vfarcic/dot-agent-deck/blob/main/prds/333-orchestration-tab-status-color.md)

**Priority**: Medium

prageethw:
## What

Colors an orchestration tab's label using the existing per-pane status palette (`palette::status_color()`), reflecting the single highest-priority state among that tab's panes: Error (red) > Needs Input (yellow) > Working (green) > Thinking (blue) > Idle (dark gray). No new colors are introduced — this is a new aggregation over the five colors `palette.rs` already emits.

## Why

Closes #333

A user running several orchestration tabs has no way to tell, without switching to each one, whether anything inside it needs attention. The sidebar already shows per-role status once a tab is focused, but that signal is invisible from the tab bar itself. This surfaces the single most urgent pane state per tab at a glance.

## Changes

- `src/palette.rs`: aggregate-priority resolver (M1) — pure function mapping an orchestration tab's pane statuses to the single highest-priority `SessionStatus`, with the Compacting→Thinking and Unknown→Idle aliasing already established in the palette.
- `src/ui.rs`: `render_tab_bar_to_buffer` now colors an orchestration tab's label via the resolver's output (M2). Non-orchestration tabs are unaffected.
- `tests/render_tab_strip.rs`: L1 `insta` snapshot coverage for the resolver's priority ordering and aliasing, and for the rendered tab-bar color under representative multi-pane mixes (M3); `layout_004` now models a real zero-pane orchestration tab.
- `docs/orchestration.md`: documents the new tab-label behavior (M4).
- `prds/333-orchestration-tab-status-color.md`: PRD marked Complete, all milestones checked.

## Gate results

- `cargo fmt --check` / `cargo clippy --all-targets -- -D warnings`: clean.
- Full `cargo test-fast`: 1431/1431 passed (confirmed twice — once by the implementer, once by an independent rerun).
- Full `cargo test-e2e` (with `DOT_AGENT_DECK_RECORD=1`): 2812/2815 passed. The 3 failures are all pre-existing and unrelated to this PRD, confirmed by reproducing against stock `origin/main`:
  - `manager_010_blank_default_command_falls_back_to_claude` — tracked as #349.
  - `dispatch_005_respects_max_per_run` — parallel-load flake (passes in isolation), same class as #351.
  - `restore_001_no_flag_startup_restores_panes_from_snapshot` — **not** pre-existing on stock `main`; the verification ran against `4f188f9`, this branch's own PRD #336 commit, not clean `origin/main`. Cause: the PRD #311 change interacting with a fixture-name literal (`restored-beta` truncated to `restored-be…` in the collapsed title bar), already fixed on `main` by #334's fixture rename. #355 has been closed as invalid.
- Review: reviewer found no blockers (one test-fidelity suggestion, already fixed). Auditor found no security findings.


prageethw:
**Note on scope/sequencing:** this PR's diff against `main` currently includes all commits from #334 (PRD #311) and #342 (PRD #336) as well, not just PRD #333's own changes — e.g. `src/tab.rs`, `src/keybindings.rs`, `changelog.d/311.feature.md`, `changelog.d/336.feature.md`, and the #311/#336 test files all appear here too.

This isn't unintentional bundling — it's because this branch was built on top of my fork's `main`, which already has #311 and #336 merged locally, while `vfarcic/dot-agent-deck`'s `main` doesn't yet (still pending your review on #334 and #342). So this PR is effectively stacked on top of those two.

**Suggested merge order:** #334, then #342, then this one (#356) — once the first two land, this PR's diff should shrink down to just PRD #333's own changes (or I can rebase it at that point if useful). Flagging so the overlap doesn't cause confusion during review.

vfarcic:
Thanks for this — and thank you especially for flagging the stacking yourself and proposing a merge order. That note saved a lot of confusion, and it's exactly the right thing to do with a stacked branch.

A few other things worth calling out before the requests: PRD #333's core is genuinely nice work. The priority resolver is a clean pure function with the Compacting→Thinking / Unknown→Idle aliasing handled deliberately, the `Scenario:` comments are thorough, the PRD explicitly reasons through the rule 9 experimental-flag question instead of skipping it, and `orchestration/layout/003`'s spawn-order guard is a thoughtful test for a bug that would have been easy to miss. Docs follow the no-hard-wrap rule throughout.

I verified this locally by merging current `main` into your head commit and running the full gate suite including the e2e tier. Good news first: **most of what was red is now green.** `mode_001`, `restore_001` and `tabstrip_003` all pass on the merged tree, and `windows-cross` passes now too. The full run is 3409/3412, and of the three remaining reds two (`dispatch_013`, `chain_smoke_pi_002`) pass in isolation — parallel-load flakes, not yours.

Here's what I'd like changed.

---

## On this PR

### 1. Tab-label coloring — one change fixes two defects

`src/ui.rs`, `render_tab_strip`:

```rust
let style = match orchestration_statuses.get(i).copied().flatten() {
    Some(statuses) => style.fg(palette::status_color(&palette::highest_priority_status(statuses))),
    None => style,
};
```

**On the active tab**, this stacks an absolute foreground onto `Modifier::REVERSED`. Under reverse video the status color becomes the *background* and the label text is drawn in the terminal's background color. Measured on the merged build:

```
ACTIVE orch [idle] : fg=DarkGray bg=Reset mod=BOLD | REVERSED
ACTIVE orch [error]: fg=Red      bg=Reset mod=BOLD | REVERSED
ACTIVE plain       : fg=Reset    bg=Reset mod=BOLD | REVERSED   <- full contrast by construction
```

Open an orchestration tab with all roles idle — the resting state — and on a dark terminal you get near-black text on a dark-grey block, with the reverse-video active-tab cue gone. The comment three lines above the change says this is deliberate: the active tab "inverts the terminal's own fg/bg in place … needs no absolute color."

**For Idle**, `Color::DarkGray` is an absolute grey on read-critical text. Before this PR every tab label went through `text_primary()` = `Color::Reset`. PRD #13 removed exactly this pattern ("Gray/DarkGray labels … poor contrast on light backgrounds"), its milestone claims zero remaining hardcoded greys in `ui.rs`, and `palette.rs` calls greys "the exact light-background hazard PRD #13 exists to prevent." `STATUS_IDLE` was fine on card and pane *borders* — decoration — but PRD #13's rule reserves faintness for "purely-decorative, non-read elements only" and renders all actual text at full contrast. A tab label is text.

The two defects hit opposite audiences:

| | Light theme | Dark theme |
|---|---|---|
| Inactive Idle tab | DarkGray on white — poor | acceptable |
| Active tab | near-white on DarkGray — acceptable | near-black on DarkGray — unreadable |

Suggested fix for both: skip the status foreground when `i == active_index`, and let Idle fall through to the base style rather than painting DarkGray. That also reads better semantically — colour then means "something needs attention", and an idle tab looks like an ordinary tab.

### 2. Add an active-tab test case

Every case in `orchestration_009` passes `active_index = 0` with the orchestration tab at index 1, so the active orchestration tab is never rendered by any test — which is why the above slipped through. Worth noting no theme guard would have caught it either: `theme/contrast/001` covers five overlay surfaces and `theme/guard/003`'s lint covers the deck-card / embedded-pane / stats-bar paths, so the tab strip is unguarded.

### 3. Add `changelog.d/333.feature.md`

Fragments are here for 311 and 336 but not for this PR's own headline feature, so the tab coloring would be invisible in the next release notes.

### 4. Rebase onto current `main` — after #342 lands

#334 merged, so PRD #311 is on `main` now and the branches collide on test identity. When I did this locally it needed: dropping this PR's duplicates of PRD #311's tests in favour of main's `tabs/orchestration/006` and `orchestration/layout/002`; renumbering #336's tests around them; porting the `Ctrl+l` tab-scoping guard into `handle_key_event` (main extracted it out of the inline event loop); renaming `open_orchestration` (it now collides with main's focus-lifecycle helper); and re-adding `orchestration_statuses` to the restored test's `TabBarInfo`. Not mechanical, so worth doing deliberately rather than resolving conflicts in a hurry.

One thing the rebase also fixes on its own: the bundled #336 content here is an older snapshot of #342 and is missing #342's `help_key_line("Toggle orch tab split ratio", …)` plus its two help-overlay snapshots. I confirmed in the running app — `Ctrl+t` is listed under `?`, `Ctrl+l` isn't, even though `docs/keyboard-shortcuts.md` documents it. `test-fast` passes because the overlay genuinely doesn't render it, so the un-updated snapshots are self-consistent.

---

## On #342 (worth landing first)

- **`orchestration_008` fails deterministically.** The behaviour is fine — the grid shows `CTRLL> ^L`, so the byte does reach the PTY and your scoping guard works. What fails is the assertion's premise: bash echoed the control character literally instead of readline running `clear-screen`. Needs a more reliable assertion.
- **Same rebase collision** — #342 will hit it too, since `main` now owns `tabs/orchestration/006` and `orchestration/layout/002`.
- **`layout_002_ctrl_l_toggles_orchestration_split_narrow`** asserts only `action.is_some()`, so any action satisfies it. Worth tightening to `Some(Action::ToggleOrchestrationSplit)`.

---

## One correction on #355

`restore_001` was not failing on `main`. #355 records it as "reproduced on plain `origin/main` (`4f188f9`)", but `4f188f9` is this branch's own PRD #336 commit, so the check ran against a tree that already contained the PRD #311 change that causes it. Measured: merge-base passes 3/3, `483fe3d1` (the PRD #311 commit) fails, this head fails.

The underlying cause is interesting and worth knowing: the test greps the grid for the literal `restored-beta`, and that string only ever appeared in the collapsed 1-row pane title bar PRD #311 removes — the dashboard card truncates it to `restored-be…`. #334 hit the same thing during its own review and fixed it by renaming the fixture panes to `r-alpha` / `r-beta`, which is why it passes on `main` today.

So #355 should be closed, and the "pre-existing and unrelated" line in the PR description doesn't hold for that one. Easy mistake to make with a stacked branch — verifying against `4f188f9` looked reasonable.

---

Critical path is short: **#342 lands → rebase this → items 1–3.** After that this is ~60 lines of product code with clean tests behind it. Happy to re-run the full suite once it's rebased.

prageethw:
Thanks — the measurements made this easy to act on.

**Items 1–3 are in** (`42160ad`, `b5c4b49`).

**1.** Fixed as you suggested — `render_tab_strip` now tints only *inactive, non-idle* orchestration tabs:

```rust
let style = match orchestration_statuses.get(i).copied().flatten() {
    Some(statuses) if i != active_index => {
        let color = palette::status_color(&palette::highest_priority_status(statuses));
        if color == palette::STATUS_IDLE { style } else { style.fg(color) }
    }
    _ => style,
};
```

The active tab falls through untouched, so it keeps `REVERSED | BOLD` and renders identically to an active non-orchestration tab. The idle carve-out tests the *resolved colour* rather than the status variant, so it follows `status_color`'s `Unknown → Idle` aliasing automatically instead of duplicating that match in the renderer.

**2.** `orchestration_010` now renders an orchestration tab at the active index and asserts its `fg` and modifiers match an active Dashboard tab exactly. `orchestration_009`'s case (c) also had to change — it was asserting `Idle ⇒ DarkGray`, i.e. defect B encoded as an expectation. Both compare against a freshly rendered ordinary tab rather than hardcoding `Color::Reset`, so they survive a future change to what `text_primary()` resolves to.

**3.** `changelog.d/333.feature.md` added — and reworded, since the draft promised "a tab whose roles are all idle stays gray". The PRD and `docs/orchestration.md` are updated to match; the PRD's priority table no longer claims Idle paints DarkGray.

**Item 4 is not done.** #342 is still open, so the rebase hasn't happened. What has changed: this branch no longer carries #336 at all — both CI shadows are dropped, and the diff is now just PRD #333's seven files. So the rebase should be smaller than you mapped out: no #336 renumbering, no `Ctrl+l` scoping guard to port into `handle_key_event`, no stale help-line snapshot. The `open_orchestration` rename and re-adding `orchestration_statuses` to the restored test's `TabBarInfo` will still need handling.

**On the theme guard** — you noted nothing would have caught this: `theme/guard/003`'s lint covers the deck-card, embedded-pane and stats-bar paths, `theme/contrast/001` covers the overlay surfaces, and the tab strip is unguarded. I've confirmed that and haven't closed it yet. Happy to extend `theme/guard/003` to cover the tab-strip path, either here or as a follow-up — your preference.

**On #355 — you're right, and it's closed.** The repro ran against `4f188f9`, which is this branch's own #336 commit rather than clean `main`, so the tree already contained the PRD #311 change that causes it. I've also corrected the "pre-existing and unrelated" line in this PR's description.

**On #342** — the `orchestration_008` assertion premise, the same rebase collision, and `layout_002` asserting only `action.is_some()`: all noted, and that PR is what we're picking up next.

CI here is 10/10 green and Greptile is 5/5 with no findings.

prageethw:
Two open questions for you on this one — pulling them out of the thread so they're not buried.

**Question 1 — is item 4's rebase still needed?**

You asked for a rebase onto current `main` after #342 lands, and listed what it would involve: dropping this PR's duplicates of PRD #311's tests, renumbering #336's around them, porting the `Ctrl+l` scoping guard into `handle_key_event`, renaming `open_orchestration`, and re-adding `orchestration_statuses` to the restored test's `TabBarInfo`.

Most of that list was about the bundled #336 content, and this branch no longer carries any of it — the #311 and #336 CI shadows are both dropped, `main` was merged in, and the diff is now seven files, all PRD #333:

```
changelog.d/333.feature.md
docs/orchestration.md
prds/333-orchestration-tab-status-color.md
src/palette.rs
src/ui.rs
tests/CATALOG.md
tests/render_tab_strip.rs
```

It currently reports `MERGEABLE` / `CLEAN` against `main`, with CI 10/10 green and Greptile 5/5 with no inline findings.

So: do you still want a proper rebase, or is the merge that's there sufficient? Happy to rebase if you'd prefer the linear history — just don't want to rewrite the branch if the reason for it has gone away.

**Question 2 — the theme-guard gap: this PR or a follow-up?**

You noted that no existing guard would have caught this bug — `theme/guard/003`'s lint covers the deck-card, embedded-pane and stats-bar paths, `theme/contrast/001` covers the overlay surfaces, and the tab strip is unguarded. I've confirmed that: there is no coverage of the tab-strip path in either.

That gap is the reason the grey-on-tab-label defect needed a human eye rather than a failing test, so I'd like to close it. Extending `theme/guard/003`'s source lint to the tab-strip render path looks like the natural fix.

Your call on scope: fold it into this PR, or file it as a follow-up so this one stays a focused ~60 lines? I'll do either — just don't want to widen a PR you've already reviewed without asking.

prageethw:
Following up on my two questions from a couple of days ago — rather than leave this parked waiting on answers, I'm taking the conservative default on both so neither blocks you. Say the word if you'd prefer otherwise on either.

**Question 1 (the rebase) — I'll leave the merge as it is.** The rebase you asked for was scoped around the bundled #336 content, and this branch no longer carries any of it: both CI shadows are dropped and the diff is seven files, all PRD #333. It reports `MERGEABLE` / `CLEAN` with CI green. If you want linear history anyway it's a cheap rebase and I'll do it on request — I just don't want to rewrite a branch you've already reviewed for a reason that's gone away.

**Question 2 (the theme-guard gap) — I'll file it as a follow-up**, so this PR stays the focused ~60 lines you reviewed rather than growing after review. The gap is real and confirmed: neither `theme/guard/003`'s source lint nor `theme/contrast/001` covers the tab-strip render path, which is why the grey-on-tab-label defect needed a human eye instead of a failing test. Extending `theme/guard/003`'s lint to that path looks like the natural fix.

So there's **nothing outstanding from my side on this one** — it's ready to merge or to send back, whichever you prefer.

---

One piece of context that might affect how you sequence things, offered as information rather than a nudge: **#342 merging is what un-gates the most work on our side.** Our fork's PRD #387 (the unified deck-global `Ctrl+L` split toggle) has its upstream follow-up gated on #342 landing, and this PR's remaining question was gated on it too. #390 is also sitting green if it's useful to batch them.

No rush intended — I know you're weighing these against your own roadmap, and I'd rather they wait than be merged before you're comfortable with them.

Issue #42, opened 2026-04-04
Native Windows support


vfarcic:
## PRD: Native Windows Support

**Problem**: dot-agent-deck only runs on macOS and Linux. Users on Windows must use WSL, which adds friction and limits discoverability. The codebase has three Unix-specific dependencies: Unix domain sockets for daemon IPC, `/dev/tty` for OSC 52 clipboard writes, and Unix-specific PTY signal handling.

**Solution**: Replace Unix domain sockets with cross-platform IPC (named pipes on Windows, Unix sockets elsewhere), add `conout$` fallback for clipboard, and handle platform-specific PTY differences. Ship Windows binaries (amd64, arm64) in releases.

**Detailed PRD**: See [prds/42-native-windows-support.md](https://github.com/vfarcic/dot-agent-deck/blob/main/prds/42-native-windows-support.md)

**Priority**: Medium

vfarcic:
### 🚧 In progress — Foundation implemented, PR open, awaiting review

Work on the **Foundation phase** (platform abstraction + cross-platform IPC + `windows-latest` CI) is implemented and under review:

- **Branch:** `prd-42-native-windows-support`
- **PR:** #167 — CI fully green (build, **build-windows 640/0**, label, security), Greptile advisory "safe to merge". Held open intentionally; **not merging yet** pending maintainer review.

**Scope of this issue (#42):** the foundation only — `src/platform/` seam, `cfg`-gated `libc`, the IPC abstraction (Unix sockets ↔ Windows named pipes), and the Windows CI job. The Windows daemon is deliberately hard-failed (`Unsupported`) until the security descriptor is hardened.

**Remaining Windows work is tracked separately:**
- #163 — process/path/filesystem platform backends (incl. the release-gating pipe security descriptor).
- #164 — release `.exe`/Scoop, interactive e2e on a Windows VM, and docs.

PR #360, opened 2026-08-03
feat: add Devin CLI as a first-class agent

mreparaz:
Devin ships a Claude-Code-compatible hooks engine, so its native command hooks post the same stdin JSON shape Claude's do and ride the existing hook socket — no new wire format, no PROTOCOL_VERSION bump, no wrapper, no trust ceremony. This makes it the cheapest possible registry addition: a typed identity, a registry entry, one hook.rs ingestion arm, and a dedicated installer module.

The installer (src/devin_hooks_manage.rs) borrows Codex's safety discipline rather than Claude's, because Devin's config is a shared file holding the user's model, permissions, MCP servers, and theme, and Devin documents it as JSON with comment support. Only NotFound is treated as empty; malformed/JSONC content is backed up and the install errors rather than silently destroying the user's config. The read-modify-write is serialized by an in-process mutex and published atomically.

Devin reads Claude's hook files by default, so a user with both the deck's Claude and Devin hooks installed gets duplicate events (badge flaps between Devin and ClaudeCode). The installer detects this and warns with the one-line remedy (read_config_from: { claude: false }) rather than writing that key itself, since it also governs Claude rules/commands/subagent imports.

The new AgentType::Devin variant is forward-compatible: the existing deserializes "devin" to the neutral placeholder rather than failing.

Generated with [Devin](https://devin.ai)


vfarcic:
Thank you for this — and I'm sorry it sat for a couple of days before getting a proper review. That delay is on me, not on you.

I want to say up front that this is a **very well-researched contribution**. I installed the Devin CLI to verify it, and every factual claim in the PR held up: the config location, the exact set of eight hook events, `PostCompaction` rather than `PreCompact` (Devin genuinely has no pre-compaction event), `PermissionRequest` rather than `Notification`, the JSONC parsing hazard, and the `%APPDATA%` path on Windows. Several of those are easy to get wrong, and I could only confirm them by running the binary.

Rather than send you a change list, I've pushed commits on top — two of the issues needed the real CLI to find, so it seemed unfair to hand them back as homework.

**Fixes**

1. **`XDG_CONFIG_HOME` is honoured by Devin.** The reasoning in the code for skipping it was sound, but the actual behaviour is inverted: with the variable set, Devin reads `$XDG_CONFIG_HOME/devin/config.json` and never touches `~/.config/devin`. Affected users would have gotten a successful install and no events at all — exactly the silent no-op the comment set out to avoid. Verified end to end against the real CLI.

2. **Config permissions were being widened.** Greptile flagged this and it turned out to be real, not theoretical: a real Devin install ships `config.json` at `0600` and it holds `devin.org_id`, while the atomic publish republished it at `0644` (or `0664` under a `002` umask). Now preserves the destination's mode and creates owner-only.

3. **The Windows path is now a documented no-op**, mirroring `codex_home()`. Your path was right — it *is* documented — but the hook command written there is single-quoted for a POSIX shell, which Windows parses literally, and the deck has no native Windows support yet (#42).

4. **The duplicate-events detection is removed.** This one is only visible with the CLI in hand. With both hook sets installed and the Claude import left at its default, a real Devin session emits exactly one event per lifecycle step, all stamped `devin` — confirmed in print mode and in an interactive pane. The warning therefore fired on the deck's own default configuration and advised disabling Claude rules, skills, commands and MCP imports to fix a symptom that doesn't occur. Your analysis is preserved in the module docs and `agent-adapters.md` in case it ever does show up.

**Tests**

- `tests/devin_hook_ingestion.rs` — fast-tier coverage for the new `hook.rs` code (the `"devin"` arm, `PostCompaction`, and the `exec` tool-detail arm). Worth noting the generic fallback arm stamps `AgentType::ClaudeCode`, so losing the `"devin"` arm would mislabel every event rather than error.
- `tests/e2e_devin_hooks.rs` (`devin/live/001`) — a real-agent PTY test that runs Devin live in a pane and asserts the dashboard reacts to its own hooks. A few non-obvious things it needs: `--respect-workspace-trust false` (print mode can't show the trust prompt), no pinned `--model` (a free-tier account rejects every explicit model), and a wait for Devin's UI to finish booting before typing, since earlier keystrokes are silently dropped.

**Also**

- Restored an unrelated paragraph in `docs/orchestration.md` that had been removed.
- Merged `main` — the conflict was our merge cadence while this sat, not anything you did.

Follow-up #382 tracks the same permissions bug in `codex_hooks_manage`, which this code was originally copied from.

Thanks again for a genuinely high-quality first contribution.

vfarcic:
Merged — this will ship in the next release.

One thing worth being upfront about: I only set up a free-tier Devin account in order to review this PR, and I don't use Devin as part of my normal workflow. So while I verified the integration end to end — hooks installing into the real config, a live session driving the dashboard, the event lifecycle — I won't be dogfooding it the way I do the other agents day to day, which is how most rough edges here actually get found.

So if you hit anything — events not arriving, a status that doesn't settle, the config merge behaving unexpectedly, anything at all — please do open an issue. You're likely to be the first person to run this in earnest, and I'd genuinely rather hear about a problem than have it sit undiscovered.

Two areas I'd flag as most likely to surprise, since I could only test them in one configuration:

- **`XDG_CONFIG_HOME`.** Devin honours it, so the deck now does too. If you set it, hooks go to `$XDG_CONFIG_HOME/devin/config.json`.
- **Duplicate events.** I removed the conflict detection because I measured exactly one event per lifecycle step with both hook sets installed and the Claude import at its default. If you ever *do* see events reported twice, that finding was wrong and I'd want to know — the analysis is preserved in `devin_hooks_manage`'s module docs so it can be restored quickly.

Thanks again for the contribution.

Issue #164, opened 2026-06-14
Windows release artifacts, e2e validation & docs


vfarcic:
## PRD: Windows release artifacts, e2e validation & docs

**Phase 3 of native Windows support** — depends on the foundation in #42 and the platform backends in #163. See #42 for the canonical architecture and decisions.

**Problem**: Once #42 and #163 land, dot-agent-deck compiles and runs on Windows and `windows-latest` CI proves the code paths build and pass tests — but Windows isn't yet a *usable, supported* platform. There are no shipped `.exe` binaries (release.yml produces none; Scoop has no Windows URLs), no interactive end-to-end proof of the things CI can't assert (clipboard via `CONOUT$` in Windows Terminal, ConPTY resize repaint, daemon-survives-shell-exit, the full install→hook→dashboard→pane→agent workflow), and no Windows documentation.

**Solution**: Add `x86_64`/`aarch64-pc-windows-msvc` matrix rows to release.yml (`.exe` artifacts, portable-shell the build/package steps, Scoop Windows URLs). Run the full interactive workflow on a cloud Windows VM with Windows Terminal (aarch64 covered by CI build). Add a README platform-support section + Windows install instructions, documenting the v1 behavior differences (OSC 52 needs Windows Terminal, best-effort graceful shutdown, per-user isolation).

**Detailed PRD**: See [prds/164-windows-release-e2e-docs.md](https://github.com/vfarcic/dot-agent-deck/blob/main/prds/164-windows-release-e2e-docs.md)

**Priority**: Medium

vfarcic:
## Windows *desktop* unblocks via #741's transport, not via this issue — and they are separate products

Recorded 2026-09-03, because the intuitive dependency is the wrong one and PRD #740 Decision 1 already made this point in a PRD that is now closed and therefore easy to miss.

There are two distinct Windows stories and they have **different prerequisites**:

| story | what it needs |
| --- | --- |
| **Windows GUI driving a Linux daemon** (the shape actually wanted) | **#741** — a transport a Windows client can speak — plus **#754**. This issue is **not** needed at all. |
| **Windows GUI driving a local Windows daemon** | **This issue** (#164): a published `*-pc-windows-*` daemon binary. |

[...]

PR #375, opened 2026-08-04
fix(368): repair the Linux-native Windows cross-check

### LINKED ISSUE #368: scripts/windows-cross-check.sh has been broken since the reqwest 0.13 / aws-lc-sys migration (#269)
The Linux-native Windows type-check documented in [`docs/develop/windows-cross-check.md`](docs/develop/windows-cross-check.md) fails on every branch, including bare `main`. It dies inside a transitive dependency's build script, **before type-checking a single line of this repo's code** — so the pre-PR Windows gate that doc prescribes currently provides **zero** signal.

Found while verifying PR #342, where `scripts/windows-cross-check.sh` failed identically at the PR head and at its merge-base — i.e. it is not that PR's defect.

## Root cause

devbox exports `CC=gcc` and `AR=ar` globally, and `cc-rs` honours them even for an MSVC target. `scripts/windows-cross-check.sh` shims **`AR` only** — its comment names `ring`'s build script as the reason.

That sufficed while rustls used `ring`. PR #269 (`fix(deps): update rust crate reqwest to 0.13`, lockfile synced in `f8ee9c0`) moved rustls onto its `aws-lc-rs` default provider, which pulls `aws-lc-sys 0.43.0` — whose build script compiles ~600 C files. Those get compiled by Linux `gcc` with Linux system headers against a Windows target, so it aborts on:

```
error: unknown type name 'pthread_rwlock_t'; did you mean 'pthread_cond_t'?
  552 |   pthread_rwlock_t lock;
```

Dependency path: `reqwest 0.13.4` → `rustls` → `aws-lc-rs 1.17.3` → `aws-lc-sys 0.43.0`.

## Shims alone will not fix it

Verified: adding a `CC_x86_64_pc_windows_msvc` shim (emitting stand-in objects, on the same "`cargo check` never links" logic as the existing `AR` shim) clears the C-compile stage, but the archiver step then rejects those objects. Getting further requires a real cross toolchain, not another shim.

## Options

1. **`cargo-xwin`** — fetches the Windows SDK/CRT headers and drives `clang-cl`, so `aws-lc-sys` compiles for real. Rigorous, and keeps the check on CI's exact `x86_64-pc-windows-msvc` target. Costs a new tool dependency plus an SDK download.
2. **Retarget the check to `x86_64-pc-windows-gnu`** with mingw. Cheap, and *sufficient for the check's documented purpose*: it exists to catch fast-tier test files calling `#[cfg(unix)]`/`libc` helpers (the PRD #126 shape), `cfg(unix)` is false on both Windows targets, and `cargo check` never links — so the msvc-vs-gnu distinction is not exercised. Diverges from CI's target for anything genuinely ABI-specific. **Unverified** — mingw was not installed on the machine where this was diagnosed.
3. **Retire the script** and update `docs/develop/windows-cross-check.md` to say CI owns Windows, so nobody trusts a gate that cannot fire.

Any of the three beats the status quo. A permanently-red documented gate costs a fresh diagnosis on every PR review that runs it, and trains readers to ignore the failure.

## Notes

- **CI is unaffected.** `build-windows` compiles natively on `windows-latest` and never invokes this script — there is no reference to `windows-cross-check` anywhere in `.github/workflows/`. That is precisely why this rotted unnoticed.
- Reproduce with `scripts/windows-cross-check.sh` on any branch, or `bash .claude/skills/verify-pr/checks.sh --only windows-cross`.

vfarcic:
Fixes #368.

## The problem

`scripts/windows-cross-check.sh` — the pre-PR Windows gate `docs/develop/windows-cross-check.md` prescribes — had been failing on **every branch, bare `main` included**, since #269 moved reqwest to 0.13 and rustls onto its `aws-lc-rs` default provider. It died inside `aws-lc-sys 0.43.0`'s build script, which compiles ~600 C files: devbox exports `CC=gcc` globally and cc-rs honours it even for an MSVC target, so Linux gcc read Linux system headers against a Windows target and aborted on `unknown type name 'pthread_rwlock_t'` — before type-checking a single line of this repo. The documented gate provided **zero** signal, and nothing in CI runs the script, so it rotted unnoticed.

## The fix

Extend the shim the script already relied on. Its `AR` shim rests on "`cargo check` never links, so the archive only has to exist"; the same logic one step earlier means **an object only has to be a valid archive member**. `CC_x86_64_pc_windows_msvc` now hands back one prebuilt empty object per compile (one real `gcc` run, then ~600 `cp`s), skipping the C build entirely.

Both overrides are per-target, so a build script compiling for the **host** still gets the real toolchain. `aws-lc-sys` still emits its cfgs (`cargo:rustc-cfg=universal`) and its pre-generated bindings still type-check for real — nothing is bypassed on the Rust side.

**Why not the three options in the issue:**

| | Verdict |
|---|---|
| `cargo-xwin` | Rejected — new tool dependency plus a Windows SDK download. This needs neither. |
| retarget to `x86_64-pc-windows-gnu` | Rejected — still hits this same bug (Linux `gcc` at a Windows target) without a mingw cross-gcc, which isn't installed; and it diverges from CI's target. |
| retire the script | Rejected — the gate has real signal, demonstrated below. |

This keeps CI's exact `x86_64-pc-windows-msvc` target and adds no tooling.

## Verification

- **Cold cache: exit 0 in 15.2s. Warm: ~5s.** Reaches our own test targets (only the pre-existing `delegate_prompt_injection` Windows warnings the doc already notes).
- **It has real signal** — injecting the PRD #126 break shape (a fast-tier test calling `#[cfg(unix)]` `common::spawn_daemon_serve`) produced exactly `error[E0425]: cannot find function spawn_daemon_serve in module common`, exit 101. Probe reverted.
- The issue's repro command `bash .claude/skills/verify-pr/checks.sh --only windows-cross` → `[PASS] windows-cross (5s)`.
- Shim arg parsing probed across `-o X` / `-oX` / `-FoX` / `/FoX`, no-output probe runs, parent-dir creation, `ar` acceptance of the produced objects, and loud failure when the empty-object path is absent.
- `cargo fmt --check` clean · `cargo clippy -- -D warnings` clean · `cargo test-fast` **1501/1501 passed** · `bash -n` + `sh -n` on both generated shims · `ci.yml` parses.

## Two stale claims also corrected

Both are the same failure mode this issue is about — documentation describing a mechanism that isn't real:

1. The doc sold `--features e2e` as extra coverage. It is **guaranteed red**: no `tests/e2e_*.rs` carries a file-level `#![cfg(unix)]` (59 of 61 open with a bare `#![cfg(feature = "e2e")]`) while the L2 helpers they call are per-item `#[cfg(unix)]`. Now documented as not-a-gate, pointing at #164.
2. `ci.yml` claimed `build-windows` excludes the L2 suite via `#![cfg(unix)]`. It's actually excluded because that job omits `--features e2e`. Same outcome, wrong reason — and it's precisely why the local `--features e2e` run is red.

The doc also gains an "If it breaks again" section recording *why* this rotted silently (nothing in CI runs the script) and how to recognise the next occurrence.

## Scope / risk

Developer tooling only. No Rust changed; the `ci.yml` edit is comment-only. CI's `build-windows` compiles natively on `windows-latest`, never invoked this script, and was unaffected throughout — which is exactly why this went unnoticed. Changelog fragment is `368.misc.md`, following the `322.misc.md` dev-tooling precedent.

Rule 12 (cross-version contract): not applicable — no daemon, protocol, orchestration, or hook code is touched.


vfarcic:
## Follow-up: the root cause is now fixed too (`87d0147`)

The first commit repaired the script but left the reason it *rotted* in place — and on a second look, that was fixable after all.

Nothing in CI has ever run `scripts/windows-cross-check.sh`, which is why it stayed red on every branch for months with no red signal anywhere. `build-windows` compiles natively and never invokes it.

**This repo already settled this argument once.** The `cargo xtask linkage-check` step in the `build` job exists because linkage-check *"was failing on main and no gate noticed, because it only ever ran when somebody invoked it by hand"* — the identical failure mode. So: identical remedy.

### What was added

A parallel `windows-cross-check` job on `ubuntu-latest`. It is deliberately **not** a second Windows code gate — `build-windows` owns that and does it properly (real build + clippy + tests on a real Windows runner). This job answers one narrow question: *does the script itself still work?* A dependency bump can break it without breaking anything else, and today that breakage is paid for by whoever next runs `/verify-pr`.

### Cost

Close to zero. It runs in parallel and the critical path is `build-macos` (3m56s). It needs neither an MSVC toolchain (the script shims `cc`/`lib.exe`, and `cargo check` never links) nor devbox.

### Verified before pushing

- The script exits 0 in 15.5s with `CC`/`AR`/`CXX`/`LD` **entirely unset** — the shims *supply* a toolchain rather than merely overriding devbox's exports, so a bare runner is enough. This was the one fact that decided feasibility.
- The job's `run:` block was executed verbatim locally → exit 0.
- The job resolves the toolchain dir via `rustup show` rather than trusting the script's local-machine default, since CI's toolchain name is whatever `dtolnay/rust-toolchain` resolved `stable` to.
- `ci.yml` parses; `fmt`/`clippy` clean.

The doc section that said *"nothing in CI runs this script"* is updated in the same commit — this change makes that claim false, and leaving it would reproduce the stale-claim problem the rest of this PR exists to fix.

One thing for a maintainer to decide: whether `windows-cross-check` should be added to **branch protection** as a required check. It has a real pending state, so it is safe to require.

PR #378, opened 2026-08-04
ci(376): devbox-native task entrypoints and a Semaphore pipeline spike

### LINKED ISSUE #376: PRD: Devbox-native CI entrypoints and a Semaphore pipeline spike
## PRD: Devbox-native CI entrypoints and a Semaphore pipeline spike

**Problem**: The Rust toolchain is pinned in `devbox.json` but floats in CI (`rust-toolchain@stable`, `install-action@nextest`), so the local gate and CI compile with different versions. Separately, no `task` entrypoints exist for the CI steps, so nothing outside GitHub Actions — including a local agent — can run them.

**Solution**: Add `task` entrypoints wrapping the six existing CI commands, source the toolchain (plus `cargo-audit`) from `devbox.json`, and stand it up as a Semaphore pipeline for the three jobs Semaphore Cloud can host. Measure against the GHA baseline, then decide whether to retrofit GHA. `ci.yml` and `release.yml` are not touched.

**Detailed PRD**: See [prds/376-devbox-ci-entrypoints-semaphore-spike.md](https://github.com/vfarcic/dot-agent-deck/blob/main/prds/376-devbox-ci-entrypoints-semaphore-spike.md)

**Priority**: Medium

vfarcic:
## What it does

Fixes a latent version-skew bug: `devbox.json` pins `rustc/cargo/clippy/rustfmt@1.97.1` and `cargo-nextest@0.9.140`, while `ci.yml` uses `dtolnay/rust-toolchain@stable` and `taiki-e/install-action@nextest`, which float — so the local gate and CI compile with different toolchains and diverge silently. This PR adds `task` entrypoints wrapping each CI step, moves `cargo-audit` into `devbox.json`, and stands up a Semaphore pipeline spike to be measured against the GHA baseline.

## `ci.yml` and `release.yml` are deliberately untouched

`git diff main -- .github/` is empty. They are the working merge gate Renovate automerges against **and** the experimental control for the measurement — editing them here would invalidate the comparison. This is the PRD's single largest stated risk, and it is intentional, not an oversight. Please don't flag it as an omission.

## Milestone status — honest accounting

- **M1** (task entrypoints + `cargo-audit` from devbox) — done.
- **M2** (Darwin toolchain gate) — done, verdict **GO**. Every pinned package has a prebuilt `aarch64-darwin` binary at nixpkgs `a5cbcfe954791221bfffe2307f7d1a1bf61a871e`; verified by narinfo HTTP 200 on all roots, a 30-path closure walk with zero misses, and `nix path-info` against the remote store. The `gcloud` flake was checked separately: its base SDK substitutes, and its five built derivations are tar/symlink/cp/sed with no compiler.
- **M3, M4, M5 — blocked, not skipped.** They need a Semaphore Cloud project connected to `vfarcic/dot-agent-deck`. None exists, there is no `sem` CLI and no credentials. The pipeline *file* is delivered; "green on Semaphore" and the measurement numbers cannot be produced from here.
- **M6** — docs done; the retrofit-or-not decision is deliberately **not** made, because it depends on M5's numbers.

**The pipeline has never executed.** It is unvalidated by design — validated only to the extent that the YAML parses, every key was checked against Semaphore's documented v1.0 schema, and the cache logic was exercised against a purpose-built stand-in `cache` CLI.

## Security review happened and changed the design

An auditor and a reviewer went over it across two rounds. Three findings were security defects in the never-executed pipeline, all now fixed:

- A privileged `sudo tar -C / -xpf -` plus `nix-store --load-db` restore of an unverified project-cache archive — a root arbitrary-write primitive. Removed entirely; nix-store caching is gone and every job substitutes from the already-signed `cache.nixos.org`. (It had also never worked: Semaphore's `cache store` strips a leading `/`, so the restore probe never matched and every job cold-provisioned.)
- Unauthenticated cache producers — keys now carry a trust namespace, only a default-branch push writes, and restores adopt only allowlisted members after validation.
- Two `curl | sh` installers of floating releases — both pinned and sha256-verified before execution.
- A symlink escape in the restore adoption itself, found on re-audit and demonstrated against the pre-fix code, where an archive with a symlinked ancestor got an external directory adopted into the checkout.

**Known residual risk, documented not hidden:** the scratch-plus-allowlist mechanism bounds what an archive can get *adopted into place*, but it does not confine the provider's extractor, which runs first. If Semaphore's `cache restore` honours absolute or `..` members, those files land before any validation. That behaviour is unverified and is an explicit item on the activation checklist in `docs/develop/ci-entrypoints.md`.

## Follow-ups worth filing (not created here)

- `ci.yml:289` still says devbox pins `rustc 1.97.0`; it pins `1.97.1`. Left wrong on purpose because `ci.yml` is off limits in this PRD.
- The new `nix-installer` / `devbox` version pins in `.semaphore/ci.sh` are not Renovate-managed; a `customManagers` regex entry is the fix, deliberately not added because it changes a live automerge bot's config.
- Retrofitting `ci.yml` to use the `task` entrypoints — gated on M5's numbers, which is the whole point of the spike.

## Gate results

- `cargo fmt --check` clean, `cargo clippy -- -D warnings` clean.
- `cargo test-fast`: **1501/1501 passed, 0 skipped**.
- `cargo xtask linkage-check`: ok (441 catalog ids, 368 annotations, 100 allowlisted, 7 rules).
- `cargo test-e2e` (rule 5 pre-PR gate, run once with `DOT_AGENT_DECK_RECORD=1`): **3406 tests, 3405 passed, 1 failed.** The failure was `chain_smoke_pi_002_worker_receives_delegate_and_signals_work_done` — a real-agent test on `openai/gpt-5-nano` where the model did the work and created the sentinel correctly but then *offered* to run `dot-agent-deck work-done` instead of executing it. Classified as a pre-existing real-agent flake: this branch changes no Rust source, no `tests/`, no harness and no agent config. It **passed on isolated retry.**
- No `#[spec]` tests added or changed, so no catalog or Scenario-comment changes. No `tests/` file is touched, so there is no demo reel for this PRD.

Closes #376


vfarcic:
Closing this unmerged and deleting the branch. PRD #376 is abandoned for now; the issue stays open and the PRD file on `main` is unchanged from its original state.

**Why:** every hosted Linux job was refused by Semaphore before executing a single command — `start_time: 0`, no agent assigned, no logs, and `"Selected machine type is not available in this organization"`. Four runs, three configurations, with both halves of the `(machine, os_image)` selection varied independently and verified against the API:

| run | machine | os_image | Linux | macOS |
|---|---|---|---|---|
| 1, 2 | `f1-standard-4` | `ubuntu2404` | refused | passed |
| 3 | `e1-standard-4` | `ubuntu2404` | refused | passed |
| 4 | `f1-standard-4` | `ubuntu2204` | refused | passed |

No field explains it, and the asymmetry is unexplained: macOS at $0.09/min runs while Ubuntu x64 4 vCPU at $0.015/min is refused, which no published plan tier accounts for. Nothing in `semaphore.yml` or `ci.sh` was ever implicated — the blocker was account-level machine availability, answerable only through Semaphore's org settings or support.

**What was proven, and is worth knowing if this is revisited:**

- **macOS worked, 4/4, first attempt, with no macOS-specific fix.** The pinned Determinate installer, the APFS `/nix` volume, `devbox run` realising the closure on `aarch64-darwin`, 8 GB sufficing for build + clippy + nextest, and the cache trust check correctly declining to write from a non-default branch — all verified on a real agent. The PRD had darwin as the highest-risk leg; it was the only one that ever ran.
- **One M5 data point:** macOS 5m45s–5m57s cold with no cache, against GHA's 4m0s with `Swatinem/rust-cache`. Worst case, since a branch push stores nothing by design.
- **Three security defects were found and fixed in never-executed code** — a `sudo tar -C /` root arbitrary-write via unverified cache restore, unauthenticated cache producers, and two `curl | sh` installers — plus a symlink escape in the restore adoption, demonstrated against the pre-fix script.
- **Two real bugs found by auditing provider behaviour against source:** `SEMAPHORE_GIT_DEPTH=0` hard-fails a PR-type checkout (`git clone --depth 0` is rejected outright), and Semaphore's `cache` CLI preserves absolute paths via `tar -P` despite docs saying it strips them.
- **`ci.yml` and `release.yml` were never touched**, as the PRD required.

If this is picked up again, start fresh with a working org. The M1 work — `task` entrypoints wrapping each CI step, and `cargo-audit` pinned from `devbox.json` instead of compiled from source every run — is independently valuable and unrelated to Semaphore; it addresses the real latent bug, which is `devbox.json` pinning the toolchain while `ci.yml` floats on `dtolnay/rust-toolchain@stable`.

Two findings from this work were kept: PR #380 (merged) fixed a pi-worker e2e flake, and PRD #381 tracks a hook-install bug that writes transient `target/` build paths into global agent config.

Issue #370, opened 2026-08-04
Bug: when a shell being executed in claude code , the pane still displays it as idle, not technically accurate in orchestration model?


prageethw:
## PRD: Treat underlying shell activity as Working status inside a worker pane

**Problem**: Pane status (Idle/Working/Thinking/etc.) is driven entirely by agent-emitted signals — Claude Code hooks, OpenCode/Pi event surfaces, or wrapped-agent stdout-line pattern matching. There is no process-tree or PTY-foreground-process inspection anywhere. When a role's agent shells out to run something long-running (e.g. `cargo build`, a release script) and its own status events go quiet in between, the pane incorrectly shows Idle while a command is visibly still running.

**Solution**: Poll the pane's PTY for its foreground process group (tcgetpgrp/getpgid). If it differs from the shell's own pgid, a child command is actively running — feed that as a supplementary signal into the existing status pipeline so the pane reflects Working even without an agent-emitted event in that window.

**Detailed PRD**: See [prds/370-shell-activity-status.md](https://github.com/vfarcic/dot-agent-deck/blob/main/prds/370-shell-activity-status.md)

**Priority**: Medium

vfarcic:
Closing as implemented — verified against `origin/main` at `c701932`.

The behaviour asked for here shipped as **PRD #386** (the process-table primitive) carrying **PRD #370's** mechanism, merged via PR #500:

- The daemon runs a shell-activity monitor that synthesizes **`ShellBusy` / `ShellIdle`** events for a pane whose agent is quietly running a long shell command. `ShellBusy` is wired through `src/daemon.rs`, `src/daemon_protocol.rs`, `src/event.rs`, `src/state.rs` and `src/ui.rs`.
- `src/state.rs` carries the implementation under this issue's own number — e.g. the comment at `src/state.rs:5843`, "PRD #370 M2: any REAL event other than `ShellBusy` clears the …".
- It required an attach-protocol bump, recorded in `CHANGELOG.md`: **"Attach-protocol bump: shell-activity event types (`PROTOCOL_VERSION` 6 → 7)"**.

The reported symptom is covered end to end by a real-agent test, not just a mechanism test — `tests/CATALOG.md`, `status/shell-activity/006`:

> A real interactive Haiku Claude agent's Bash call that crosses Claude Code's 120s default timeout keeps the pane's rendered badge on `Working`, with the command genuinely still running — the reported bug, reproduced as the user actually sees it (PRD #386 M6b).

[...]

PR #388, opened 2026-08-06
fix(prd-387): unified deck-global split toggle and Ctrl+L chord fix

### LINKED ISSUE #387: Ctrl+L is swallowed on orchestration tabs; propose a unified deck-global, command-mode-scoped split toggle
This is two things: a **bug** that is live on `main` today, and a **proposal** that generalises the fix already sitting in #342. They are separated below because the bug stands on its own and deserves fixing regardless of what happens to the proposal.

## Part 1 — the bug: `Ctrl+L` is swallowed on orchestration tabs and never reaches the agent

`Ctrl+L` is claimed by the deck's global keybinding resolver and, on an **orchestration tab**, is claimed *regardless of mode*. So a focused role pane's PTY never receives `0x0c`, and **Claude Code's own `Ctrl+L` (clear screen) never reaches the agent**. This is a daily annoyance for anyone running an interactive agent in a role pane — which is the deck's primary use case.

Evidence, from the fork's `main`:

```
src/keybindings.rs   ToggleOrchestrationSplit → section: Section::Global, default "Ctrl+l"
src/ui.rs:6545       matched in the global resolver → Some(Action::CycleSplitStage)
src/ui.rs            scope_orchestration_split → NOT PRESENT
```

The claim decision lives inline at `src/ui.rs:9077-9086`:

```rust
if matches!(action, Some(Action::CycleSplitStage)) {
    let claims_ctrl_l = match tab_manager.active_tab() {
        Tab::Orchestration { .. } => true,                  // <-- mode-independent: THE BUG
        Tab::Dashboard { .. } => ui.mode == UiMode::Normal,  // <-- already command-mode scoped
        Tab::Mode { .. } => false,
    };
    if !claims_ctrl_l {
        action = None;
    }
}
```

Worth stating precisely, because it is easy to get backwards: **Dashboard tabs are already guarded** — PRD #371 scoped them to `UiMode::Normal`, so a focused Dashboard pane's `Ctrl+L` does forward to its PTY, and there is a passing e2e test pinning exactly that (`tests/e2e_orchestration_pane_column.rs::orchestration_007_ctrl_l_forwards_to_pty_on_non_orchestration_tab`, which drives a real interactive bash pane and asserts readline's clear-screen actually runs). The unguarded case is **orchestration tabs**, and it has no equivalent test.

That is also precisely where it hurts most, and #342 already says so — its `scope_orchestration_split` doc comment calls a role pane "the most likely place to want a screen clear".

**#342 already fixes this half.** `scope_orchestration_split(action, is_orchestration_tab, mode)` un-resolves the chord outside command mode, mirroring how `close_pane` was scoped in PRD #241 M1 (so `Ctrl+W` still reaches the PTY as word-delete). It is not on `main` — #342 is unmerged, and the three-stage work in PRD #371 was developed independently of it. Nothing anywhere covers the general case.

## Part 2 — the proposal: one deck-wide split stage, command-mode-scoped on every tab type

Offered as a **follow-up to #342, explicitly not a re-scope of it.** #342 should merge as-is: it is the orchestration-only global toggle already reviewed twice. This would then go upstream as a second PR.

1. **Command-mode only, everywhere.** The deck claims `Ctrl+L` only in command mode, on every tab type; in pane input it passes through to the agent untouched. Generalise `scope_orchestration_split` rather than inventing a parallel mechanism. Consistent with `close_pane` (PRD #241 M1). Note that once the rule is uniform, the tab-type dimension mostly collapses — the predicate becomes "command mode, on a tab that has a sidebar split" — so this is a simplification of the current inline `match`, not another special case.

2. **One split stage across the entire deck.** Dashboard and every orchestration tab share a single value; toggling anywhere changes everywhere, and a newly opened tab of any kind adopts the current stage. The reasoning: sidebar width reflects how someone reads, not which tab they happen to have opened, and per-tab state means anyone who prefers a narrow sidebar re-toggles forever. There is a precedent for the shape — `pane_layout` is already a deck-global field on `UiState` with a global chord (`Ctrl+T`).

3. **Three stages** — Default (34/66) → Narrow (25/75) → Hidden (sidebar collapsed) → Default. Note the shared value is the *stage*, not the percentages: `Default` resolves to each tab type's own ratio (Dashboard 33/67, Orchestration 34/66), which is already how `split_stage_percents` works. Upstreaming two stages now and adding the third later would be two review cycles for one feature.

4. **Upstream path.** #342 first, then this as a follow-up PR.

## Prior art, and being straight about it

This fork has been running the three-stage Dashboard+Orchestration toggle in daily use — that is PRD #371, whose issue **was closed upstream as not-planned**. The fork continued it independently. This proposal folds that experience back upstream rather than relitigating that decision: the part worth carrying over is not "three stages are essential" but the two things daily use actually surfaced — that per-tab state is the wrong granularity, and that a globally-claimed `Ctrl+L` eats a chord agents genuinely want.

If the answer is still no on the toggle itself, **Part 1 is independent and still worth fixing** — the swallowed `Ctrl+L` is a bug in the current binding no matter how many stages it ends up having.

prageethw:
## Summary

Two problems, one chord:

1. **Bug fix**: `Ctrl+L` was claimed by orchestration tabs regardless of UI mode, so a focused role pane's agent never received the keystroke. Claude Code's own clear-screen never reached the agent — the deck's primary use case. A pure `scope_split_stage(action, has_split_sidebar, mode)` now claims the chord only in command mode, only on tab types with a sidebar split.

2. **Unification**: Per-tab `split_stage` fields on `Tab::Dashboard` and `Tab::Orchestration` are replaced by one deck-global `split_stage` on `UiState`. Toggling anywhere changes everywhere; a newly opened tab adopts the current stage.

## What Changed

- Replace per-tab `split_stage` with deck-global `split_stage` on `UiState`
- Introduce `scope_split_stage()` to gate `Ctrl+L` to command mode on tabs with sidebar splits
- Dashboard, Orchestration, and Mode tabs now follow uniform scoping rules
- No on-disk migration needed: `split_stage` is not persisted

## User-Visible Changes

- `Ctrl+L` now reaches agents in role panes (forwarded as `^L`, byte `0x0c`)
- **Tradeoff**: toggling split from an Orchestration tab while focused requires `Ctrl+D` first (same as `Ctrl+W` already does)
- Split stage is now deck-global: toggling anywhere changes all tabs; per-tab stages removed

## Cross-Version Compatibility

**The cross-version check was run and is PARTIAL.** Hooks demonstrably routed across the version boundary (work-done and status events). The **delegate leg is UNVERIFIED — not failed**: the test fixture's worker role runs `bash`, which the deck does not recognize as an agent, so there was nothing to inject a task into. No evidence of breakage; the branch touches no daemon/protocol/hook code. Per PRD #387 rule 12, a skipped or partial check is stated in the PR rather than left ambiguous.

**Baseline note**: The cross-version check ran against the newest pre-#387 build (`66f2c8d`, protocol 7), not the literal previous release (v0.35.7, protocol 6) — that protocol bump is PRD #370's and landed unreleased on `main`, so a v0.35.7 daemon would be refused at the handshake for a reason predating this branch. The build-version handshake *silently restarts* a mismatched daemon when no agents are running, so the check only tests a real boundary if an orchestration is left detached under the old daemon.

**Protocol decision**: No `PROTOCOL_VERSION` bump and no `.breaking.md` fragment, deliberately. Rule 12 scopes "breaking" to a TUI↔daemon interoperability break. This is client-side key resolution, `UiState`, `Tab`, and layout — no wire field, no `EventType`, no handler. Neither `src/daemon.rs` nor `src/hook.rs` is touched. Filing a user-visible behavior change as `.breaking.md` would wrongly signal a compatibility break.

## Implementation Notes

- `d1391d8` (`docs/develop/fork-sync-workflow.md`) is intentionally on this branch. The PRD's Open Question 1 recommended correcting the fork-sync stack table "now, as a standalone docs commit"; that is what `d1391d8` is.
- **No experimental flag**: This adds no new user-visible surface, and gating a key-dispatch bug fix would ship a build that still swallows `Ctrl+L` by default.
- **Semgrep audit**: Semgrep could not run during the audit (CLI was denied writing `~/.semgrep/semgrep.log` outside the workspace), so the audit is static diff review plus `git diff --check`.
- **Demo reel**: Skipped. The branch changed e2e tests, but no local full-tier e2e run was performed under exception (b), so `DOT_AGENT_DECK_RECORD=1` never wrote `.cast` files locally. CI's runner has them, not us. There is nothing local to stitch.

## Testing

All progression tests pass:
- `layout_001/002/003/005/006`
- `orchestration_role_pane_dims_*` (6 tests)
- `orchestration_006/007/024`
- `dashboard_001`

No full-tier local run performed (regression testing is CI's job on this fork, per CLAUDE.md rule 5).

## Linting & Checks

- ✅ `cargo fmt --check`
- ✅ `cargo clippy -- -D warnings`
- ✅ `cargo clippy --all-targets --features e2e -- -D warnings`
- ✅ `cargo xtask linkage-check` (476 catalog ids, 403 annotations, 100 allowlisted, 7 rules)

Closes #387


prageethw:
This PR was opened against the wrong base repository by mistake. PRD #387's M1–M6 implementation targets the fork's `main` branch (`prageethw/dot-agent-deck`), not the upstream. The cross-repository diff includes the entire fork divergence (68 files vs the intended 11).

Per PRD #387 decision 4 / M7, the upstream path is a separate future PR that will be authored *after* PR #342 merges and will delete `split_narrow` in the same PR. That upstream PR is gated on #342 and is not part of this implementation cycle.

This PR is being closed in favor of a corrected PR opened against `prageethw/dot-agent-deck#main`. The branch will not be deleted, as it will become the head of the corrected PR.

PR #390, opened 2026-08-06
fix(shell-activity): a pane reads Idle while its agent's shell command is still running (#386)

### LINKED ISSUE #386: Pane reads Idle while a long shell command is still running (Claude backgrounds it at the 120s Bash cap); PRD #370's backstop never fires
## What a user sees

A role pane whose agent is running a long shell command — e.g. a worker that typed `cargo test-e2e` — shows **`Working` for the first ~2 minutes and then `Idle` for the rest of the run**, while the command is demonstrably still executing. On a ~700-second run that is ~9.7 minutes of a pane advertising `Idle` while it is busy.

This is not a race and not a dropped hook. Every component in the chain does exactly what it was written to do; the *assumption* is what is wrong.

## Measured cause

Claude Code's Bash tool has a **120-second default timeout**. A command that exceeds it is **not killed — it is moved to a background shell**. `PostToolUse` fires at the 120s mark, the agent then ends its turn, and the `Stop` hook fires. On the deck side:

- `src/hook.rs:105` — `"Stop" => Some(EventType::Idle)`
- `src/state.rs:3614-3617` — `EventType::Idle` sets `SessionStatus::Idle`

So the deck is told, correctly by Claude Code's own model, that the *agent* is idle — because it is. The *work* it launched is not, and a background shell emits no hooks at all.

### Event stream (reproduction, daemon log)

Rig: a real logged-in interactive `claude --model claude-haiku-4-5-20251001 --allowedTools Bash` on a real PTY, with `DOT_AGENT_DECK_SOCKET` pointed at a sandbox daemon started with `DOT_AGENT_DECK_LOG=<file>`, `DOT_AGENT_DECK_PANE_ID=sbx4`. Prompt: run `ping -c 200 127.0.0.1 > /dev/null; echo PINGDONE` — 200 s of real work, **default** Bash settings (no `run_in_background`, no `timeout` parameter). Exactly the shape of a worker typing a long `cargo` command.

```
01:20:46.199  SessionStart   pane_id=Some("sbx4")
01:20:48.837  Thinking       pane_id=Some("sbx4")
01:20:53.307  ToolStart      pane_id=Some("sbx4")   <- Bash starts -> Working
01:22:53.695  ToolEnd        pane_id=Some("sbx4")   <- +120.4 s (the cap)
01:22:58.574  ToolStart      pane_id=Some("sbx4")   <- 80 ms blip (Monitor on the output file)
01:22:58.654  ToolEnd        pane_id=Some("sbx4")
01:23:00.372  Idle           pane_id=Some("sbx4")   <- Stop -> SessionStatus::Idle
```

Claude Code's own transcript for that tool call, verbatim:

> `"Command did not complete within its 120s timeout and was moved to the background (ID: bye2ifewm). Output is being written to: … You will be notified when it completes."`

The next tool call is a `Monitor` on that output file — itself a background watcher — so the agent ends its turn and waits. The work really does keep running: `ps` still showed `ping -c 200 127.0.0.1` alive at 02:23 elapsed, long after the pane had gone `Idle`.

### Control run that isolates the 120 s cap as the sole cause

Identical 200 s `ping`, but with `timeout: 300000` passed to the Bash tool so it never hits the default cap:

```
01:20:39.400  SessionStart   pane_id=Some("sbx3")
01:20:42.089  Thinking       pane_id=Some("sbx3")
01:20:47.749  ToolStart      pane_id=Some("sbx3")   <- Working
        …    (200.4 s of complete silence — no Notification, nothing)
01:24:08.128  ToolEnd        pane_id=Some("sbx3")
01:24:10.296  Idle           pane_id=Some("sbx3")
01:24:17.811  SubagentStop   pane_id=Some("sbx3")
```

`Working` holds for the full 200 s. A long tool call emits `PreToolUse` once and nothing after, but **nothing clears `Working` either** — so "long call" is not the trigger. The 120 s cap is the entire difference between the two runs.

A short call is also fine end to end: a 5 s `sleep` produced the healthy `SessionStart → Thinking → ToolStart → ToolEnd → Idle`, and `tests/e2e_card_layout.rs` (`dashboard/card-stats/005`) already asserts a real 4-second Bash call paints a live `Working` badge. The basic wiring is not in question.

### Not a regression

`git log -S` shows both halves of the mapping landed in the first commit and have never changed:

```
8a17388  feat(prd-1): Agent Status Dashboard …
   "Stop" => Some(EventType::Idle)      (src/hook.rs)
   EventType::ToolStart => …Working     (src/state.rs)
```

What changed is on Claude Code's side: the 120 s Bash timeout that *moves* a command to a background shell instead of killing it. Before that behaviour existed, the same command would have finished or errored inside one tool call.

### Which agents are affected

The **deck-side** assumption is agent-agnostic — every adapter maps "the agent's turn ended" onto `SessionStatus::Idle`: `Stop` for Claude and Codex (`src/hook.rs:105`, shared arm), `session.idle` / `session.status=idle` for OpenCode (`src/hook.rs:348,354`), `--type finished` for Pi. Any of them will read `Idle` while work they launched continues.

The **trigger** measured here is Claude Code specific. Only Claude was measured; Codex ships a Claude-compatible hooks engine and its own shell timeout, so the same shape is likely there, but that is inference.

## The intended backstop does not fire

PRD #370 (issue #370) added a shell-activity signal precisely so a pane would not sit at a stale `Idle` while a foreground command ran. Measured on a live deck, **it never fires in any real pane configuration**:

- **Agent panes.** Claude's Bash tool spawns its child on **pipes, in a new session**, off the pane's PTY. Measured with `ps -eo pid,ppid,pgid,tty,stat,command` during a live run:

  ```
    PID  PPID  PGID TTY  STAT COMMAND
   7074  6444  7074 ??   Ss   /bin/zsh -c source …/.claude/shell-snapshots/… && eval 'cd … && cargo test-fast …'
   7076  7074  7074 ??   S    cargo-nextest nextest run …
  ```

  TTY is `??`. Meanwhile the pane's own PTY child and every process on that tty share one pgid (`6387`), so `tcgetpgrp(pane_pty)` returns `6387` unchanged for the entire life of the command. `RunningAgent::shell_foreground_busy` (`src/agent_pty.rs`) computes `fg_pgid != shell_pid` = `6387 != 6387` → **`Some(false)`, permanently**. No `ShellBusy` is ever emitted for an agent pane. The same shape held on every role pane measured (5 panes, 5 tty's).

- **Bare shell panes.** The pgid mechanism genuinely works there (an interactive `sh` does `tcsetpgrp` a child into a new foreground pgroup), but the pane is dropped earlier, by the `pane_hook_session_id` gate in `run_shell_activity_monitor` (`src/daemon.rs:857-863`). That map is populated only from real agent hook events (`src/hook.rs`), so a pane that never emitted one is skipped before the signal is consulted.

- **Why it shipped green.** Its one end-to-end test (`shell_activity_monitor_reflects_a_real_foreground_command`, `src/daemon.rs`) spawns `/bin/sh`, types `sleep 2` **directly into the pane's PTY**, and hand-seeds a synthetic `SessionStart` carrying `pane_id: "pane-370"` purely so the gate resolves. No real pane performs either step. PRD #370's M5 and M6 are both unchecked; the PTY-attached L2 test was attempted and reverted.

So the two failures are independent: the `Stop → Idle` mapping above is a real bug on its own, and the mechanism meant to backstop it does not run.

## Proposed fix (brief — not part of the bug report)

Replace `tcgetpgrp` with a **descendant-process scan**: does the pane's PTY child have a descendant that is a session leader on pipes running the agent's bash-tool shell (`/bin/zsh -c source …/.claude/shell-snapshots/… && eval '<command>'`)? Measured discriminator: the long-lived children every Claude pane carries — `npm exec @upstash/context7-mcp`, `engram mcp`, `caffeinate` — do **not** match that shape, which is what the #370 notes feared was missing. Foreground and background Bash calls have the identical process shape, so one signal covers both.

This is attractive because `apply_event`'s existing `ShellBusy` arm (`src/state.rs:3628-3638`) already promotes exactly `Idle`/`Unknown` → `Working`, which is precisely the state this bug leaves the pane in — so a signal that actually fires repairs it **with no hook-semantics change at all**, and covers all agents plus bare shell panes (the `pane_hook_session_id` gate would need dropping for the latter).

The narrower alternative is to teach the deck about outstanding background shells from structured hook fields (`tool_input.run_in_background`, and the `tool_response` "moved to the background" text). Cheaper, but Claude-specific, partly English-text matching against another product's message, and it has no reliable clearing edge — Claude notifies the *agent* on completion, which produces no hook.

Either way, process-shape or message-text matching is coupled to another product's internals and can rot; that tradeoff belongs in the fix's own design discussion.

## Environment

- macOS (Darwin 25.5.0), `dot-agent-deck` at `main`.
- Claude Code with the deck's standard hook block installed in `~/.claude/settings.json` (`PreToolUse` with `matcher=None`, so no tool filtering).

## Note for anyone reproducing this

`daemon serve` writes no log at all unless **`DOT_AGENT_DECK_LOG=<file>`** is set (`src/main.rs`). With it set, `src/daemon.rs`'s `Received event` line gives one line per event and answers this in minutes. A deck started in the foreground without it keeps no event record.

prageethw:
Closes #386

## The bug

A pane reading `Idle` while its agent's shell command was still running confused me hell out of it, and I had to click all panes to figure out what was going on since everything seemed idle. Claude Code's Bash tool caps a command at 120 s and then **backgrounds** it rather than killing it, ending the turn — so `Stop` fires and maps to `Idle` while the command runs on. For a ~700 s command, the pane is right for 2 minutes and wrong for the remaining ~10.

## The fix

A descendant-process scan. Claude `setsid`-detaches its Bash-tool child into a POSIX session of its own, while MCP servers and `caffeinate` stay in the agent's — so `getsid(descendant) != getsid(agent_pid)` separates a real shell command from the long-lived children every agent pane carries, with one libc call. The daemon polls that and synthesizes `ShellBusy`/`ShellIdle`. **Hook semantics are untouched — nothing about `Stop -> Idle` changes.** The synthesized event promotes only `Idle`/`Unknown`, so a real `WaitingForInput`/`Error`/`Thinking` is never overridden.

## Why the PR is bigger than the bug

Upstream has none of #370's mechanism, so this carries it too, plus the `PROTOCOL_VERSION` 6 -> 7 bump owed by the new wire-serialized `ShellBusy`/`ShellIdle` variants. It supersedes #370's `tcgetpgrp` primitive, which cannot work for an agent pane; #370's fate is yours to decide.

## Tests

`status/shell-activity/001`–`007`. `001`–`004` are fast tier: descendant walk, cycle termination, the session-id discriminator, the busy flip. `005`–`007` are real-agent PTY tests — a rot canary on the process shape, an end-to-end reproduction of the bug, and a false-positive guard.

## Known limitations

- Only Claude Code's process shape has been measured; no other agent.
- The signal is coupled to that shape and fails silently in both directions if it changes.
- Windows returns `None`.
- `005`–`007` self-skip without credentials, so a green CI run says nothing about this signal.
- The `ps` call runs synchronously without a timeout on a Tokio worker, so a wedged `ps` stalls hook ingestion and shutdown, not just this signal. Disclosed rather than fixed here.

Full record — measurements, process tables, rejected alternatives, the cross-version note — in `prds/386-descendant-scan-shell-activity-signal.md`, which ships in this PR.


src/platform/proc/unix.rs
```diff
@@ -181,6 +181,69 @@ pub fn send_sigterm_to_child_group(
...
+/// implementations, and is only worth taking if the measurement says so.
+pub fn process_table() -> Option<Vec<super::ProcessInfo>> {
+    // `-w -w` disables `ps`'s width truncation on both macOS and Linux, so the
+    // argv column survives whole; the trailing `=` on every `-o` field
+    // suppresses the header line entirely.
+    let output = std::process::Command::new("ps")
+        .args(["-A", "-w", "-w", "-o", "pid=,ppid=,tty=,args="])
+        .stdin(std::process::Stdio::null())
+        .stderr(std::process::Stdio::null())
+        .output()
```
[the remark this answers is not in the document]
prageethw: Valid, and thank you — this sharpens a limitation the PR body already lists ("`ps` runs without a timeout") into something more precise than I had it.

The finding is right on both counts, and the second is the more serious one: `process_table()` runs a synchronous `Command::output()` inside `AgentPtyRegistry::shell_foreground_busy_snapshot`, which the monitor calls directly from an async task. So it is not merely untimed — it occupies a Tokio **worker** thread for the duration, and a `ps` stuck in D-state (NFS, a wedged filesystem) would stall hook ingestion, client requests and shutdown, not just this signal. The 500 ms cadence makes it a recurring exposure rather than a one-off.

I have deliberately **not** fixed it in this PR, and the reason is process rather than disagreement. This branch is a curated port of code that shipped and was proved against a real agent downstream; the same defect is tracked there, and fixing it only here would leave the two trees running different code with the fix unproved on either. The two halves also want different treatment — the worker-blocking half is a small `spawn_blocking` boundary around the snapshot call, while a genuine timeout needs an async `Command` with a `tokio::time::timeout` and a decision about what a timed-out sample means (almost certainly `None` / "no opinion", never `Some(false)`).

If you would rather this land fixed than land disclosed, say so and I will add it here — it is a small change and I would rather match your preference than my tracking convenience.

I have updated the limitations section of the PR description to name the worker-blocking dimension explicitly, since "no timeout" undersold it.

vfarcic:
Verified this locally: fresh worktree at your head with `origin/main` merged, full gate run plus the e2e tier.

## First, the good news

The feature is genuinely proven. My e2e run had 12 failures, but **all 12 were resource contention** — nine unrelated tests that each timed out at 180s finished together in 168s when rerun serially. Yours pass in isolation: `004` 8/8, `005` and `006` both green, `007` green even in the loaded run.

`006`'s isolated log is the real proof:

```
ToolStart observed at 02:55:22.975
native Idle observed at 02:57:27.027, 124052ms after ToolStart — this run genuinely exercises the bug path
command confirmed still running at the sample point (shell pid 2575003, a live `ping` descendant found beneath it)
```

That's the reported bug reproduced — Claude backgrounding at the 120s cap, `Stop` → `Idle` — with the badge holding `Working` anyway, verified as a user actually sees it. `005` showed `ShellBusy` synthesized 596ms after `ToolStart`. The `PRECONDITION NOT MET` guard also did its job: in the loaded run it failed loudly on an inconclusive run rather than passing silently. That's better discipline than most tests in this repo.

## Answering your open question on Greptile's P2

Land it disclosed — your reasoning about keeping the two trees running identical code is right, and I'd rather not take an unproved fix here either. I've filed #429 to track it upstream, since your fork's tracking doesn't travel with the port. When you do fix it, please put the "a timed-out sample means `None`, never `Some(false)`" note at the call site as Greptile suggested; it isn't recoverable from the code alone.

## Cross-version check — done, against v0.35.8

The one you explicitly left to me. Clean: the branch TUI met a v0.35.8 daemon in an isolated sandbox and the old daemon took a graceful SIGTERM with the matching one respawning — no silent miscommunication. Local path guards on `build_version`, ssh path on `PROTOCOL_VERSION` via `probe_remote_protocol`.

I also confirmed v0.35.8's `EventType` has 12 variants and **no `#[serde(other)]`**, so a `KIND_EVENT` frame carrying `shell_busy` fails its whole-frame decode at a released reader. The 6 → 7 bump is genuinely owed. Good call carrying it.

## Two things before merge

### 1. Needs `changelog.d/386.breaking.md`

`analyze.sh` decides the bump purely from fragment suffixes, so with only `370.feature.md` this cuts as a **patch** (0.35.9) when a protocol break must be **minor** (0.36.0). The minor digit is our signal that an older peer can no longer safely talk to this build — that's the whole point of the 0.x recalibration in `docs/develop/versioning.md`.

#201 and #140 both shipped `.breaking.md` **and** `.feature.md`; `has_breaking` wins in `analyze.sh`, so keep both here too.

I think this slipped for a traceable reason rather than an oversight: the "Bump policy: patch" conclusion was reasoned in the fork context where 6 → 7 had already landed via #370, and the upstream amendment correctly updated the `PROTOCOL_VERSION` half but never revisited the bump-policy half derived from the same premise.

### 2. M5's overhead number — I measured it, and it should land in this PR

I ran what M5 would have. Daemon with **zero panes**, 20s, reaped-child CPU (`/proc/<pid>/stat` fields 16+17), against a v0.35.8 control in the same sandbox:

| Build | Reaped-child CPU / 20s |
|---|---|
| This branch | **181 ticks — 1.81s, ~9% of one core** |
| v0.35.8 control | **0** |

That's a floor, not a ceiling: `cutime`/`cstime` count only children, so the ~714 `getsid()` calls and the line parsing — which run in the daemon itself — aren't included.

It also isn't an idle-only cost. The table is sampled once per tick regardless of pane count, so zero panes costs what ten panes costs, and the daemon polls for as long as a client is attached (the 30s idle-shutdown only covers the no-client case). 714 processes on my box, so a lighter machine pays proportionally less — but a busier one pays more.

**The useful part: ~75% of it is argv rendering.**

| `ps` invocation | CPU |
|---|---|
| `-o pid=,ppid=,tty=,args=` (what ships) | 0.04s |
| `-o pid=,ppid=,sess=` (no argv) | 0.01s |

Since the discriminator only reads argv for descendants that already failed the session test — a handful, not 714 — deferring argv and `getsid` to candidates should get ~8% → ~2% while staying on `ps`, with no new platform code and no new `unsafe`. I'd rather do that than Route B (native `/proc` + `sysctl(KERN_PROC_ALL)`) for now: Route B means two hand-written platform implementations where `ps` currently gives us one, and it loses `parse_ps_table` as a single fixture-testable surface.

Whether the argv optimisation lands here or as a follow-up is your call — but I'd like **the number itself on the record in this PR**, in three places:

- **`docs/develop/shell-activity-signal.md`** — a new `## What the signal costs` section. That page is already the durable record for this mechanism's non-obvious properties, and cost is the one of that character it's missing. Draft below.
- **The `POLL_INTERVAL` comment in `src/daemon.rs`** — it currently says "PRD #386 M5 is where that cost gets measured and the cadence confirmed or revised," which is now stale. Carry the number and point at the doc, so whoever considers 500ms → 2s finds it without hunting.
- **A Work Log line in `prds/386-…md`** — the PRD records M5 as deferred and names itself the durable record for this work's numbers.

<details>
<summary>Draft for the doc section</summary>

```markdown
## What the signal costs

Measured 2026-08-08 on Linux/x86_64, 714 processes, daemon with **zero panes**, over 20 s,
comparing reaped-child CPU (`/proc/<pid>/stat` fields 16+17) against a v0.35.8 control in the
same sandbox:

| Build | Reaped-child CPU / 20 s |
|---|---|
| With the monitor | 1.81 s (~9% of one core) |
| v0.35.8 control | 0 |

This is a **floor**: `cutime`/`cstime` count only children, so the ~714 `getsid()` calls and the
line parsing — which run in the daemon itself — are excluded.

It is not an idle-only cost. The table is sampled once per tick regardless of pane count, so zero
panes costs what ten panes costs, and the daemon polls for as long as a client is attached (the
30 s idle-shutdown only covers the no-client case). Cost scales with the machine's process count.

**Most of it is argv rendering.** `ps -A -w -w -o pid=,ppid=,tty=,args=` costs 0.04 s; dropping
`args=` costs 0.01 s. The discriminator only reads argv for descendants that already failed the
session test — a handful, not the whole machine — so deferring argv and `getsid` to candidates
should reach ~2% without leaving `ps` or adding platform code. Native enumeration (Route B) is the
larger lever and remains unexercised.

Method note: do not try to observe this by polling `/proc` for live `ps` children — each lives
~40 ms and a polling loop reports zero. Use reaped-child CPU with a previous-release control.
```

</details>

The method note matters as much as the number — polling `/proc` for live `ps` children reported **zero** on my first attempt, because each `ps` lives ~40ms. Without that written down the next person repeats the dead end and concludes there's no cost at all.

## One question back

What did you want to do with #370 — close it alongside #386, given this supersedes its `tcgetpgrp` primitive and carries its mechanism?

---

Everything else checks out: `fmt`, `clippy -D warnings`, release build, `test-fast`, `linkage-check` (rule 7), `windows-cross` all pass, CI is green including `build-macos` / `build-windows` / `security`, and rules 3/4/9/10/11 are satisfied — rule 4 comfortably exceeded. The `docs/develop/` note is genuinely good: documenting both silent failure modes, what's measured vs. inferred, and that a green CI run says nothing about this signal is exactly the right instinct.

src/daemon_protocol.rs
```diff
@@ -192,7 +192,18 @@ pub const KIND_STREAM_REJECT: u8 = 0x17;
...
+/// `ShellIdle` variants (the daemon-synthesized "a foreground shell command
+/// is running" signal), following the exact precedent PRD #201 set for
+/// `AgentType::Pi` above — a pre-#370 reader has neither variant nor a
+/// `#[serde(other)]` catch-all, so a `KIND_EVENT` frame carrying one fails
+/// its whole-frame decode. The bump forces `probe_remote_protocol` to refuse
+/// the old-reader/new-daemon pairing with a clean `ProtocolMismatch` instead
+/// of a mid-session crash. `EventType` now also carries `#[serde(other)]`
+/// (mirroring `AgentType`'s retrofit), so future event-type additions need
+/// no further bump.
+pub const PROTOCOL_VERSION: u32 = 7;
```
vfarcic, line 206: **Blocking — this bump needs a `changelog.d/*.breaking.md` fragment.**

By this repo's own definitions the 6 → 7 step is a `breaking` change:

- `docs/develop/versioning.md:9` — "Breaking = a change to the TUI↔daemon protocol/handler contract such that an older and a newer build cannot safely interoperate", explicitly including "the obvious structural cases (a new request variant, a changed field shape)".
- `pyproject.toml:19-24` defines the `breaking` towncrier type in those exact words, noting "While major is 0 a `breaking` fragment bumps the MINOR digit".
- `.claude/skills/dot-ai-tag-release/analyze.sh:56,84-90` sets `has_breaking` **only** from a `*.breaking.md` filename; without one, `bump_type=patch`.

The only fragment on the branch is `changelog.d/370.feature.md`, so the next release would cut `v0.35.9 → v0.35.10` (patch) while the minor digit — which this project has deliberately made mean "compatibility broke" — never moves. That matters here: a pre-#370 reader has neither variant nor a `#[serde(other)]` catch-all and fails its whole `KIND_EVENT` frame decode, and `probe_remote_protocol` (`src/connect.rs:516`) now refuses an old-TUI/new-daemon `connect` with `ProtocolMismatch`.

The precedent this comment cites confirms it — PRD #201 (`f5f53ba`), which bumped 5 → 6 for `AgentType::Pi`, shipped **both** `changelog.d/201.breaking.md` and `changelog.d/201.feature.md`.

The PRD already contains the reasoning: `prds/386-descendant-scan-shell-activity-signal.md:240` says "**The `PROTOCOL_VERSION` bump IS owed upstream** … Upstream is at **6**, so this change carries the 6 → 7 step itself." Line 269's "no `.breaking.md` owed … Bump policy: patch" was written while the PRD was still scoped fork-only (where the 6 → 7 step had already landed) and wasn't revised when the destination changed. Only the file is missing.

**Fix:** add `changelog.d/386.breaking.md`.

tests/e2e_shell_activity_real_agent.rs
```diff
@@ -0,0 +1,580 @@
...
+    // The precondition this test exists to gate on. Roughly a third of runs
+    // reach the moment that actually exercises Defect A: the rest end with
+    // Claude Code's `ToolEnd` and no `Stop`, in which case the card never
+    // leaves `Working` and the badge assertion below would hold whether or
+    // not the fix works at all. Waiting for the real, native `Idle` (mapped
+    // from the agent's own `Stop` hook, never fabricated by this test) turns
+    // the old implicit timing assumption (ToolStart + a fixed offset) into an
+    // explicit causal one: this run only counts as a PASS of the bug path if
+    // the Idle that Defect A's fix has to correct actually landed.
+    let idle_wait_timeout =
```
vfarcic, line 346: **Blocking — `006` has no nextest timeout override, so it dies in the full-tier run.**

`.config/nextest.toml:3` sets the default profile to `slow-timeout = { period = "60s", terminate-after = 3 }` → a hard **180s** kill, and `cargo test-e2e` (`.cargo/config.toml:3`) uses that default profile with no `-P`. This branch doesn't touch `.config/nextest.toml`. Observed in my full-tier run:

```
SLOW        [> 60.000s] shell_activity_006_...
SLOW        [>120.000s] shell_activity_006_...
TERMINATING [>180.000s] shell_activity_006_...
TIMEOUT     [ 180.027s] (3718/3719) shell_activity_006_...
```

To be fair about severity, since I then measured it: standalone on a quiet machine `006` finishes in **129.3s**, so it isn't unconditionally impossible — about 50s of headroom. But its own declared waits budget far past that: `:270` prompt-ready **120s**, `:310` `ToolStart` **120s**, and this `Idle` wait `127 + 30 =` **157s** — up to **397s**, more than 2× the window. Any run that actually uses that margin, or any full-tier run under normal contention, is killed. Rule 5 makes the whole-tier `cargo test-e2e` the pre-PR gate, and that's exactly where it died.

The failure shape is the bad kind: a bare `TIMEOUT` with no diagnostics, so the `PRECONDITION NOT MET` vs `BADGE WRONG` distinction you carefully built never prints — and the `[reel]` clip captures a terminated run (a `full-stream.cast` is still written).

Every comparable long real-agent test here already has a named carve-out for exactly this — `route_001` (×8), `new_pane_016` (×5), `orchestration_dispatch_002` (×7), `version_001` (×10), `devin_live_001` (×6) — and those blocks carry the warning that a missing filter "silently drops this test back to the 3 × 60s default, killing it mid-wait."

**Fix:** add to `.config/nextest.toml`

```toml
[[profile.default.overrides]]
filter = 'test(shell_activity_006_real_claude_bash_call_crossing_the_cap_keeps_the_badge_working)'
slow-timeout = { period = "60s", terminate-after = 8 }
```

plus the matching `[[profile.e2e.overrides]]` (the `e2e` profile doesn't inherit the default's override list). Worth considering for `005` too — its `120s + 120s` waits can also exceed 180s on a slow boot, even though its happy path is ~26s.

src/daemon.rs
```diff
@@ -988,6 +1001,270 @@ async fn run_idle_monitor(
...
+            .shell_foreground_busy_snapshot(crate::platform::proc::MEASURED_SHELL_TOOL_SHAPES);
+        let seen: std::collections::HashSet<&str> = snapshot
+            .iter()
+            .map(|(pane_id, _)| pane_id.as_str())
+            .collect();
+        // Drop panes that disappeared from the registry since the last poll
+        // (closed / respawned) so a later reuse of the same pane id starts
+        // edge-detection from a clean slate instead of inheriting a stale
+        // busy/idle reading.
+        last_known.retain(|pane_id, _| seen.contains(pane_id.as_str()));
```
vfarcic, line 1116: A failed `ps` sample is indistinguishable from "no panes" here, and that re-emits a spurious edge.

When `process_table()` returns `None`, `shell_foreground_busy_snapshot` returns an empty `Vec` (`src/agent_pty.rs:4104`), so `seen` is empty and this `retain` **clears every pane**. On the next good tick every pane looks new (`:1109`, `insert(...) != Some(busy)` → `changed = true`), so each idle pane emits a `ShellIdle`.

Harmless for *status* — `apply_event`'s `ShellIdle` arm no-ops while `shell_synthetic_working` is false (`src/state.rs:3814`) — but the event is still broadcast to every attached TUI, pushed into `recent_events`, and bumps `session.last_activity` (`src/state.rs:3687`). That resets the card's elapsed display (`src/ui.rs:16506`) and the idle-notification timer (`src/ui.rs:16791`). Concretely: a pane sitting at `WaitingForInput` for nine minutes would show `0s` and have its idle nudge deferred.

Low probability (needs `ps` to actually fail), but this is the one follow-up with user-visible consequences, and it's cheap to prevent by distinguishing "the sample failed" from "there are no panes" — e.g. have the snapshot return `Option<Vec<_>>` and skip the whole tick (leaving `last_known` intact) when the sample is `None`.

src/event.rs
```diff
@@ -17,6 +17,25 @@ pub enum EventType {
...
+    /// gone, i.e. the previously-running foreground command has finished. See
+    /// [`ShellBusy`](Self::ShellBusy).
+    ShellIdle,
+    /// PRD #370 / precedent PRD #201 (`AgentType`'s identical retrofit):
+    /// forward-compat catch-all for a future/unknown `event_type` string on
+    /// the wire, so a build newer than THIS one can add further variants
+    /// without another `PROTOCOL_VERSION` bump. Deserialize-only — never
+    /// produced by this build. Treated as a no-op wherever `EventType` is
+    /// matched (never proof of agent activity, never changes `SessionStatus`).
+    #[serde(other)]
```
vfarcic, line 37: The forward-compat catch-all is right and matches `AgentType`'s precedent — but it silently swallows a hook typo that used to be reported.

Before this, an unrecognized `event_type` failed the decode and the daemon logged `warn!("Malformed event: {line}")`. Now it decodes to `Unknown`, lands in `recent_events`, and bumps `last_activity` while changing nothing visible — so a typo in a hand-written hook looks like it works. `reject_invalid_event_type` having its assertion inverted is the deliberate part; the lost diagnostic is the incidental part.

A `debug!` (or `warn!`) at ingestion when `event_type == EventType::Unknown` would restore it without giving up the compatibility win.

tests/shell_activity.rs
```diff
@@ -0,0 +1,670 @@
...
+/// exercised. Polls `shell_foreground_busy_snapshot(&[CLAUDE_BASH_TOOL_SHAPE])`
+/// and asserts the pane reads idle before the detached child appears, busy
+/// while it lives — independently confirmed via `process_table()` +
+/// `descendants()` that the found descendant has no controlling terminal, is
+/// its own session leader, and carries a session id different from the
+/// pane's own shell — and idle again once the child is killed.
+#[cfg(unix)]
+#[spec("status/shell-activity/004")]
+#[test]
+fn shell_activity_004_shell_foreground_busy_flips_for_a_real_detached_pipe_child() {
```
vfarcic, line 503: This test's pre-edge assertion races a ~300 ms window and fails under load.

The pane script is `sleep 0.3; python3 -c "…setsid…"; sleep 5`, and the first assertion polls up to 2s for `Some(false)`. The idle window is only ~300 ms, so once load delays `spawn_agent` + registry setup past it, the first successful sample already sees the detached child, the loop spins out its full 2s, and it fails:

```
assertion `left == right` failed: an idle shell with no detached descendant yet must not read busy
  left: Some(true)
 right: Some(false)
```

The production code is correct here — a shape-matching detached descendant genuinely existed by then. I confirmed it's purely a race: **5/5 passes in isolation** (0.77–1.87s), and it failed only in the contended full-tier run.

Suggest widening the leading `sleep`, or sampling the idle state before the script can launch the child at all, so the assertion doesn't depend on winning a 300 ms race.

tests/e2e_shell_activity_real_agent.rs
```diff
@@ -0,0 +1,580 @@
...
+/// Scenario: launch a real interactive Haiku Claude agent through the normal Ctrl+N new-pane flow (per-folder trust pre-seeded, `--allowedTools Bash`), then type a directive prompt naming a uniquely-named sentinel fixture file that instructs the agent to run `ping -c 20 127.0.0.1 > /dev/null` as its one Bash tool call. While that ~20s foreground command is in flight, assert — over a `SubscribeEvents` connection, never against the rendered badge — that the daemon's shell-activity monitor synthesizes a `ShellBusy` broadcast event for the pane.
+#[spec("status/shell-activity/005")]
+#[test]
+fn shell_activity_005_real_claude_bash_child_trips_the_descendant_scan() {
+    skip_unless!(common::check_claude_available());
+
+    let deck = TuiDeck::builder()
+        .with_imported_claude_credentials()
+        .launch_with_fixture("minimal");
+    deck.wait_for_string("No active sessions");
```
vfarcic: `005` and `007` kept the shared 10s startup bound that `006` was explicitly given 30s to avoid.

This line and `:448` (`007`) call `deck.wait_for_string("No active sessions")`, which uses the shared `WAIT_TIMEOUT` of **10s** (`tests/common/mod.rs:53`). `006` at `:234` instead uses `wait_for_grid_string_within(…, Duration::from_secs(30))`, with a comment naming this exact hazard — "`wait_for_string`'s shared 10s bound is tight on a loaded machine (this one routinely runs at load average 20+) and this startup race is unrelated to anything this test asserts".

That diagnosis is right, and it's precisely how both of these failed in my run: `did not see "No active sessions" within 10s` with a **completely empty** captured grid, while `006` cleared startup under the same conditions. The fix is already written for one test — worth copying to the other two.

vfarcic (commented):
## Verification of #390

Ran the full local gate against `d6f4bf1` merged with `origin/main` (clean merge, 0 behind): `fmt`, `clippy --workspace --all-targets --features e2e -D warnings`, release build, `cargo test-fast`, `linkage-check`, `windows-cross`, and the whole `cargo test-e2e` tier. CI is green on all 11 checks, including the three a local run can't replace (`build-macos`, `build-windows`, `security`).

**The feature works, and I verified it rather than taking the PRD's word for it.** All three real-agent tests pass on a quiet machine, against a live **Claude Code 2.1.226** (the PRD measured 2.1.220, so the coupling holds at the newer version too):

- **`005`, the rot canary** — passes (25.6s). Claude Code still `setsid`-detaches its Bash-tool child, and the daemon genuinely synthesizes `ShellBusy` for the pane through the real pipeline.
- **`007`, the false-positive canary** — passes (7.4s) against that machine's real MCP-server process table.
- **`006`** — passes (129.3s) and genuinely exercised the bug path, not a shortcut past it: `ToolStart`, then the **real `Stop`-mapped `Idle` 123.5s later**, after which the rendered badge still read `Working` and `process_table()` confirmed the sentinel-tagged shell with a live `ping` beneath it. That's the M6b re-emit working end to end, as a user sees it.

So both silent failure modes `docs/develop/shell-activity-signal.md` worries about are checked and clear. The diagnosis, the structural discriminator, and the documentation are all excellent — the `getsid` separation is the right instrument, and the precedence rules never clobber a real status.

### About the e2e run — please don't read the raw numbers as yours

My full-tier run reported `3693 passed, 23 failed, 3 timed out`. That run was contaminated: **five concurrent `verify-pr` sessions** were running e2e suites on one 16-core box, load peaked at **206** with **0 GB free memory**, and `/tmp` (a tmpfs) held 7.7 GB of `dad-tests-*` roots. Rerun serially per rule 6, **every one of this PR's failures cleared** — `004` 5/5, `005`, `007`, and `006`. Twenty of the 23 failures are in unrelated suites.

### Two things to fix before merge

Neither is a production correctness defect — I found none. Both are shipping artifacts, one file each, commented inline:

1. **A `.breaking.md` fragment is owed for the `PROTOCOL_VERSION` 6 → 7 bump** — without it the release cuts as a *patch* and the minor digit never moves. (`src/daemon_protocol.rs`)
2. **`006` needs a nextest `slow-timeout` override** — it is killed at 180s in the full-tier run that rule 5 makes the pre-PR gate. (`tests/e2e_shell_activity_real_agent.rs`)

Four smaller items are also inline (`src/daemon.rs`, `src/event.rs`, `tests/shell_activity.rs`, `tests/e2e_shell_activity_real_agent.rs`) — all in code this PR introduces, so they're cheaper to fix here than to track.

One note on sequencing: `dismiss_stale_reviews_on_push: true` on this repo, so please batch all of these into a single push before asking for review, otherwise an approval gets voided.

Greptile's P2 on the untimed `ps` is already well answered and I'm not reopening it — deferring it is reasonable and it's disclosed in both the PR body and the PRD.

src/daemon_protocol.rs
prageethw: **Agreed — fixed in `b84135d`.**

Added `changelog.d/386.breaking.md`. It states the two concrete consequences a user needs: an old TUI cannot talk to a new daemon (`probe_remote_protocol` refuses with `ProtocolMismatch`), and a pre-#370 reader fails its whole `KIND_EVENT` frame decode because it has neither the new variant nor a `#[serde(other)]` catch-all — so TUI and daemon must be upgraded together.

You were right about the origin of the mistake, and I fixed that too: `prds/386-descendant-scan-shell-activity-signal.md:269` still said "no `.breaking.md` owed … Bump policy: patch", which directly contradicted line 240 of the same file. That line was written while the PRD was scoped fork-only and never revised when the destination changed to upstream. Line 269 now matches; line 240 was already correct and is untouched.

src/daemon.rs
prageethw: **Agreed — fixed in `b84135d`.**

`shell_foreground_busy_snapshot` now returns `Option<Vec<(String, bool)>>`, which makes the two cases distinguishable at the type level rather than by convention: `None` means the `ps` sample failed, `Some(vec![])` means there genuinely are no panes. The tick in `src/daemon.rs` skips entirely on `None` — `last_known` is left intact and nothing is emitted — while `Some(vec![])` still runs the normal `retain`/diff path.

One call site in `tests/shell_activity.rs` needed `.unwrap_or_default()` to compile. That test samples a real live process table, so `None` is not a path it exercises; the edit is purely to satisfy the type and changes nothing it asserts.

Your read on the consequence was the part that made this worth doing rather than deferring — the status no-op hid it, but the `last_activity` bump reaching the card's elapsed display and the idle-notification timer is genuinely user-visible.

src/event.rs
prageethw: **Agreed — fixed in `b84135d`.** The `#[serde(other)] Unknown` catch-all is unchanged; only the lost diagnostic is restored.

Went with **`warn!`** rather than `debug!`, for two reasons. The closest neighbouring diagnostic in the same ingestion block — the `SessionStart for a pane this daemon did not spawn` case — is itself a `warn!`, and it is the same shape of anomaly: the frame parsed fine, its content is semantically off. And this line is the direct replacement for the pre-#386 `warn!("Malformed event: {line}")` that the catch-all silenced, so keeping the level puts the diagnostic back where operators were already used to seeing it.

It logs `session_id`, `pane_id` and the raw line, so the offending `event_type` string — the actual typo — is visible without a second lookup.

tests/e2e_shell_activity_real_agent.rs
prageethw: **Agreed — fixed in `4f042b1`.** Added the override for `006` in both `[[profile.default.overrides]]` and `[[profile.e2e.overrides]]`, following the house style: a comment block carrying the declared-wait arithmetic (120s + 120s + 157s → up to 397s) and the same "the filter tracks the TEST FUNCTION name, not the `#[spec]` id" warning the existing entries carry. The exact function name was confirmed by reading `tests/e2e_shell_activity_real_agent.rs` rather than copying it from the comment — it matched.

**On `005`: added it too**, since you said "worth considering" rather than blocking and it deserved an explicit answer either way. After the Finding 6 fix below, `005`'s declared gating waits sum to up to 315s (30s startup + 120s prompt-ready + 120s `ToolStart` + 15s `ShellBusy`), plus a 30s soft sentinel-echo wait that still burns wall clock — and the 120s + 120s pair alone already clears the 180s window. Gave it `terminate-after = 6` (360s) rather than `006`'s 8, since its happy path is ~20-26s and the margin only needs to cover the declared budget.

**One correction on the profile-inheritance point.** I checked the file rather than taking it on trust, and the claim is not quite established by it: only 2 of the 5 existing `[[profile.default.overrides]]` entries (`devin_live_001`, `version_001`) have a matching `e2e`-profile counterpart — `route_001`, `new_pane_016` and `orchestration_dispatch_002` have none. That is consistent with your claim rather than evidence against it, because `cargo test-e2e` is `nextest run --features e2e` with no `-P`, so it selects the **default** profile and the `e2e` profile is never consulted by the alias at all. Those three simply never needed a counterpart. I added both anyway, matching the two entries that already have both — but if the `e2e` profile is genuinely dead config, that is worth knowing separately.

**Honest scope note:** this branch was cut from `upstream/main` and carries no `e2e:` job, and upstream CI's `build` runs the fast tier only, so **nothing here exercises this override**. CI confirms it parses; it does not confirm it prevents the timeout. The evidence behind it is your measured numbers for `006`, the declared `Duration::from_secs(…)` values I read for `005`, and structural consistency with the five existing entries. I did not run the tier locally.

tests/shell_activity.rs
prageethw: **Agreed — fixed in `4f042b1`**, and I took the second of your two suggestions rather than the first.

Widening the leading `sleep` would have made the race less likely without removing it, and it adds wall clock to every run. Instead the pane script now busy-waits on a ready-marker file before running the `python3`/`setsid`/`execv` sequence. The Rust side samples and asserts the pre-edge `Some(false)` state exactly as before — same 2s-deadline poll, same assertion message — and only writes the marker once that assertion has passed, releasing the gated shell.

That makes the pre-edge read structurally incapable of racing the child's existence: before the marker exists the only observable states are `None` (pane not yet registered) or `Some(false)` (idle), and `Some(true)` is unreachable. The assertion still tests what it tested — an idle shell with no detached descendant must not read busy — and was not weakened. Added a stale-file guard (pids wrap) and best-effort cleanup.

This one **is** covered: it is a fast-tier test, so upstream CI's `build` job genuinely ran it and it passed. Though to be precise about what that proves — a single green run only says this run was not unlucky. What establishes the race is gone is the structural argument, not the passing run.

tests/e2e_shell_activity_real_agent.rs
prageethw: **Agreed — fixed in `4f042b1`.** Mirrored `006`'s existing fix at both sites verbatim, including the explanatory comment, so the three tests now agree: `deck.wait_for_string("No active sessions")` → `wait_for_grid_string_within("No active sessions", Duration::from_secs(30))` with the same panic-message shape.

Small correction to the reference for anyone reading later: `006`'s site is now around `:245` rather than `:234`, since the file grew earlier in the branch.

Same honest scope note as the nextest override — these are real-agent-gated e2e tests, this PR's CI does not run the e2e tier, and they self-skip in CI for lack of credentials in any case. What is verified is that they compile and lint clean under `--features e2e`, and that the edit is a line-for-line mirror of the fix you had already reviewed on `006`. The 30s bound clearing under load on these two specifically is not something I can show you from CI, and I did not run the tier locally.

prageethw:
All six review findings are addressed across two commits, and CI is green on `4f042b1` (11/11).

| Finding | Commit | What changed |
|---|---|---|
| `daemon_protocol.rs:206` — missing `.breaking.md` **(blocking)** | `b84135d` | Added `changelog.d/386.breaking.md`; also corrected the PRD's stale line 269, which contradicted line 240 |
| `daemon.rs` — failed `ps` sample re-emits a spurious edge | `b84135d` | `shell_foreground_busy_snapshot` → `Option<Vec<_>>`; the tick is skipped on `None`, leaving `last_known` intact |
| `event.rs:37` — swallowed hook-typo diagnostic | `b84135d` | `warn!` at the ingestion site on `EventType::Unknown`; catch-all unchanged |
| `nextest.toml` — `006` has no timeout override **(blocking)** | `4f042b1` | Override added for `006` **and** `005`, in both profiles, house style |
| `shell_activity.rs:503` — 300 ms race | `4f042b1` | Ready-marker gate; the pre-edge sample now structurally cannot see the child |
| `e2e_..._real_agent.rs:69,:448` — shared 10s startup bound | `4f042b1` | `006`'s 30s `wait_for_grid_string_within` mirrored to `005` and `007` |

Two things in the thread replies are worth surfacing here rather than leaving buried:

**A correction on the `e2e` profile.** The inheritance claim holds, but the file does not demonstrate it — only 2 of the 5 existing overrides have an `e2e` counterpart. The reason the other three never needed one is that `cargo test-e2e` is `nextest run --features e2e` with no `-P`, so it selects the **default** profile and the `e2e` profile is never consulted by that alias at all. I added both anyway for consistency with the two entries that already have both, but if the `e2e` profile is dead config that seems worth knowing on its own.

**What is and is not verified.** This branch was cut from `upstream/main`, so it carries no `e2e:` job, and upstream CI's `build` runs the fast tier only. So the `shell_activity.rs` race fix genuinely ran in CI and passed — the other two test-side fixes are confirmed to compile and lint under `--features e2e` and nothing more. Their confidence rests on your measured numbers and on mirroring the existing in-repo precedent. I did not run the tier locally, and I would rather say that plainly than let green checks imply coverage that is not there.

Threads left unresolved for you to close — deciding whether a fix answers the finding is the reviewer's call, not the author's.

prageethw:
@vfarcic — all six findings are addressed and replied to inline, and CI is green on `4f042b1` (11/11).

The PR is `BLOCKED` on `required_review_thread_resolution` with all six threads still open. I've deliberately not self-resolved them: whether a fix actually answers a finding is your call, not mine — same principle you applied on #480. Could you take a look and close the ones you're satisfied with?

Two replies have something in them beyond "fixed", if you're skimming:

- **[nextest override]** — a correction on the `e2e` profile claim, and my reasoning for adding `005` as well as `006`.
- **[the 300 ms race]** — I took your second suggestion (sample before the child can launch) rather than widening the sleep, since a wider sleep is still a race.

Happy to change any of it if you'd rather it went a different way.

vfarcic (approved):
Approving. All six findings from my 2026-08-10 review are fixed in `b84135d` + `4f042b1`, and I verified each fix rather than taking the replies on trust.

**Rule 12 — the classification is right, and it's not either/or.** The wire shape moved (two serialized `EventType` variants riding `KIND_EVENT`), so `PROTOCOL_VERSION` 6→7 is owed; separately the towncrier `breaking` type is what drives the 0.x minor digit. Both are present, and the bump resolves to **v0.36.0**. Verified empirically: the released v0.35.10 binary self-reports `server_version: 6`, this branch reports `7`, and `git show v0.35.10:src/event.rs` confirms the fragment's central claim — that reader has neither variant nor a `#[serde(other)]`, so a `shell_busy` frame genuinely fails its whole frame decode.

**Cross-version test run**, with sockets, `HOME`, state dir *and* `DOT_AGENT_DECK_LOG` isolated: an old protocol-6 daemon under this branch's TUI took `SIGTERM` at 18:09:37.97 and a new daemon came up at 18:09:38.19 — PRD #161's `SilentRestart`. No mismatched frame is exchanged locally. Worth recording for later: `daemon_protocol.rs:16-18` scopes the `Hello` version check to the SSH path only, so the `ProtocolMismatch` guard the fragment cites is the remote path — which is what the fragment says.

**The feature works against a real agent.** You flagged twice that you hadn't run the tier locally; that gap is now closed. In isolation: `005` PASS (27.1s, `ShellBusy` 478ms after `ToolStart`), `007` PASS (22.0s, no false positive with live MCP servers), and `006` PASS (134.2s) genuinely reproducing the bug — `native Idle observed 129165ms after ToolStart`, command confirmed still running, badge stayed `Working`.

The nextest overrides are warranted and correctly targeted — I confirmed both filters resolve to real function names, and `006`'s measured 134.2s against a 180s kill is exactly the thin margin you described.

Three follow-ups, none blocking:

1. The blocking `ps` on a Tokio worker (Greptile's P2) — disclosed and deferred by agreement.
2. **New:** `process_table()` is called before the registry is filtered (`src/agent_pty.rs:4116`), so the daemon forks `ps -A` ~2x/second **even with zero panes** — measured on a daemon with no panes at all, plus a `getsid(2)` per process row. An early return when no live panes exist is the cheap win; fits PRD #386's M5.
3. Your profile-inheritance question, answered: `[profile.e2e]` is dead config. `.cargo/config.toml:3` defines `test-e2e` with no `-P`, and no `-P e2e` appears anywhere in `.github/`, so the override list is inert.

Separately and not yours: `idle_worker_011` and `card_stats_005` fail at the merge-base too — pre-existing on `main`.

Issue #34, opened 2026-04-03
Extensible Modes System — project-local workspace layouts with side panes


vfarcic:
## PRD: Extensible Modes System

**Problem**: When AI agents execute commands (kubectl get, cargo test, etc.), the output is buried in the agent's conversation. DevOps/SREs and developers want to see live command output in dedicated side panes alongside the agent.

**Solution**: A fully config-driven modes system where projects define workspace layouts via `.dot-agent-deck.toml`. Each mode creates a Zellij tab with an agent pane and configurable side panes — both persistent (always-running commands like test watchers) and reactive (populated when the agent executes matching commands).

**Detailed PRD**: See [prds/34-extensible-modes-system.md](https://github.com/vfarcic/dot-agent-deck/blob/main/prds/34-extensible-modes-system.md)

**Priority**: High

PR #392, opened 2026-08-06
WIP: fix hydration idle-edge race (#36)

prageethw:
Closes the reconnect **snapshot/subscribe window** — fork issue #36, found by Greptile while reviewing PR #34.

## The defect

On attach the TUI rebuilt its world from two independent daemon connections that nothing ordered:

| | where |
|---|---|
| `ListAgents` snapshot | `hydrate_from_daemon`, on the blocking TUI thread (`src/ui.rs`) |
| `SubscribeEvents` stream | a detached `tokio::spawn` from `main` |

Whichever the daemon accepted first won. When the snapshot won, every event broadcast between the two was delivered to nobody — no receiver was registered yet, and nothing replays.

Invisible for most events; **permanent for an edge**. The shell-activity monitor emits exactly one `ShellBusy` when a foreground command starts and one `ShellIdle` when it ends. Snapshot captures the synthetic `Working`, the paired `ShellIdle` falls in the gap, hydration seeds `Working` from an already-stale snapshot — and nothing ever corrects it.

Distinct from #21 (fixed in PR #34, `0784ff4`): there the `ShellIdle` *was* delivered and hydration dropped the provenance marker that entitled it to revert. Here it is never delivered.

## The fix — subscribe first, buffer across hydration

The daemon registers a subscriber's broadcast receiver **before** writing the subscription's OK `RESP` (`handle_subscribe_events`, `src/daemon_protocol.rs:1827-1829`), so a client that has read that RESP can no longer miss anything. That makes the gap closable entirely client-side, **with no wire change**:

1. `reconnect::run_event_subscriber` (moved out of `main.rs` so tests can drive it) signals a `HydrationGate` once the stream is confirmed.
2. `hydrate_from_daemon` waits for that signal before `ListAgents` — the snapshot is then strictly newer than the subscription. Expiry after 5s degrades to today's behaviour, never stalls.
3. Events arriving while hydration is still seeding are **buffered**, not applied: `apply_event` drops any non-`SessionStart` event whose pane is not yet registered, so reordering alone would still lose the edge.
4. `run_tui` releases the gate once every hydrated pane is registered and seeded; the subscriber drains in arrival order. A 30s watchdog in `main` opens the gate if hydration never signals.

Replay is safe: the daemon broadcasts and applies each event under one write-lock acquisition (`ingest_event`, PRD #386), so the snapshot can never be captured *between* an event's broadcast and its application.

## Rule 12

**No `PROTOCOL_VERSION` bump and no `changelog.d/36.breaking.md`.** No field was added, removed, or reinterpreted; the change is purely the order in which one client opens its own two connections. An old TUI against a new daemon (and the reverse) interoperates exactly as before. Shipped as `changelog.d/36.bugfix.md` (patch).

## Scope — what this does NOT fix

Only the **bootstrap** gap. If the subscription later dies (daemon restart, `KIND_STREAM_END "lagged"`), the TUI re-subscribes but does **not** re-hydrate, so events broadcast during that outage are still lost. Pre-existing, out of scope for #36.

## Test

`session/live/012` pins the **race**, not the symptom: a mock daemon deliberately slow to acknowledge `SubscribeEvents` (receiver registered immediately before the OK RESP, exactly as the real one does) that broadcasts the paired `ShellIdle` the instant it has served the snapshot. The delay makes the window deterministic rather than scheduler-dependent, so reverting *either* half of the fix fails it rather than flaking. `session/live/011` and the daemon seam test still pin #21's half.

Revert-check evidence follows in the thread before this leaves draft.


prageethw:
Opened against the wrong repository by mistake — this is a fork-only fix (prageethw/dot-agent-deck issue #36). Reopening on the fork.

Issue #330, opened 2026-08-01
delegate exits 0 with no output, so a confirmatory re-run silently arms a duplicate delegation


vfarcic:
Observed live while orchestrating #303. Same knock-on effect as the one documented in #303, but a **different root cause** — and one that #303's fix does not address.

## What happened

An orchestrator ran:

```bash
dot-agent-deck delegate --to coder --task-file .dot-agent-deck/<slug>.md
```

The command printed **nothing at all** and exited 0. With no output, there was no way to tell a successful delivery from a silently dropped one, so the orchestrator re-ran the identical command to check. That armed **two** delegation records for the same worker.

## Why that matters

Per `src/state.rs:2690-2730`, each `delegate` arms its own record with a sequence number, and a single `work-done` retires only one — the other is deliberately kept armed:

- `SilenceWatchRetirement::KeptNewer { seq, remaining }` — "work-done: credited to a superseded delegation; the newest silent-worker watch stays armed"
- `DelegationRetirement::RetiredSuperseded { role, seq, remaining }`

That behaviour is correct by design for a genuine re-delegation (a re-delegated worker that then goes silent should still be reported). But when the second delegate was only a *duplicate of the first*, the surviving record has no real work behind it, and PRD #126's idle-worker detector eventually injects a **false STUCK notice** for work that was already finished and committed.

#303 documents this exact end state, reached via shell-mangled text prompting a corrective re-delegation. This is a second, independent path into it: a silent success prompting a confirmatory re-delegation. Fixing #303's advice does not close it.

## Proposed fix

Make `delegate` confirm delivery on success — for example:

```
delegated to coder (pane <pane-id>)
```

One line naming the role and the resolved target is enough. An orchestrator that sees positive confirmation never re-runs to find out whether the command worked, which removes the duplicate-arming path at its source.

Worth considering the same treatment for `work-done`.

## Related

- #303 — the shell-mangling path into the same knock-on effect.
- The companion issue about the daemon overwriting a work-done input file at its own output path — another silent-failure mode in the same delegation loop, also observed live in this run.

prageethw:
Fixed on our fork if it is useful to you — #309 and #330 turned out to be one fix, and the design has a couple of decisions worth surfacing rather than just handing over a diff.

## The shape

[...]

Issue #309, opened 2026-08-01
`delegate` from a non-orchestrator pane is rejected silently — the CLI exits 0 and the agent believes the task shipped


vfarcic:
## The defect

The daemon rejects a `delegate` issued from any pane that is not the orchestrator, and that rejection reaches nobody:

- `src/state.rs:2523-2531` — `handle_delegate` logs `warn!("delegate from non-orchestrator pane")` and returns. Nothing is written back to the caller.
- `src/main.rs:657-661` — the CLI is fire-and-forget: `hook::send_to_socket` returns as soon as the line is written, and the command returns `ExitCode::SUCCESS`.

So an agent that runs `dot-agent-deck delegate --to … --task …` from a worker pane sees **exit 0, no stderr, and no effect**. From the agent's point of view the delegation shipped. The sibling early return at `src/state.rs:2519-2522` (`delegate from unknown pane`) is silent in exactly the same way.

## Why it matters

Surfaced in #305, where a user asked about cross-agent requests "and vice versa". Trying the worker→orchestrator direction is indistinguishable from the feature silently not existing: no error to search for, nothing in the pane, and the only evidence is a `warn!` in a daemon log nobody is watching.

It also blunts the anti-spoofing check it implements. A rejection no one can observe teaches no one — the agent will simply try again the same way.

`docs/orchestration.md:415` documents the rejection and the log line, which is the best that can be done while the CLI reports success.

## Fix shape

Have `delegate` await an acknowledgement instead of firing and forgetting. The precedent already exists: `hook::request_from_socket` (`src/hook.rs:432`) does exactly one request → one reply line on the same connection. It was added for `get-seed` in PRD #201 and already documents how to degrade when the daemon never answers (older build, or no daemon) without hanging.

Sketch:

- daemon replies `{"ok":true}` on accept, `{"ok":false,"error":"…"}` on either early return;
- the CLI prints the error to stderr and exits non-zero;
- an absent reply preserves today's behaviour, so an older daemon does not break the caller.

Worth deciding at the same time whether `work-done` (also fire-and-forget, `src/main.rs:690`) should get the same treatment — a work-done from a pane the daemon cannot route is equally invisible.

## Cross-version note (CLAUDE.md rule 12)

[...]

Issue #17, opened 2026-03-31
Aggregate Stats Bar


vfarcic:
## PRD: Aggregate Stats Bar

**Problem**: When running many agents, there's no quick way to see overall status — users must visually scan all cards to understand how many agents are active, waiting, or erroring.

**Solution**: Add a persistent status bar at the bottom of the dashboard showing real-time aggregate metrics across all sessions.

**Detailed PRD**: See [prds/17-aggregate-stats-bar.md](https://github.com/vfarcic/dot-agent-deck/blob/main/prds/17-aggregate-stats-bar.md)

**Priority**: Medium

PR #408, opened 2026-08-07
test(delegate): RED tests for caller-visible delegate feedback (#309, #330)

prageethw:
## Summary

RED tests only — no production code touched. Pins the caller-visible contract that upstream #309 and #330 both describe as missing: `delegate` is fire-and-forget end to end (`src/main.rs:675` + the silent early returns at `src/state.rs:3005`/`3014`), so a rejected delegation and a successful one are both exit 0 with no output.

- `orchestration/delegate/016` — a non-orchestrator pane's `delegate` must exit non-zero with a stderr reason (#309).
- `orchestration/delegate/017` — an unknown pane's `delegate` must do the same, exercising the sibling early return (#309).
- `orchestration/delegate/018` — a successful `delegate` must print a non-empty stdout confirmation naming the delegated role (#330).

Each invokes the real `dot-agent-deck delegate` CLI as a subprocess against a real daemon socket (`common::spawn_inprocess_daemon`), not the in-process `handle_delegate` shortcut the rest of `delegate_prompt_injection.rs` uses — that's the only way to observe the CLI's own exit code/stdout/stderr.

Assertions deliberately do not pin exit code value, stderr/stdout wording, or wire shape (the fix likely needs a daemon→CLI reply, an open `PROTOCOL_VERSION` design decision per CLAUDE.md rule 12) — mirrors fork #17's `result.is_err()` precedent.

## Test plan

- [ ] CI fast tier runs `orchestration/delegate/016`/`017`/`018` and reports RED (today's fire-and-forget CLI passes exit-code-0 / empty-output through unchanged).
- [ ] `cargo fmt --check`, `cargo clippy -- -D warnings`, `cargo clippy --all-targets --features e2e -- -D warnings`, `cargo xtask linkage-check` all pass locally (done).


prageethw:
Opened in error — this branch belongs on my fork (prageethw/dot-agent-deck), not here. Closing immediately, no review requested.

PR #411, opened 2026-08-07
fix(codex-hooks): preserve destination mode in write_atomic (#382)

### LINKED ISSUE #382: Preserve destination file mode in codex_hooks_manage::write_atomic (and extract the shared helper)
`src/codex_hooks_manage.rs::write_atomic` publishes through `File::create`, which applies `0666 & !umask` — 0644 under a typical 022 umask, **0664 (group-writable) under 002**. The subsequent `rename` then replaces the destination with that wider file, so the deck silently widens a config the user had kept private.

Affected call sites, all in `codex_hooks_manage.rs`:

- line 326 — `hooks.json`
- line 457 — `hooks.json`
- line 831 — the user's `config.toml` (model choice, hook-trust records, and any hand-written settings)

Codex credentials themselves are not affected: `auth.json` is written by a separate path that already forces 0600.

## Why now

The identical defect was found and fixed for the Devin adapter in #360, where it was **not** theoretical — a real `devin` install ships `~/.config/devin/config.json` at 0600 and it holds `devin.org_id`. The Devin code had been copied verbatim from `codex_hooks_manage`, so the original still carries the bug. Greptile flagged it there as a P1.

## Fix

Preserve the destination's existing mode, and create owner-only when the file is new — the patch applied in #360:

```rust
#[cfg(unix)]
{
    use std::os::unix::fs::PermissionsExt as _;
    let mode = std::fs::metadata(dest)
        .map(|meta| meta.permissions().mode() & 0o777)
        .unwrap_or(0o600);
    file.set_permissions(std::fs::Permissions::from_mode(mode))?;
}
```

Rather than copy it a third time, extract `write_atomic` (and the `shell_quote_if_needed` / `build_command` pair, which are also duplicated verbatim between the two modules) into a shared helper both adapters use. That is why this was kept out of #360: the clean fix spans two adapters and needs the Codex trust/hooks e2e tier re-verified, which is outside the scope of "add Devin".

Add a regression test mirroring `devin_hooks_manage::tests::install_never_widens_config_permissions`.

prageethw:
## Summary
- Fixes vfarcic/dot-agent-deck#382: `codex_hooks_manage.rs::write_atomic` created its temp file with `File::create` (Unix default `0666 & !umask`, typically `0644`) regardless of the destination's existing mode, and `rename(2)` replaces the destination inode outright — so `CODEX_HOME/hooks.json` and, via `edit_trust_state`, the user's real `CODEX_HOME/config.toml`, were silently widened the first time the deck wrote through them (and stayed widened afterwards, since nothing repairs it).
- Mirrors the fix already present in `devin_hooks_manage.rs::write_atomic`: set the temp file's mode to the destination's existing mode (`& 0o777`), or `0600` for a new file, before writing/renaming.
- Adds `changelog.d/382.bugfix.md`.

Fork-first: this is upstream-contributable but not opened upstream per this repo's fork-first workflow.

Rule 12: local file-permissions-only change — no wire, no `PROTOCOL_VERSION`, no cross-version check needed.

## Test plan
- [x] `cargo fmt --check`
- [x] `cargo clippy -- -D warnings`
- [x] `cargo clippy --all-targets --features e2e -- -D warnings`
- [x] `cargo xtask linkage-check`
- [ ] CI (fast tier) — added two `#[cfg(unix)]`-gated unit tests in `codex_hooks_manage`: `write_atomic_keeps_an_existing_destinations_restrictive_mode`, `write_atomic_creates_a_new_destination_owner_only`


prageethw:
Opened by mistake against the wrong remote (gh defaulted to upstream instead of my fork). Closing — the real PR is against my fork, not here. Apologies for the noise.

Issue #407, opened 2026-08-07
cargo clippy --all-targets type-checks nothing in e2e files without --features e2e, so rule 2's gate is a no-op for e2e-only changes


prageethw:
CLAUDE.md rule 2 requires `cargo fmt --check` and `cargo clippy -- -D warnings` before every commit. Neither of those — nor `cargo clippy --all-targets -- -D warnings` — actually type-checks any `tests/e2e_*.rs` file.

Every one of those files opens with `#![cfg(feature = "e2e")]`. Without `--features e2e` the whole file compiles to nothing, so clippy reports clean over a build that never contained the changed code.

**Consequence: an e2e-only change committed under the current gate has never been type-checked locally.** It is caught later by whoever runs the tier, or not at all.

## How it surfaced

Found on a change touching 19 `e2e_*.rs` files and **nothing else**. The standard gate passed instantly and meaninglessly. The command that actually checked the work was:

```
cargo clippy --all-targets --features e2e -- -D warnings
```

This was verified rather than assumed — inspecting `--message-format=json` output confirmed the `e2e_*` test targets are present in the feature-enabled build and absent otherwise.

## Suggested fix

Add the feature-enabled variant to rule 2's pre-commit list. Options, roughly in increasing order of reliability:

1. **Document it in CLAUDE.md rule 2** as a conditional extra command for changes touching `tests/e2e_*.rs`. Cheapest, but relies on the author noticing their change is e2e-only — which is exactly the failure mode that just occurred.
2. **Make it unconditional** in rule 2. One list, no judgement call, at the cost of a slower pre-commit gate.
3. **Enforce it in `cargo xtask linkage-check`**, which already runs pre-commit and already knows the e2e test files: fail if e2e files changed and the feature-enabled clippy has not been run.

Option 2 seems the best cost/benefit for the rule text; option 3 is the durable version if it turns out cheap to wire up.

Issue #398, opened 2026-08-07
build_pane_status: duplicate pane_id sessions collide non-deterministically, so a pane can show Needs Input while the lock keeps dropping keys


prageethw:
`build_pane_status` (`src/ui.rs:12317`) collects `state.sessions.values()` into a `HashMap` keyed by `pane_id`, with **no dedupe**. When two sessions carry the same `pane_id`, the surviving status is whichever session `HashMap` iteration happens to reach last — which is unspecified and can differ between runs.

Found while implementing PRD #393 M3 (fork PR #51). Filing separately because it **predates that work** and affects consumers that shipped long before it.

## Why two sessions can share a `pane_id`

`AppState::apply_event`'s card-reuse guard only reuses an existing card when the incoming `agent_id` matches. An event carrying `agent_id: None` therefore creates a **new** session record even when a session for that `pane_id` already exists. Observed directly in an e2e failure grid — a two-role fixture rendering three cards, two of them labelled `worker`:

```
┌ 1 orchestrator ───────────   No agent ┐   Dir:  .tmpfhuTbX
┌ ▸ 2 worker ────────────   Needs Input ┐   Dir:  —
┌ 3 worker ─────────────────   No agent ┐   Dir:  .tmpfhuTbX
                                             1 active  │  1 waiting  │  0 tools
```

## Who reads this join, and what the ambiguity costs them

Three consumers, all reading the same map:

- **PRD #333 — orchestration tab-label colouring.** A tab could colour for the wrong session's status.
- **PRD #373 — `auto_focus_waiting_pane`.** Focus steering could ignore a genuinely waiting pane, or steer toward one that has already resolved.
- **PRD #393 M3 — the `WaitingForInput` command-entry carve-out** (the new one). This is the sharpest case, because the ambiguity is now user-visible in a confusing way: **a worker pane can render `Needs Input` on its card while the gate reads the other, stale session and keeps dropping the human's keystrokes.** The user sees an agent asking a question, types the answer, and nothing happens — with a status message telling them the pane is locked.

The first two degrade gracefully — a wrong colour, a focus move that self-corrects on the next status change. The third looks like the feature is simply broken.

## Suggested direction

Two independent things, either useful alone:

[...]

Issue #393, opened 2026-08-06
PRD: A command-mode-scoped, deck-global command-entry lock — and focus that follows the lock


prageethw:
**PRD**: `prds/393-command-mode-lock-and-lock-governed-focus.md` (fork branch `prd-393-command-mode-lock-and-lock-governed-focus`)

Two changes to the Orchestration-tab command-entry lock and the focus behaviour on top of it. Both reverse decisions recorded in #374 and #373 respectively.

**Read this first if you are coming from upstream**: neither #373 nor #374 exists on `upstream/main` — both were closed here as not-planned and continued on the fork. Verified: `command_entry_locked`, `auto_focus_*`, `ToggleOrchestrationLock` and `gate_pane_input_key` all have **zero** occurrences on `upstream/main`. So this issue describes fork behaviour and there is nothing here to port. See "Upstream" at the bottom.

## Why the lock exists at all

#374 built the lock and described *what* it does, but never recorded *why* anyone would want it — which is very likely part of why it was declined, and which leaves a reader seeing only an odd restriction on typing. Recording it now, because everything else here follows from it.

**An orchestration deck is not a set of terminals that happen to share a window. It is one workflow with a single coordinator.** When every pane accepts human keystrokes, the human becomes a second, uncoordinated actor inside that workflow. Interrupting a worker directly — typing into it, answering its question, redirecting it — mutates state the orchestrator believes it owns, and the orchestrator has no way to learn it happened. The deck still looks coherent; the model behind it has quietly diverged.

The second half is about the human. **Open panes invite a reflex.** A prompt appears in a worker pane and the natural response is to answer it right there, because the cursor is already available and the question is in front of you. That reflex bypasses the coordinator and commits before the wider context is considered. **The lock converts that reflex into a decision** — the deliberate `Ctrl+D`, `Ctrl+E` is not overhead, it is the pause. The friction is the feature. A lock you must remember to engage protects nothing, so locked-by-default is load-bearing rather than an opinionated default.

**Scope note, because it is the obvious first objection**: this is Orchestration tabs only. Dashboard and Mode tabs are untouched. Nothing becomes read-only.

**Boundary, settled deliberately**: the lock's subject is the *unsolicited* interruption — typing at an agent that is working, while the orchestrator believes it owns that state. An agent that has stopped and asked is a different case: the workflow is already blocked on a human, so answering is a response to a request, not an intrusion into one. Hence decision 4 below.

## Problems

[...]

Issue #243, opened 2026-07-28
Pre-prompt readiness signal: retire the load-bearing 30s SESSION_START_WAIT_TIMEOUT and 1000ms DELEGATE_READINESS_BUFFER


vfarcic:
Follow-up from PRD #225 (PR #237). This is the fix #225 deliberately scoped out, and it is load-bearing.

**Covers two agents, not one.** Originally filed for Codex (wrapper strategy); #146 established that OpenCode (plugin strategy) has the identical defect from a different cause, and closed into this issue. See "Scope: which agents this covers" below — the two need *different* mechanisms, and a fix that only does the wrapper half leaves OpenCode paying the full timeout.

## The problem

For Codex the delegate readiness gate **never fast-paths**. codex-cli 0.145.0 posts its native `SessionStart` when the first *turn* starts — i.e. **after a prompt is submitted** — not when the TUI is ready. Measured on a live delegate:

| time | event |
|---|---|
| T+0.000s | `SessionStart` `origin=wrapper_fork` (respawn; the fork-time card event) |
| T+29.999s | `SessionStart` `origin=-` (native — exactly the 30s `SESSION_START_WAIT_TIMEOUT` fallback) |
| T+30.004s | `Thinking` `user_prompt=Y` (the injected prompt's own `UserPromptSubmit`) |

So the signal the gate waits for is a *consequence* of the prompt it is gating. Every `clear = true` Codex delegate pays the full 30s timeout and then injects.

### The same defect on OpenCode (from #146)

OpenCode emits **no** pre-prompt readiness signal on its plugin bus at all: `session.created` (→ `EventType::SessionStart`, `src/hook.rs:371`) fires when the prompt is accepted, ~16ms *after* submission. Measured on `opencode 1.18.16` — a 35s idle cold boot produced zero `session.*` events. So the gate cannot fast-path for OpenCode either, and a cold OpenCode spawn pays **30s** (scheduler, `src/spawn.rs:806`) or **31s** (delegate `clear = true`, `src/state.rs:2143` plus the 1000ms buffer). Full evidence, method and version pins in [#146's triage comment](https://github.com/vfarcic/dot-agent-deck/issues/146#issuecomment-5256620103).

Same shape as Codex, different cause: Codex *has* a native `SessionStart` that arrives too late; OpenCode has none to wait for.

## Why it matters

[...]

PR #416, opened 2026-08-08
Add visual agent control deck

KeerthiK12:


vfarcic:
Hi @KeerthiK12 — I saw you closed this, and I want to make sure that wasn't because it looked like nobody was paying attention. That was on us, not you.

Two things went wrong on our side: the only thing posted here for the first hour was a bot summary, and the CI runs never actually started — GitHub holds workflow runs on fork PRs from first-time contributors until a maintainer approves them, so those checks were waiting on me, not failing on you. I'm sorry that's the experience you got for a contribution this size.

**I've now reviewed it properly, and it's good work.** Some specifics:

- **No security issues.** I went through the whole diff carefully, since I don't merge first-contribution code unread. No network calls, nothing reading credentials or environment secrets, no obfuscated payloads, every npm and Cargo dependency resolving from the default registries, and no install hooks. The Rust side is careful in ways I noticed and appreciated: the webview can't supply an executable path, `verify_endpoint_trusted` gates the socket, POSIX quoting rejects NUL, the Windows guard rejects a crafted `StartWorkflow` IPC before spawning anything, partial workflows roll back, and there's a test asserting daemon error text doesn't leak into the UI.
- **The fast tier is green** — 3,475 tests pass.
- **The two e2e failures are not yours.** `sigterm_001` and `sigterm_002` fail on clean `main` in my own checkout too. That's a pre-existing daemon bug I'll file separately.
- **The architecture is right.** Zero daemon changes, `PROTOCOL_VERSION` untouched, `src/ui.rs` behaviour-identical — a real client of the existing protocol rather than a fork of it, which is exactly what PRD #176 asked for.

For context on why this landed on a PRD that says "Not Started": I'd started a desktop GUI myself months ago and stalled. That branch is now six weeks stale and hundreds of commits behind, in files that have changed heavily since — it isn't worth rebasing. **Yours is current and further along on the parts users actually see, so I'd rather build on it than on mine.** I'll port the design notes from my branch into the PRD so you have the reasoning behind some of the decisions there.

If you're up for reopening, here's what I think it needs to merge — all small:

1. **`cargo fmt --check` fails** — import ordering in `desktop/src-tauri/src/daemon_bridge.rs:10`. One `cargo fmt` run.
2. **Revert `.dot-agent-deck.toml`.** I understand why you changed it — the comment you added says it, that devbox-wrapped role commands aren't usable when launched from the GUI. That's a real problem and I'm glad you spotted it. But that file is the project's shared orchestration config, so switching every role to a specific CLI and model changes the setup for everyone working on the repo, and it isn't something the desktop PR should decide. Let's revert it here and solve the underlying issue properly — either the desktop resolves commands the same way the TUI does, or we make the launch environment explicit. Happy to open a separate issue for it if you'd rather not carry it.
3. **Get the new crate under CI.** Nothing here uses `--workspace`, so `clippy` and `nextest` never see `desktop/src-tauri` — about 2,850 lines of Rust and all its tests are currently unchecked. Either `--workspace` or explicit `-p dot-agent-deck-desktop` steps.

And two small additions I'd love, if you're willing:

4. **A visible "early developer preview" note in the app chrome.** Nothing on screen currently says the app is incomplete, so the known gaps read as bugs. In fixture mode it should also be explicit that none of the data is real.
5. **Surface fixture mode in the README.** `pnpm dev` in a browser with no Rust, no daemon and no API cost is a genuinely great way to show this off, and right now it's buried in a developer doc that isn't published.

After that I'd like to keep going in separate PRs rather than one big one — the app being opt-in and out of the release artifacts means an incomplete version on `main` costs nothing, and small PRs are far easier for both of us. Roughly what I have in mind:

- **Protocol crate extraction** — I'll take this one. It touches the TUI's internals and it's the risky piece. Worth doing early, before the desktop grows more dependencies on the main crate.
- **Keyboard navigation / TUI parity** — I'll write up the decisions I'd already made, if you want to implement.
- **Decks and tabs, live-mode model/cost fields, profile write-back** — self-contained in `desktop/`, yours if you want them.
- **The agents-communication graph** — its own PRD; it needs new daemon events, so it's the one piece that isn't free.

No pressure on any particular item — I'd just like you to know the work is wanted and has somewhere to go.

And if something else was behind closing it — scope, time, or it just not being fun — that's completely fine, and I'd still rather hear it than guess. Thanks for building this.

KeerthiK12:
@vfarcic Reopened — thanks for this. Good to know the silence was an approval gate and not a pile of red.

All three blockers are done:

- `cargo fmt --check` is clean.
- `.dot-agent-deck.toml` is reverted to main. You're right that it wasn't this PR's call — I'd been using it as a local workaround and shouldn't have carried it up.
- CI now runs clippy and nextest with `--workspace` (plus Tauri's Linux deps), so the desktop crate is actually gated. That found a pre-existing clippy failure in `xtask/linkage-check`, fixed here — invisible for the same reason the desktop crate was.

Both additions are in too: a persistent PREVIEW tag in the chrome, the fixture banner now says outright that all the data is simulated, and the README has a Desktop GUI section leading with fixture mode (`pnpm dev` + `?fixture=1` — no Rust, no daemon, no cost).

Gates locally: nextest `--workspace` 1613/1613, clippy clean, fmt clean, desktop tests and build green.

Heads-up: the branch picked up some desktop-only work since the close — a readable view for long CLI output, per-agent message boxes, and terminal rendering fixes. All in `desktop/`, but it's a bigger diff than what you read.

One thing on the TOML worth passing on. I ran the deck against itself for a day and hit your problem from the other side: delegation *respawn* reads role commands from `.dot-agent-deck.toml`, while a GUI launch builds them from the profile fields. When those disagree, workers start fine and then die on the first handoff, silently. Same root cause you're describing.

Same dogfooding turned up three small daemon-side things — happy to file any of them as issues or PRs:

- agents can't find `dot-agent-deck` by bare name when the daemon runs from `target/`, and the delegation protocol tells them to call it that way, so delegations are `command not found`;
- `delegate --to <role>` exits 0 and prints nothing when no pane exists for that role, so a whole run can go into the void with no signal (that one took me a day to find);
- the post-respawn readiness buffer defaults to 1s, under Claude Code's boot time, so task prompts get swallowed.

On the roadmap: yes to decks/tabs, live-mode model/cost fields, and profile write-back, and I'd like keyboard nav once your notes are up. I'll stay out of the protocol crate. On the agents-communication graph — I have a rough version running locally (daemon emits dispatched/delivered/failed/work-done, GUI draws it as a live chain with failures shown in place). It's where those three findings came from. It bumps `PROTOCOL_VERSION`, so it probably belongs after your protocol work, but the event shape is there if it helps the PRD.

Thanks again for the review, especially the security pass.

prageethw (commented):
Hi @KeerthiK12 — apologies this sat so long without a review. That's on us, not you.

This is **Part A of two**: everything *outside* `desktop/`. A second review of the `desktop/` application itself is in progress and will follow. Read-only, at `ca45d1ee`, compared against `upstream/main` = `0339366d`. No tests run, no `npm`/`pnpm`/`tauri` and no `desktop/` script executed.

**Verdict: approve the direction; needs a rebase and one careful conflict resolution before merge.**

The important thing to say up front: **the blockers below are almost entirely staleness, not defects in your design.** The branch is about a month behind `main`, and in that month upstream independently landed two of the very things this PR proposes. Your reasoning was right; you just got there in parallel.

---

## What's done well

Worth stating plainly, because a diff-shaped review would skip over it:

- **The workspace isolation is exactly right.** `Cargo.toml` gains one `members` entry and *nothing else* — no new root dependency. With no `default-members` and a root package present, a plain `cargo build --release` still targets the root package alone, so **the shipped binary never pulls Tauri**. That's the correct shape and it was clearly deliberate.
- **The dependency addition is clean.** All 180 new lockfile packages resolve to crates.io; **no** git or path sources beyond the workspace's own five. Nothing unusual in the set — the one crate that might raise an eyebrow on a supply-chain skim, `bs58`, arrives transitively via `serde_with`.
- **The rule 9 (experimental flag) question was asked and its answer recorded properly.** `prds/176-desktop-gui.md:100`, Design Decision #6, explains why the flag is N/A for a separate binary and notes *"Recorded per CLAUDE.md permanent instruction #9; the user chose 'inherently opt-in'."* That's the process being followed, not skipped.
- **Rule 12 is correctly not triggered** — and I verified rather than assumed. `daemon_protocol.rs`, `daemon.rs` and `hook.rs` are untouched. The desktop app is a fourth *client* of an existing wire, so no `PROTOCOL_VERSION` bump and no `.breaking.md` is right.
- **Rule 10 fully satisfied** across all added Markdown. I checked every added non-code line.
- **`docs/develop/desktop-gui.md` is unusually honest documentation.** The "Security and ownership model" and "Current milestone limits" sections state what the preview *cannot* do — Windows launch blocked, Pi coordinator blocked, model/cost/token fields deliberately shown as unavailable rather than guessed, and *"Do not treat the fixture as proof that a real agent or daemon path works."* Documentation that volunteers its own limits is rarer than it should be.
- **The fixture-mode on-ramp is genuinely good.** `pnpm dev` → `?fixture=1`, no Rust, no daemon, no API cost. I verified the port — `desktop/vite.config.ts:8` sets `port: 1420, strictPort: true`, so the documented URL is accurate.
- **Doc placement follows rule 11 exactly** — `docs/develop/`, linked from `CONTRIBUTING.md`, with the README link being the explicitly-permitted optional one.

---

## Blockers

### B1 — resolving the `ci.yml` conflict toward this PR would regress `main`

**This is the single thing that must not be got wrong.**

```yaml
# this PR, ci.yml:126
- run: cargo clippy --workspace -- -D warnings

# upstream main, ci.yml:170
- run: cargo clippy --workspace --all-targets --features e2e -- -D warnings
```

- When you cut the branch the line read `cargo clippy -- -D warnings`, so adding `--workspace` was a strict improvement.
- Since then upstream landed **#407** (`--all-targets --features e2e`) and **#436** (`--workspace` — the same flag you added).
- A merge tool sees "both sides changed this line". A human resolving in the PR's favour gets `--workspace` but **loses `--all-targets --features e2e`**.
- That is not cosmetic. Upstream's comment at `ci.yml:142-148` records why: every `tests/e2e_*.rs` opens with `#![cfg(feature = "e2e")]`, so without both flags those **64 files compile to empty crates and are type-checked by nothing**. Upstream verified it by injecting an `E0308` into `tests/e2e_handshake.rs` and watching bare `cargo clippy` report clean.

**Fix:** on rebase, keep upstream's line verbatim and **drop this PR's clippy edit entirely**. Your intent is already satisfied by `main`; nothing is lost.

### B2 — the branch is ~1 month behind `main`; six files conflict

Merge base is `4c7fa7b8`. `git merge-tree --write-tree --name-only upstream/main HEAD` reports content conflicts in exactly six files:

| File | Resolution | Mechanical? |
|---|---|---|
| `.github/workflows/ci.yml` | Take `main`'s clippy line; re-apply only the `apt-get` step and the nextest decision | **No** — see B1, B3 |
| `CLAUDE.md` | Renumber 14/15 → 15/16 | Yes — B4 |
| `Cargo.lock` | Regenerate | Yes |
| `README.md` | Re-apply the Desktop GUI section at the new insertion point | Yes |
| `src/ui.rs` | Re-apply the visibility widening onto main's current `ui.rs` | Yes, but re-verify |
| `xtask/linkage-check/src/clean_tmp.rs` | **Drop your hunk** — main already has the identical fix | Yes |

**Rebase rather than merge**, so the phantom `src/state.rs` hunk and the duplicated `clean_tmp.rs` fix drop out cleanly.

### B3 — `cargo nextest run --workspace` needs an explicit decision

`ci.yml:146`. This is the one flag upstream did **not** adopt: #436 widened *clippy* to `--workspace` and deliberately left `cargo nextest run` unwidened. So this is genuinely new, on the job every PR waits on, and it deserves deciding on its merits rather than slipping in as conflict fallout.

It newly *runs*:
- **48 xtask test functions** no gate has ever executed — `clean_tmp.rs` (6), `list_tests.rs` (14), `main.rs` (16), `xtask/docs/src/lib.rs` (12);
- the desktop crate's Rust tests, which is the actual goal.

I checked the deletion-path risk specifically, since `clean_tmp.rs` reaps directories: **its tests are hermetic** — they build a `tempfile::tempdir()` and call `collect()` against that scoped root, never the real `std::env::temp_dir()`. So there's no danger here, only a decision.

**Suggested:** either **(a)** land `--workspace` on nextest as its own small PR against `main`, so it goes green independently of a 13k-line feature PR, or **(b)** narrow this one to `cargo nextest run -p dot-agent-deck -p dot-agent-deck-desktop`. (a) is better for the repo; (b) is smaller for this PR.

### B4 — `CLAUDE.md` rule numbering collides

`CLAUDE.md:39-40` adds rules **14** and **15**. The merge base ended at 13 — but upstream has since added **its own rule 14**, *"Never Put Build Artifacts or Worktrees on the Agent Scratchpad / a tmpfs"* (`main`'s `CLAUDE.md:57`). Both sides define a 14.

**Fix:** renumber to **15** and **16**.

Worth a cross-reference while you're there: upstream's rule 14 and your rule 15 are complementary advice about worktree placement. Yours should point at rule 14's `../<repo>-<suffix>` disk-backed sibling convention rather than proposing `../<repo>-agents` in isolation — they agree, and saying so once stops them drifting apart.

---

## Two notes for anyone else reviewing this

- **`src/state.rs` in the GitHub diff is a phantom.** The file at PR head is **byte-identical** to current `main`. It shows as +577 lines only because GitHub diffs against a stale merge base, and it's upstream's own work (issue #398, PRD #393). It vanishes on rebase. Skip it — I nearly spent real effort on it.
- **CI cost is real but smaller than it looks.** `rust-cache` amortises the Tauri compile. The parts worth changing are the **uncached `apt-get` on every run** and the fact that a **cross-platform desktop crate is linted on Linux only** — `build-macos` and `build-windows` stay narrower by design, so desktop Rust gets one platform's lint.

---

## Also checked and clean

`Cargo.toml` isolation · lockfile source hygiene (all crates.io) · rule 12 non-triggering (verified, not assumed) · rule 10 across all added Markdown · rule 11 doc placement · rule 9 decision recording · `.gitignore` correctness · `prds/176-desktop-gui.md` conventions.

Part B — the `desktop/` application itself, the daemon seam, the Tauri command surface and the front-end — is being reviewed separately and will follow in its own comment.

prageethw (commented):
@KeerthiK12 — **Part B of the review**: the `desktop/` application itself (49 files). Part A covered everything outside it. Read-only at `ca45d1ee`; nothing was executed — no `npm`/`pnpm`/`vite`/`tauri`/`vitest`, no scripts.

**Verdict: don't merge as-is — but this is good work, and the blockers are small and local.**

Two defects make a **live** deck display **false** state. Both are contained fixes in one file each. The Rust/IPC layer is genuinely well built and I found nothing exploitable in it. The main structural problem is **size**, not quality — see the split proposal at the end, which is probably the most useful thing in this comment.

---

## What's done well

Stated first because on a 13.5k-line PR this is information, not courtesy.

- **It reuses the real protocol instead of reimplementing it.** `daemon_bridge.rs:5-12` imports `daemon_attach`, `daemon_client`, `daemon_protocol`, `platform::ipc` from the crate; `hello()` (`:118`) sends a genuine `AttachRequest::Hello` with the real `PROTOCOL_VERSION`. This really is a fourth client of the wire, not a fork of it.
- **Endpoint trust is verified before connecting** — `daemon_bridge.rs:139` calls `platform::fsperm::verify_endpoint_trusted`, the same gate `daemon_attach.rs:219` uses. Easy to omit; not omitted.
- **The Tauri capability surface is tight.** `capabilities/default.json` grants `core:default` and nothing else — no shell plugin, no fs plugin, no dialog plugin. CSP is `default-src 'self'` with `connect-src` limited to `ipc:`. Every command is additionally scoped to the main webview via `ensure_main_webview`. I specifically looked for a command that shells out with front-end input and **there isn't one**.
- **Validation reuses the daemon's own validators** rather than inventing parallel ones — `is_valid_cwd`, `is_valid_display_name`, `is_valid_pane_id_env`, `is_valid_orchestration_cwd`, `PTY_RESIZE_DIM_MAX`. `mint_desktop_pane_id` even `debug_assert!`s its output passes the daemon's validator.
- **"Replace daemon" is guarded three times over.** I went looking for an unguarded destructive button and did not find one: the button only renders when `runningAgentCount === 0` (`App.tsx:357`), `requestRestartDaemon` re-checks and returns early (`:228`), and `run_daemon_stop(force=false)` independently refuses with `StopError::LiveAgents`. The UI never sends `force: true` anywhere.
- **Delivery semantics are kept honest.** `action_result_ok` (`lib.rs:599`) treats only `Applied`/`Queued` as success, so `Stale`/`WrongSession`/`HistoryOnly`/`NoLiveTarget` correctly surface as *not delivered* — with a test enumerating all of them. Many GUIs would have reported "sent".
- **The workflow launcher is unusually rigorous for a preview.** Subscribes to readiness *before* the first spawn so a fast `SessionStart` can't be lost (`lib.rs:498`); reconciles a lost `StartAgent` response by pane+orchestration identity before rolling back, so a spawned agent isn't leaked because its id never arrived (`:526-547`); rolls back in reverse order and reports cleanup uncertainty distinctly.
- **Per-agent live/fixture honesty in the UI is meticulous** — TOKENS/COST/CONTEXT only in fixture, `LEASE · UNAVAILABLE` in live, a distinct caption on every empty panel (*"…is not exposed by the daemon"*), `SPEND` as `—`. Somebody thought hard about not inventing numbers. B2 below is the one place that discipline slipped.
- **Type hygiene:** zero `any`, zero `as any`, zero `@ts-ignore`/`@ts-expect-error` across ~28 files. The one `eslint-disable` is a justified `no-control-regex`. Every `catch {}` is a guarded `localStorage`/xterm path with a comment explaining the swallow.
- **Performance was measured, not guessed** — the snapshot-coalescing rationale and the PTY-bytes-bypass-React-state decision both read like someone watched it be slow and fixed it.

---

## B1 (High) — an `unknown` status renders a live agent as **stopped** and locks its terminal read-only

`desktop/src/lib/bridge.ts:116-124`:

```ts
function statusFromDaemon(status: string): AgentStatus {
  ...
  return "stopped";        // <-- "unknown" lands here
}
```

- `dto.rs:241` serialises `SessionStatus::Unknown` as `"unknown"`. It matches no branch and falls through to `"stopped"`.
- `AgentTile.tsx:204` then sets `readOnly` for `stopped` — so **a live, running agent gets a terminal the user cannot type into.**
- **Trigger:** a user runs a newer daemon than the desktop app, which emits a status string added after this build. Per PRD #162 that needs **no protocol bump** — `#[serde(other)]` exists precisely so a new status doesn't break older clients — so `PROTOCOL_VERSION` matches, the handshake succeeds, the deck connects normally, and affected agents silently grey out as stopped. They are running fine.
- **The crate documents the opposite in as many words**, `src/state.rs:214-221`: *"`Unknown` only ever originates from an unrecognized wire value on a newer daemon. **Rendered neutrally (like `Idle`) so it never masquerades as an active state.**"* The TUI honours it — `src/ui.rs:15995` → `("Idle", style)`. The desktop is the only client rendering it as a terminal state.

**Fix:** map `"unknown"` → `"waiting"` (closest to the TUI's neutral Idle), and never let an unrecognised status imply the agent is dead. Fold in with M3 so the default case can't recur.

## B2 (High) — live mode shows fixture telemetry, with no DEMO DATA banner, whenever the daemon is unreachable

`desktop/src/hooks/useDeckRuntime.ts:11-18`:

```ts
const initial = createFixtureSnapshot("empty");
return { ...initial,
  worktree: mode === "live" ? "No active project" : initial.worktree,
  connection: { status: "loading", message: "Connecting to local daemon…" } };
```

- `createFixtureSnapshot("empty")` empties `stages`/`agents`/`evidence` but **not the scalars** — it still returns `runId: "run_7f24a"`, `branch: "codex/visual-control-deck"`, `elapsed: "16:42"`, `currentNode: 4`, `totalNodes: 6` (`fixture.ts:271-281`).
- Those render **ungated** in live mode: `App.tsx:315` `{snapshot.runId}`, `:316` `{snapshot.branch}`, `:320` `NODE 04/06`, `:322` `ELAPSED 16:42`. `SPEND` **is** gated at `:323` — so the instinct was there, it just didn't reach these four.
- **It isn't a startup flash.** Both error paths spread `current` rather than resetting (`:59-63`, `:87`), so with no daemon running **the fixture values are the steady state**. The user sees an error banner *and* a top bar reporting a branch, run id, elapsed time and node count that were invented.
- **The safety net can't fire:** the DEMO DATA bar (`App.tsx:345`) is keyed on `mode === "fixture"`, not on where the data came from.

This is exactly what your own `docs/develop/desktop-gui.md` warns against — *"Do not treat the fixture as proof that a real agent or daemon path works."*

**Fix:** give live mode its own honest empty seed — `runId: "—"`, `branch: "Unavailable"`, `elapsed: "—"`, `currentNode: 0`, `totalNodes: 0` — built literally in `useDeckRuntime`, leaving `createFixtureSnapshot` reachable only from `FixtureDeckBridge`. Belt and braces: key the DEMO DATA bar on snapshot **provenance** (the DTO already carries `source`) rather than on `mode`.

---

## Medium

- **M1 — fixture state leaks into live through unnamespaced `localStorage`.** `useProjects.ts:4`/`:45-56`. Run the fixture once: the seed guard is only `defaultCwd.startsWith("/")`, which the fixture's `/dev/active/dot-agent-deck-gui` satisfies, so a project is persisted under a key **not namespaced by mode**. Live mode reads it back, it becomes `activeProject`, and `App.tsx:432` passes `activeProject?.cwd` as the default working directory for launching **real agents** — a path that never existed. (Live mode's own guard correctly prevents seeding, so this is purely a fixture→live leak.) Namespace the keys by mode, or skip persistence in fixture mode. Same class applies to `WORKFLOW_STORAGE_KEY` and the prompt-library key.
- **M2 — no CI job runs the TypeScript tests.** `grep -rn 'vitest' .github/workflows/` → **no match anywhere.** Six files, ≈795 lines, never executed. And the Rust side is asymmetric: `ci.yml:146` runs `--workspace` on Linux, but `build-macos` (`:247`) and `build-windows` (`:280`) run a bare `cargo nextest run` — so the desktop crate's Rust tests **never run on macOS**, the primary target for a Tauri `.app`. Also `ci.yml:125` has no `--all-targets`, so the desktop crate's test code is never linted. Suggest a `desktop:` job running `npm ci && npx vitest run && npx tsc --noEmit`; if that's out of scope here, say so in the PR body — right now the test count reads as coverage CI doesn't provide.
- **M3 — `statusFromDaemon` substring-matches a closed 8-member union.** An exhaustive `Record<>` would have made B1 a **compile error** instead of a silent fallthrough.

---

## Split proposal — the most useful thing I can offer

13.5k lines is a lot to land at once, and the branch also needs a rebase (it's `dirty`, ~1 month behind). There's a clean seam:

- **First slice — the read-only deck:** `daemon_bridge.rs` + `terminal.rs` + `dto.rs` + the observe-only commands. This is bounded, it's the strongest part of the work, and **it contains all three blockers** — each a contained fix.
- **Second slice — the workflow launcher:** the ~520 lines at `lib.rs:65-585`, plus `profileCommands.ts` and `ConfigurationPanels.tsx`. This carries nearly all the remaining risk and most of the preview caveats.

Splitting also makes the rebase far cheaper, and gets the good part of this merged while the rest iterates.

---

## Coverage gap, stated so it isn't mistaken for a clean bill

**`profileCommands.ts` (208) and `ConfigurationPanels.tsx` (410) — POSIX-shell command construction from user-editable profile fields — were not reviewed in depth here.** That's the highest-value area left, and a dedicated pass on it is running now; I'll post it separately.

Also not read closely: `TerminalViewport.tsx`, `OutputReader.tsx`, `AgentComposer.tsx`, `daemonEvents.ts`, the six test files' assertions, `pnpm-lock.yaml`, and the PRD/docs line-by-line.

prageethw (commented):
@KeerthiK12 — **Part C**, the last of three. This is the narrow pass on the one area Parts A and B explicitly left uncovered: `profileCommands.ts` and `ConfigurationPanels.tsx`, where shell commands get built from user-editable profile fields.

**Verdict: clean. No injection found.** The quoting is correct for the shell that actually executes it, and it is applied on every generated path. Three low-severity items below, none blocking.

That is a good result and worth saying plainly — this is the classic place injection lives, and it holds up.

---

## First, where the text actually goes — because it changes what "clean" means

I traced this rather than assuming it, and it corrects a nuance in Part B. Part B said *"no Tauri command shells out with front-end input"* — that is **true**, but incomplete. **The daemon does.** The full chain:

```
ConfigurationPanels.tsx:348
  → App.tsx:271 (IPC)
  → DesktopAction::StartWorkflow
  → dto.rs:479 validate_command   (non-empty, ≤64 KiB, no NUL — that is all)
  → order_workflow_roles          (constrains role names/order/start marker only)
  → StartAgentOptions.command
  → agent_pty.rs:985
  → <$SHELL> -c "<string>"
```

Two things follow:

- **The command passes through verbatim.** The project TOML is *not* a command whitelist — `order_workflow_roles` constrains role names, ordering and the start marker, and nothing else.
- **So the quoting is load-bearing**, which is precisely what makes the clean result meaningful rather than incidental.

**Provenance is local-only.** Commands come from `DEFAULT_PROFILES` (a source constant) or the user's own typing in `localStorage`. I checked specifically: `bridge.ts:209` resolves live-snapshot profiles back to `DEFAULT_PROFILES`, so **the daemon never supplies a command**, and there is no import/export or network path. Residual risk is self-inflicted, so everything below is graded as robustness rather than a privilege boundary.

---

## What's done well here

- **`quoteShellExecutable` always quotes argv[0]** — no fast path, no exception. I checked whether that breaks agent-type detection and it does **not**: `event.rs:240` tokenizes quote-aware, and I traced the `'"'"'` case by hand.
- **The front-end byte limit mirrors the backend exactly** — same constant, same `>` comparison, same UTF-8 semantics — and it is pinned by a test. Duplicated limits usually drift; this one cannot.
- **Windows is refused explicitly**, because the quoting is POSIX-specific. Refusing a platform you cannot quote for is the right call and it is rarer than it should be.
- **`effort` and `permissionMode` are normalised against closed lists before interpolation**, so they cannot carry arbitrary text at all.
- **The existing test genuinely attacks the builder** — it fires `$( )`, `;` and a bare `'` at it and asserts exact output. That is a real quoting test, not a happy-path test wearing one's clothes.

---

## Low-severity findings

**1. Quoted for POSIX `sh`, executed by `zsh`.** `workflow_start_options` pins no `SHELL`, so the daemon's `$SHELL` runs it — `/bin/zsh` on macOS, which is the only platform this path allows. Single-quoting is safe in both, so this is not an injection. The divergence is `quoteShellWord`'s unquoted allowlist (`profileCommands.ts:99`), which permits a **leading `=`**. A model value of `=codex` emits `--model =codex`, and zsh's default-on EQUALS expansion rewrites it to a path — or aborts the launch with `zsh: codex not found`. **Expansion, not evaluation; no code execution.** Fix: forbid a leading `=`, or drop the fast path and always quote.

**2. `--effort` on the Anthropic path (`profileCommands.ts:147`) — unverified.** I could not check this against a real CLI (no-run prohibition, and nothing in-repo pins it). Flagging it because the failure mode is total rather than partial: an unrecognised flag makes **every** Anthropic role fail at startup, inside a PTY. Worth confirming against the CLI you target.

**3. `roleId` is unvalidated (adjacent, `useAgentProfiles.ts:13-23`).** The loader validates `id`, `role` and `command` as strings but not `roleId` — which `ConfigurationPanels.tsx:348` sends as `role`. A non-string value yields an opaque serde deserialization error with no correctable UI field, so the user gets a failure they cannot act on. Note Part B found a real fixture→live `localStorage` leak in a sibling area, so persisted state reaching a live path is a demonstrated pattern here, not hypothetical.

---

That completes the review — Parts A, B and C. Thanks again for your patience with how long this sat, and for a PR that documents its own limits honestly; that made all three passes considerably easier to do.

vfarcic:
@KeerthiK12 — you're owed an update, and the fact that you're getting it 19 days late is on us again.

Here's what happened. You wrote on 10 August that all three blockers were done, and you were right — they were. About seven hours later a three-part review landed with new findings. But it landed as GitHub *review* objects rather than comments, and the review request was removed the same day, so from where you're sitting this PR has shown no activity and no reply since your message. That's twice now that our process has looked like indifference to you. It isn't, and I'm sorry.

**We'd like this to land. The project would benefit greatly from having an alternative UI** — the TUI is the only way in today, and that's a real limitation, not a stylistic one. A desktop deck built on the existing daemon protocol is the right shape for it, and the read-only core of what you wrote is the strongest part of the work.

Where it actually stands, so you can decide with full information:

- Three reviews carry the findings — [Part A](https://github.com/vfarcic/dot-agent-deck/pull/416#pullrequestreview-4895606180) (everything outside `desktop/`, four blockers), [Part B](https://github.com/vfarcic/dot-agent-deck/pull/416#pullrequestreview-4895698076) (the `desktop/` app, two High), [Part C](https://github.com/vfarcic/dot-agent-deck/pull/416#pullrequestreview-4895751465) (the shell-command surface, low severity). Happy to distill all three into one flat checklist if that's more useful than three long reviews.
- Two things you can't see from your side: the four required CI checks (`build`, `build-macos`, `build-windows`, `security`) have **never run** on this PR, because they're `pull_request` jobs and fork runs need a maintainer to approve them — the only two checks you've seen are a `pull_request_target` labeler and an app, which is why it looked like CI was fine. So it reads as unmergeable no matter how good the code is. And the branch is now **406 commits** behind `main`, well past the "six files conflict" the review recorded.

So, plainly: **are you still interested in finishing this?** No wrong answer, and no deadline attached to the question.

If yes, Part B proposes a split that we think makes it much easier — first slice is the read-only deck (`daemon_bridge.rs`, `terminal.rs`, `dto.rs`, the observe-only commands), which contains all three of that review's blockers as contained fixes and makes the rebase far cheaper; the workflow launcher follows separately. You'd take whichever slice you want.

If you'd rather not, or you've simply moved on — say so and that's completely fine. We'd pick up the first slice ourselves with you credited as co-author, and you'd have lost nothing by telling us.

KeerthiK12:
@vfarcic @prageethw — still interested, yes. And an apology of my own to balance yours: I never saw the three review parts until your update pointed at them — they landed as review objects while I was watching the conversation tab, so from my side the PR looked quiet too. Having read them now: that's some of the most careful review my code has ever gotten, especially Part A verifying rule 12 rather than assuming it and Part C tracing where the quoted text actually executes. Thank you.

The branch is updated and mergeable again. What's in the push:

- **Merged current `main`** (the 406 commits). `src/` and `tests/` are byte-identical to main except the three one-word `pub` widenings Part A asked to re-apply. Daemon, protocol, and the shared `.dot-agent-deck.toml` untouched.
- **Part A**: B1 resolved exactly as directed — main's clippy line kept verbatim; my only `build`-job addition is the Tauri apt deps that `--workspace` now needs. B3 got decided by main itself in the meantime (#489). B4: rules renumbered 15/16, cross-referenced to your new 14 as suggested. The `SPAWN_TIME_READINESS_TIMEOUT` casualty of #243 became a desktop-local constant rather than a crate re-addition.
- **Part B**: both Highs fixed — `statusFromDaemon` is now an exhaustive `Record` with a neutral `waiting` fallthrough (an unknown status can't render a live agent dead anymore), and live mode has its own honest empty seed, so fixture scalars can never reach the live top bar. M1: all four localStorage keys are mode-scoped. M2: a new `desktop-web` CI job runs vitest + tsc/vite build. M3 folded into the B1 fix as you suggested.
- **Part C**: leading-`=` removed from the unquoted fast path (the zsh EQUALS case), `roleId` type-validated in the loader. On finding 2: `--effort` is confirmed against a live CLI — claude v2.1.226 accepts it, and six roles on my machine launch with it daily.
- Since the last review, desktop-only: a Reader view for long CLI output, per-agent message composers, and terminal rendering fixes (PTY bytes bypass React state; buffer trims no longer reset xterm scrollback). The handoff-graph UI ships fixture-only — its daemon events belong to the agents-communication PRD, as discussed.

Gates on my machine: `cargo nextest run --workspace` 3373/3378 — the five failures are `verify_pr_stream`, which fail bit-identically on a pristine `main` checkout here, so I believe that's my environment rather than either branch. Clippy `--workspace` clean, fmt clean, desktop 46/46 + build green. The four required checks still need a maintainer to approve fork runs.

On the split: I'm happy either way. The rebase cost that motivated it is already paid, so if it's easier to review as one PR now, it's ready; if you'd still rather land the read-only deck first, say so and I'll cut that slice — the Part B blockers inside it are already fixed on this branch either way.

And yes to the roadmap as before: decks/tabs, live-mode model/cost fields, profile write-back, and keyboard nav once your notes are up. Staying out of the protocol crate.

One process note for anyone else in my position: it took two rounds of mutual apology to discover we were each waiting on the other. If review parts land as review objects on a fork PR, a one-line conversation comment saying "reviews posted, see Files tab" would have saved us both nineteen days.

prageethw (dismissed):
Follow-up to my three-part review (2026-08-10, at `ca45d1ee`). Re-verified against the current head (`bf37b0b7`), which has since been rebased onto current main (406 commits) and pushed four commits explicitly addressing every finding:

**Part A** (outside `desktop/`): B1 — `ci.yml`'s clippy line now matches main's `--workspace --all-targets --features e2e` verbatim (confirmed the PR's own edit was dropped rather than merge-resolved over main's). B2 — resolved via the upstream-main merge. B3 — `cargo nextest run --workspace` now matches main's own current convention. B4 — CLAUDE.md rules renumbered to 15/16, no collision with main's own rule 14.

**Part B** (`desktop/` app): B1 (High, live agent renders as stopped) — `statusFromDaemon` now defaults unrecognized/`unknown` statuses to `"waiting"` via `?? "waiting"` rather than falling through to `"stopped"`. B2 (High, fixture telemetry leaking into live mode with no banner) — live mode now seeds an honest empty state (`runId: "—"`, `branch: "Unavailable"`, etc., commented as "PR #416 review B2") built independently of `createFixtureSnapshot`; both error paths preserve that honest seed rather than reverting to fixture scalars. M1 (localStorage leak) — all three storage keys (projects, prompts, agent-profiles) now go through a shared `modeScopedKey`. M2 (no CI for the TS suite) — new `desktop-web` CI job runs `pnpm test` (vitest) + `pnpm build` (tsc --noEmit). M3 (non-exhaustive status mapping) — addressed pragmatically via the safe default rather than a compile-time-exhaustive union; acceptable given B1's fix is now robust to any future/unrecognized status value.

**Part C** (shell command construction): low-severity finding 1 (leading `=` zsh glob-expansion) and finding 3 (unvalidated `roleId`) are both fixed with comments citing the finding directly.

One caveat for whoever merges this: the `CI`, `Docs`, and `aarch64 cross-build check` workflow runs on the current head show `action_required` rather than a real conclusion — as an external contributor's PR this needs a maintainer to approve the workflow run before those checks actually execute. Please don't merge on the strength of this approval alone until that gate has actually run and gone green.

vfarcic (approved):
Approving after a full review pass and a first-ever green CI run on this branch.

CI had never executed on this PR (fork PR held at `action_required`), and running it surfaced three real defects, all now fixed on the branch:

- **`build-windows`** — `icons/icon.ico` was missing; `tauri-build` requires it to compile a Windows Resource file.
- **`build-windows`** again — `error[E0277]: *mut c_void cannot be sent between threads safely`. Tauri requires a `Send` future for every `#[tauri::command]`, and `desktop_run_action` reaches `run_daemon_stop`, which holds a `PinnedProcess` across awaits by design. Fixed with a narrowly scoped `unsafe impl Send for PinnedProcess`.
- **`nix`** — `desktop/src-tauri` was absent from the flake's `lib.fileset`, so crane could not resolve the workspace. Verified with a real local `nix build`.

Plus two pre-existing repo issues found along the way: a rustfmt violation in this PR's own `lib.rs` (CI checks this), and an ungated `InheritedMasterFds` in `tests/wrap_io.rs` that made CLAUDE.md rule 2's mandatory clippy command fail on **macOS** for everyone — invisible to CI because Linux constructs the struct and `build-macos` runs bare `cargo clippy`.

And one gate gap worth noting: `windows-cross-check` was passing on the same commit where `build-windows` failed to compile, because its script ran `cargo check --tests` without `--workspace` — the same omission rules 2 and 5 document elsewhere. Now widened, with an `llvm-rc` shim so `tauri-winres` can run on a Linux runner.

Scope caveat recorded rather than fixed: the GUI is **local-daemon-only**. Remote attach works over a forwarded socket (measured end to end — agents listed, PTYs streamed, hook events arriving), but project resolution stays client-side, so a remote workflow launch reads the wrong `.dot-agent-deck.toml` and writes coordinator context to the wrong host. PRD #176 amended to say so; follow-ups filed as #741 and #742.

Merging as a starting point, with the remaining milestones tracked in #740–#746.

vfarcic:
@KeerthiK12 — thank you for this, and welcome. 👋

This is an excellent first contribution, and I want to say that clearly, because the review I left above is a list of defects and that is a misleading way to summarise a PR of this quality.

What stood out while going through it:

- **The daemon boundary is respected exactly.** No business logic leaked into the client, no HTTP or TCP listener opened, the same per-user IPC endpoint and the same `Hello` handshake the TUI uses. That is the hard part of adding a second front-end, and it was done right.
- **The safety instincts are the ones we would have asked for.** Refusing to attach across a build mismatch, refusing to recycle a daemon that owns live agents, bounding terminal input and launch commands, constraining dimensions, sanitising errors before they reach the webview, and blocking a launch path rather than guessing at unsafe quoting. Several of these anticipated problems rather than reacting to them.
- **The honesty in the UI is unusual and appreciated.** Rendering model, cost, token and lease fields as *unavailable* because the daemon does not expose a reliable value — instead of inventing something plausible — is the right call and not the tempting one.
- **The documentation carries its reasoning**, including `docs/develop/desktop-gui.md` explicitly warning not to treat the fixture as proof the daemon path works. That kind of self-skepticism in a doc is rare.

The things I fixed on the branch were almost all invisible from a Mac or Linux checkout: a missing `icon.ico` that only `tauri-build` on Windows asks for, an `E0277` reachable only via the MSVC target, and a Nix fileset entry. Finding them needed CI, and CI had never actually run on this PR — GitHub holds fork workflows at `action_required` until a maintainer approves them, so nothing was red because nothing had reported. That is on our side, not yours. Along the way it also surfaced two pre-existing problems in our own repo that had nothing to do with your work.

@vfarcic has opened a set of issues and PRDs from the review, covering where this goes next: #740 (release artifacts), #741 (connecting to a daemon anywhere), #742 (several daemons at once), #743 (light/dark), #744 (zoom), #745 (an agent overview as the landing screen) and #746 (a project logo). Several of them exist *because* this PR made the questions concrete — the remote-connection work in particular came out of testing your bridge against a real remote daemon and watching it stream live agents into the window.

**If you would like to keep going, we would be very glad to have you.** Any of those issues is a reasonable next step, and #743 or #744 are self-contained if you want something smaller to pick up. Comment on whichever interests you so we do not duplicate effort, and please open issues of your own for anything you noticed while building this — you have spent more time in this code than anyone else has.

Thanks again. 🚀

Issue #425, opened 2026-08-08
Write the worktree ownership marker at creation time, so the deck can recognise its own worktrees


prageethw:
Follow-up to #422.

## Problem

#422 adds `worktree reclaim`, which removes a worktree without asking only when the deck can **prove it created it**. Ownership is a marker file (`dot-agent-deck-owner`) in the worktree's own git metadata dir — `<repo>/.git/worktrees/<name>/`, which sits outside the working tree so it can never make the tree dirty, and which `git worktree remove` deletes along with the worktree.

#422 implements **reading** that marker. Nothing writes it.

The consequence is correct but hollow: every worktree the deck itself creates reads as *foreign*, so it can only ever reach the `ask` verdict, never `remove`. That is the right fail-safe default — unknown origin must resolve to foreign — but it means the unattended reclamation #422 exists for never actually triggers for the deck's own worktrees, which is most of the value.

## Proposal

Write the marker at worktree-creation time, so the deck can recognise its own work later. The natural site is the existing worktree-add path in `src/issue_dispatch_run.rs`; anywhere else the deck creates a worktree should do the same.

Requirements worth stating up front, because each one is a way to get this subtly wrong:

- **Write it into the worktree's git metadata dir**, resolved via `git rev-parse --git-dir` from inside the worktree — never anywhere in the working tree. A marker inside the tree makes every marked worktree permanently dirty, and therefore permanently unreclaimable: the gate would defeat itself. This was verified empirically, not assumed.
- **Best-effort, never fatal.** A failure to write the marker must not fail worktree creation or the dispatch run. The cost of a missing marker is one extra confirmation prompt later, which is exactly the fail-safe direction.
- **Idempotent**, so a re-created or re-attached worktree does not accumulate state.
- **Say what created it.** A marker that records only "the deck" is weaker than one naming the creating task or orchestration — the same argument #421 makes about recording a claimant rather than a bare claim. Worth deciding here rather than discovering later.

## Not in scope

Retro-marking worktrees that already exist. They stay foreign and get asked about, which is correct — the deck genuinely cannot prove it created them.

prageethw:
Field evidence for this, from a real cleanup session tonight (2026-08-10).

`dot-agent-deck worktree list` across five live worktrees returned **`OWNED=no` for every one of them**:

```
PATH        BRANCH                             PR      CLEAN OWNED VERDICT REASON
-fork-only  fork-only                          no_pr   yes   no    keep    ...
[...]

PR #427, opened 2026-08-08
feat: reclaim merged worktrees behind a PR-state + clean + ownership gate

### LINKED ISSUE #422: Reclaim merged worktrees automatically: gate on PR state + a clean tree, not git ancestry
## Problem

Worktrees accumulate. The convention is already "remove it once its branch is merged", but that is prose, so it is forgotten — measured on one working copy just now: **8 worktrees sitting on MERGED PRs, every one of them clean**, i.e. 8 that could have been reclaimed automatically with zero risk.

Once a PR is merged, the code is in `main`. There is nothing to lose and nothing to ask a human about. This should not need a person to remember it.

## The trap, because it is not as simple as it sounds

**"Is this branch merged?" cannot be answered with git ancestry**, and both failure directions are live on a real working copy today:

| Branch | PR state | `git branch --merged origin/main` |
| --- | --- | --- |
| `ci/codeql-action-v4` | **MERGED** | says **no** |
| `fix/102-path-taking-request-helper` | **MERGED** | says **no** |
| `ci/87-sha-pin-actions` | **MERGED** | says **no** |
| `chore/orchestrator-scratch` | **no PR** | says **yes** |
| `fork-only` | **no PR** | says **yes** |

Squash-merging is why: the branch's commits never enter `main`'s ancestry, so an ancestry check **under-detects 3 of the 8** genuinely merged trees and reclaims nothing for them.

The other direction is worse. Ancestry reports *merged* for branches that were simply never advanced past `main` — including `chore/orchestrator-scratch`, which has no PR at all and is an actively-used scratch worktree. A naive "git says merged, delete it" rule destroys that one.

## Proposal

Gate reclamation on **two** conditions, both required:

1. **The PR's state is `MERGED`**, queried via `gh` — not inferred from git ancestry.
2. **The worktree is clean** (`git status --porcelain` empty).

The clean check is not redundant with the merge check. A merged branch's worktree can still hold **uncommitted or untracked files that were never part of the PR** — files that are genuinely *not* in `main`, which is the one case the "the code is already merged" reasoning does not cover. Removing those is data loss.

**Remove the worktree; keep the branch.** The branch costs nothing and keeps committed work recoverable.

**Encode the gate in a tested command, not in documentation.** The whole reason those 8 trees accumulated is that the rule lived in prose, and the squash-merge subtlety above is exactly what someone working from prose gets wrong. A command makes the safety logic testable, and reduces the instruction to "run this" — which cannot be misremembered.

Shape suggestion: list reclaimable worktrees and remove them, following the conventions of the existing read-only status command (human-readable table plus versioned JSON, meaningful exit codes), with the removal explicit rather than implicit.

## Relationship to #236

#236 covers worktree removal policy broadly, including the destructive `git worktree remove --force` still on the tab-close path (`remove_worktree`, `src/issue_dispatch_run.rs:133`) — verified present, with no dirty check anywhere in `src/`.

This is deliberately narrower and does not overlap that hazard: it **only ever removes provably-merged, provably-clean trees**, and it adds no new removal path to the close flow. If #236's Phase 2 reclamation ever lands it would subsume this; filing separately so the smaller, strictly-safe win is not blocked behind the larger policy question. Worth keeping the two aligned so they do not diverge on the same code path.

prageethw:
Adds `dot-agent-deck worktree list [--json]` and `worktree reclaim [--yes]` — reclaiming git worktrees whose work is finished, behind a gate that is deliberately conservative.

Closes vfarcic/dot-agent-deck#422.

## Why

Worktrees accumulate. Measured on one working copy: **8 sitting on MERGED PRs, every one of them clean** — eight that could have been reclaimed with zero risk, and were not, because the convention to remove them lived only in prose.

## The gate

A worktree is removed **unattended** only when all three hold:

1. **Its PR's state is `MERGED`** — via `gh`, never git ancestry
2. **The tree is clean** — `git status --porcelain` empty
3. **The deck can prove it created it** — otherwise it *asks*

The worktree is removed; **the branch always survives.** Every kept worktree reports why.

## Why ancestry is not the signal

This is the part worth reading, because the obvious implementation is wrong in **both** directions, and both were observed on a real repository:

| Branch | PR state | `git branch --merged main` |
| --- | --- | --- |
| three genuinely merged branches | **MERGED** | says **no** |
| an active scratch worktree | **no PR** | says **yes** |

Squash-merging means a merged branch's commits never enter `main`'s ancestry, so an ancestry check **misses 3 of the 8**. The other direction is destructive: ancestry reports *merged* for a branch simply never advanced past `main` — including a scratch worktree with no PR at all, which a "git says merged, delete it" rule would destroy.

`worktree/reclaim/002` and `/004` pin both directions, and each **asserts a fixture precondition first** — that `git branch --merged` does (or does not) list the branch — so neither can pass by accidentally agreeing with ancestry.

## Cleanliness is not redundant with merged-ness

A merged branch's worktree can still hold uncommitted or untracked files that were never part of the PR — genuinely absent from `main`. That is the one case "the code is already merged" does not cover, and removing it is data loss. A dirty tree is kept even with `--yes`.

## Ownership, and why the marker lives where it does

Ownership is a marker in the worktree's **own git metadata dir** (`<repo>/.git/worktrees/<name>/`), not the working tree. That is load-bearing rather than incidental: a marker inside the tree makes every marked worktree permanently dirty and therefore permanently unreclaimable — **the gate would defeat itself.** It is also removed by `git worktree remove`, so it cannot go stale.

This PR implements **reading** the marker. Nothing writes it yet, so in practice every worktree currently reads as foreign and reaches *ask* rather than *remove*. That is the intended fail-safe default; writing markers at creation time is filed separately as #425.

## Fail-closed throughout

An unresolvable PR state, an unparseable remote, a failed `git status`, or an ambiguous PR match all resolve toward **keep**, never remove. `gh` is not invoked at all when the repository cannot be derived, so it can never silently infer the wrong one. Removal uses `git worktree remove` **without `--force`**, so git independently re-refuses a tree that turned dirty after the check.

## Two defects worth knowing about, caught after the first "all green"

The implementation initially passed its whole test suite while being **unable to reclaim anything in reality**:

- **no `--state`** — `gh pr list` defaults to `--state open`, so every merged PR resolved to "no PR" and was kept
- **no `--repo`** — from a fork checkout the query resolved against the *upstream* repository

Neither was catchable, because the test stub ignored unrecognised flags: it verified `gh` was *called*, never that it was called *correctly*. The stub now models `--state`, requires `--repo`, and rejects unknown flags. The corrected invocation was verified against real GitHub, not only the stub.

## Coverage

`worktree/reclaim/001`–`007`, fast tier: the real binary as a subprocess against real git repositories in tempdirs, with a synthetic `gh` on `PATH` — no PTY, no daemon, no network, no `e2e` gate.

## Provenance and what backs this port

Developed and merged on my fork (`prageethw/dot-agent-deck`), where it went through a reviewer and an auditor pass in addition to CI. Ported here on top of `upstream/main` and adapted: fork issue numbers, a fork-only docs path, and references to tests that do not exist in this tree were all removed.

**Update — the tests have now run here, in this repository's CI.** When this PR was opened, this section said plainly that no test run backed the port: a branch cut from `upstream/main` carries none of my fork's CI, so I had only local lint and build gates to point at and would rather have said so than let passing lint imply otherwise. That caveat no longer applies, and this replaces it.

The branch has since been rebased onto current `upstream/main` — `#### daemon/status` had landed at the same catalog anchor as `#### worktree/reclaim`, which was the only conflict — and this repository's own CI has executed the fast tier on the result:

| Job | Result | `worktree/reclaim` coverage |
| --- | --- | --- |
| `build` (Linux) | `1652 tests run: 1652 passed, 0 skipped` | `worktree_reclaim_001`–`007` plus the `worktree_reclaim::tests::*` unit tests, all PASS |
| `build-macos` | `1652 tests run: 1652 passed, 0 skipped` | same seven integration tests, all PASS |
| `build-windows` | `1298 tests run: 1298 passed, 0 skipped` | the integration tests do not run here, matching their declared `Platform coverage: mac+linux` |

`cargo fmt --check`, `cargo clippy --workspace --all-targets --features e2e -- -D warnings`, `cargo build --release`, `cargo xtask linkage-check`, `nix`, `devbox`, `security` and `windows-cross-check` are green alongside them.

One thing that is still **not** exercised: the L2 PTY/e2e tier, which this repository's CI excludes by design. That costs this PR nothing in particular — every test it adds is fast-tier and ungated, so none of its own coverage sits behind that exclusion.


prageethw:
Good catch, and I think it is worth more than P2 — this parser feeds a path to `git worktree remove`, so misparsing it is a wrong deletion target, not just a failed inspection.

Confirmed the mechanism: `git worktree list --porcelain` emits Git's *display* representation, which C-quotes paths containing quotes, newlines or non-ASCII bytes. The parser treated that display text as a literal filesystem path.

Fixing it with `--porcelain -z` and a byte-level parse: a path can never contain NUL, so the record boundary is unambiguous regardless of path content. On Unix the `PathBuf` is built straight from raw bytes via `OsStr::from_bytes` rather than a lossy UTF-8 round-trip, which also fixes non-UTF-8 paths that the old parser would have mangled even without quoting. Verified `-z` support against git 2.55.0.

Also adding a regression test for a worktree path containing a literal newline — precisely the input the old parser would have silently misparsed.

The fix is validated on my fork first and I will push it to this branch, so please hold off merging until it lands here.

Worth noting for anyone reading later: my own audit of this module examined path handling but scoped it to argv-safety — whether the path could be read as a flag or reach a shell — and never questioned whether the path had been parsed correctly to begin with. This review caught a class my audit was not looking at.

prageethw:
The parsing fix has landed on this branch (`853a974`) — **my earlier request to hold off merging is lifted.**

What changed in `src/worktree_reclaim.rs`:

- `git worktree list --porcelain` → `--porcelain -z`, so records are NUL-delimited. A path can never contain NUL, so the boundary is unambiguous regardless of path content.
- The parser now works on raw bytes rather than text. On Unix the `PathBuf` is built via `OsStr::from_bytes`, which also fixes non-UTF-8 paths the old code would have mangled even without quoting. On other platforms it falls back to a lossy conversion, which yields a *non-matching* path rather than a successful wrong target — it fails closed.
- Added a regression test for a worktree path containing a literal newline: exactly the input the old parser misread.

Verified `-z` against git 2.55.0 before relying on it, and confirmed the record grouping matches its actual output (NUL after every field, an empty field between records) rather than assuming the delimiter was the only difference.

Two things I checked because they would be serious if wrong, given this path is handed to `git worktree remove`:

- **The main worktree can still never be a removal candidate.** Exclusion relies on git listing the main working tree first, which is documented — and I confirmed it holds even when the command runs from inside a *linked* worktree, where the ordering might plausibly have differed.
- **Bare and detached worktrees** carry no `branch` field, so they resolve to unresolvable and are kept, not removed.

The fix was reviewed and merged on my fork first, where it is CI-green including the e2e tier. This branch carries no CI of its own, so what backs the port here is a clean build with `--features e2e`, fmt, clippy and linkage-check — I would rather state that than let the fork's green imply coverage this branch has not run.

Thanks for the catch. It was rated P2, but I think it earned more than that: a misparsed path here is a wrong deletion target, not a failed inspection. Worth noting that my own audit of this module examined path handling and scoped it to argv-safety — whether the path could be read as a flag or reach a shell — and never questioned whether the path had been parsed correctly to begin with. That is the gap this review closed.

vfarcic (approved):
Verified locally on Linux: fmt, clippy (`--workspace --all-targets --features e2e`), release build, fast tier (1653/1653, including all 18 new `worktree_reclaim` tests), linkage-check and windows-cross all green. The e2e failures in my run were unrelated — two flakes that passed on isolation rerun, and `idle_worker_011`/`focus_007`/`card_stats_005`, which fail identically on a clean `main` checkout and are already tracked in #460.

Also exercised the release binary directly against real git repos: the fail-closed paths hold (missing `gh`, non-GitHub remote, detached HEAD, locked worktree, dirty tree, open/no PR all keep), the main working tree is unreachable as a candidate, and the branch survives removal. Confirmed the `-z` porcelain parser against real git 2.55.0 output.

Thanks for the thorough treatment of the ancestry trap and the two `gh` invocation defects.

Issue #401, opened 2026-08-07
Hook-socket status is unauthenticated, and PRD #393 makes it authority-bearing: an agent can report status for a pane it does not own


prageethw:
Agent-reported status arriving over the hook socket is **unauthenticated**, and PRD #393 is the first change to make it authority-bearing — it now decides whether a human's keystrokes are delivered to a pane. Raised by a security audit of fork PR #51; the immediately-reachable defect is being fixed there, and this issue covers the underlying trust question, which is larger than one PRD.

## The chain

- `WaitingForInput` originates as `EventType::WaitingForInput` (`src/event.rs:7-19`). Native Claude/Codex `Notification` hooks map to it (`src/hook.rs:96-105`); `dot-agent-deck agent-event --type waiting` constructs a raw `AgentEvent` with `pane_id` from `DOT_AGENT_DECK_PANE_ID` (`src/main.rs:716-772`).
- The hook loop accepts newline-delimited JSON — first as `DaemonMessage`, then as a bare `AgentEvent` — with **no signature, capability, token or peer-credential check** (`src/daemon.rs:1153-1168`, `1229-1297`).
- `AppState::apply_event` validates only that `pane_id` is in `managed_pane_ids`, and auto-registers an unknown `SessionStart` to cover a startup race (`src/state.rs:3133-3166`). It does **not** verify that the sender owns the pane it names, nor bind the status to the registry's current agent generation.

## Why `0600` is not the answer

The socket is owner-only (`src/platform/ipc/unix.rs:120-130`), which excludes other Unix uids. It does not exclude *other processes of the same uid* — and **the daemon deliberately injects the socket path into every spawned child** as `DOT_AGENT_DECK_SOCKET` (`src/agent_pty.rs:2190-2199`, `2766-2777`). Every agent, and everything an agent launches, is therefore a writer by design. The endpoint is also predictable (`$XDG_RUNTIME_DIR/dot-agent-deck.sock`, else `/tmp/dot-agent-deck-<uid>.sock`, `src/platform/paths.rs:256-280`).

The consequence worth stating plainly: **an agent can report status for a pane that is not its own.** There is nothing in the protocol tying an event to the pane its sender actually occupies.

## What that currently buys

Before #393, agent-reported status drove presentation and focus — tab-label colouring (#333) and `auto_focus_waiting_pane` (#373). Wrong status meant a wrong colour or an odd focus move, both self-correcting.

[...]

Issue #373, opened 2026-08-04
Auto-return focus to the orchestrator pane (all-clear + 30s inactivity), Orchestration tabs


prageethw:
Split out from #361. When all WaitingForInput panes in an orchestration tab resolve, focus moves to the orchestrator's own pane; if the human manually moves focus elsewhere and 30s pass with no activity, focus snaps back. Developing fork-only; closing immediately, same as #361 — may reopen if it graduates to an upstream PR.

prageethw:
Closing for now — developing fork-only. Will reopen or file a fresh issue if/when it's validated and ready to raise as a PR here.

Issue #110, opened 2026-05-24
Fix session reuse conflicting with F9 clear=true respawn


vfarcic:
## PRD: Fix session reuse conflicting with F9 clear=true respawn

**Problem**: After a clear=true delegate respawn (PRD #92 F9), the new agent emits SessionStart with a fresh session_id and new agent_id, but the TUI's session-reuse logic in `apply_event` silently remaps it back to the old session — making it appear no new session started.

**Solution**: Track `agent_id` in `SessionState` and only skip session-reuse when the incoming agent_id matches the existing session's agent_id. Different agent_id means intentional respawn — create a new session.

**Detailed PRD**: See [prds/110-fix-session-reuse-vs-f9-respawn.md](https://github.com/vfarcic/dot-agent-deck/blob/main/prds/110-fix-session-reuse-vs-f9-respawn.md)

**Priority**: High

PR #443, opened 2026-08-08
fix(state): let an untagged hook event adopt the pane's session, not fork a rival card (#398)

### LINKED ISSUE #398: build_pane_status: duplicate pane_id sessions collide non-deterministically, so a pane can show Needs Input while the lock keeps dropping keys

vfarcic:
Fixes #398.

## The defect

An `AgentEvent` carrying `agent_id: None` matched neither path that would route it onto a pane's existing session, so it fell through and created a **second session sharing that `pane_id`**:

- `apply_event`'s reuse guard requires `session.agent_id == event.agent_id`, and `Some(x) != None`;
- the retire block immediately after skips `None` events by construction — PRD #110's deliberate carve-out so a pre-F9 hook cannot wipe a tagged session's history.

Nothing downstream dedupes. `build_pane_status` keys a `HashMap` by `pane_id`, so **which of the two statuses survived was decided by iteration order and could differ between runs**, and the deck rendered two cards for one pane.

Three features read that join — PRD #333 tab-label colouring, PRD #373 focus steering, and the pane borders. The sharpest symptom was PRD #393's command-entry lock: a worker pane could render `Needs Input` while the gate read the other, stale session and kept dropping the human's keystrokes, with a status message insisting the pane was locked.

The untagged shape is **not malformed input**. It is what PRD #110 preserves for pre-F9 hooks, and what any producer emits when `DOT_AGENT_DECK_AGENT_ID` did not reach it — a hand-written hook, a wrapper that scrubbed its env, or `dot-agent-deck agent-event` from a subprocess that lost it. Every daemon-spawned agent *does* get the var injected (`AgentRegistry::spawn_agent`), so this is the legacy/un-envied path rather than the default one. #401 (open) is how it can be triggered deliberately rather than accidentally.

## The fix

An untagged event now **adopts** the pane's existing session instead of minting a sibling.

PRD #110 framed this as a choice between a visible duplicate and silent data loss. It was a false dilemma: the retire block still skips `None`, and adoption only remaps *where the event is recorded*, so `recent_events` / `tool_count` / `first_prompts` / `started_at` are untouched and the card keeps its `Some` `agent_id`. Only the duplicate goes away.

Adoption requires **exactly one candidate**. An untagged event carries nothing that says which generation it belongs to, so on an already-ambiguous pane the guard declines rather than guessing — which would be the same coin-flip this fix removes.

`build_pane_status` keeps its plain `collect()` deliberately, now documented at the function: there is no tie left to break, and a tie-break rule there would read as if duplicates were still expected. `build_pane_status_for_gate` (#404) keeps its own fail-closed guard regardless — the lock's safe answer (deny) is the opposite of what a colouring consumer wants, and it must hold against a hostile status wire however clean this seam is.

## Tests

Four unit tests in `src/state.rs`, including `pre_f9_hook_with_no_agent_id_does_not_wipe_tagged_session` — **the regression test the retire block has cited by name for months without it ever existing**, so the half PRD #110 meant to guarantee was in fact untested until now. Plus L1 `dashboard/pane/010`, which drives the real `AppState` transition and asserts the grid draws one card.

Verified the new arm is load-bearing: with it disabled, the two behaviour tests fail while the two invariant tests still pass.

`cargo fmt --check` · `cargo clippy --workspace --all-targets --features e2e -- -D warnings` · `cargo test-fast` (1559 passed, 0 skipped) · `cargo xtask linkage-check` — all clean. `cargo test-e2e` running locally.

## Cross-version note (rule 12)

**No `PROTOCOL_VERSION` bump.** The wire shape is unchanged and this is purely TUI-side event folding — an older daemon paired with this TUI is unaffected, and the change alters no daemon behaviour, handler contract, or field meaning. Not a semantic break either: it changes how the TUI folds an event it already received, not what any field means. Bump policy therefore reads **patch**.

The manual old-daemon sandbox run has **not** been performed; flagging it rather than implying it was.


vfarcic:
Both P1 findings were valid and are fixed in b87cb3b. Both were regressions introduced by this PR's adoption fallback, not pre-existing.

## P1 — Terminal adoption erases session history ✅ fixed

Confirmed exactly as described. `SessionEnd` never reaches the status path — it hits the terminal branch, which removes `event.session_id` and rebuilds a bare placeholder. Adoption handed that branch the pane's *tagged* session, so an untagged `SessionEnd` destroyed the `recent_events` / `tool_count` / `first_prompts` that PRD #110's `None` carve-out exists to protect. My own commit message claimed the retire block skipping `None` made history safe; that was wrong, because the terminal branch removes sessions by a different route.

Adoption now requires `claims_generation` (hoisted above the reuse guard). Before this PR an untagged `SessionEnd` resolved to no session and was a silent no-op, and that is now restored precisely — so the fix cannot lose history on any path. This is also the same rule the retire block already applies one screen down: an untagged end cannot name a generation, so it is not evidence that this one ended.

## P1 / security — Pane-only adoption bypasses input lock ✅ fixed

Also valid, and the more interesting of the two. Before #398 an untagged `WaitingForInput` could not reach a tagged session at all — it minted a rival, and `build_pane_status_for_gate` then denied the pane for being ambiguous. So #404's fail-closed guard was incidentally blocking this attack, and removing the duplicate removed that protection, leaving a route to the carve-out needing only a pane id — a small integer — on the unauthenticated hook socket (#401). That is strictly easier to reach than the tagged path it replaced, so this was a real weakening even though #401 leaves the socket open regardless.

`AppState::untagged_status_panes` now records whether a pane's current status was written by a producer that named a generation, and the gate omits panes where it was not. Design notes:

- **A tagged frame clears the mark**, so one legacy event cannot poison a pane for the rest of the session.
- **`SubagentStart`/`Stop` leave provenance alone** — they assert no status, so they can neither launder an untagged mark away nor invent one.
- **Only the gate consults it.** Cards, borders and tab colours still display an untagged report: being unable to name a generation makes a status untrustworthy to *act* on, not wrong to *show*.
- **The duplicate rule stays.** It covers two *tagged* sessions, which provenance says nothing about.

Fails closed for legacy setups: a deck running entirely on pre-F9 hooks gets no carve-out and reaches its panes with `Ctrl+d`, `Ctrl+e` — the same trade the lock already makes for any agent that never reports `WaitingForInput`. The changelog fragment claimed the exemption would now apply *more* widely, which was wrong in the other direction; corrected.

## Tests

`pre_f9_hook_with_no_agent_id_session_end_does_not_adopt_and_wipe`, `untagged_status_marks_the_pane_and_a_tagged_frame_clears_it`, `subagent_frames_do_not_change_status_provenance`, and L1 `orchestration/lock/013` proving an unidentified `WaitingForInput` does not open the lock while an identified one still does.

Both new guards verified load-bearing — disabling either fails its test rather than passing silently.

`cargo fmt --check` · `cargo clippy --workspace --all-targets --features e2e -- -D warnings` · `cargo test-fast` (1563 passed, 0 skipped) · `cargo xtask linkage-check` (474 catalog ids) — all clean.

The prior `cargo test-e2e` run had one unrelated failure, `devin_live_001`, from Devin's backend (`failed to start ACP agent session: Failed to load team settings: Failed to fetch`); re-running the tier against these commits.

vfarcic:
## P1 / security — Provenance cleared by a frame that only preserved the status ✅ fixed in 72c8df9

Valid, and the subtlest of the three. `ToolStart` does not assert a status when one is already `WaitingForInput` — its arm deliberately *preserves* that value instead of overwriting it with `Working`. My provenance update was keyed on event type from outside the match, so a tagged `ToolStart` cleared the mark while the untrusted `WaitingForInput` it had declined to overwrite stayed on the card, and the gate then acted on exactly the status an unidentified producer had planted.

Cheap to exploit, too: plant `WaitingForInput` on the unauthenticated hook socket (#401), then wait for the pane's real agent to run any tool — the agent itself silently vouches for the planted status.

Provenance now tracks the writer of the **current** status rather than the last frame to arrive. Each match arm reports whether it asserted a status, and only an assertion moves the mark. Classifying from outside the match is what caused this, so the arms report for themselves and can't drift when one is edited.

This exposes an asymmetry worth naming, now pinned in both directions:

| frame | effect on an existing `WaitingForInput` | asserts? | mark |
|---|---|---|---|
| `ToolStart` | preserved | no | stands |
| `ToolEnd` | overwritten with `Thinking` | yes | cleared |

`a_tagged_frame_that_preserves_an_untagged_status_does_not_clear_the_mark` reproduces the laundering path; `a_tagged_frame_that_overwrites_an_untagged_status_clears_the_mark` holds the legitimate case so the fix isn't over-applied. Neither arm can be collapsed into the other without failing a test.

`cargo fmt --check` · `cargo clippy --workspace --all-targets --features e2e -- -D warnings` · `cargo test-fast` (1565 passed, 0 skipped) · `cargo xtask linkage-check` — all clean.

## e2e tier — resolved, no regressions from this PR

The full `cargo test-e2e` run against b87cb3b reported 4 failures, all in the real-agent tier. Investigated each rather than waving them off as flake:

- `devin_live_001`, `dispatch_013`, `delegate_015` — **pass on re-run** (`devin_live_001` in 12s versus the 125s timeout it hit before; its earlier failure was Devin's backend returning `Failed to load team settings: Failed to fetch`).
- `card_stats_005_real_agent_card_narrows_without_restructuring` — **fails identically on `main`**, which I ran as a baseline: 50.3s on `main` versus 50.7s here, same assertion. Pre-existing, not from this PR. The grid shows the real Haiku agent still mid-turn (`Running 1 shell command`, `Wrangling…`) with the sentinel not yet printed inside the 30s window; the card machinery itself worked, reaching `Idle` with `1 tools`.

Neither could plausibly be reached by this change in any case: adoption fires only for untagged events, and daemon-spawned agents always get `DOT_AGENT_DECK_AGENT_ID` injected, while the provenance mark is read only by the orchestration lock's gate — `card_stats_005` is a dashboard card test with no orchestration tab.

Issue #445, opened 2026-08-08
graduate-command-entry-lock: make the orchestration command-entry lock visible to all (PRD #393)


vfarcic:
The orchestration command-entry lock (PRD #393, merged in #404) is gated behind the `experimental` feature flag while it is evaluated in real use. This issue tracks the decision to graduate it — and the open questions that should be answered first.

## Why it was gated

It shipped un-gated and **locked by default**, and it is a sharp default: it changes a reflex users already have, silently drops keystrokes until a chord is learned, and it landed without CLAUDE.md rule 9's experimental-flag question being asked. It was still unreleased when the flag was added (not in v0.35.8), so nothing that ever shipped changed.

## To graduate

`grep show_command_entry_lock` finds every gate. Three seams in `src/ui.rs`:

1. the `Ctrl+E` binding resolution (expressed through the orchestration-tab term)
2. the `PaneInput` keystroke gate
3. the auto-focus chain that steers toward waiting panes

Then:

- delete the wrapper from `src/features.rs` and inline the `true` branches
- remove the row from `docs/develop/experimental-flag.md` (move it to **Graduated**)
- drop the experimental admonitions from `docs/keyboard-shortcuts.md` and `docs/orchestration.md`
- drop the flag paragraph from the changelog fragment, or write a new one if it has already been released
- remove `.with_env("DOT_AGENT_DECK_EXPERIMENTAL", "1")` from the five L2 tests in `tests/e2e_orchestration_lock.rs`, and **delete `orchestration/lock/014`** — it exists only to pin flag-off behaviour, so it becomes meaningless (and will fail) once the gate is gone

## Decide before graduating

**1. Should the unlock persist across restarts?** Today it does not — `command_entry_locked` starts `true` every launch. So a user who deliberately works alongside their agents re-pays the unlock **every session, forever**. The field docs justify the default with *"a lock you must remember to engage protects nothing"*, but that argument cuts both ways: a lock you must remember to *disengage* every session mostly teaches people to disengage it reflexively — at which point they have lost the focus steering and kept none of the protection. This is probably the single most important question here, and it is largely independent of the default.

**2. Should the default be locked at all, or should there be a config key?** e.g. `[orchestration] command_entry_lock = false` alongside the existing `[bell]` / `[dashboard]` / `[issue_dispatch]` tables. Note that flipping the default to unlocked also silently disables the focus steering, since that only runs while locked — so "default off" is a bigger change than it first appears.

[...]

Issue #139, opened 2026-06-07
PRD: Experimental feature flag (single-toggle gating for in-flight features)


vfarcic:
## PRD: Experimental Feature Flag

**Problem**: New features developed against `main` ship to anyone running `dot-agent-deck`, including the maintainer's day-to-day usage. There is no way to land work-in-progress code on `main` while keeping the user-facing surface hidden during testing.

**Solution**: A single `experimental` boolean flag, read from `.dot-agent-deck.toml` (with a `DOT_AGENT_DECK_EXPERIMENTAL` env override) and watched for live changes by both TUI and daemon. New user-visible surfaces (panes, fields, commands) are gated through per-feature wrapper functions (`features::show_<name>() -> bool`) so each feature stays greppable for graduation. Off by default; opt-in for testing.

**Detailed PRD**: See [prds/139-experimental-feature-flag.md](https://github.com/vfarcic/dot-agent-deck/blob/main/prds/139-experimental-feature-flag.md)

**Priority**: Medium

PR #446, opened 2026-08-08
feat(orchestration): gate the command-entry lock behind the experimental flag (#393)

vfarcic:
Gates PRD #393's command-entry lock behind the `experimental` flag so it can be evaluated in real use before it reaches everyone. Graduation tracked in #445.

## Why

The lock shipped un-gated and **locked by default** in #404, and CLAUDE.md rule 9's experimental-flag question — which explicitly covers a new keybinding — was never put. It is a sharp default: it changes a reflex users already have, silently drops keystrokes until a chord is learned, and the unlock is not persisted, so the cost is re-paid every session.

**Nothing that has shipped changes.** The lock is not in v0.35.8, so no released version ever exposed it and no user is disrupted.

## What is gated

One wrapper, `features::show_command_entry_lock()`, read at three live-loop seams in `src/ui.rs`:

1. **the `Ctrl+E` binding resolution** — expressed through the existing orchestration-tab term, so one place decides whether the chord is claimed at all
2. **the `PaneInput` keystroke gate** — with the flag off this also skips building the status join, since the lock is its only reader
3. **the auto-focus chain** that steers toward waiting panes

The third seam is the one worth naming. The focus steering is part of the same surface rather than a separate feature, because it only ever ran while the lock was engaged. Gating it too is what makes flag-off behaviour identical to v0.35.8, rather than "no lock, but the deck still moves focus on its own for a lock you cannot see or reach".

## What is deliberately NOT gated

`UiState::command_entry_locked` still starts `true`, and `scope_command_entry_lock` / `gate_pane_input_key` stay flag-free. The flag decides whether that state is **consulted**, not what it is — which keeps this a presentation switch per rule 9 / PRD #139 M3.2, and lets the helpers' unit tests keep exercising the real logic rather than the gate.

## Docs

Both published pages presented the lock as default behaviour, which the gate makes wrong. `docs/keyboard-shortcuts.md` and `docs/orchestration.md` now carry the house-style experimental admonition (matching `docs/scheduled-tasks.md`) and say how to enable it; `docs/develop/experimental-flag.md` lists the new wrapper.

Both pages also now state that unlocking leaves you in command mode, so returning to the pane takes a third keypress — `Ctrl+d`, `Ctrl+e`, `Ctrl+d`. The docs previously described a two-chord flow that stops one key short of actually typing. The status **message** is untouched; that fix is recorded in #445 rather than changing a contributor's copy on a feature that has not been evaluated yet.

## Tests

All five existing L2 lock tests opt the flag ON via `.with_env("DOT_AGENT_DECK_EXPERIMENTAL", "1")` — without it they would assert against un-gated behaviour and quietly test the wrong thing.

New **`orchestration/lock/014`** launches *without* the flag and pins the default: a keystroke reaches the focused worker with no unlock chord, no `Pane locked` message appears, and `Ctrl+e` is not claimed. So the gate is covered from both sides and a regression that ships the lock unconditionally fails here rather than reaching every user silently.

Verified load-bearing: forcing the wrapper to `true` fails `lock/014` (and it passes again on restore).

## Gates

`cargo fmt --check` · `cargo clippy --workspace --all-targets --features e2e -- -D warnings` · `cargo test-fast` (1601 passed, 0 skipped) · `cargo xtask linkage-check` (479 catalog ids) — all clean.

`cargo test-e2e lock_008 lock_009 lock_010 lock_011 lock_012 lock_014` — **6 passed**, including `lock_012` against a real Claude Haiku worker (not skipped; credentials present).

## Cross-version note (rule 12)

No `PROTOCOL_VERSION` bump. This touches no daemon behaviour, protocol, or hook handling — it is a TUI-side presentation gate only.

### DIFF
--- a/changelog.d/orchestration-command-entry-lock.feature.md
+++ b/changelog.d/orchestration-command-entry-lock.feature.md
@@ -1,4 +1,6 @@
-## Typing Into a Worker Pane Is Now a Decision, Not a Reflex
+## Typing Into a Worker Pane Can Now Be a Decision Rather Than a Reflex (Experimental)
+
+**This ships behind the `experimental` feature flag and is off by default.** Nothing about how you work today changes unless you turn it on: set `experimental = true` under a `[features]` table in your `.dot-agent-deck.toml`, or launch with `DOT_AGENT_DECK_EXPERIMENTAL=1`. It is gated precisely because it changes a reflex you already have — the default is deliberately restrictive, and that deserves real use before it reaches everyone. Everything below describes what you get once it is on.
 
 An orchestration is one workflow with a single coordinator, but every worker pane sits there with a cursor in it, inviting you to answer its question on the spot. Doing that puts a second, uncoordinated actor inside the workflow: you change state the orchestrator believes it owns, with no path for it to learn that you did. The deck does not look broken afterwards — it looks fine, and the model behind it has quietly diverged. Most often it is not even deliberate: you open a worker pane to check on it, get distracted, and type your next instruction into the pane in front of you instead of the one you meant.
 
--- a/docs/develop/experimental-flag.md
+++ b/docs/develop/experimental-flag.md
@@ -58,6 +58,9 @@ This lets work-in-progress code merge to `main` without exposing unfinished UI d
 |---|---|---|---|
 | `show_experimental_footer()` | The experimental dashboard footer | #139 | — |
 | `show_issue_dispatch_authoring()` | The new-pane `schedule: issues` modal authoring option (PRD #120 creation UX) | #120 | `graduate-issue-dispatch` |
+| `show_command_entry_lock()` | The orchestration command-entry lock: the `Ctrl+E` binding, the keystroke gate on a focused worker pane, and the waiting-pane focus steering | #393 | `graduate-command-entry-lock` |
+
+> **`show_command_entry_lock()` was added AFTER the feature merged (#404), not before it.** PRD #393 shipped un-gated and locked-by-default; the flag was added while it was still unreleased, so no version ever exposed it on. Three seams in `src/ui.rs` read the wrapper — the `Ctrl+E` binding resolution, the `PaneInput` keystroke gate, and the auto-focus chain. Note the third: the focus steering is part of the same surface rather than a separate feature, because it only ever ran while the lock was engaged, so gating it off is what makes flag-off behaviour identical to v0.35.8. The helpers themselves (`scope_command_entry_lock`, `gate_pane_input_key`) stay flag-free so their unit tests exercise the real logic rather than the gate; `UiState::command_entry_locked` also still starts `true`, since the flag decides whether that value is *consulted*, not what it is.
 
 ## Graduated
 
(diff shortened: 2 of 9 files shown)


vfarcic:
Valid finding — fixed in 597fc3f.

The part that makes this real is that the flag is **live-reloaded**: `features::spawn_watcher` re-reads `.dot-agent-deck.toml` roughly every 2s, so unlike a compile-time gate it can flip mid-session. That turned the new gate into a *second* way to stop observing waiting panes, alongside the `Ctrl+E` unlock — and the latch it guards (`had_waiting_pane` / `all_clear_pending`) is edge-triggered, so skipping observation freezes it rather than resetting it.

Sequence: a worker latches while the flag is on → the user turns the flag off → the worker resolves unobserved → the flag goes back on → the first observing frame reads the frozen latch as a fresh `true` → `false` all-clear and yanks focus to the orchestrator for an episode already dealt with.

This is exactly the failure the `Ctrl+E` handler already compensates for by calling `clear_waiting_pane_latch` on its locked→unlocked half. The flag needed the same compensation and didn't have it — my gate skipped observation without it.

Fixed in the flag's `false` arm unconditionally rather than by tracking an edge, because holding no latch at all is the honest state while the surface doesn't exist, and the clear is two bools per orchestration tab so there's nothing to save by tracking one.

It's also worth noting this is more reachable than it might look given the workflow this PR sets up: the whole point is that the maintainer runs with `DOT_AGENT_DECK_EXPERIMENTAL=1` for a while to evaluate the lock, and toggling it off and on to compare is exactly the thing someone doing that evaluation would do.

**Test:** `orchestration/focus/009`, modelled directly on `orchestration/focus/006` (which pins the same contract for the `Ctrl+E` unlock). Its local `frame()` helper mirrors the real per-frame call site's shape including the new flag arm. Verified load-bearing — dropping the `clear_waiting_pane_latch` call makes it fail with precisely the stale focus steal described above. The real `src/ui.rs` call site stays out of L1 reach, as `orchestration/focus/008` already records for its own uncovered half.

`cargo fmt --check` · `cargo clippy --workspace --all-targets --features e2e -- -D warnings` · `cargo test-fast` (1602 passed, 0 skipped) · `cargo xtask linkage-check` (480 catalog ids) — all clean. `cargo test-e2e lock_008…lock_012 lock_014` — 6 passed, including `lock_012` against a real Claude Haiku worker.

vfarcic:
This one is a **false positive** — but I've taken the suggestion anyway, for a different reason. Pushed as 3f16d9d.

## Why the finding doesn't hold

The condition is:

```rust
if !crate::features::show_command_entry_lock() {
    tab_manager.clear_waiting_pane_latch();
} else if ui.command_entry_locked {
```

That is **one** read of the flag, not two. Rust evaluates the `if` condition once; the `else if` tests `ui.command_entry_locked` — a `UiState` bool that has nothing to do with the flag. So there is no window between two reads for the watcher to land in, and the suggested diff is semantically identical to what was already there.

## And a second reason it couldn't bite even if there were two reads

The clear is **level-triggered**, not edge-triggered: it runs on *every* frame the flag is off, rather than once on the on→off transition. So a frame that somehow observed the wrong value self-corrects on the very next frame (~16ms) — orders of magnitude faster than a human can resolve a waiting pane, which is the only thing that could turn a stale latch into a visible focus steal.

That is precisely why the previous commit compensated unconditionally instead of tracking the edge, and it's now written down at the call site so the choice isn't mistaken for an oversight later.

## Why I applied it regardless

The hoist costs nothing and makes the single-read property obvious, so a later edit can't quietly introduce a second read that disagrees with this one mid-frame. Worth having as documentation even though it fixes no defect — I just don't want it recorded as having fixed one.

`cargo fmt --check` · `cargo clippy --workspace --all-targets --features e2e -- -D warnings` · `cargo test-fast` (1602 passed, 0 skipped) — clean. No behaviour change, so `orchestration/focus/009` is unaffected.

Issue #442, opened 2026-08-08
Selected deck card is hard to distinguish: DIM magenta reads as an idle card


vfarcic:
## Problem

On a dark theme the selected deck card is almost indistinguishable from a card holding an idle agent.

The cause is narrow and specific: `selected_card_border_style` (`src/ui.rs:16254`, added by PRD #341 M4) renders the selection accent as

- **command mode** → `palette::SELECTED` (Magenta) + `Modifier::BOLD`
- **`UiMode::PaneInput`** → `palette::SELECTED` (Magenta) + `Modifier::DIM`

DIM Magenta on a dark background lands right next to `palette::STATUS_IDLE` (`Color::DarkGray`, `src/palette.rs:87`). So the selected card *darkens* into the same visual band as every idle card on the deck. This only bites while the keyboard is driving a pane rather than the deck — which on the Dashboard is most of the time, since the pane overlay sits off to the right and the deck is where your eyes already are.

The existing code comment at `src/ui.rs:16251` already flags that DIM is not honoured by every terminal. What it did not anticipate is that where DIM *is* honoured, it collides with an existing semantic role.

## Deeper cause

The card border currently multiplexes three separate facts onto one channel — colour:

1. selection (Magenta, wins outright)
2. focus (Cyan)
3. status (green / blue / yellow / red / dark-gray)

Because selection wins, **the selected card's border stops reporting status entirely** — only the corner badge still carries it. Adding a fourth meaning (mode) to the same channel, as a modifier, is what produced the collision above.

The embedded-pane render path already solved this the other way round: `src/terminal_widget.rs:179` uses `BorderType::Thick` for focus precisely so colour is free to answer a different question. `src/palette.rs:59-64` spells out the reasoning — thickness was chosen over a fourth accent colour because the 16-colour-safe palette is full, and `BorderType` never feeds `Block::inner` so it costs no layout. Deck cards never adopted that split.

## Proposed fix

Give the deck cards the same channel separation the panes already have:

- **Colour = status, always.** Selection no longer overrides the border colour, so a selected working agent stays green, a selected waiting agent stays yellow, and so on. `palette::SELECTED` stops being applied to card borders.
- **Glyph weight = selection + keyboard locus**, as a three-step ladder:
  - unselected → `BorderType::Plain`
  - selected, `UiMode::PaneInput` → `BorderType::Thick`
  - selected, command mode → `BorderType::Thick` + `Modifier::BOLD`
[...]

PR #465, opened 2026-08-09
test(e2e): fix three consistently-red e2e tests (idle_worker_011, focus_007, card_stats_005)

### LINKED ISSUE #460: Three e2e tests are consistently red on main: idle_worker_011, focus_007, card_stats_005
Found while verifying #420, but **unrelated to that PR** — these fail on `main` itself.

## The failures

```
idle_worker_011_silent_worker_prompt_is_visible_in_attached_tui   (tests/e2e_idle_worker_detector.rs:191)
focus_007_lock_governed_focus_contract_on_real_binary             (tests/e2e_orchestration_focus.rs:212)
card_stats_005_real_agent_card_narrows_without_restructuring      (tests/e2e_card_layout.rs:272)
```

`focus_007` reports: *injecting `WaitingForInput` for the non-orchestrator alpha role never steered focus onto its expanded pane box (needle `┌alpha`) — expected the locked deck's auto-focus chain to pull focus onto a waiting role.*

## They are not flakes, and not a worktree artifact

- Rule 6 isolation rerun: each fails **consistently** when run alone.
- Run in the **canonical main checkout** (`/home/vfarcic/code/dot-agent-deck` at `d3ea031 [main]`, the short path) they fail identically, with near-identical timings to the worktree run: 21.65 vs 21.66s, 11.67 vs 11.64s, 49.06 vs 49.83s.

The short-path control matters because of the #352 lesson — a test can be deterministically red under a long `../dot-agent-deck-*` checkout and green in the main checkout. That is ruled out here: these are red in the main checkout too.

## Why this went unnoticed

Per CLAUDE.md rule 5 the e2e tier is pre-PR only and never runs in CI, so `main` being green in CI says nothing about these.

Likely three independent causes (focus chain, card layout, idle-worker detection) — worth triaging separately.

vfarcic:
## What

Fixes three e2e tests that were consistently red on `main` — `idle_worker_011`, `focus_007`, `card_stats_005` — with a **test-only** change. Nothing under `src/`, `xtask/`, or `tests/fixtures/` is touched.

Fixes #460

## Why they were red

There was no production defect. Triage established the production code was correct in all three cases; two earlier, deliberate production commits changed behaviour that three tests were never updated for:

- `eff6256` (PRD #393) gated the command-entry-lock surface — including the auto-focus chain — behind the `experimental` flag (off by default). It updated all 5 decks in `tests/e2e_orchestration_lock.rs` but not the sibling `tests/e2e_orchestration_focus.rs`, so `focus_007` asserted the flag-on contract while launching flag-off.
- `a861c8d` (PRD #442) changed the selected card border from Plain `│` to Thick `┃`. Two tests' observation helpers knew only the plain glyphs, so `idle_worker_011`'s whole-grid squeeze let sidebar borders interpose in the needle, and `card_stats_005`'s structural parser could not find a selected card at all.

## The fixes

- Scope the experimental flag to `focus_007`'s deck.
- Crop to the pane's column range before matching (`idle_worker_011`).
- Make the card parser border-weight-agnostic while *requiring* coherent same-weight borders at identical columns (`card_stats_005`).
- Harden one latent same-cause predicate in `tests/e2e_dispatcher_mode.rs`.

```
82f1047 test(orchestration): enable gated focus contract
85ff47d test(tui): tolerate thick card borders
28f71a1 test(tui): address e2e review findings
```

## Gates run before this PR

- `DOT_AGENT_DECK_RECORD=1 cargo test-e2e` (full L2 tier, the rule-5 pre-PR gate): **3,637 selected / 3,637 passed / 0 failed / 0 skipped**, exit 0, 222.52s wall-clock.
- `cargo fmt --check`: PASS
- `cargo clippy --workspace --all-targets --features e2e -- -D warnings`: PASS
- `cargo test-fast`: PASS — 1,603 tests, 0 skipped
- `cargo xtask docs --tests`: PASS
- Reviewer + auditor ran in parallel; 6 findings, all resolved in `28f71a1`.

Independently confirmed during review, worth stating here because it's what makes this diff trustworthy: border coherence is still enforced (a plain top with a thick bottom is rejected), byte-boundary/multibyte handling is correct (column indices are never used as byte offsets), no process-global env mutation was introduced (`with_env` only), and `idle_worker_011`'s 21.68s → 2.17s speedup is a genuine earlier match rather than a weakened predicate.

CI never caught any of this because per CLAUDE.md rule 5 the e2e tier is pre-PR only and never runs in CI.

## CLAUDE.md rule 12 classification

Neither a protocol-shape break nor a semantic break. No production code changed, so `PROTOCOL_VERSION` in `src/daemon_protocol.rs` is untouched, no wire shape moved, and no field changed meaning. No `changelog.d/460.breaking.md` fragment. No cross-version manual test is warranted. Both causal commits (`eff6256`, `a861c8d`) are presentation-layer only. Per the bump policy this is a **patch**, not a minor.

## Changelog

`changelog.d/460.misc.md` — filed as `misc` (internal/test-only, no user-visible behaviour change), not `bugfix`, since this restores e2e coverage rather than changing what the product does.


vfarcic:
## No demo reel for this PR

By deliberate decision, no demo reel was produced for this PR. This branch is test-only and changes no user-facing behaviour, so there is nothing to demo — a reel would show behaviour that never changed.

Of the tests this branch touches, only `dashboard/card-stats/005` is `[reel]`-marked, so any reel here would have been a single clip anyway.

That clip **was** recorded successfully during the pre-PR `cargo test-e2e` gate and is on disk — it's available for a future reel, it was just not published as part of this PR. Worth calling out: that cast had been missing for as long as `card_stats_005` was red, and this PR is what restores it.

vfarcic:
**Review routing note — independent of #467 and #472, and can land in any order relative to them.**

No file overlap with either: this touches the e2e tests, they touch `xtask/linkage-check/src/clean_tmp.rs` and the harness temp-root placement.

Worth knowing that the three failures fixed here were hit independently by two other in-flight branches during their pre-PR e2e gates, and both reached the same conclusion this issue did. On #467's branch they were re-measured against unmodified `main` at `d02111bc` in a clean disk-backed worktree and failed identically — same assertion, same source line, same observable final-grid symptom — so the diagnosis in #460 is now confirmed from three independent directions.

Landing this first would be useful regardless of the #467/#472 ordering: it gives every subsequent pre-PR gate a clean baseline, instead of each run having to re-derive that these three are pre-existing before it can trust its own result.

vfarcic:
@prageethw — welcome, and thanks for taking a look. Some orientation since this is your first review here:

**What this is**: test-only, no `src/` or `xtask/` change. It repairs three e2e tests that were consistently red on `main` (issue #460), plus one latent same-cause predicate.

**There was no production defect.** Two earlier, deliberate, *correct* production commits changed behaviour that three tests were never updated for: `eff6256` (#393) gated the auto-focus chain behind the `experimental` flag, and `a861c8d` (#442) switched the selected card border from plain `│` to thick `┃`.

**Why CI never caught it**: per `CLAUDE.md` rule 5, the e2e tier is pre-PR only and never runs in CI.

**Gates that did run**: full e2e 3,670/3,672 passing on the merged tree (the 2 failures were unrelated LLM-variance flakes in tests this PR doesn't touch, each passing on first isolated rerun); CI 10/10; `cargo test-fast` 1,635/1,635; fmt and clippy clean.

**The most useful thing to look at**: the `card_titled` predicate in `tests/e2e_dispatcher_mode.rs` and its new negative test `card_titled_rejects_labels_outside_card_span`. An auditor found the old predicate could accept a role label from anywhere on a row-joined grid line, so a card whose title had regressed to a UUID could still pass on aligned pane text. The fix restricts matching to each card's own column span, and the new test proves the old predicate accepted the adversarial grid while the new one rejects it.

One honest caveat: that new test lives in the e2e-gated file, so it runs pre-PR and *not* in CI — a known wart, accepted rather than fixed by refactoring shared harness code on an otherwise finished PR.

prageethw (commented):
Read-only review at `31709d9c` in a detached worktree. No tests run, no GitHub writes beyond this. Behavioural claims below were verified by replicating the Rust predicates character-for-character against the PR's own fixtures — not by executing anything.

**Verdict: approve with suggestions.** Nothing blocks.

The thing I went looking for on a test-repair PR is whether a red test was made green by weakening it. **It wasn't** — and the two changes that *look* like weakening are strictly stronger. Detail in the "checked and clean" section, because on this kind of PR the negative result is the finding.

---

## Headlines

- **No assertion weakened, anywhere.** The two deleted `card_stats_005` corner asserts were already tautological under the *old* extractor. The new extractor is strictly stronger. `card_titled` went from "label anywhere on a row containing a `┌`" to "label inside one same-weight border span" — run against your own `FALSE_PASS_GRID`: old `true`, new `false`.
- **No timeout widened.** Zero `Duration` literals change in the diff. Every fix is a deterministic gate on observable state, not a longer sleep. `idle_worker_011` got *faster* (21.68s → 2.17s), which is what an earlier genuine match looks like.
- **But the class is patched three times, not fixed once** — my highest-value finding, and it's the same shape as the bug being fixed. See S1/S2.
- `.config/nextest.toml` correctly untouched; no override filter went stale. Rule 12 classification and the `misc` fragment type both right.

---

## S1 — `orchestration_pane_column` duplicates a helper that already exists

**Suggestion.** `tests/e2e_idle_worker_detector.rs:50-69` (new) vs `tests/e2e_orchestration_pane_column.rs:170-184` (pre-existing).

Same three corner literals, same byte-index→scalar conversion, neither lifted into `tests/common/mod.rs`. `grep -n "┏orchestrator" tests/*.rs` returns exactly those two sites.

- **What breaks:** the next presentation change of the `a861c8d`/`eff6256` kind — a fourth border weight, or renaming the fixture's start role away from `orchestrator` — updates one copy. The other silently returns `None`.
- **And because of S4**, that failure surfaces as *"the daemon-authored idle prompt never became visible"*, pointing at the idle detector when the actual break is the crop anchor.
- **This is the answer to "does it fix the class or the instance?"** — instance. A production change that two sibling test files needed, and only one got. Which is precisely the bug this PR exists to fix.
- The two copies already differ in tie-break: the pre-existing one returns the first corner in array order, the new one takes `.min()` of byte positions. They agree today; they're two things to keep in sync.

**Suggested:** lift into `tests/common/mod.rs`, which already hosts shared grid helpers (`nonblank_rows`, `joined_rows` at `:1889`/`:1901`), as `pub fn orchestration_pane_left_edge(grid: &str) -> Option<usize>`, and call it from both.

## S2 — three inconsistent encodings of "border weight" land in one PR

**Suggestion.**

| File:line | Encoding | Weights |
|---|---|---|
| `tests/e2e_card_layout.rs:42-66` | `const CARD_BORDERS: [CardBorder; 2]` | plain, thick |
| `tests/e2e_dispatcher_mode.rs:576-580` | inline `match` | plain, thick |
| `tests/e2e_idle_worker_detector.rs:54, 83` | array literal + `matches!` | plain, thick, **double** |

- **The double branch is dead code today.** `card_border_glyph` (`src/ui.rs:16372-16385`) returns only `Plain` or `Thick`; `src/terminal_widget.rs:180-182` likewise; `grep -rn "BorderType::Double" src/*.rs` finds nothing (only unrelated "Double-click" comments).
- So one file defends a weight that cannot occur, while the other two would break if it ever did. **Whichever way the product moves, exactly one of these three is wrong.**

**Suggested:** one shared table next to the S1 helper. Include `Double` everywhere or nowhere — today's split is the worst of both.

## S3 — `wait_for_wrapped_grid_string` now describes the old behaviour

**Suggestion.** `tests/e2e_idle_worker_detector.rs:87-94`. The doc says *"Wrap-tolerant `wait_for_grid_string_within`"*, but the function no longer searches the grid — it searches only the orchestration pane column, and returns `false` outright when no orchestrator box is rendered. Both are new preconditions the name and doc actively deny.

A future test waiting on a needle rendered in the *sidebar* calls this, gets a permanent `false`, and the doc offers no hint why. **Suggested:** rename to `wait_for_wrapped_pane_string` and state the two preconditions — the fixture's start role must be literally named `orchestrator`, and its pane must render as an expanded box.

## S4 — "anchor missing" and "needle absent" produce identical failure text

**Suggestion, and the one real regression here.** `tests/e2e_idle_worker_detector.rs:90-93` — `is_some_and` collapses two different failures into one `false`.

- If the orchestrator pane renders **collapsed**, it draws a `Block` with `Borders::TOP` and a padded title — **no corner glyph at all**. Your own sidebar-collision doc at `tests/e2e_orchestration_pane_column.rs:15-29` spells that layout out.
- The crop then returns `None` on every poll, and the operator is told *"the daemon-authored idle prompt never became visible in the attached orchestration pane"* (`:222`) — when the prompt may be rendering perfectly and only the anchor is gone.
- That's a genuine diagnosability regression against the old whole-grid search, which had no anchor to lose.
- **It also silently applies to `idle_worker_012`**, a credential-gated real-agent test (`skip_unless!`, `:240`) that CI never runs — so the misleading message lands exactly where it's hardest to diagnose.

**Suggested:** assert the anchor once before the needle waits, or have the helper report which of the two conditions failed.

## S5 — the one new regression guard sits in the tier CI never runs

**Suggestion.** `tests/e2e_dispatcher_mode.rs:599-635`, `card_titled_rejects_labels_outside_card_span`.

It's a **good** test — genuinely RED against the old predicate and GREEN against the new; I verified that. But it's pure string logic — no PTY, no binary, no daemon — in a file gated `#![cfg(feature = "e2e")]`. Per rule 5 the L2 tier is pre-PR only, so the only guard against this exact predicate regression fires when a human runs the full tier locally.

It's also invisible to `cargo xtask linkage-check`: rules 4, 6 and 7 all key off `#[spec]` (`xtask/linkage-check/src/main.rs:8-23`) and this carries none. Precedent exists — 7 other non-`#[spec]` tests live in `tests/e2e_*.rs` — but every one genuinely drives a PTY. This is the only pure-logic one, and the only one with no doc comment.

**Suggested:** move `card_label`, `card_titled` and this test to the fast tier, so CI runs the guard on every push. If it stays, give it a doc comment matching the other seven.

---

## Nits

- **N1 — two of the three new fixture grids aren't rectangles.** `tests/e2e_dispatcher_mode.rs:622-628`. Functionally fine (`card_titled` reads only the top row, and all three assertions hold). But measured: `PLAIN_CARD_GRID`'s top-right corner is at column 28 while its middle and bottom right edges are at column 31 — three adrift; `THICK_CARD_GRID` identical. (`FALSE_PASS_GRID` is correctly formed.) These constants read as documentation of a rendered card, and two document something the renderer can't produce — anyone reusing them for a `try_first_card`-style parser, which requires exact column matches, gets a confusing failure.
- **N2 — `try_first_card`'s leftmost assumption has a concrete counterexample.** `tests/e2e_card_layout.rs:194-196`. The caveat is honest but abstract, and a real case exists: `ActiveTabView::Mode` (`src/ui.rs:12134-12141`) is *"50/50 horizontal split: agent pane left, side panes right"*. `Dashboard` and `Orchestration` both go through `split_cards_area` returning `(dashboard_area, panes_area)`, which is what makes the assumption hold. Worth naming `Mode` in the comment.
- **N3 — the `focus_007` scenario rewrite drops the mechanism.** `tests/e2e_orchestration_focus.rs:155`. The old comment explained *why* focus can't move while unlocked — *"the per-frame call site never even calls `observe_waiting_panes`"*. The new form restates the assertion instead. Not a rule 7 violation, and the single-line style matches the newer house style; only the `observe_waiting_panes` clause is worth keeping.
- **N4 — `border_segments` pairs only *consecutive* verticals** (`tests/e2e_card_layout.rs:108-131`). With `left == right == '│'` the segments are `(p0,p1), (p1,p2)…` — never `(p0,p2)`. A card content row carrying a `│` between its own borders would fail to extract. **Fails loud** (`first_card` panics), so no false green, and I couldn't construct a reachable case. **Unverified** whether it's reachable at all — recording it because the property is non-obvious and the panic message wouldn't suggest it.

---

## Checked and clean — where I spent the review

**No weakening.** Item by item:

- **`card_stats_005`, the two deleted corner asserts** — already tautological under the old extractor. `corner_segment(line, '└', '┘')` returned a slice beginning at `└` and ending at `┘` *by construction*, so `starts_with('└') && ends_with('┘')` could never fail. Nothing lost.
- **`card_stats_005`, the new extractor is strictly stronger.** Old `try_first_card` took `top` = first line containing any `┌…┐` and `bottom` = first later line containing any `└…┘` **at any columns**, then built each row from that row's own first corner pair — so a snapshot could splice segments from different columns and pass. New requires an identical `(start, end)` scalar-column span on the top row, *every* middle row, and the bottom row, all from one weight's glyph set. A plain top with a thick bottom is now rejected.
- **`bottom_label_is_right_aligned`** expands to exactly the old predicate for a plain card. The one relaxation — accepting `┛` on a selected/thick card — is correct product behaviour after `a861c8d`, and the weight contract is still pinned **in the fast tier** by `tests/render_dashboard.rs:622-626`. So going weight-agnostic gives up nothing that isn't asserted in a tier CI actually runs.
- **`card_titled`** — replicated both predicates. Old against `FALSE_PASS_GRID` → `true`; new → `false`. Against `THICK_CARD_GRID`: old → `false` (it missed thick borders entirely — the latent bug), new → `true`. The new negative test is a real guard, not a restatement.
- **`idle_worker_011`** — narrows the haystack and adds a precondition. Narrowing can only turn a failing assertion green by removing the interposing sidebar text, which is exactly the documented root cause: with the sidebar in the haystack, squeezing and joining yields `card1·pane1·card2·pane2…`, so a needle spanning `pane1`+`pane2` is broken by `card2`. Cropping the prefix is the correct minimal fix.

**Root cause vs symptom.** All three fixes are deterministic gates. `focus_007` enables the flag the surface is actually gated behind (`features::show_command_entry_lock`) rather than working around it. `idle_worker_011` crops on a rendered anchor, not a sleep. `card_stats_005` matches the glyph the product now paints.

**`.config/nextest.toml`** — correctly untouched. None of the three is in the >3×60s class, and `idle_worker_011` moved the other way. No filter went stale: filters track test *function* names and this PR renames none.

**Copy-paste errors between the three fixes** — none. The only cross-fix inconsistency is the border-weight set (S2).

**Experimental-flag scoping** — complete and follows precedent. `focus_007` is the only test in its file, so scoping to its deck covers the file. `.with_env("DOT_AGENT_DECK_EXPERIMENTAL", "1")` is the exact idiom the five sibling decks in `tests/e2e_orchestration_lock.rs` use. `src/features.rs` reads the override in both TUI and daemon, and the daemon inherits the spawned deck's environment, so one `with_env` covers both. The flag-off test at `e2e_orchestration_lock.rs:578-623` correctly left alone.

**Rule 12** — agreed. Six files, none under `src/` or `xtask/`, `PROTOCOL_VERSION` untouched, no wire shape or field meaning moved. `misc` is the right towncrier type for a test-only change. Patch bump.

**Catalog accuracy** — both `tests/CATALOG.md` edits match the code. `dashboard/card-stats/005`'s *"matching-weight intact bottom corners"* is exactly what `try_first_card` now enforces.

**Rule 10, dead code** — both clean. All `CardBorder` and `BorderSegment` fields are read; `card_label` still used at `e2e_dispatcher_mode.rs:804`. The only semantically dead path is S2's double branch.

**Context, not a finding:** Greptile genuinely ran here — check-run present, passed in 1m46s, summary comment at 5/5 confidence, no inline findings. Worth noting because an empty Greptile review and a clean one look identical unless you check the check-run exists.

prageethw (approved):
Re-reviewed at `40fcdbcb` — delta only (the two commits since my last review), code read, no build and no test run.

**Approving.** All nine prior asks are addressed, the generalisation is correctly scoped, and — the thing I most wanted to check — **no assertion anywhere in the delta is weaker than before**.

| Prior ask | Verdict |
|---|---|
| S1 — duplicated pane-column helper | ADDRESSED — one copy in `common`, both consumers delegate, the tie-break divergence is gone by construction |
| S2 — three border-weight encodings | ADDRESSED — one `BORDER_WEIGHTS` table, `Double` listed everywhere, reasoning written down |
| S3 — doc/name drift | ADDRESSED — renamed to `wait_for_wrapped_pane_string`, both preconditions stated |
| S4 — indistinguishable failures | **ADDRESSED, better than asked** — `Result<(), String>`, with `anchor_seen` tracked across the whole poll loop rather than re-checked after the timeout |
| S5 — guard sitting in a tier CI never runs | **ADDRESSED DIFFERENTLY, correctly** — a new un-gated `tests/grid_box_helpers.rs` |
| N1–N4 | all ADDRESSED |

**On S5 I verified execution mechanically rather than taking the commit message's word**, since "CI actually runs this" *is* the entire value of that ask — a guard that does not execute is the exact failure the ask existed to fix. `Cargo.toml` has no `autotests = false` and no `[[test]]` stanza, so Cargo auto-discovers the target; the file carries no `#![cfg(feature = "e2e")]`; and `mod common;` from a non-e2e test binary has 15 precedents in this repo. It runs.

It also picks up coverage nobody asked for: the pane crop's splice bug. I replicated its negative assertion — that the *uncropped* grid must still miss the needle — and confirmed it holds for all three border weights.

**On whether the three red tests were fixed for the right reason** — they were. `idle_worker_011`, `focus_007` and `card_stats_005` were each fixed by correcting a genuine test bug, not by loosening an assertion until it could not fail. Zero `Duration` literals move in the delta, and `squeeze_wrapped_text` is a byte-for-byte reimplementation of the local `squeeze` it replaced, so `idle_worker_011`'s match predicate is unchanged.

I established the behavioural claims by replicating the Rust predicates character-for-character against your own fixtures in a scratch interpreter, rather than by running the suite.

---

### One relaxation, which I am recording rather than blocking on

**NIT 1.** N4's fix changes `border_segment_at` from consecutive-pair matching to direct column indexing. That was my own ask, and it is the right shape — but it costs something the doc comment claims it does not. A stray `│` inside a card content row used to make the rectangle check fail loudly; it now parses fine. So a renderer regression of that exact shape would ship green.

I could not make it reachable — card content is role and status text, and the glyph is U+2502 rather than ASCII `|` — which is why it is a nit and not a finding. It is worth a sentence in the doc comment, because the comment currently promises strictness the function no longer has, and that is what the next person will rely on.

**NIT 2** — the top row and bottom row are now checked to different strictness.

**NIT 3** — `assert_is_rectangle` hardcodes `RIGHT_EDGE = 31`, so the adversarial `FALSE_PASS_GRID` at column 29 is the one fixture still verified by eye rather than by the helper.

All three are the same class: the helpers are now good enough that the remaining gaps are in what they *promise*, not in what they *do*.

Issue #435, opened 2026-08-08
IpcClient::connect is uninterruptible on Unix, so both sync IPC paths can hang past their advertised deadline


vfarcic:
## What

Both synchronous IPC paths establish the connection with a blocking, uninterruptible call and only apply a deadline afterwards:

- `src/hook.rs` — `IpcClient::connect(path)`, then `set_timeouts(remaining)` (the `get-seed` request path)
- `src/ui.rs:669-671` — `IpcClient::connect(&path)?`, then `stream.set_timeouts(timeout)?` (`send_daemon_request_blocking`, on the attach socket)

On Unix, `IpcClient::connect` is a straight lift of `std::os::unix::net::UnixStream::connect` (`src/platform/ipc/unix.rs:160`), which is blocking and honours no timeout. So the deadline each caller advertises does not actually cover connect.

## Is it reachable?

The mechanism is real, and I confirmed it directly: a **blocking** `connect()` to a Unix socket whose listener has a full backlog and never calls `accept()` blocks indefinitely — it does not return `EAGAIN` or `ECONNREFUSED`. (A non-blocking socket gets `EAGAIN` instead, which is why this only bites the blocking path.)

Reachability is narrow, though. The hook socket's accept loop at `src/daemon.rs:1000` `tokio::spawn`s each connection and returns to `accept()` immediately, so the backlog only fills if the entire tokio runtime is starved *and* enough connections queue behind it. This is a hardening gap, not an everyday hang.

Windows is already bounded — `IpcClient::connect` there retries `ERROR_PIPE_BUSY` against `CONNECT_RETRY_BUDGET` and gives up (`src/platform/ipc/windows.rs`).

## Suggested fix

Add a `connect_timeout`-style constructor to `IpcClient` and use it on both sync paths. `std::os::unix::net::UnixStream` has no `connect_timeout` (unlike `TcpStream`), so the Unix side needs the manual form: create the socket non-blocking, `connect`, poll the fd for writability with the remaining budget, check `SO_ERROR`, then restore blocking mode. Windows can keep its existing retry-budget behavior behind the same signature.

With that in place, `hook.rs`'s "bounds the **whole exchange**" doc comment becomes literally true rather than approximately true.

## Context

[...]

Issue #163, opened 2026-06-14
Windows process, path & filesystem platform backends


vfarcic:
## PRD: Windows process, path & filesystem platform backends

**Phase 2 of native Windows support** — depends on the platform-abstraction foundation in #42 (the `src/platform/` seam, `cfg`-gated `libc`, the IPC abstraction, and the `windows-latest` CI job). #42 holds the canonical architecture, the per-category Unix→Windows mapping, the behavior-equivalence risk analysis, and the locked v1 decisions.

**Problem**: After #42 lands, the `platform/` seam exists and IPC is cross-platform, but five Unix-only mechanism categories still have no Windows backend: paths & shell (`$HOME`/`/bin/sh -c`), daemon detach & spawn lock (`setsid`+`pre_exec`, `flock`), peer-PID & lifecycle signals (`SO_PEERCRED`/`LOCAL_PEERPID`, `kill`/`killpg`), the filesystem security model (`umask`/mode-bits/`O_NOFOLLOW`), and clipboard (`/dev/tty`). So the daemon can't fully spawn, stop, signal, secure, or share-clipboard on native Windows.

**Solution**: Implement the Windows `windows.rs` backend for each `platform/` submodule, preserving the Unix backend byte-for-byte. Per #42's locked decisions: `dirs` crate + `%LOCALAPPDATA%`; `cmd /C`; `DETACHED_PROCESS`+Job-Object breakaway; named mutex spawn lock; `GetNamedPipeServerProcessId`; `KIND_SHUTDOWN`→`TerminateProcess` + Job-Object agent teardown; current-user-only pipe security descriptor + dir ACLs; `CONOUT$` clipboard. v1 ships best-effort graceful shutdown (`CTRL_BREAK_EVENT`→`TerminateJobObject`), differences documented.

**Detailed PRD**: See [prds/163-windows-platform-backends.md](https://github.com/vfarcic/dot-agent-deck/blob/main/prds/163-windows-platform-backends.md)

**Priority**: Medium

PR #466, opened 2026-08-09
fix(orchestration): let a dispatched orchestration delegate, and make a failed delegate loud

vfarcic:
Fixes the reported bug: an orchestration started with `dot-agent-deck dispatch --orchestration` cannot delegate to its worker roles, while the same orchestration started with `Ctrl+N` can.

Two stacked defects, both real. Each change below is labelled with the defect it addresses.

## Defect 1 (primary) — pane identity: the daemon never learned who these panes were

**Changes:** `src/state.rs` (new `AppState::register_orchestration_role`), `src/spawn.rs`, `src/daemon_protocol.rs`, `src/dispatch.rs`, `src/daemon.rs`, `src/issue_dispatch_run.rs`.

The daemon's `pane_role_map` / `orchestrator_pane_ids` / `pane_orchestration_map` — the maps `handle_delegate` routes on — were populated in exactly one place: the `AttachRequest::StartAgent` handler. That is the path a TUI-initiated (`Ctrl+N`) orchestration takes, and the **only** path that went through it. An orchestration the daemon spawns *itself* (`dispatch --orchestration`, a scheduled fire, issue dispatch) reaches `AgentPtyRegistry::spawn_agent` directly via `spawn::spawn` and never touched that state, so `handle_delegate` dropped every delegate at its first check with `delegate from unknown pane`.

The pane id's **string shape was incidental**, not the cause. The brief's hypothesis was that `sched-dispatch-<name>-<n>-r0` could not be resolved where a numeric `14` could; in fact nothing ever *recorded* the dispatched id, so any id shape would have been equally unknown. `handle_delegate` has always been shape-agnostic — it does a `HashMap` lookup on a `String`.

Both spawn paths now register through **one** shared function, which is the point: a second inlined copy is how they drifted apart. The daemon's `AppState` is threaded into `spawn::spawn`, fixing the dispatch, scheduler and issue-dispatch paths together. Registration happens synchronously before `spawn` returns and before the orchestrator's prompt is delivered, so a fast orchestrator's first delegate cannot beat its own registration.

One deliberate difference from the `StartAgent` path: the orchestrator is registered by `orch_idx`, not the raw `is_start_role` flag. `orch_idx` is already this path's authority on which role receives the orchestrator context, and it falls back (`orchestrator` by name → any `start = true` → role 0) where `is_start_role` alone is false for *every* role of an orchestration whose toml sets no `start` — which would have left this same bug for a narrower input. The `StartAgent` path knowingly still uses the raw flag; unifying the rule is **#523** (see round 2 below).

## Defect 2 (secondary) — silent failure: `delegate` could not tell success from failure

**Changes:** `src/event.rs` (new `DelegateResponse`), `src/state.rs` (`handle_delegate` now returns it), `src/daemon.rs` (replies on the connection), `src/hook.rs` (new `send_and_await_reply`), `src/main.rs`.

`delegate` was fire-and-forget: it exited 0 without waiting to hear whether anything happened. So an orchestrator whose delegation had been dropped was told it succeeded, announced that its worker was now working, and waited forever for a `work-done` that could never arrive — which is what turned a routing bug into a hung orchestration instead of an error message.

The daemon now answers on the same hook-socket connection (the `GetSeed` / `ListTargets` pattern) and the CLI reports it. Non-zero exit plus a stderr message naming the pane id and what could not be resolved, for both failure shapes:

- a sender the daemon holds no orchestration role for, or one that is not its orchestration's orchestrator;
- a `--to` role set that resolves to **no** worker pane at all — which also used to exit 0 having reached nobody.

```
$ dot-agent-deck delegate --to worker --task probe   # pane with no role
Error: delegate from pane xcaller failed: the daemon holds no orchestration role
for pane xcaller, so this delegate was routed nowhere. Only a pane spawned as
part of an orchestration can delegate.
$ echo $?
1
```

A delegate that reaches **some** of its roles is a distinct outcome and is reported as one — see round 2.

## Review round 2 (@prageethw + greptile)

Seven threads. Every finding is answered in its thread; the code changes are:

| finding | resolution |
|---|---|
| **BLOCKER** — a partially-delivered delegate reported total failure, and the retry double-dispatches | non-zero now only when **nothing** landed; a half-landed delegate exits 0 and names both sides |
| the message asserted a cause it had not established | it now states all three causes (absent role / the orchestrator itself / closed worker pane) |
| a fully-defaulted `DelegateResponse` + `Err(_) => SUCCESS` fails open | `kind` discriminator; a reply that does not identify itself is treated as `NoReply`, not as success |
| no read deadline on Unix | `send_and_await_reply` now shares `get-seed`'s bounded transport (5s **total-operation** deadline) |
| the registrar was unified, the orchestrator rule was not | documented at the site, split out as **#523** |
| greptile P1 — stale targets report successful delivery | adjudicated in-thread, split out as **#524**; `delivered`'s doc comment now states the caveat |
| nothing that blocks a merge would fail if the rejection went silent again | **10 unit tests** in the fast tier, plus the half-landed case in `dispatch/001` |

Also in this round: the branch was merged with `main`, which surfaced a **semantic merge conflict** GitHub reported as `MERGEABLE` throughout. #419 landed its own `hook::SocketReply` while this branch was open; the two merge textually without a conflict marker and then fail to compile (`SocketReply` defined twice, conflicting `Debug` impls). This branch's copy is deleted and `send_and_await_reply` returns main's enum — whose doc comment already anticipated this exact caller ("as the not-yet-submitted `delegate` wiring will"). Sharing that implementation is also what gives the Unix read its deadline, so the two resolutions are one change.

⚠️ **#435 is in flight in the same file** (`IpcClient::connect` is uninterruptible on Unix, so both sync IPC paths can hang past their advertised deadline). This PR deliberately adds **no** new timeout machinery — it routes `delegate` through the transport `get-seed` already uses — so it should not conflict; whichever lands second should rebase.

## CLAUDE.md rule 12 — cross-version contract

**Decision: no `PROTOCOL_VERSION` bump and no `changelog.d/*.breaking.md`.** The attach protocol is untouched; the hook-socket change is purely additive (the daemon writes a reply line where it previously wrote none), and old/new builds interoperate in both directions.

`send_and_await_reply` exists specifically for this: `request_from_socket` folds "no daemon" and "old daemon that does not answer this verb" into one `None`, and reporting those the same way would make a newer CLI call every delegate against an older daemon a failure. Only a genuinely undelivered signal is reported as one.

**Re-run for round 2**, because the round-2 changes moved the reply shape again (the new `kind` field) — against **v0.36.0** this time, in a sandbox with sockets, `HOME`, state dir **and `DOT_AGENT_DECK_LOG`** all isolated (the rule's own warning about the log path; verified by the two daemons' lines landing in the sandbox logs):

| case | exit | ms | result |
|---|---|---|---|
| new CLI → new daemon | 1 | 7 | loud message naming the pane |
| new CLI → **old** daemon (v0.36.0) | 0 | 6 | Unix half-close ⇒ prompt `NoReply`, no false failure, no stall |
| **old** CLI → new daemon | 0 | 5 | signal received and handled — unchanged |
| old CLI → old daemon | 0 | 6 | control |
| new CLI → new daemon, **after** an old caller hung up | 1 | 6 | the daemon's reply-write to a departed caller does not wedge its hook loop |

The reply on the wire, read straight off the socket:

```json
{"kind":"delegate","delivered":[],"unresolved_roles":[],"error":"the daemon holds no orchestration role for pane xcaller, …"}
```

`kind` is `#[serde(default)]` on the way in, so an older reader is unaffected — and no older reader exists anyway: before this PR the daemon wrote no reply at all and the CLI read none.

Bump policy while `0.x`: bugfix → **patch**. The `delegate` exit-code change is user-visible behaviour, but rule 12 defines "breaking" as old/new builds no longer interoperating, which is explicitly not generic user-facing change.

## Tests — reproduce-first

Per the skill, existing coverage was **extended**, not duplicated — and both entries' `Does not assert` lines named this exact gap ("a real delegation round-trip", "a delegation round-trip"), which is how the bug shipped green.

**`orchestration/dispatch/001`** (cat roles, zero tokens) — runs the real `delegate` CLI from the dispatched orchestrator and requires the daemon-authored pointer to reach the worker's PTY. The **control** the skill asks for is the same orchestration opened via `Ctrl+N`, in the same deck, same fixture, same roles, same CLI — only the spawn path differs — and it runs **first**, so a broken control is reported as a harness fault rather than read as a product result. Also pins defect 2's exit codes, and (round 2) the half-landed case.

**`orchestration/dispatch/002`** (three real Haiku agents) — asks the *real* orchestrator to delegate a sentinel-file task and requires the *real* `coder` to actually create the file in the dispatched worktree. That is the user's altitude: "I dispatched an orchestration and the team got something done."

**Round 2 added 10 unit tests in the tier that gates a merge**, because the two above are behind `#![cfg(feature = "e2e")]` and no CI build job passes `--features e2e` — so a refactor returning `DelegateResponse::default()` from the unknown-pane path, restoring the exact silent success being fixed, passed every gate. Five sit beside the `delegate_targets_*` tests in `src/state.rs` (unknown pane, registered non-orchestrator pane, unresolvable role, partial delivery, `orch_idx` registration) and five on the CLI's `delegate_verdict` / `parse_delegate_reply` seams in `src/main.rs`.

Evidence the e2e tests are honest:

- Confirmed **red for the reported reason** before any fix, with the daemon log showing `delegate from unknown pane pane_id=sched-dispatch-team-probe-0-r0` while the control's `pane_id=2` was accepted in the same run.
- Each fix **reverted independently** turns them red again. Reverting defect 1 now fails at the CLI's exit code rather than 90s later at the pointer, because the two fixes compose.
- The first harness attempt was itself wrong and was fixed before being trusted: after a `clear=true` respawn two records briefly share a pane id, and `.find()` could return the dead one — a coin flip that failed the *control* once.
- `dispatch/002` needed `wait_until_panes_settled` before injecting. Without it, it passed in ~17s alone and failed at its full 300s budget inside a full `cargo test-e2e`: on a loaded machine the directive lands in a claude TUI's mid-init lull and is dropped. Confirmed fixed by a full-suite run.

## Gates

- `cargo fmt --check` and `cargo clippy --workspace --all-targets --features e2e -- -D warnings` — clean.
- `cargo test-fast` — 2533 passed, 0 failed.
- `cargo xtask linkage-check` — ok.
- `cargo test-e2e` — see the comment thread for the round-2 run.


src/state.rs
```diff
@@ -2968,6 +3061,18 @@ impl AppState {
...
+        // Reported once the fan-out is QUEUED, not once each worker has answered.
+        // The dispatches are deliberately detached (see above), so waiting here
+        // would re-introduce the multi-second stall that async-dispatch removed.
+        // "Queued to a resolved worker pane" is the strongest claim this call can
+        // honestly make, and it is precisely the claim the old silent exit-0 made
+        // falsely.
+        crate::event::DelegateResponse {
+            delivered,
+            unresolved_roles,
+            error: None,
```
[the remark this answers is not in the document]
vfarcic: Verified — this is valid, and I am deliberately leaving it for human review rather than fixing it in this PR (author's call). Flagging it here so @prageethw can weigh in with full context.

**The mechanism checks out.** The only daemon-side caller of `unregister_pane` is the `StopAgent` handler (`src/daemon_protocol.rs:1465`); the `src/ui.rs` call sites operate on the TUI's own `AppState`, not the daemon's. So a worker that exits *without* going through the explicit pane-close path — a crash, or an agent that quits on its own — leaves `pane_role_map` / `pane_orchestration_map` behind. `delegate_targets` then still matches that role, it lands in `delivered`, and the CLI exits 0. The actual PTY write happens later inside `dispatch_one_owned`, which runs in a detached `tokio::spawn`, so its `NotFound` cannot reach the response.

**Scope, to be precise about what this PR does and does not claim.** The staleness itself predates this PR — `delegate_targets` has always been able to resolve a dead pane. What changed is that it now has a *consequence*: before, `delegate` exited 0 unconditionally, so there was no success claim to be wrong. This PR makes the claim, and the claim is "queued to a resolved worker pane", which is what the code comment on the return value says. A stale mapping makes *resolved* wrong, so the claim is wrong with it.

So this is a genuine gap in the fix, not in the original bug. The reported defect (a dispatched orchestration whose roles were never registered at all) is fixed and covered by `orchestration/dispatch/001` + `002`; this is the adjacent case of a role registered but no longer live.

**Two directions a fix could go**, which is exactly the judgement call worth a human on it:

1. *Liveness-check at resolution* — filter `delegate_targets` on `registry.pane_is_live(...)` (already used by `spawn_or_reuse`'s reuse gate) so a dead pane is reported as unresolved, alongside a role that has no pane at all. Cheap and synchronous, but it is a second source of truth about liveness sitting beside the role maps.
2. *Clean up on unexpected exit* — call `unregister_pane` when the daemon observes a pane's process die, not only on `StopAgent`. Fixes the root cause and would also stop stale panes participating in `handle_work_done` routing, but it is a broader change with its own blast radius.

Happy to take either in a follow-up, or in this PR if the reviewer would rather it not ship as-is.

prageethw (commented):
Read-only review at `0935b27b` in a detached worktree. Full +846/−33 across 12 files. No tests run, no binary started, no daemon, no `delegate` issued, no GitHub writes.

**Verdict: merge-worthy.** The diagnosis is right — only `AttachRequest::StartAgent` populated the daemon's role maps, so daemon-spawned orchestrations were unroutable — the shared `register_orchestration_role` is the correct shape, and the loud-failure half is a real improvement over a genuinely dangerous silent success.

Two things I'd want addressed before or immediately after merge.

---

## 1. A partial delegate reports total failure — and a retry double-dispatches · **blocker, small fix**

`src/main.rs:748-761`.

**The scenario:** an orchestration with roles `orchestrator`, `coder`. The orchestrator runs `delegate --to coder --to tester`.

- `handle_delegate` produces `delivered = ["coder"]`, `unresolved_roles = ["tester"]`, `error = None` (`state.rs:2957-2973`).
- The fan-out loop at `state.rs:2998` **then runs for `coder`** — `arm_idle_worker_watch_for_delegation` arms its record, `dispatch_one_owned` is spawned, and the task genuinely reaches the coder pane.
- `handle_delegate` returns at `state.rs:3070`. The CLI prints an error naming only `tester`, and exits **1**.
- The orchestrator applies **the contract this PR establishes** — non-zero means it did not land — and re-issues the delegate.
- The coder now receives the same task twice, arming a second idle-worker record for one pane.

That last step is a hazard the codebase already names, at `state.rs:4604-4617`: *"the second superseding the first, so a single `work-done` would leave one armed"* — with a unit test that exists specifically to stop one signal producing two dispatches into a pane. The stderr text gives the orchestrator no way to know `coder` must be excluded from the retry.

**Worth noting:** `resp.delivered` is **already on the wire and unread by the CLI**. `grep DelegateResponse src` shows `main.rs` only ever inspects `error` and `unresolved_roles`. The fix needs no protocol change.

**Suggested minimum:**

```
Error: delegate from pane {pane} reached no worker for role(s): tester.
Delivered to: coder — do NOT re-send to those roles; re-issue only the unresolved ones.
```

A distinct exit code for partial vs total (2 = partial) would be better still, but naming `delivered` on stderr is the floor.

## 2. None of this is exercised by CI — reverting either fix stays green · **blocker, cheap**

- `tests/e2e_dispatcher_mode.rs:1` is `#![cfg(feature = "e2e")]`.
- `.github/workflows/ci.yml:166` is `cargo nextest run` — no `--features e2e`. Same at `:267` and `:300`.
- So `orchestration/dispatch/001` and `/002` — the only two tests that would catch a regression — **are never executed by any CI job**.

Your evidence that both tests are load-bearing (*"reverting either fix alone turns this test red"*) is real, but it was established on your machine and CI cannot repeat it. `grep -rn "DelegateResponse\|register_orchestration_role\|send_and_await_reply" src tests` returns **no test file at all** — there is no fast-tier reference to any of the three new APIs.

**Suggested:** three or four fast-tier unit tests beside the existing `delegate_targets_*` tests in `src/state.rs` — they already build an `AppState` by hand and construct `AgentPtyRegistry::new()` (`:4545-4617`, `:4686`), so each is a few lines:

- `handle_delegate` from a pane absent from `pane_role_map` → `resp.error.is_some()`, message contains the pane id;
- from a registered non-orchestrator pane → `resp.error` names the role;
- from a valid orchestrator with `--to` a role having no pane → `unresolved_roles == ["x"]`, `error.is_none()`;
- `register_orchestration_role(pane, role, true, identity, cwd)` → pane lands in `orchestrator_pane_ids`, `pane_role_map`, `pane_orchestration_map` — this one pins the registrar contract both callers now depend on.

Keep the e2e tests as they are: they cover the integration, the unit tests cover the regression.

---

## Suggestions

**3. `is_start_role` vs `orch_idx` — the two spawn paths still diverge, so the same bug survives on `Ctrl+N`.**

```rust
// spawn.rs:567-575
st.register_orchestration_role(&agent.pane_id, &role.role_name, idx == orch_idx, …);
// daemon_protocol.rs:1373-1379
state.write().await.register_orchestration_role(pane_id, &role_name, is_start_role, …);
```

For a toml with two roles and **no** `start = true`: the interactive path sends `is_start_role: role.start` — `false` for every role (`tab.rs:870`) — while the tab separately computes `start_role_index = …unwrap_or(0)` (`tab.rs:919`) and delivers the orchestrator prompt to role 0. Nothing enters `orchestrator_pane_ids`. Role 0 then runs `delegate` and is rejected with — and this is worth reading aloud — *"pane 7 is the `orchestrator` role, not this orchestration's orchestrator, so it may not delegate."* The same input through `dispatch --orchestration` works, because `orch_idx` falls back.

Your own justification for using `orch_idx` — that the raw flag *"would have left this same bug for a narrower input"* — is exactly right, and is applied to only one of the two callers.

**Reachability, honestly:** `config_validation.rs:112-118` reports an Error for an orchestration without exactly one `start = true`. But validation is **lint-only** — `grep -rn config_validation src` shows the sole caller is the `validate` subcommand (`main.rs:1238`); nothing on the load or spawn path enforces it. Reachable, but not a well-formed config, which is why this is a suggestion.

**Suggested:** have `register_orchestration_role` take the orchestrator **index** (or the role list plus index) rather than a `bool`, so neither caller can supply a different rule. A shared registrar that still lets each caller decide *who the orchestrator is* has moved the drift one argument to the left rather than removing it.

**4. Registration also switches on the `work-done` return edge — correct, but untested and unmentioned.**

`state.rs:3161-3166`. Before this PR, a worker in a dispatched/scheduled/issue-dispatched orchestration running `work-done` hit the `None` arm — no `work-done-<role>.md`, no feedback line to the orchestrator. After it, `pane_role_map` is populated for those panes, so **the whole return edge activates for the first time**: `std::fs::write` into `<pane_cwd>/.dot-agent-deck/work-done-<role>.md` (`:3182-3189`, where `pane_cwd_map` is now the dispatch **worktree**) plus `write_to_pane_and_submit` into the orchestrator pane (`:3221`).

That's desirable and almost certainly what a user expects. But it's a **second user-visible path this PR turns on**, with zero assertions: `tests/CATALOG.md` explicitly disclaims it for *both* new tests (*"Also not asserted: the `work-done` return edge"*), and neither the PR body nor `changelog.d/dispatched-orchestration-delegate.bugfix.md` mentions it. If it misbehaves — a worktree path that no longer exists, a stale orchestrator pane — the first report will be confusing, because nothing documents that it started working.

**Suggested:** one sentence in the changelog fragment, and cheaply, extend `orchestration/dispatch/001` to run `work-done` from the dispatched worker's pane and assert the `Worker worker has completed their task.` line reaches the orchestrator's PTY. The roles are `cat`; it costs no tokens and closes a round trip the test already sets up.

**5. The unresolved-role message misdiagnoses self-exclusion.** `main.rs:750-752`. `delegate_targets` excludes the sender's own orchestrator pane by design (`state.rs:2835`), pinned by `delegate_targets_still_excludes_the_sending_orchestrator` (`:4595`). So an orchestration whose orchestrator role is named `lead`, where the orchestrator mistakenly runs `--to lead`, is told: *"reached no worker for role(s): lead. This orchestration has no pane for it — check the role names in .dot-agent-deck.toml."* The agent checks the toml, finds `lead` right there, and has been sent to the one place that cannot explain the failure. Same for `--to orchestrator`. Given the PR's goal is loud *and* informative, worth a distinct message: *"`lead` is this orchestration's orchestrator; an orchestrator cannot delegate to itself."*

**6. `send_and_await_reply` has no read deadline on Unix** (`hook.rs:474-498`) — a wedged daemon hangs the orchestrator, which is the failure class this PR exists to remove.

**7. A fully-defaulted `DelegateResponse` plus `Err(_) => SUCCESS` means any JSON object parses as success.**

---

## Nits

Three, recorded in full in my notes; none worth holding the PR for.

---

## On the open Greptile thread (`src/state.rs:3074`)

**Not re-reporting it** — I verified it, I agree with it, and I agree with deferring it. It's a gap in the *fix* rather than in the original bug.

For what it's worth, between your two proposed directions I'd ship **option 1** (the `pane_is_live` filter at resolution), because it converts a wrong claim into an `unresolved_roles` entry the caller already handles — no new reporting path. And if it stays deferred, the code comment the bot suggested is worth taking: finding 1 above makes the `delivered` list **caller-visible for the first time**, so its exact meaning starts to matter in a way it didn't before.

## Checked and clean

Sender identity and spoofing · cross-orchestration isolation · worker-cannot-delegate (verified the new path does not make it reachable) · pane-id reuse after close · the reuse path · swallowed errors · stale comments · rule 10 in the touched docs.

**Rule 12 — I agree with you: no `PROTOCOL_VERSION` bump and no `.breaking.md`.** The wire shape doesn't move and no existing field changes meaning; `register_orchestration_role` is an internal consolidation, and the newly-populated maps are daemon-local state rather than a contract an older TUI reads. Reasoning written out rather than asserted, since a semantic break behind a stable wire is exactly the case rule 12 exists to catch and it deserved checking rather than assuming.

prageethw:
Supplementary to my review above — a second, independent pass ran over the same diff and reached the same verdict (merge-worthy) by a different route. Rather than repost the overlap, here is the one thing it found that mine did not.

## The changelog owes one sentence about mixed versions

Your fragment's final paragraph covers old-CLI→new-daemon and new-CLI→old-daemon, and frames both as "keeps working". Both claims check out:

- **Old CLI → new daemon:** `DelegateResponse` is a new line written where the daemon previously wrote nothing, on a connection the old CLI closes without reading. The daemon's write fails with `EPIPE`, absorbed by the deliberate `let _ =` in the Delegate arm. Rust installs `SIG_IGN` for `SIGPIPE`, so this is an ignorable error, not a crash. Unaffected, as your matrix says.
- **New CLI → old daemon:** `send_and_await_reply` half-closes, the old daemon's `lines.next_line()` hits EOF, its write half drops, the client reads EOF → `NoReply` → exit 0. Correct — and the reason this needed its own function rather than `request_from_socket`.

**What the fragment does not say** is that in the **new daemon + old CLI** combination, *the fix's whole point is absent*. The daemon computes the response and the CLI discards it, so delegation failures stay silent until the CLI is upgraded too. Someone who upgrades the daemon alone and still sees a hung orchestration has no way to know why — and "keeps working" reads as reassurance that the fix is in effect.

One line would close it:

> The loud failure arrives with the CLI, not the daemon: an older `delegate` binary against a newer daemon still exits 0 — upgrade both.

## Two independent corroborations, for what they are worth

- **Rule 12** — the second pass reached your conclusion independently from source: no `PROTOCOL_VERSION` bump, no `.breaking.md`. Wire shape unchanged, no field changed meaning, additive reply.
- **The Unix read deadline** (`send_and_await_reply`, `hook.rs:474-498`) was found separately by both passes. `IpcClient::connect` seeds a 5s timeout only on Windows, so the half-close bounds an *old* daemon but not a live-but-wedged one — `delegate` can now block where it previously could not. Two reviewers landing on it independently is probably a signal about its priority.

The second pass also read the open Greptile thread first and did not re-report it, but recorded one disagreement with how it was settled — happy to pass that along if useful.

src/state.rs
```diff
@@ -2882,6 +2948,33 @@ impl AppState {
...
+        let delivered: Vec<String> = {
+            let mut seen: Vec<String> = Vec::new();
+            for (role, _) in &targets {
+                if !seen.iter().any(|r| r == role) {
+                    seen.push(role.clone());
+                }
+            }
+            seen
+        };
+        let unresolved_roles: Vec<String> = {
```
prageethw, line 2991: **A partially-delivered delegate reports total failure — and the retry double-dispatches.**

`--to coder --to tester` where only `coder` has a pane:

- `delegate_targets` resolves `coder`, so the fan-out **runs**: the task is written into the coder's PTY and its idle-worker record is armed.
- `delivered = ["coder"]`, `unresolved_roles = ["tester"]`, `error = None`.
- `main.rs` sees a non-empty `unresolved_roles`, prints a message naming **only `tester`**, and returns `ExitCode::FAILURE`.

So the CLI reports failure for a delegation that half-landed, and the stderr never mentions that `coder` did receive it. An orchestrator applying this PR's own new contract — non-zero means it did not land — retries, and the coder gets the same task twice. That arms two idle-worker records for one pane, the second superseding the first, so a single `work-done` leaves one armed: the exact hazard `delegate_targets_de_duplicates_a_repeated_target_role` exists to prevent, reached by a different route.

This is the one finding I'd call a blocker, because it is created by the fix rather than merely uncovered by it: before, everything exited 0, so there was no failure claim to be wrong.

`resp.delivered` is already on the wire and simply unread by the CLI. Suggested:

- exit non-zero only when `delivered.is_empty()` (nothing landed at all), and
- when `delivered` is non-empty *and* `unresolved_roles` is non-empty, report the partial outcome explicitly — naming both sides — so a retry can be aimed at just the roles that missed.

Whatever the exit code ends up being, the message has to name what *did* land, or a retry cannot be made safe.

src/main.rs
```diff
@@ -702,8 +707,56 @@ fn main() -> ExitCode {
...
+                // contract: delivered, unverifiable. See `SocketReply::NoReply`.
+                SocketReply::NoReply => return ExitCode::SUCCESS,
+                SocketReply::Reply(line) => line,
+            };
+            let resp = match serde_json::from_str::<dot_agent_deck::event::DelegateResponse>(&line)
+            {
+                Ok(r) => r,
+                // Same reasoning as `NoReply`: a line we cannot parse is a daemon
+                // we do not understand, not a proven failure.
+                Err(_) => return ExitCode::SUCCESS,
```
prageethw: **A fully-defaulted `DelegateResponse` plus `Err(_) => SUCCESS` makes the success check fail open.**

All three fields of `DelegateResponse` are `#[serde(default)]` and the struct derives `Default`, so `serde_json::from_str::<DelegateResponse>` **succeeds** on any JSON object — `{}`, or a reply belonging to a different verb such as `{"seed":null}` — yielding `delivered: []`, `unresolved_roles: []`, `error: None`. That is indistinguishable from a clean success, so the CLI exits 0.

Combined with `Err(_) => return ExitCode::SUCCESS` on the line above, every failure mode in the parse step resolves to "it worked": malformed line, wrong verb's reply, truncated write, future daemon with a renamed field. The verb whose entire purpose is answering "did this land?" answers yes when it cannot tell.

The `Err(_) => SUCCESS` arm itself is defensible — an unparseable line from an unknown daemon build is not proof of failure, same reasoning as `NoReply`. The gap is that a *parsed* response carries no positive evidence that it is a delegate response at all.

Suggested: make success affirmative rather than residual. A `#[serde(deny_unknown_fields)]`-style tightening is too brittle across versions, so cheaper: require one non-defaulted discriminator — e.g. a `#[serde(default)] kind: Option<String>` set to `"delegate"` by the daemon, or make `delivered` non-`default` so a response that omits it fails to parse and falls into the (correct) `NoReply`-equivalent branch.

src/spawn.rs
```diff
@@ -530,6 +538,53 @@ pub async fn spawn(
...
+                        // orchestrator context and the caller's task below — and
+                        // it falls back (role named `orchestrator` → any
+                        // `start = true` → role 0) where `is_start_role` alone
+                        // would be false for EVERY role of an orchestration whose
+                        // toml sets no `start`. Registering on the raw flag would
+                        // leave such an orchestration with a context-bearing
+                        // orchestrator that is still not in
+                        // `orchestrator_pane_ids`, i.e. this same bug for a
+                        // narrower input.
+                        idx == orch_idx,
```
prageethw, line 599: **The registrar was unified; the rule for *who is the orchestrator* was not — so the two paths still disagree.**

This path passes `idx == orch_idx`. The `StartAgent` path passes the raw `is_start_role`, which comes from `src/tab.rs` as `is_start_role: role.start` — the bare toml flag, no fallback.

For a toml whose role is **named `orchestrator` but sets no `start = true`**:

- **Dispatched** — `orchestrator_role_index` matches on the name, this pane enters `orchestrator_pane_ids`, `delegate` works.
- **`Ctrl+N`, same toml, same directory** — `is_start_role` is `false` for every role, nothing enters `orchestrator_pane_ids`, and the orchestrator's `delegate` is rejected with *"pane 14 is the `orchestrator` role, not this orchestration's orchestrator, so it may not delegate"* — a sentence that contradicts itself on the way past.

The comment right here makes the argument for why the raw flag is wrong (*"`is_start_role` alone is false for EVERY role of an orchestration whose toml sets no `start`… i.e. this same bug for a narrower input"*), and that argument applies verbatim to the path it leaves alone. Same-config-different-spawn-path divergence is the exact drift `register_orchestration_role`'s docstring says the shared function exists to end.

Not a blocker — the PR *improves* the interactive case by making that rejection loud instead of silent, and `validate` flags such a config (lint-only, so it does not prevent the config existing). But a reader of the new comment will fairly conclude both paths now agree, and they do not.

Suggested: decide it once — have `tab.rs` compute the flag with `orchestrator_role_index` before sending `OrchestrationSpawnMeta`, or move the decision inside `register_orchestration_role` (take the role list plus index rather than a pre-computed bool) so neither caller can get it wrong. Failing that, say in this comment that `StartAgent` knowingly still uses the raw flag.

prageethw (commented):
## Review — merge-worthy, with two worth fixing first

Independent read-only review pass over the full +846/−33 diff at `0935b27`, plus a separate audit pass. No tests run, no binary started; every claim below is read out of the blob at that SHA.

**Verdict: merge.** The diagnosis is right and well-evidenced — only `AttachRequest::StartAgent` ever populated the daemon's role maps, so anything reaching `spawn_agent` via `spawn::spawn` was unroutable — and the single shared `register_orchestration_role`, called synchronously before `spawn` returns, is the right shape and the right moment. The loud-failure half is a real improvement: a worker's `delegate`, previously dropped with only a `warn!`, now exits non-zero and says why.

Two findings I'd take before merging, both inline:

1. **A partially-delivered delegate reports total failure** (`src/state.rs`) — the one with a live failure mode: it can cause a double-dispatch.
2. **Any JSON object parses as success** (`src/main.rs`) — a fully-defaulted `DelegateResponse` plus `Err(_) => SUCCESS` means the CLI's "did it land?" answer fails open.

Then four more inline: the `orch_idx` vs `is_start_role` asymmetry, no read deadline on Unix, the unresolved-role error text, and CI coverage.

### Two things not filed inline

**The `work-done` return edge silently switches on for daemon-spawned orchestrations.** `handle_work_done` routes through `orchestrator_for_worker`, which reads the same `pane_orchestration_map` / `orchestrator_pane_ids` this PR now populates from the dispatch path. Before, a dispatched worker's `work-done` resolved no orchestrator and wrote no feedback; after, it resolves and writes into the orchestrator pane. That is almost certainly wanted — an orchestration where delegation works but completion never returns would be half-fixed — but it is a behaviour change on a path the PR does not mention, and both new CATALOG entries explicitly disclaim the return edge (`Does not assert: the work-done RETURN edge`). Worth a sentence in the changelog and, ideally, one assertion.

**Rule 12 — I agree with your call: no `PROTOCOL_VERSION` bump, no `.breaking.md`.** Reasoning reached independently: `DaemonMessage::Delegate` and `DelegateSignal` are untouched; the reply is a new line on the hook socket (not the attach protocol `PROTOCOL_VERSION` governs); no existing field changed meaning; and an old CLI closing without reading leaves the daemon's write to fail `EPIPE` into the deliberate `let _ =`, which is ignorable rather than fatal since Rust installs `SIG_IGN` for `SIGPIPE`. The cross-version matrix in the PR body matches the code.

One line still owed in the changelog fragment, though: it says mixed versions "keep working" in both directions, but not that **new daemon + old CLI has none of the fix** — the daemon computes the response and the old CLI discards it, so failures stay silent until the CLI is upgraded too. Someone who upgrades only the daemon and still sees a hung orchestration has no way to know why.

### Checked and clean

Stating these so the absence of findings is readable rather than assumed:

- **Pane-id reuse** — `next_pane_id` draws from a process-global `PANE_COUNTER.fetch_add`, so no two panes of one daemon share an id and a re-fire cannot overwrite a prior fire's entries.
- **Orchestration matched by identity, not by string** — `mint_orchestration_id()` runs once per spawn and the registered identity is `Instance { id, name }`, so `delegate_targets`' equality compares an instance token. Two dispatches of the same orchestration in the same directory stay separate routing groups, matching what `StartAgent` derives.
- **"Unknown pane" is an error, never a default** — `handle_delegate` returns before touching any routing, and `register_orchestration_role` is called only from the `SpawnTarget::Orchestration` branch, so single-agent and dashboard panes are never registered.
- **Workers still cannot delegate, and cross-orchestration delegation is still blocked** — the `orchestrator_pane_ids` gate is unchanged and `delegate_targets` still scopes by identity; `delegate_targets_never_cross_delivers_between_same_name_cwd_tabs` still applies untouched.
- **Sender spoofing** — `pane_id` is attacker-controlled (`DOT_AGENT_DECK_PANE_ID`), but the hook socket is created owner-only (`with_socket_umask` + `set_endpoint_mode_owner_only`), so this needs same-uid local execution — inside the agent's trust boundary already, and not widened here. Minor new side-effect worth knowing rather than changing: the loud errors are a success oracle for enumerating valid orchestrator pane ids, where the old unconditional exit 0 leaked nothing.
- **No re-entrant lock from the new `state.write().await` inside `spawn`** — the daemon's `Dispatch` arm does the worktree+spawn work outside any `AppState` lock (as its own comment says), the scheduler callbacks capture `state.clone()` rather than a guard, and the write holds no `await` inside its loop.
- **Empty `--to`** — rejected client-side before the signal is built, so it is not a silent-success hole. **Duplicate `--to`** — de-duped in `delegate_targets` and again in `delivered`, so no spurious failure.
- Swallowed errors (`let _ =` on the reply write and the half-close) are all deliberate and commented. Rule 10 holds in the changelog fragment and both CATALOG entries.

The open Greptile thread on `src/state.rs:3074` (stale registration → false `delivered`) is **not** re-reported — verified, agreed, and reasonably deferred. One small thing: the thread's own fallback, annotating the `DelegateResponse` return with the staleness caveat, costs a comment line and wasn't taken.

Nice work on the test structure, incidentally — running the `Ctrl+N` control *first* and reporting its failure as a harness fault rather than a product result is the right shape for this bug, and the `wait_for_delegate_pointer` docstring documenting the respawn/`.find()` coin flip is the kind of thing that usually has to be discovered twice.

src/hook.rs
```diff
@@ -438,6 +438,66 @@ fn build_opencode_event(input: OpenCodeHookInput) -> Option<AgentEvent> {
...
+}
+
+/// Send a line to the daemon hook socket and classify what came back.
+///
+/// [`request_from_socket`]'s `None` cannot distinguish "no daemon" from "old
+/// daemon that does not answer this verb", which is fine for `get-seed` (both
+/// degrade to "no seed") and wrong for `delegate`, where the first is a failure
+/// the orchestrator must see and the second must stay a success or every
+/// delegate against an older daemon starts reporting a phantom error.
+pub fn send_and_await_reply(json: &str) -> SocketReply {
```
prageethw, line 477: **No read deadline on Unix — `delegate` can now block where it previously could not.**

`send_and_await_reply` never calls `set_timeouts`, and the two platforms are not symmetric:

- **Windows** — `IpcClient::connect` seeds `timeout_ms` with `SYNC_CLIENT_DEFAULT_TIMEOUT = Duration::from_secs(5)`, so `SocketReply::NoReply`'s docstring is accurate there.
- **Unix** — `IpcClient::connect` is `Ok(Self(UnixStream::connect(endpoint)?))` and nothing else. `set_timeouts` exists but is opt-in, and this function does not use it. The `read_line` below is unbounded.

The half-close covers the case the docstring reasons about — an *old* daemon's line reader hits EOF, its task ends, the write half drops, the client reads EOF. It does not cover a **live** daemon that accepts and then does not answer promptly. `handle_delegate` runs under `state.read().await`, and tokio's `RwLock` is write-preferring, so a queued `state.write().await` — including the one this PR adds inside `spawn` for a large orchestration — parks subsequent readers behind it. In that window `dot-agent-deck delegate` blocks with no ceiling, where `send_to_socket` previously wrote, flushed and exited.

An orchestrator whose `delegate` hangs is the same hung orchestration this PR set out to remove, reached through a different door. The Windows backend names the hazard in as many words: *"an unbounded read against a wedged — or merely older, non-answering — daemon would hang the caller forever… this is the floor that keeps the ones that don't from wedging."* Unix has no floor.

Suggested — one line after connect, matching the Windows budget:

```rust
// Match the Windows client's floor. The half-close bounds an OLD daemon;
// nothing bounds a live one that is slow to answer.
let _ = stream.set_timeouts(std::time::Duration::from_secs(5));
```

and name `ErrorKind::TimedOut | ErrorKind::WouldBlock` explicitly in the match, so a timeout is visibly a deliberate "delivered, unverifiable" rather than falling through the catch-all.

Minor, same function: `// Past this point the daemon HAS the signal` overclaims — `write_all` + `flush` puts bytes in the kernel buffer, not in the daemon. A daemon killed between accept and read yields EOF → `NoReply` → exit 0, a silent success in the class this PR removes. Unavoidable given the old-daemon ambiguity and the right way to err, so not a defect; but the comment should say the bytes have left *this process*, and `NoReply` should read "handed to the socket, unverifiable" rather than "Delivered".

src/main.rs
```diff
@@ -702,8 +707,56 @@ fn main() -> ExitCode {
...
+            };
+            if let Some(error) = resp.error {
+                eprintln!("Error: delegate from pane {pane_id_for_report} failed: {error}");
+                return ExitCode::FAILURE;
+            }
+            if !resp.unresolved_roles.is_empty() {
```
prageethw: **The message asserts a cause it hasn't established.**

Three different conditions land in `unresolved_roles`, and *"check the role names in .dot-agent-deck.toml"* is right for only one:

1. the role really is absent from the toml — correct;
2. the named role **is the sending orchestrator itself** — `delegate_targets` excludes orchestrator panes, so `--to orchestrator` from the orchestrator sends the agent to inspect a role name that is sitting correctly in the file;
3. the role's worker pane is gone — same message, same wrong instruction.

An orchestrator that fans out to every role name it read in its own context file, including its own, gets told to go fix a config that is not wrong. That relocates the confusion rather than removing it, which is most of the value of making this loud.

Suggested — say what *did* resolve, which separates all three causes for free and needs no new field, since `delivered` is already computed and already on the wire:

```
Error: delegate from pane {pane} reached no worker for role(s): {unresolved}.
Roles with a live worker pane in this orchestration: {delivered}.
(An orchestrator cannot delegate to itself, and a closed worker pane stops
being a target.)
```

This overlaps with the partial-delivery finding on `src/state.rs` — both want `delivered` in the output — so one change likely closes both.

tests/e2e_dispatcher_mode.rs
```diff
@@ -753,6 +1077,113 @@ fn orchestration_dispatch_002_every_real_agent_role_comes_alive() {
...
+    );
+    let coder_pane = role_panes_in(deck.attach_socket_path(), ORCH, |_| true)
+        .get("coder")
+        .cloned()
+        .expect("the dispatched orchestration has a `coder` role pane (asserted above)");
+    let orchestrator_pane = role_panes_in(deck.attach_socket_path(), ORCH, |_| true)
+        .get("orchestrator")
+        .cloned()
+        .expect("the dispatched orchestration has an `orchestrator` role pane");
+    let _ = &coder_pane; // named for the failure message below
```
prageethw: **Nothing that blocks a merge would fail if the rejection went silent again.**

The tests themselves are good — this is about the tier they live in:

- All three CI build jobs run `cargo nextest run` with **no** `--features e2e` (`ci.yml:166`, `:267`, `:300`), and every `tests/e2e_*.rs` opens `#![cfg(feature = "e2e")]`, so both new tests compile to nothing in CI. The workflow's own comment at `ci.yml:236-237` says so.
- `orchestration/dispatch/002` additionally needs real Claude credentials, so it self-skips even where `e2e` is enabled.
- `grep -rn "DelegateResponse\|unresolved_roles\|register_orchestration_role" tests/` returns **nothing**.

So a later refactor returning `DelegateResponse::default()` from the unknown-pane path — restoring the exact silent success being fixed — passes every gate that gates a merge. The `dispatch/001` exit-code assertions do pin this properly; they just aren't run where it counts.

Suggested: three cheap unit tests beside the existing `delegate_targets_*` tests in `src/state.rs`. `handle_delegate` is `async` but needs only an `Arc<AgentPtyRegistry>` and a `broadcast::Sender`, both constructible in-process (`AgentPtyRegistry::new()` is already used in that module's tests):

- unknown pane → `error.is_some()`;
- registered non-orchestrator pane → `error.is_some()`;
- valid orchestrator, `--to nonexistent` → `unresolved_roles == ["nonexistent"]`, `error.is_none()`;

plus one for `register_orchestration_role` asserting `idx == orch_idx` lands the pane in `orchestrator_pane_ids`. A partial-delivery case (`--to coder --to tester`, only `coder` present) would also pin the `src/state.rs` finding above.

Tiny nit while here: `let _ = &coder_pane;` on this line is a leftover — `coder_pane` is genuinely used in the assertion's format string below (`"Orchestrator pane {orchestrator_pane}, coder pane {coder_pane}."`), so the discard and its comment can go. Same test resolves `role_panes_in(deck.attach_socket_path(), ORCH, |_| true)` twice back to back; binding it once would read better.

src/state.rs
vfarcic: Fixed as suggested, and thank you — this was the right call. The retry chain is the part that makes it a blocker rather than a cosmetic message bug, and you are right that the fix *created* it: before, everything exited 0, so there was no failure claim to be wrong.

`resp.delivered` is now read. The rule is exactly the one you proposed:

* `error` set → **failure** (nothing was dispatched);
* `unresolved_roles` non-empty and `delivered` **empty** → **failure** (every named role missed);
* `unresolved_roles` non-empty and `delivered` **non-empty** → **not a failure**, exit 0, with the partial outcome named on stderr.

```
$ dot-agent-deck delegate --to worker --to nonexistent-role --task 'Half-landed.'
Warning: delegate from pane 14 reached no worker for role(s): nonexistent-role.
It WAS delivered to: worker. Retry only the roles that missed — re-sending the
whole delegate would dispatch the delivered roles a second time. (A role reaches
no worker when it is absent from .dot-agent-deck.toml, when it is the delegating
orchestrator itself — an orchestrator cannot delegate to itself — or when its
worker pane has been closed.)
$ echo $?
0
```

The decision moved out of the `Delegate` arm into a pure `delegate_verdict(pane_id, &resp) -> DelegateVerdict` in `src/main.rs`, so the contract is unit-testable — see the tests thread; `delegate_verdict_does_not_fail_a_partial_delivery` is the one that pins this finding, and `handle_delegate_reports_a_partial_delivery_as_partial` in `src/state.rs` pins the daemon half (`--to coder --to tester`, only `coder` present → `delivered == ["coder"]`, `unresolved_roles == ["tester"]`, `error == None`). `orchestration/dispatch/001` gained the same case from the real CLI.

One thing I did *not* do: I kept the daemon side unchanged. `delivered` and `unresolved_roles` were already both computed and both on the wire — the defect was entirely that the CLI read one and ignored the other — so the fix is one-sided by design, and the comment in `handle_delegate` now says so rather than leaving the next reader to rediscover why both fields exist.

src/main.rs
vfarcic: Fixed, and merged with the partial-delivery finding as you predicted — one change closed both, because both wanted `delivered` in the output.

The message no longer asserts a cause. It names all three, since the CLI cannot tell them apart and should not pretend to:

```
Error: delegate from pane 14 reached no worker for role(s): reviewer. No role in
this orchestration received it. (A role reaches no worker when it is absent from
.dot-agent-deck.toml, when it is the delegating orchestrator itself — an
orchestrator cannot delegate to itself — or when its worker pane has been closed.)
```

I went with stating the three causes rather than your "Roles with a live worker pane in this orchestration: {delivered}" line, for one reason: in the *total* failure case `delivered` is empty by definition, so that sentence would render as an empty list exactly when the user most needs to know why. `delivered` does appear in the partial case, where it is non-empty and load-bearing (it is what makes a retry aimable) — see the other thread. If you would rather have both — the causes *and* an explicit "roles that did resolve: …" even when empty — say so and I will add it.

`delegate_verdict_fails_when_nothing_landed` asserts all three causes are present in the string, so a future edit that quietly reverts to "check the role names" fails a test rather than relocating the confusion again.

src/main.rs
vfarcic: Fixed with the discriminator, and you are right about the shape of the gap: the `Err(_) => SUCCESS` arm is defensible, and the problem was that a *parsed* response carried no positive evidence of being a delegate response.

`DelegateResponse` gains `#[serde(default)] pub kind: Option<String>`, and the daemon stamps `"delegate"` on every reply. Read straight off the sandbox socket:

```json
{"kind":"delegate","delivered":[],"unresolved_roles":[],"error":"the daemon holds no orchestration role for pane xcaller, …"}
```

The CLI's parse step is now `parse_delegate_reply(&line) -> Option<DelegateResponse>`, which returns `None` both for a line that fails to parse **and** for one that parses without the marker. `None` takes the `NoReply`-equivalent branch — handed to the socket, unverifiable, exit 0 — which is the branch you identified as correct. `{}`, `{"seed":null}` and `{"kind":"get-seed"}` are all covered by `parse_delegate_reply_requires_the_delegate_marker`.

I preferred the discriminator over making `delivered` non-`default` because it says what it means: a *future* daemon that renames `delivered` (your fourth failure mode) would also fall out of the non-default version, but so would a daemon that legitimately omits an empty vector, and the reader could not tell which. `kind` is unambiguous.

Two details worth calling out:

* **`Default` is hand-written now**, not derived. It has to be: the daemon builds its early returns with `..Default::default()`, so a derived `Default` would leave `kind: None` on exactly the rejection paths this PR exists to make loud, and the CLI would treat its own daemon's error as an unrecognised reply and exit 0. The serde field default stays `None` (that is the whole point of the check); only the Rust `Default` stamps the marker.
* **Rule 12 re-run, since you flagged it.** Redone against **v0.36.0** (the PR body's original run was v0.35.9), sandboxed sockets + `HOME` + state dir + `DOT_AGENT_DECK_LOG`:

  | case | exit | ms | result |
  |---|---|---|---|
  | new CLI → new daemon | 1 | 7 | loud message naming the pane |
  | new CLI → **old** daemon | 0 | 6 | half-close ⇒ prompt `NoReply`, no false failure, no stall |
  | **old** CLI → new daemon | 0 | 5 | signal received and handled — unchanged |
  | old CLI → old daemon | 0 | 6 | control |
  | new CLI → new daemon, **after** an old caller hung up | 1 | 6 | the reply-write to a departed caller does not wedge the hook loop |

  The last row is the one the extra field could plausibly have broken and did not. Analysis unchanged: additive, no `PROTOCOL_VERSION` bump, no `.breaking.md`. The PR body carries the new table.

src/hook.rs
vfarcic: Fixed — and merging with `main` turned this into a bigger finding than either of us thought, in a good way.

**Your diagnosis was exactly right.** `send_and_await_reply` set no timeouts at all; Windows got its floor from `IpcClient::connect`'s 5s default and Unix got nothing, so the `read_line` was unbounded against a live-but-slow daemon — and `handle_delegate` running under `state.read().await` against a write-preferring `RwLock` is precisely the window where that bites.

**But rather than add the one line, I deleted the function's body.** #419 landed on `main` while this branch was open, and it introduced its own `hook::SocketReply` — with a doc comment that reads:

> richer than `request_from_socket`'s `Option<String>` because a caller that needs to tell "never even reached the daemon" apart from "reached it, but got no confirmation back" **(as the not-yet-submitted `delegate` wiring will)** can report each honestly

That is this caller. The two enums merge **textually without a conflict marker** and then fail to compile — `SocketReply` defined twice, conflicting `Debug` impls — while GitHub reported the PR `MERGEABLE` throughout. So this PR would have merged green and broken `main`. Worth knowing independently of the timeout.

The resolution is one change for both: this branch's duplicate is gone, and `send_and_await_reply` is now

```rust
pub fn send_and_await_reply(json: &str) -> SocketReply {
    request_from_socket_inner(json, Some(DELEGATE_REPLY_TIMEOUT))
}
```

which is `get-seed`'s transport. That is strictly better than the idle timeout you suggested, and it is your own reviewer's work doing the lifting:

* the deadline is **total-operation**, not per-read, so the slow-drip peer `error/socket/005` exercises cannot outlast it either;
* the clock starts **before connect**, so connect and the write count against the budget;
* the half-close and the read-exactly-one-line behaviour PRD #163 M4 needed for Windows come along unchanged;
* `error/socket/003` and `error/socket/005` already prove the bound, so `delegate` is covered by tests of the *shared* implementation rather than by a bespoke test of a duplicate one.

`DELEGATE_REPLY_TIMEOUT` is 5s, in `src/hook.rs` — which also repairs a dangling reference: `GET_SEED_REQUEST_TIMEOUT`'s comment on `main` already argues its own value by reference to "the `delegate` reply already bounded at `DELEGATE_REPLY_TIMEOUT` (5s, `src/main.rs`)", a symbol that did not exist anywhere. It exists now, and the parenthetical points at the right file.

**On #435** — agreed, and deliberately handled: this change adds **no** new timeout machinery and touches no part of `IpcClient`, so there is nothing here for the connect-side fix to collide with. When #435 lands, `delegate` inherits it for free, because it is now the same code path as `get-seed`. Noted in the PR description too, so whichever of us lands second knows to rebase.

**The two comment nits, both taken.** `// Past this point the daemon HAS the signal` is gone — you are right that `write_all` + `flush` only proves the bytes left this process. And `NoReply`'s doc now reads "handed to the socket, unverifiable", and says explicitly that a daemon killed between accept and read lands here too, so it is deliberately *not* a claim of delivery.

I did not name `ErrorKind::TimedOut | ErrorKind::WouldBlock` in a match, because there is no longer a match to name them in — `read_reply_line` folds every read error into `None` → `NoReply` one layer down, and re-splitting it there would be a change to `get-seed`'s path for a distinction neither caller acts on. If you would rather the timeout be *visibly* distinct end to end, the honest way is a fourth variant (`SocketReply::TimedOut`) that `delegate` reports as "handed over, daemon did not answer in 5s" — happy to do it, but it is a wider change than this thread and I did not want to make it unasked.

src/spawn.rs
vfarcic: Valid, and taken as the documentation-only resolution you offered as the floor — plus a follow-up issue, **#523**, rather than a bare comment.

The comment at this site now says the `StartAgent` path knowingly still uses the raw flag, names the config shape that diverges, and points at the issue.

**Why not unify it here.** I started on your first suggestion (have `tab.rs` compute the flag with `orchestrator_role_index` before sending `OrchestrationSpawnMeta`) and stopped when I found there are **three** rules, not two:

| where | rule |
|---|---|
| `spawn::orchestrator_role_index` | name `orchestrator` → any `start = true` → role 0 |
| `tab.rs` → `is_start_role: role.start` | the bare flag |
| `tab.rs` → `start_role_index` (`position(|r| r.start).unwrap_or(0)`) | the bare flag, else role **0**, no name fallback |

The third one drives default focus, the all-clear focus move, and **orchestrator-prompt delivery**. So for your example config — role named `orchestrator`, no `start = true` — the `Ctrl+N` path does not merely fail to register an orchestrator: it delivers the orchestrator context to whichever role happens to be first. Changing the membership flag alone would leave `tab.rs` internally inconsistent (one field meaning "the routing orchestrator", another meaning "the prompt target"); changing both is a user-visible TUI change that owes tests under rule 4, and it is not what this PR is about.

I also rejected the cheap version — also treating `role_name == "orchestrator"` as the orchestrator in the `StartAgent` handler — because for a config with **both** a `start = true` role and a differently-positioned role named `orchestrator` it registers **two** orchestrator panes, and `orchestrator_for_worker` then picks between them by `HashSet` order. That re-opens exactly the non-determinism PRD #140 closed, which is a bad trade inside a routing fix.

Your second suggestion (move the decision inside `register_orchestration_role`, taking the role list plus index) is in the issue as the preferred shape. The `StartAgent` handler has only one pane's membership at hand, so it needs the list plumbed to it — worth doing, still not local.

One thing this PR does improve for the interactive case, as you noted: the rejection is now loud instead of silent, so a user hitting the divergence gets *"pane 14 is the `orchestrator` role, not this orchestration's orchestrator"* rather than a hang. Self-contradictory, and #523 has it verbatim as the symptom to fix.

tests/e2e_dispatcher_mode.rs
vfarcic: Fixed — this was the most valuable finding in the review, and your diagnosis of the tier is exactly right: `grep -rn "DelegateResponse\|unresolved_roles\|register_orchestration_role" tests/` returning nothing while both new tests sit behind `#![cfg(feature = "e2e")]`, with no CI job passing that feature, means the rejection could have gone silent again with every gate green.

**Ten unit tests, all in the fast tier.** Five in `src/state.rs`, beside the `delegate_targets_*` tests exactly as you suggested — `handle_delegate` needs only an `Arc<AgentPtyRegistry>` and a `broadcast::Sender`, both constructible in-process, and the detached fan-out means an empty registry does not affect the response under test:

* `handle_delegate_rejects_an_unknown_pane` → `error.is_some()`, and the message names the pane (the literal `sched-dispatch-…-r0` shape from the report);
* `handle_delegate_rejects_a_registered_non_orchestrator_pane` → `error.is_some()`, naming the pane and the role it actually holds;
* `handle_delegate_reports_a_role_that_resolves_to_no_pane` → `unresolved_roles == ["nonexistent"]`, `error.is_none()`;
* `handle_delegate_reports_a_partial_delivery_as_partial` → your `--to coder --to tester` case, which also pins the blocker;
* `register_orchestration_role_makes_orch_idx_the_orchestrator` → a no-`start`-anywhere role list, asserting the pane at `orch_idx` is the *only* member of `orchestrator_pane_ids` and can then delegate.

Five more in `src/main.rs`. Pinning `handle_delegate` alone would have left the other half of the contract — what the CLI *does* with the response — untested, and that is where the blocker lived, so the decision moved into a pure `delegate_verdict(pane_id, &resp)` plus `parse_delegate_reply(line)`. Those cover the full-delivery, routing-error, nothing-landed, half-landed and unrecognised-reply cases.

Both of your nits are gone: `let _ = &coder_pane;` (and its comment), and the doubled `role_panes_in(...)` call — now bound once and read twice.

`dispatch/001` also gains the half-landed case from the real CLI (`--to worker --to nonexistent-role` → exit 0, stderr naming both sides), so the e2e tier still pins the contract end to end where it *is* run; it is simply no longer the only thing that does.

src/state.rs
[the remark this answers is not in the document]
vfarcic: Adjudicated, and split out as **#524** rather than fixed here — but with a correction to the analysis in my own earlier reply, because option 1 turns out to be the wrong fix, not merely the cheaper one.

**Evidence gathered since.** `write_to_pane_internal` does filter on `!exited` (round-11 reviewer #A), so a write to a dead worker really does return `AgentPtyError::NotFound` — greptile's mechanism holds. But that path is only reached for a `clear = false` role. For `clear = true` — the default — `dispatch_one_owned` calls `respawn_agent_for_pane` first, and that function deliberately does the opposite:

> The `exited` filter is deliberately omitted: a dead-but-not-yet-reaped agent should also be replaced … `clear = true` on a crashed agent should still produce a fresh worker.

So for the common config the delegate genuinely lands: the crashed worker is replaced and the fresh one receives the task.

**Which kills option 1.** Filtering the resolved set on `pane_is_live` would report a crashed `clear = true` worker as an unresolved role and fail the delegate — converting a delivery that works today into a reported failure, in exactly the direction this PR is trying to improve. Liveness is not decidable at resolution time without also knowing the role's `clear` flag, and that lives in `.dot-agent-deck.toml`, loaded per-target inside the detached dispatch task. I had offered option 1 as the cheap ship-it-now fix; it is not safe, and I would rather say so than take it.

That leaves option 2 (clean up on unexpected exit) as the right fix, which is what #524 describes — with the blast radius you both named, plus one more: it must not race a `clear = true` respawn, which legitimately has a window where the old child is dead and the new one is not yet registered.

**And greptile's third suggestion is taken.** `DelegateResponse::delivered`'s doc comment now states the caveat outright — that it means "queued to a pane that resolved at delegate time", that the detached fan-out means no synchronous reply can promise the worker read it, and that a worker which exited without `StopAgent` still resolves here — so the next reader does not have to reconstruct this analysis. The residual false success is now narrower than it looks (`clear = false` roles, and panes gone from the registry entirely) and it is written down rather than implied.

@prageethw — flagging in case you disagree with deferring; the scope call is "the reported defect is fixed and covered; this adjacent case is a pre-existing staleness that the fix gives a consequence to".

vfarcic:
## Round-2 gates

Run on the merged branch (this branch + current `main`, v0.36.0):

| gate | result |
|---|---|
| `cargo clippy --workspace --all-targets --features e2e -- -D warnings` | clean |
| `cargo test-fast` | **2533 passed**, 0 failed |
| `cargo xtask linkage-check` | ok (504 catalog ids, 432 annotations, 8 rules) |
| `cargo test-e2e` (full tier) | **7822 / 7824 passed**; the 2 failures **both pass in isolation** |

The two:

* `dispatch_013_orchestration_surfaces_and_delegates` — failed its own *precondition* under full-suite load ("the daemon must clone from GitHub + worktree + spawn BOTH role agents within 120s … Records: []"). No agent ever spawned, so nothing delegate-related was reached. **Passes in 10.2s alone.**
* `shell_activity_006_real_claude_bash_call_crossing_the_cap_keeps_the_badge_working` — self-diagnosing panic: "*This is a harness flake (prompt adherence), NOT a badge regression: rerun rather than reading this as BADGE WRONG*" — the model never issued the Bash tool call. **Passes alone (128s).**

Both are real-agent, load-dependent, and unrelated to this change; `dispatch_013` was already named as such in the PR description. The three failures the original description reported as pre-existing (`idle_worker_011`, `card_stats_005`, `focus_007`) **now pass** — they were fixed on `main` in the meantime, which is one more reason the merge was worth doing.

`orchestration/dispatch/001` (including the new half-landed case from the real CLI) and `orchestration/dispatch/002` (three real Haiku agents) both pass in the full run and in isolation.

### Rule 12 cross-version check — re-run for this round

Redone because round 2 moved the reply shape again (the `kind` discriminator). Against **v0.36.0**, sandboxed sockets + `HOME` + state dir + `DOT_AGENT_DECK_LOG`:

| case | exit | ms | result |
|---|---|---|---|
| new CLI → new daemon | 1 | 7 | loud message naming the pane |
| new CLI → **old** daemon | 0 | 6 | half-close ⇒ prompt `NoReply`, no false failure, no stall |
| **old** CLI → new daemon | 0 | 5 | signal received and handled — unchanged |
| old CLI → old daemon | 0 | 6 | control |
| new CLI → new, **after** an old caller hung up | 1 | 6 | the reply-write to a departed caller does not wedge the hook loop |

Verdict unchanged: additive, no `PROTOCOL_VERSION` bump, no `.breaking.md`.

### One thing worth knowing beyond the review

Merging `main` in surfaced a **semantic merge conflict that GitHub reported as `MERGEABLE` throughout**: #419 landed its own `hook::SocketReply` while this branch was open, and the two merge textually with no conflict marker and then fail to compile. This PR would have merged green and broken `main`. Resolved by deleting this branch's duplicate and routing `delegate` through the transport #419 built — which is also what closes the read-deadline finding, so the two are one change.

prageethw (approved):
Approving `3f272af`. I had one question worth more than a re-read — whether a new CLI against an old daemon can hang — so I had that dimension audited specifically. It is bounded.

## New CLI → old daemon: bounded, no hang

`send_and_await_reply` carries a five-second total-operation budget (`src/hook.rs:458, 477-479`). It starts **before** connect (`:605-613`), is installed on the socket (`:618-630`), and is re-checked before every read (`:659-671, 696-700`). On the expected path the client half-closes its write side so the old daemon's reader reaches EOF (`:636-640`); the old daemon closes without a response, `n == 0` with an empty line returns `None` (`:700-715`), which becomes `SocketReply::NoReply` (`:653-656`) and exit success (`src/main.rs:875-880`). If EOF somehow never arrives, the same deadline still produces `NoReply`. There is no unbounded wait on either path.

## Old CLI → new daemon: fully best-effort

Confirmed the `let _ =` covers **both** the `write_all` and the `flush` (`src/daemon.rs:1578-1582`), and that a serialization failure only skips the reply under `if let Ok(json)` (`:1578`). An `EPIPE` from an old CLI closing without reading cannot propagate out of the per-connection task, let alone terminate the daemon.

## `PROTOCOL_VERSION` correctly stays at 7

The discriminated response is a JSON line on the **hook socket**, not the versioned attach framing (`src/daemon_protocol.rs:5-18, 144-155`), and this PR's `daemon_protocol.rs` change is the internal consolidation of `StartAgent` state registration into `register_orchestration_role` (`:1386-1396`). No attach-wire field changed shape, and none changed meaning behind a stable wire — so rule 12 wants neither a bump nor a `.breaking.md`. Worth stating explicitly given how easy that call is to get wrong in the other direction.

## Partial delegation reports honestly

This is the case the PR exists for, so I checked it rather than assuming. Empty `--to` is rejected before any signal is emitted (`src/main.rs:823-826`). Both `delivered` and the de-duplicated `unresolved_roles` derive from the routing result (`src/state.rs:2982-2999`) before any target is queued (`:3021-3085`). The client fails when unresolved roles exist and nothing was delivered (`src/main.rs:652-659`); where some landed it exits success but names both the misses and the deliveries and tells the caller to retry only the misses (`:661-669, 889-897`). No false total-success, and no retry that would double-dispatch the ones that did land.

## Merge integrity

Diffed `909a2de` against both parents for `daemon.rs`, `daemon_protocol.rs` and `spawn.rs`. Synchronous role registration before prompt delivery (`src/spawn.rs:541-567`) and the discriminated response (`src/event.rs:748-811`) both survived intact.

The changelog's mixed-version paragraph is accurate in both directions and matches what I traced.

PR #467, opened 2026-08-09
fix(xtask): reap e2e temp roots by owning-PID liveness, not age alone (#461)

### LINKED ISSUE #461: clean-e2e-tmp: reap by owning-PID liveness, not age alone — 280 provably-dead roots (6.2 GB) refused while swap sat at 5 MiB free
Follow-up to #322, which remains open and is the umbrella for e2e tmpfs exhaustion. This is a narrower, independently fixable defect in the reaper that #322 proposed and that has since landed: **`clean-e2e-tmp` reaps by age alone, so it refuses dirs it can prove are dead.**

## What happened tonight

Ten dispatched agents were running, one about to build and run the full e2e tier. State of the machine:

- `/tmp` (a **14 GB tmpfs**) at **78% — 11 GB used, 3.2 GB free**
- **280** leftover `/tmp/dad-tests-<pid>-*` roots, **6.2 GB** total
- **Swap completely exhausted: 8.0 GiB of 8.0 GiB used, 5 MiB free**
- Every one of the 280 roots had a **dead owning PID** — verified with `kill -0` against the PID embedded in each name. Zero belonged to a live test process.

`cargo xtask clean-e2e-tmp` was run against exactly that state and reported:

```
nothing to reap in /tmp (no owned dirs older than 6h)
```

The oldest root was 4h09m old, so all 280 fell under the 6-hour default. The tool declined to free 6.2 GB of resident memory that provably belonged to no one, at a moment when swap had 5 MiB left and an e2e compile was about to start — the precise conditions under which CLAUDE.md rule 14 warns `rustc` dies with `signal: 9, SIGKILL` and nothing points at the filesystem.

Passing `--older-than 2 --apply` reaped 223 dirs / 5.8 GB, and the effect was not merely disk:

| | Before | After |
|---|---|---|
| `/tmp` used | 11 G (78%) | 4.8 G (35%) |
| tmpfs resident in RAM (`shared`) | 2.8 GiB | 755 MiB |
| Swap free | **5 MiB** | **3.8 GiB** |

## Why it declined

`xtask/linkage-check/src/clean_tmp.rs` filters on age only. `Options` carries `max_age`, `apply`, `include_untagged`; `Candidate` carries `path`, `bytes`, `age`. The module header documents the naming scheme — "The harness nests everything it creates under one `dad-tests-<pid>-*` root per test process" — but that PID is never read back. `DEFAULT_MAX_AGE_HOURS = 6` is the sole guard against deleting a live run's root.

So the tool has a **definitive** ownership signal sitting in the filename and instead uses a **time-based proxy** for it. Age is a proxy for "is anyone still using this"; PID liveness answers it directly.

## Proposed fix

Read the PID out of the `dad-tests-<pid>-*` name and reap any root whose PID is no longer alive, regardless of age. Keep `--older-than` as an additional filter for the cases PID can't settle:

- **PID dead** → reap (this is the common case and needs no age grace at all)
- **PID alive** → never reap, at any age — strictly safer than today, where a long-running suite past the threshold is currently eligible
- **PID unparseable / recycled** → fall back to the existing age rule

Note the PID-alive branch closes a real hazard in the current behaviour: today a genuine e2e run lasting over 6 hours can have its own live temp root reaped out from under it.

Worth considering alongside:

- **Reap automatically at the start of a test run.** The failure is self-amplifying — the more the suite is interrupted, the less headroom the next run gets — and a dead-PID sweep at startup is cheap and needs no human to notice first.
- **The leak rate itself deserves a look.** 280 roots accumulated in roughly **four hours**. The `atexit` hook covers normal exit, so that count is a proxy for how often the suite is being SIGKILLed (nextest `slow-timeout terminate-after`, interrupted runs). That is a lot of abnormal termination and may be worth its own investigation.
- **#322's `TMPDIR`-under-`target/` suggestion still stands** and would make this class of leak land on disk that `cargo clean` already owns, rather than in RAM. This issue is not a substitute for it — it makes the existing tool trustworthy in the meantime.

## Note on the naming scheme

#322's measurements describe **521 `/tmp/.tmp*` dirs**, the `tempfile` crate's default prefix. Tonight's leftovers are all `dad-tests-<pid>-*`, so the tagged-prefix scheme has landed since that report. The counts in the two reports are not directly comparable, and `.tmp*` is now opt-in behind `--include-untagged` for good reason.

---

Measured on Linux, 27 GiB RAM, `/tmp` a 14 GB tmpfs, on `main` at `d3ea031`.

vfarcic:
## The defect

`cargo xtask clean-e2e-tmp` reaped leftover e2e temp roots by **age alone**. On this box it refused 280 provably-dead roots (6.2 GB of a 14 GB tmpfs — i.e. resident RAM) because the oldest was 4h09m against a 6h default, while swap sat at 5 MiB free and an e2e compile was about to start. The harness names every root `dad-tests-<pid>-*`, so ownership was knowable all along — the tool just never read it.

## The fix — three branches, decided by owning-PID liveness

- **PID not alive** → reap, at any age; `--older-than` cannot suppress it.
- **PID alive** (including `EPERM`) → keep, at any age, unconditionally, with no timestamp read anywhere. This also closes a pre-existing hazard where a run lasting past the 6h threshold could have its own live root deleted out from under it.
- **PID unparseable / untagged / no `kill(2)` facility** → the existing age rule, unchanged.

Reporting now attributes every decision (`dead-pid` / `live-pid` / `untagged`) and explains why survivors survived, replacing the misleading `nothing to reap … (no owned dirs older than 6h)`.

## A deliberate deviation from the issue

Issue #461 asked for a fourth branch: a **recycled** PID (alive, but started after the root was created) should fall back to the age rule. It was implemented twice and **deleted both times**:

- v1 derived the process start from the `/proc/<pid>` directory mtime. That is a lazily-instantiated procfs inode timestamp, not a start time, and a reviewer reproduced it **deleting a live e2e root**.
- v2 derived it authoritatively from `/proc/<pid>/stat` field 22 plus `/proc/stat` `btime`. Linux's `getboottime64()` contract says `settimeofday` affects `btime`, while inode timestamps are not retroactively shifted — so a forward wall-clock step (admin action, VM correction, time sync, suspend/resume) makes a *live* owner look recycled and its root deletable.

The comparison is unsound in principle: ordering a process start against a filesystem timestamp must bridge the wall clock, and any forward step biases toward the deletion-unsafe direction. And it buys little — a PID collision is **transient**, so the colliding process exits and the root reaps as `dead-pid` on a later run. The branch was therefore removed rather than patched a third time. Two guards prevent reinstatement: a source-level scan (`source_has_no_pid_recycling_machinery`) and `owner_of`'s timestamp-free signature. Both the module docs and `docs/develop/e2e-temp-dirs.md` state plainly that no *behavioural* test can catch reinstatement and why.

## Files changed

`xtask/linkage-check/src/clean_tmp.rs`, `xtask/linkage-check/Cargo.toml`, `xtask/linkage-check/src/main.rs`, `docs/develop/e2e-temp-dirs.md`, `CLAUDE.md` (rule 14), `Cargo.lock`, `changelog.d/461.misc.md`.

No product, daemon, TUI, protocol, orchestration, or hook code. No `tests/e2e_*.rs`.

## Gates — all green at `5bab8df`

| Gate | Result |
|---|---|
| `cargo clippy --workspace --all-targets --features e2e -- -D warnings` | pass, 0 warnings |
| `cargo test-fast` | pass — 1603/1603 |
| `cargo xtask linkage-check` | pass — 481 catalog ids, 408 annotations, 100 allowlisted, 7 rules |
| `cargo test -p xtask-linkage-check` | pass — 56/56 |
| cross-target `cargo check` (win-msvc, x86_64/aarch64-darwin) | pass, 0 warnings |

Review: three rounds. Reviewer and auditor have both signed off on `5bab8df` with no blockers and no follow-ups.

## E2E tier — run, and not fully green; stated honestly

`cargo test-e2e` on the branch: **3,637 run, 3,633 passed, 4 failed.** All four were investigated:

- `idle_worker_011_silent_worker_prompt_is_visible_in_attached_tui`
- `focus_007_lock_governed_focus_contract_on_real_binary`
- `card_stats_005_real_agent_card_narrows_without_restructuring`

  → all three **measured failing identically on unmodified `origin/main`** (`d02111bc`) in a clean disk-backed worktree: same assertion, same source line, same observable symptom. Pre-existing, not regressions.

- `delegate_015_real_opencode_worker_acts_on_clear_true_delegate` → failed in the full run, **passed in isolation**; confirmed real-model flake.

The branch introduces **no e2e regression**. CI does not run the e2e tier, so this does not affect PR checks.

## Real-world validation

Run against this box's live `/tmp` before the e2e gate:

| | Before | After |
|---|---|---|
| `/tmp` used | 11 G (73%) | 4.5 G (32%) |
| Available | 3.9 G | 9.5 G |

219 directories reaped, ≥4.9 GB freed, **every one labelled `dead-pid`** — zero `untagged` age-rule deletions, and the live-owner roots of other agents working concurrently on the box were left untouched. The old tool would have refused all 219. That is the issue's reported scenario, resolved.

## Not included

Per CLAUDE.md rule 14: no demo reel (the reel adapter selects e2e `#[spec]` tests the branch added or changed; this branch changed none). No `PROTOCOL_VERSION` bump or `.breaking.md` fragment: nothing here touches the daemon, the TUI↔daemon protocol, orchestration, or hooks (CLAUDE.md rule 12 does not apply).

Closes #461.


vfarcic:
**Review routing note — this PR overlaps #472, and the two need a deliberate landing order.**

Both rewrite `xtask/linkage-check/src/clean_tmp.rs`, and both add `libc` to `xtask/linkage-check/Cargo.toml` (here for `kill(2)`, there for `geteuid(2)`). Overlapping files: `clean_tmp.rs`, `xtask/linkage-check/Cargo.toml`, `Cargo.lock`, `docs/develop/e2e-temp-dirs.md`.

They are complementary rather than competing, and that is how #461 framed the relationship: #472 moves the harness roots off the RAM-backed tmpfs so the leak stops mattering, while this PR makes the existing reaper trustworthy in the meantime. The overlap inside the file is largely orthogonal — #472 changes *which roots are scanned* (`--root`, canonicalisation, restricting `--include-untagged`), this changes *how each candidate is classified*.

**Suggested order: this PR (#467) first, then #472 rebases onto it.** Two reasons. It is the smaller and more self-contained of the two. And it has finished review: three reviewer rounds and two security audits with every finding resolved, plus a clean Greptile pass (5/5, zero inline findings). #472's own description records that a second security audit — covering the code it added in response to its first round (`traversal_objection`, `private_dir_objection`, `create_dir_private`, and the `--root` deletion authority) — was commissioned twice and never completed, so that code has had no independent security review yet. That is worth resolving on its own schedule rather than holding this behind it.

Because the ruleset sets `dismiss_stale_reviews_on_push: true`, whichever PR lands second will need a rebase and therefore a second approval. Sequencing them avoids a wasted review pass.

**On the e2e tier**, since it is stated honestly in the description above and could otherwise read as a red flag: the three failures this branch's gate hit — `idle_worker_011`, `focus_007`, `card_stats_005` — are issue #460, and were measured failing identically on unmodified `main` at `d02111bc` (same assertion, same source line, same observable symptom) in a clean disk-backed worktree. #465 already fixes them. The fourth, `delegate_015`, passed in isolation and is a real-model flake. This branch introduces no e2e regression, and CI does not run the e2e tier in any case.

prageethw (commented):
Read-only audit at `e68dd395`, in a detached worktree. The reaper was **never invoked** — not even `--dry-run` — and no tests were run. Everything below is static, with `file:line` for each point.

**Verdict: no blocker in the deletion path.** The decision logic is unusually careful — every ambiguous input lands on the *keep* side, verified branch by branch. Two items worth addressing before merge (**F1**, **F2**); the rest are nits.

---

## F1 — Owner-PID liveness is not *root* liveness

**Suggestion, and it's the one thing I'd want changed.** `clean_tmp.rs:464-467` (`Owner::Dead` → `reap: true`).

- The harness names each root after the **test process**: `tests/common/mod.rs:3431`, `dad-tests-{}-` from `std::process::id()`.
- nextest's `terminate-after` SIGKILLs that test process. The `atexit` cleanup never runs (`:3443-3478`) — this is the leak the tool exists for.
- **But the processes the test spawned don't die with it.** Your own harness says so at `tests/common/mod.rs:1558-1560`: *"The deck's own lazy-spawned daemon setsid's into a separate session and escapes this group."* Corroborated at `src/daemon_attach.rs:250` and `src/agent_pty.rs:787-788`. The group kill at `:1566` can't reach them.
- That orphan lives up to **300s** (`DOT_AGENT_DECK_TEST_MAX_LIFETIME_SECS`, `:647` and `:3621`) — still writing its state dir, sockets and logs under the dead test's root.
- `owner_of` probes the **test** PID, which is dead → `Owner::Dead` → reaped **at any age**. `--older-than` explicitly cannot hold it back.
- The developer then runs `cargo xtask clean-e2e-tmp --apply` — which is the literal command your out-of-space message tells them to run (`:3390`), and running it *immediately* after a dead run is this PR's whole motivation.

**Net:** `remove_dir_all` deletes the working tree of a live daemon. Impact is bounded — test scratch, never user work; the orphan limps on with open fds until its 300s cap — so it's a suggestion, not a blocker. But the review question was *"can the new rule delete something the old one kept?"*, and **this is that class.** The 6h rule was 72× the orphan's max lifetime, so the hazard didn't exist before. The new rule removes protection precisely in the first five minutes, which is when orphans are alive.

**Cheapest fix:** a **~10-minute minimum-age floor on the `DeadPid` branch only**.
- Reintroduces a timestamp, but on the **keep** side — the safe bias, and the inverse of the recycled-PID branch that biased toward deletion and is rightly gone.
- Doesn't weaken #461: the headline case was 280 roots at 4h09m. A 10-minute floor still reaps every one.
- Expressible without touching `owner_of`'s signature, so it doesn't disturb the compile-time guard.

**Minimum acceptable alternative:** keep the behaviour and document the window in `docs/develop/e2e-temp-dirs.md`, so the next person debugging a bizarre orphan failure right after `--apply` finds the explanation.

## F2 — Two safety claims are overstated, and F1 is why

**Suggestion.** Three places say more than the mechanism supports:

- **PR body** — *"closes a pre-existing hazard where a run lasting past the 6h threshold could have its own live root deleted"*. True and a real improvement, but it's about the **owner**. It reads as though the live-root hazard is now closed in general.
- **`clean_tmp.rs:13-14`** — *"The `<pid>` … answers 'is anyone still using this?' directly."* It answers *"is the creator still running?"*. Those differ exactly when a `setsid`'d descendant outlives its parent — which this codebase does deliberately.
- **`docs/develop/e2e-temp-dirs.md`**, same claim in "Ownership decides, age is only the fallback".

**Why it matters:** a future maintainer reads "answers it directly", concludes a dead PID proves the tree is unreferenced, and removes the F1 mitigation — or declines to add one. One sentence in each place fixes it.

---

## Nits

**F3 — the reinstatement guard is bypassable.** `clean_tmp.rs:1494` — `include_str!("clean_tmp.rs")` scans one file. Reinstate PID-recycling detection in a **new module** and call it from `SystemProbe::is_alive` (`:227-231`), and all three guards stay green: the source scan sees nothing, the signature test is untouched because the comparison now lives behind the trait, and every behavioural test passes. Either scan `src/*.rs`, or name `SystemProbe::is_alive` in the prose as the place the guard can't see — the second costs one sentence.

**F4 — "a PID collision is *transient*" isn't, across a reboot.** `clean_tmp.rs:58-64` and the same claim in the docs. A leftover `dad-tests-1234-*` survives a reboot; the kernel re-issues low PIDs from the start; 1234 lands on a long-lived system service. `kill(1234, 0)` → `Owner::Live`, root kept **for the life of that boot**, and the next reboot can re-collide. That's #461's tmpfs exhaustion minus the ability to fix it — the tool reports `live-pid` and refuses, correctly, and the user has no override. No behaviour change wanted; just soften the prose, and consider noting in the `live-pid` report line that a survivor may be a collision rather than a running suite.

**F5 — PID namespaces (marked unverified).** `pid_is_alive` (`:525-533`) asks the **caller's** namespace. A container with `/tmp` bind-mounted from the host creates `dad-tests-42-*` where 42 is the in-namespace PID; a reaper on the **host** probes host PID 42, gets `ESRCH`, and deletes a running suite's root. I found no container-based workflow here — tooling is `devbox`, same namespace — so I could not construct a supported path to it. Recording it because the docs enumerate where liveness is *unavailable* (`:521-524`) but not where it's available and answers about a different namespace. One sentence, no code change.

**F9 — the symlink-safety comment credits the wrong mechanism.** `clean_tmp.rs:685-687` says the delete is safe because *"`collect` established was a directory and not a symlink"*. `collect` `lstat`s at `:410-413`; `sweep` deletes at `:365`, after every candidate is classified and size-walked — seconds to minutes later. In a mode-1777 `/tmp` a user can `rename(2)` their own entry to a symlink in that window, and sticky doesn't help because they're modifying their own entry. **The code is safe anyway** — `std::fs::remove_dir_all` doesn't follow a symlink at the given path and has been `openat`-based since the CVE-2022-21658 fix. But as written the comment would justify replacing `remove_dir_all` with a hand-rolled recursive delete, which is exactly where this bug class gets introduced.

**F10 — the age column is now mostly zeros.** `human_duration` (`:757-764`) truncates to whole hours, and dead-PID roots are reaped at any age, so most rows read `0h old`. Consider minutes below 1h, or dropping the column for `dead-pid` rows since age explicitly isn't why they were chosen.

---

## The merge conflict — prose only, but one trap

GitHub reports `CONFLICTING` / `DIRTY`. Measured it:

- Merge base `d3ea031b`. `main` has since moved to `0339366d`.
- Files changed by **both** the PR and `main`: **`CLAUDE.md` and `Cargo.lock`, nothing else.** No source file overlaps — `clean_tmp.rs`, `main.rs`, `xtask/linkage-check/Cargo.toml` and `docs/develop/e2e-temp-dirs.md` are untouched on `main`.
- `Cargo.lock` overlaps by file but **not by hunk** (main's last hunk ends ~3381; your `libc` line is ~3584). It auto-merges.
- `CLAUDE.md` is the real conflict and it's entirely prose: your rule-14 rewrite vs a new rule-13 paragraph main added in the same diff context.

**So resolving it cannot change what gets deleted.** The trap: the two edits are on *different lines inside one hunk*, so a reflexive `--ours`/`--theirs` is silently lossy — ours drops main's new rule-13 paragraph, theirs drops the rule-14 rewrite documenting this change. **Keep both.**

Worth re-confirming the PR body's green-gate table after the rebase — it was measured at `5bab8df`, two commits back and against a different `main`.

---

## Checked and clean

Stated explicitly so a green section isn't mistaken for an unexamined one.

- **Deletion targeting** — every candidate comes from `read_dir(temp_root)` (`:404-406`), always a direct child of `std::env::temp_dir()`. No path derived from file *contents* or attacker data; only the name, via prefix match and digit parse. Only `reap` reaches `remove_dir_all`, pinned through the real apply path by `apply_removes_only_the_reap_slice`. No cwd-relative resolution anywhere.
- **Fallback direction** — keep-biased throughout, verified branch by branch: `EPERM` → alive; non-Unix → age rule (old behaviour, not "delete"); unparseable / empty / `> i32::MAX` → age rule; PID 0 rejected with the right reason (`kill(0,0)` addresses the process group); future mtime → age 0 → keep.
- **Cross-platform** — `kill(2)` is Unix-wide so macOS/BSD match Linux; `libc` correctly gated `[target.'cfg(unix)'.dependencies]`; the `#[cfg(unix)]`/`#[cfg(not(unix))]` pair is total; PID-dependent tests are gated.
- **Dry-run and observability** — `apply: false` is the default, the dry-run path prints the re-run instruction, `dry.removed == 0` asserted through the real path. Per-decision attribution with per-reason counts and sizes is a genuine improvement over the old age-only line.
- **Hostile input — better than I expected.** Paths printed with `{:?}` not `.display()`, with the OSC-52 clipboard-rewrite reasoning documented at `:544-550` and pinned by a test. Size totals saturate against sparse-file `u64` overflow. `--older-than` saturates its `hours * 3600`. The size walk is bounded and **never reaches `classify`**, so truncation cannot move a verdict.
- **Dependency hygiene** — `Cargo.lock` diff is exactly one line, `libc` added to `xtask-linkage-check`, **no new `[[package]]`**. Your "pulls no new crate" claim is accurate.
- **Test quality** — the `ProcessProbe` injection removes a real race the previous test had. Two things I'd single out as unusually honest: `a_live_pid_is_kept_whatever_mtime_the_root_carries` carries a comment saying it must **not** be read as a guard against the deleted comparison; and `the_source_scan_strips_comments_and_literals_but_not_labels` proves the scanner's stripping is load-bearing rather than decorative. `FORBIDDEN` tokens are lower-cased to match the `to_lowercase()` comparison, with a doc comment explaining why — a plausible silent-no-op bug avoided deliberately.
- **Dead code / swallowed errors** — none. Removal failures are surfaced on stderr with a non-zero exit; the `let Ok(…) else { continue }` patterns are all in scan/size paths where skipping is correct and cannot cause a deletion.
- **Rule 10, upstream naming** — both compliant.

**On the deliberate deviation from #461** — dropping the recycled-PID branch is the right call and the reasoning at `:34-82` is sound: ordering a process start against a directory timestamp must bridge the wall clock, `settimeofday` moves `btime` while inode timestamps aren't retroactively adjusted, and every resulting error biases toward deletion. Two failed implementations is sufficient evidence. F4 only softens the *"transient"* argument used to size the cost, not the decision itself.

prageethw:
Cross-reference from the #472 review — **this PR and #472 compose into a permanent leak, and they merge cleanly with no conflict marker.**

- #472 moves temp roots to `/var/tmp`, which the FHS requires **not** to be cleared at boot.
- This PR makes a `live-pid` root unreapable **at any age**, with `--older-than` explicitly not applying.
- After a reboot the owning PID is recycled — low PIDs go to systemd units early, so this is the *expected* outcome for any small PID, not a coincidence.
- The root is then misread as live and the only remedy left is a manual `rm -rf` against a directory that holds real agent credentials.

Neither PR's tests can see it: this one injects a `LivenessProbe` and cannot manufacture a boot boundary, #472 never touches PIDs. `grep -in reboot` over this PR's full diff returns nothing.

It also falsifies `docs/develop/e2e-temp-dirs.md:185` — *"treat anything older than the 6h default threshold as something to reap"* — for the `live-pid` class, on the one filesystem where leftovers are now durable.

**This PR is already `CONFLICTING`, so it merges second and is the natural place to land the fix.** Full reasoning and a suggested shape (visibility plus an opt-in escape — deliberately *not* a boot-time comparison, which your own module docs rightly ban) are in the #472 review.

That is separate from the F1 in my review above — F1 is the orphaned-`setsid`-daemon window, this is the reboot window. Different mechanisms, both in the `DeadPid`/`LivePid` split.

vfarcic:
@prageethw — thanks, this changed the PR. Both of your headline items were real and are now fixed in `e35cc7b`, on top of a `main` merge in `da1c1c2`. Point by point.

## Fixed

**F1 — owner-PID liveness is not root liveness.** Accepted in full, including your suggested shape: a minimum-age floor on the `DeadPid` branch only, `DEAD_PID_MIN_AGE = 600s`. I verified every link in your chain before changing anything — `tests/common/mod.rs:1558-1560` states outright that the daemon "setsid's into a separate session and escapes this group — its `DOT_AGENT_DECK_TEST_MAX_LIFETIME_SECS` cap is the net for that"; `src/daemon_attach.rs:250` runs `setsid(2)` in `pre_exec`; `src/agent_pty.rs:787-788` puts a wrapped agent two levels down "in a process group the deck cannot signal"; and the cap is 300s at `mod.rs:647` and `:3621`.

The argument that decided it was yours: the old 6h rule was **72×** the orphan cap, so the hazard did not exist before, and removing the floor opened it precisely in the first five minutes — which is when orphans are alive and when `--apply` is most likely to be typed. As you noted, it costs #461 nothing (the headline case was 280 roots whose youngest was 4h09m) and it needs no change to `owner_of`'s signature, so the compile-time guard is untouched. It is also a timestamp on the *keep* side, the inverse of the deleted comparison.

**The reboot leak (your cross-PR comment).** Also accepted, and I took both halves of your suggested fix rather than only the visibility one:

- `--ignore-liveness` demotes a live-PID root to the **age rule** rather than reaping it outright — the operator supplies the one fact the program cannot measure, and it is never inferred. No boot-time comparison, per your warning and this module's own ban.
- Survivors are now **listed individually** with path, size and age, capped and counted exactly like the reap list, so a stale collision is visible instead of hiding inside a count.

You were right that I had repeated "a PID collision is transient" without testing it; it holds within a boot and not across one. That claim is now corrected in the module docs, `e2e-temp-dirs.md` and CLAUDE.md rule 14. I also confirmed your premise that no override existed — the whole flag surface was `--apply`, `--include-untagged`, `--older-than`, `-h`.

**F2 — overstated claims.** All three corrected: the module docs now say the PID answers "is the creator still running?", explicitly *not* "is anyone still using this?", with a pointer to why the floor follows from that gap. Same in `e2e-temp-dirs.md` and rule 14. Your reasoning for why this matters — a maintainer who believes a dead PID proves the tree unreferenced will delete the floor — is quoted in the code near the constant.

**F9 — symlink comment credited the wrong mechanism.** Fixed. The comment now attributes delete-time safety to `remove_dir_all` being `openat`-based since CVE-2022-21658, and says plainly that crediting the earlier `lstat` would license hand-rolling the delete.

**F10 — the age column.** `human_duration` now shows minutes below an hour. This was worse than cosmetic once the floor landed: the floor's own summary line read "older than 0h".

**F3 and F5 — documented rather than coded, as you suggested.** The module docs now name `SystemProbe::is_alive` as the seam the source scan cannot see, and record that `kill(2)` answers about the **caller's** PID namespace — a wrong answer, not an unavailable one, so no fallback triggers.

## Also worth knowing (found separately)

**The source guard runs in no CI gate at all.** Stronger than your F3: the workspace has no `default-members`, so `cargo nextest run` — which is `test-fast` and all three CI build jobs — selects the root package alone. `cargo nextest list` finds **0** `clean_tmp` tests; `--workspace` finds **25**. So `source_has_no_pid_recycling_machinery` is both bypassable *and* inert. `cargo clippy --workspace --all-targets` does compile them, which is what still enforces `owner_of`'s signature. Filed as #489 (deliberately not fixed here — flipping the selection may surface never-run failures in `xtask/docs` / `xtask/spec`).

**Also filed #490**: `verify-pr`'s `checks.sh` vacuous-skip detector greps `'^SKIP: '` while nextest indents, so it silently reported 0 skips on a run that had 4.

## Verification

Since your audit deliberately ran nothing, here is the behavioural half, driven through the real binary against an isolated `TMPDIR`:

```
   195.3 KB  dead-pid     3h old  ".../dad-tests-4243-LongDead"
reap: 1 dir(s), 195.3 KB
  dead-pid     1 dir(s)   195.3 KB  owning process is gone; older than 10m
kept:
   195.3 KB  dead-pid     0m old  ".../dad-tests-4242-FreshlyDead"
   195.3 KB  live-pid     0m old  ".../dad-tests-1740322-LiveOwner"
keep: 2 dir(s), 390.6 KB
  dead-pid     1 dir(s)   195.3 KB  owning process is gone, but younger than 10m — a spawned daemon may still be writing here
  live-pid     1 dir(s)   195.3 KB  owning process is still running — never reaped (--ignore-liveness overrides)
```

`--apply` removed only `LongDead`. Against a genuinely live PID, the root reports `live-pid` and is kept, then becomes `live-aged` and reapable under `--ignore-liveness`. Earlier, against the real `/tmp`, it selected 170 dead roots / 5.6 GB.

Gates: `cargo fmt --check` clean; `cargo clippy --workspace --all-targets --features e2e -- -D warnings` clean; **1729 tests pass** with `--workspace` (61 in this crate, up from 56 — 5 new, 5 adjusted to age their fixtures past the floor); `linkage-check` ok.

I did not re-run the e2e tier, deliberately: the branch's entire delta versus `main` outside `xtask/linkage-check` and docs is one `Cargo.lock` line adding `libc` to an xtask crate, and `xtask-linkage-check` is not a dependency of the root package, so the e2e binaries are unchanged.

## On landing order

I withdraw my earlier "#467 first" — your argument wins on the data. #472 is `MERGEABLE`; #467 was `CONFLICTING` and 29 commits behind, so it had to rebase regardless and is the natural place for the reboot fix, which lives in the classifier. That merge is now done, so #467 is mergeable again — if #472 lands first, this will need one more rebase, and the `clean_tmp.rs` conflict is 6 hunks either way.

Two notes for whoever rebases #472 onto this: the docs sentence "Directories younger than the age threshold (default 6h) are always left alone … and symlinks are never followed" is **deleted** here, so keep this side; and `e2e-temp-dirs.md`'s "treat anything older than the 6h default threshold as something to reap" still needs the `live-pid` caveat you identified, which only exists in #472's tree.

prageethw (dismissed):
Re-reviewed at `e35cc7be` — delta only (the merge from `main` plus `e35cc7be`), code read, no build, no test run. The reaper was never executed in any form: not `--dry-run`, not `--apply`, not `--root`, not `--ignore-liveness`, not against a scratch directory.

**Approving.** All seven prior findings are addressed, two of them more thoroughly than I asked.

| Prior ask | Verdict |
|---|---|
| F1 — dead-PID reap could delete a live orphan's tree | **ADDRESSED**, exactly as proposed |
| F2 — overstated safety claims | **PARTIALLY** — module docs, `e2e-temp-dirs.md` and `CLAUDE.md` all corrected; the PR body, the third site, is not (N1) |
| F3 — reinstatement guard bypassable | **ADDRESSED**, plus a volunteered disclosure |
| F4 — "collision is transient" false across a reboot | **ADDRESSED DIFFERENTLY, better** |
| F5 — PID namespaces | **ADDRESSED** — documented as a *wrong* answer rather than an unavailable one, which is the more useful framing |
| F9 — symlink comment credited the wrong mechanism | **ADDRESSED** — now credits `remove_dir_all` / CVE-2022-21658 |
| F10 — age column all zeros | **ADDRESSED** — minutes below an hour |

**F1 is right.** `DEAD_PID_MIN_AGE = 600s`, applied on the **keep** side at `clean_tmp.rs:588`, so below-floor dead roots are retained. It costs #461 nothing — that issue's headline case was 280 roots with the youngest at 4h09m. Both the code fix and the doc note I offered as a fallback were done.

**F4 is better than what I asked for.** I asked for the prose to be retracted; you retracted it, shipped `--ignore-liveness` as the operator override I had noted was missing, *and* made survivors list with size and age.

**The mechanism errs toward keeping in every branch**, which is the property that matters for a tool that deletes directories. Some specifics I checked rather than assumed:

- **`kill(pid, 0)` errs the safe way.** `EPERM` is treated as **alive**; `ESRCH` is the only value meaning dead. Getting that backwards is how a reaper deletes a running suite's directory, so it was the first thing I looked at.
- **`--ignore-liveness` is properly bounded.** It demotes a live-PID root to the age rule — it is not an unconditional delete. It does not bypass the dead-PID floor (asserted at `:1215`), the name allowlist, the dry-run default, or the symlink skip. Ownership proof — prefix match, plus direct child of `temp_dir()`, plus `symlink_metadata` — all sits upstream of `classify` and is untouched by the flag. Critically, **there is no `--root` on that branch**, so it cannot be aimed at an arbitrary directory.
- **The merge is orthogonal.** `main` never touched `xtask/`, and its only `tests/common/mod.rs` change is an unrelated `attach_path` field. The `CLAUDE.md` conflict was resolved keeping **both** sides — `main`'s new rule-13 paragraph and the rule-14 rewrite are both present — so the lossy `--ours`/`--theirs` trap I flagged last time was avoided.
- **The floor is well covered by tests**, including boundaries and the real `collect`/`sweep` path.

---

### N1 — the PR body still carries the two claims this commit exists to retract

It still says "three branches", "dead PID reaped **at any age**", and "a PID collision is **transient**" — the last one verbatim from what F4 retracted — and never mentions `--ignore-liveness`.

That matters more than a stale description normally would: **on squash-merge the PR body becomes the commit message**, so the permanent record of *why* this change was made would assert the two things it was made to correct. The changelog fragment is already accurate; this is two bullets.

### N2 — `--ignore-liveness` is never exercised end to end

`ignore_liveness` is `true` in only three direct `classify` unit calls, plus `parse_args`. No test drives `collect` / `sweep` with it.

A mis-wire at `:553` therefore fails silently in either direction: the flag becomes a no-op and the reboot leak it exists to clear persists, or — the unsafe direction — every live root is demoted to the age rule, with no test objecting. For a flag whose entire purpose is to make deletion more aggressive, one end-to-end test is cheap insurance.

Related, and the one place I would tighten the mechanism rather than the tests: **`--ignore-liveness` is the only deleting branch with no floor of its own, and it is precisely the branch where the process is known to be alive.** `--ignore-liveness --older-than 0 --apply` reaps every harness root at any age, including a running suite's. The strictly-safer dead-owner branch cannot be made to do that. If you want the flag to keep a floor, `reap: age >= max_age.max(DEAD_PID_MIN_AGE)` gets it without changing the intended use.

### N3 — the floor is measured from the wrong instant

`DEAD_PID_MIN_AGE` is measured from the root's **mtime** — its last direct-child change, i.e. the last deck launch — not from the owner's death, while the docs frame it as covering the 300s post-death orphan window.

Those coincide only when the two are close. The invariant safety actually needs is *last-daemon-spawn minus last-deck-launch < 300s*. I verified the mechanism by reading; whether an existing test happens to cover the gap I could not confirm, since I ran nothing — flagging that as unconfirmed rather than asserting it. The cheap fix is to state the real invariant in the `DEAD_PID_MIN_AGE` doc comment, so the next person does not re-derive the wrong one from the current wording.

---

### One thing worth knowing that is not this PR's fault

**Every test pinning this tool's safety invariants runs in no CI gate.** `Cargo.toml` has no `default-members` and CI runs a bare `cargo nextest run`, so all 61 tests in `xtask/linkage-check` never execute — including `apply_removes_only_the_reap_slice`, described in-source as *"the single highest-consequence invariant in this file"*, and the whole F1 floor. Clippy compiles them; nothing runs them.

You disclose this in the PR, but scope the disclosure to the source scan — it is far wider than that. It is also **pre-existing and not introduced here**, so it is no reason to hold this PR.

The good news is that it is already fixed elsewhere: **PR #498 adds `--workspace` to exactly those `ci.yml` lines**, which I approved earlier today. I checked the diff — once #498 lands, this PR's safety tests start actually running. Worth landing #498 first if the order is free, since it is the difference between these guarantees being asserted and being tested.

prageethw (dismissed):
Re-reviewed at `96473894` — the two commits since my previous approval was auto-dismissed by this push. Code read only, no build, no test run, and `clean-e2e-tmp` was never invoked.

**Re-approving.** My previous approval was dismissed by the ruleset's stale-review rule, not withdrawn on merit, and this delta earns it back.

**The specific risk I re-reviewed for did not materialise.** A commit titled "make the fixtures portable across three platforms" is the classic place to quietly shrink a fixture until it fits and call the tier green. That is not what happened: `HUGE` is unchanged and all six assertions are byte-for-byte identical — the fix **relocates the fixture to a filesystem that accepts the length** rather than reducing it. `set_len` is bounded by the filesystem's maximum file size rather than free space, so relocating is the right lever. The test still asserts saturation, which is what its name promises.

The merge from `main` did pull in `main`'s Windows-red state, and the delta then fixes it in-branch, as the commit claims.

---

### S1 — the "loud rather than silent" `SKIP:` is invisible in CI

The skip branch prints `SKIP: no filesystem here accepts a {HUGE}-byte sparse file …`, and the comment above it justifies the design as *"Loud rather than silent"*. In CI it is neither: all three jobs run `cargo nextest run --workspace`, and `[profile.default]` sets no `success-output`, so a **passing** test's stdout is discarded. The test reports `PASS` on both Linux (0.005s) and macOS (0.008s) with **zero** `SKIP: ` lines in either job log.

Worth connecting to something already in flight: the repo's own vacuous-skip detector in `verify-pr/checks.sh` greps the **e2e** log only, so it will never see this line either. That detector was just fixed (#495/#496) precisely because a skip that reports `PASS` is indistinguishable from a real run — this is the same shape, one tier over.

Cheapest fix is `--success-output=final` on the relevant gate, or an explicit marker the detector can find. Not blocking, but the comment currently promises a property the configuration does not deliver.

### S2 — the compensating-coverage comment is wrong about `freed`

The comment claims *"The group, grand and `freed` totals are covered on every platform by `group_and_grand_totals_saturate`, which needs no filesystem."*

That test builds two `u64::MAX` candidates, calls `report(...)`, and asserts only `text.contains("reap: 2 dir(s)")` and `text.contains("dead-pid")`. It **never calls `sweep` and never produces a `freed` value at all** — `freed == u64::MAX` is asserted only inside the branch that gets skipped. So on any host that skips, that accumulator has no coverage from anywhere, while the comment states the opposite.

This matters more than a stale comment normally would, because the comment is the *justification* for the skip being acceptable. Either extend `group_and_grand_totals_saturate` to cover `freed`, or narrow the comment to the two totals it really does cover.

### N1 — still unaddressed, re-checked as promised

The PR body still says "three branches", "dead PID reaped **at any age**", and "a PID collision is **transient**", and mentions `--ignore-liveness` **zero** times — though the flag now exists and `DEAD_PID_MIN_AGE` directly contradicts "at any age". On squash-merge this becomes the permanent commit message, so the record of *why* the change was made would assert three things the change itself corrects. The changelog fragment is already accurate; this is two bullets.

N2 (`--ignore-liveness` never exercised end to end) and N3 (`DEAD_PID_MIN_AGE` measured from mtime rather than the owner's death) are untouched by this delta and stand as previously described.

---

### On the overlap with #512

**These two PRs fix the same defect.** #512 makes the same change to the same test, differing only in shape — `SCRATCH_ONE`/`SCRATCH_TWO` there, `const SCRATCH_ROOTS: [&str; 2]` here — with identical values and the same `#[cfg(windows)]` split.

**My recommendation is to land #512 first.** It is the minimal unblock — one file, +18/-3, test-only, all 11 checks green including `build-windows` — for a `main` whose redness is currently inherited by every open PR on its merge commit, whereas this PR is ~1400 lines with review items still open. The extra scope here (the `EFBIG` fix) is genuinely independent and need not travel with the Windows fix.

This PR should then take `main`'s side of the conflicting hunk — delete `SCRATCH_ROOTS`, use `SCRATCH_ONE`/`SCRATCH_TWO` — and drop the now-stale paragraph 2 of `96473894`'s commit message in the same pass.

One last note, since it closes a loop from my earlier review: this commit's own message says both failures were *"exposed, not caused, by #498 putting the workspace's tests in a gate — these assertions had never run in CI on any platform."* That confirms the coverage gap I raised previously, and #498 is already approved.

vfarcic:
## Conflict resolved — re-requesting review

@prageethw the **only** change since your approval of `9647389` is the merge-conflict resolution. Nothing was redesigned, and no behaviour changed. Details below so re-approving is cheap.

### What you actually need to re-check: one doc comment

Since `9647389`, the entire hand-written diff in this PR is a doc comment on a test fixture — **zero executable lines**. Every other changed file (`CHANGELOG.md`, `Cargo.toml`, `Cargo.lock`, `flake.nix`, `site/helm/*`, the consumed `changelog.d/*` fragments, `.env.vals.yaml`) is v0.36.0's release content, taken verbatim from `main` by the merge.

```
git diff 9647389 de0d59f -- xtask/linkage-check/src/clean_tmp.rs   # <- the whole review surface
```

### The conflict, and which side each hunk came from

Exactly one file conflicted: `xtask/linkage-check/src/clean_tmp.rs`. It was a **convergent duplicate**, not a disagreement — both sides had independently fixed the *same* Windows bug in `explicit_roots_replace_the_standard_set`:

| Side | Shape | Why it existed |
|---|---|---|
| `main` (PR #512, issue #511) | two scalar consts `SCRATCH_ONE` / `SCRATCH_TWO`, each `#[cfg]`'d | make `build-windows` green on `main` |
| this branch (`9647389`) | one `#[cfg]`'d array `SCRATCH_ROOTS: [&str; 2]`, with the test's three call sites already indexing it | same fix, found while getting this PR's own gates green |

**Resolution — kept this branch's `SCRATCH_ROOTS` array** (one const for one job, and its three call sites already use it), and **folded `main`'s mechanical explanation and its issue #511 reference into the doc comment** so the Windows fix's provenance is still greppable by `#511`. `main`'s `SCRATCH_ONE`/`SCRATCH_TWO` are therefore gone as *names* — but not as *intent*: both spellings give Windows a drive-prefixed absolute path, so `parse_args`'s `is_absolute()` check passes and the test exercises the replacement rule instead of panicking in its own setup. That is precisely what #511 needed. `build-windows` passes here in 3m54s.

Hunk-by-hunk: the const block is this branch's, with `main`'s prose merged in; the three call sites (`SCRATCH_ROOTS[0]`, `SCRATCH_ROOTS[1]`, and the `assert_eq!` vec) are this branch's, unchanged.

### Neither half was dropped

- **#467's half** — the merge's *entire* delta against `9647389` is that doc comment, so every line of the owning-PID liveness machinery is **byte-identical to the revision you approved**. `git diff 9647389 de0d59f` touches no logic.
- **#511's half** — preserved as described, and additionally this branch still carries the *other* half of the portability work, `sparse_capable_tempdir` (the ext4 `EFBIG` fix for the exabyte fixture). `main` never had that one, and it did not conflict.

### Gates

| Gate | Result |
|---|---|
| `cargo clippy --workspace --all-targets --features e2e -- -D warnings` | pass, 0 warnings (`xtask-linkage-check` genuinely linted) |
| `cargo test-fast` | pass — 2549 / 2549, matching the pre-merge count exactly |
| `cargo nextest run -p xtask-linkage-check` | pass — 83 / 83 |
| `cargo xtask linkage-check` | ok (504 catalog ids, 432 annotations, 8 rules) |
| CI | 10 / 10 green, incl. `build-windows`, `build-macos`, `windows-cross-check` |

Because this file is a **deletion** tool whose safety properties exist only as runtime assertions, I confirmed by name that they actually ran rather than trusting the tier summary: `owned_prefixes_are_reaped_by_default`, `untagged_tempfile_prefix_is_opt_in`, `include_untagged_is_confined_to_the_historical_temp_dir`, `symlinks_named_like_owned_dirs_are_skipped`, `a_directory_replaced_by_a_symlink_is_not_descended` — plus `source_has_no_pid_recycling_machinery`, which reads this very file and would have caught the edit above, and `explicit_roots_replace_the_standard_set` itself.

### Post-merge dry run against live machine state

```
reap: 56 dir(s), 1.5 GB   dead-pid   owning process is gone; older than 10m
keep:  1 dir(s), 313.6 MB live-pid   owning process is still running — never reaped
```

Several reaped roots were only **21–22m old** — the age-only rule would have refused all 56. The kept root's owner was verified independently with `kill -0` rather than trusting the label: PID 486327 is alive, and it is `e2e_prompt_deli`, a **live e2e test binary running at the time**. That is exactly the root the pre-fix tool could have deleted out from under a running test. Five reaped PIDs were spot-checked as genuinely dead.

Greptile re-reviewed `de0d59f`: 5/5 confidence, no blocking issues, zero inline findings. Unresolved review threads: 0.

Merged from the newest `origin/main` (`3496e7a`, post-v0.36.0) as a normal merge commit — no rebase, no force-push. Not merging this myself per `docs/develop/governance.md`; over to you.

prageethw (approved):
Re-approving `de0d59f`. Your account of the delta checks out exactly as described.

The approved-to-head diff over `xtask/` and `src/` is a single hunk in `clean_tmp.rs` and every changed line begins `///` — the `SCRATCH_ROOTS` fixture doc comment about `is_absolute()` being platform-specific. No executable line moved.

The rest of the delta file list — `.env.vals.yaml`, `CHANGELOG.md`, `Cargo.{toml,lock}`, eight consumed `changelog.d/*` fragments, `flake.nix`, `site/helm/{Chart,values}.yaml` — reproduces `git diff --name-status 3cf4b689 3496e7ab` exactly, so it is the v0.36.0 release range arriving through the merge rather than a hand edit.

I checked the conflict resolution specifically, since that is where a hunk goes missing quietly, and all three halves survived:

- **#461's machinery** — `Owner::Dead` reaping at `age >= DEAD_PID_MIN_AGE`, `Owner::Live` kept by default, `owner_of` going through the injected liveness probe. The parent-to-head diff has no hunk near that code.
- **main's #511 portability fix** — the merged fixture carries `#[cfg(windows)] SCRATCH_ROOTS` with drive-prefixed paths while Unix keeps `/scratch/...`, and `explicit_roots_replace_the_standard_set` consumes those through `parse_args`.
- **This branch's `sparse_capable_tempdir`** — the ext4 `EFBIG` fix for the exabyte fixture. `git grep` finds its three references at `9647389` and none at `3496e7ab`, which confirms it is yours and not something main brought in.

Sorry this sat for a day — the dismissal was a formality and I should have cleared it sooner.

Issue #468, opened 2026-08-09
Split dispatch into placement and spawning: agents create the worktree, the CLI dispatches into a directory


vfarcic:
## PRD: Split dispatch into placement and spawning

**Problem**: `dispatch` couples two unrelated jobs — creating a git worktree and spawning an agent into it. Because placement is implicit in the verb, there is no way to start an orchestration in a directory that already exists, so continuing a line of work (single agent → orchestration) forces a new worktree and strands the branch, probe tests and scratch state the first agent built. The orchestrator then re-derives analysis that had already been done.

**Solution**: Split the verb in two. Agents create the worktree or branch themselves with plain `git` — they already know how, and the CLI adds nothing. `dispatch` becomes a single orthogonal primitive: spawn an agent or orchestration **into a specified directory**, with a task. Dispatching into the current directory is then in-place continuation, with no second verb needed.

**Detailed PRD**: [`prds/468-split-dispatch-placement-spawning.md`](https://github.com/vfarcic/dot-agent-deck/blob/main/prds/468-split-dispatch-placement-spawning.md) — resolves once PR #469 merges. While that PR is open, read it on the branch: [`prd-468-split-dispatch-placement`](https://github.com/vfarcic/dot-agent-deck/blob/prd-468-split-dispatch-placement/prds/468-split-dispatch-placement-spawning.md).

**Discussion**: PR #469 — opinion requested from @irizzant, who built the `dispatch` verb in #232.

**Priority**: Medium

irizzant:
**@vfarcic** — apologies for going quiet on this. You flagged me twice and I never got back to you before #469 merged; that was the one question you explicitly wanted my read on, and the timing was mine to miss. Thank you for asking in the first place, and for the way this ran — prageethw's review and your responses did a better job than I would have on a deadline. The merged PRD is stronger than the draft.

FWIW, my answers are now on the record anyway:

**On the pre-spawn boundary — you and prageethw both read it right.** The invariant #220 protects is *when* the cwd is bound, not *who* picks it. `spawn_command` still receives a directory that was chosen before spawn, and nothing running is ever relocated. A continuation spawns a *new* orchestration into an existing directory after the previous agent has finished. That stays on the right side of the line I drew.

[...]

Issue #220, opened 2026-07-19
Dispatcher mode + worktree dispatch: one-step, agent-callable isolated line of work


vfarcic:
## PRD: Dispatcher mode + worktree dispatch

**Problem**: Concurrent orchestrations are only safe across directories (worktree-per-orchestration is the supported model, per #140), but getting an orchestration into a worktree interactively is a multi-step manual chore: create a worktree (e.g. `/worktree-prd`), then manually open an orchestration pointed at that directory. There is no agent-callable, one-step entrypoint — `delegate` only targets pre-existing worker panes, and issue-dispatch (#120) creates a worktree + spawns an orchestration but is scheduler-triggered only.

**Solution**: Add an agent-callable `dispatch --worktree` CLI verb that deterministically creates a worktree and spawns a fresh, fully-isolated orchestration inside it (reusing #120's engine), plus a built-in "dispatcher" mode (a seeded single agent) that teaches — agent-agnostically, with zero ambient overhead — what to call and when. Isolation is always a pre-spawn decision and never an LLM judgment call, so a mis-timed dispatch wastes a cleanable worktree rather than re-creating #140's cross-delivery corruption.

**Detailed PRD**: See [prds/220-dispatcher-mode-worktree-dispatch.md](https://github.com/vfarcic/dot-agent-deck/blob/main/prds/220-dispatcher-mode-worktree-dispatch.md)

**Relationships**: Same-project precursor/subset of #174 (cross-project dispatch); builds on the worktree-per-orchestration model documented by #140 (not its code); #140 remains the safety/correctness floor.

**Priority**: Medium

vfarcic:
## Phase 2 is mostly retention: the delivery path already exists and runs today

Measured 2026-09-14 against `main` at `4a7ae53`, while working out what Phase 2 would actually cost. Three of the four pieces M2.0/M2.1 need are already built and exercised on every dispatch — which makes this smaller than the milestone list reads, and moves the real weight onto M2.3.

### The injection primitive is already pointed at the dispatcher

A dispatch **acknowledgement** — the `dispatch: spawned isolated agent for '<name>' in <path>, cut from main at <sha>` line — is not printed by the CLI. It is injected into the caller's pane by the daemon, via `deliver_dispatch_result` (`src/daemon.rs:2131`):

```rust
registry.write_and_submit_guarded(pane_id, message, expected_agent_id, || async { true })
```

[...]

Issue #156, opened 2026-06-14
Orchestration identity: use a unique instance ID instead of (name, cwd)


vfarcic:
## Follow-up from PRD #113 / PR #151

**Problem**: An orchestration's identity is the `(name, cwd)` tuple in `pane_orchestration_map`, with no unique per-instance ID. This is fragile on two axes:
- It depends on the name resolving consistently. A PRD #107 regression broke worker context-cleaning in **worktrees** because the new-pane form's display name overwrote the orchestration identity (the per-delegate role lookup then missed → `clear=true` respawn skipped). PR #151 fixed it by decoupling the display title from the canonical config name — a targeted fix, not the durable one.
- It cannot distinguish **two orchestrations started in the same directory** — identical `(name, cwd)` tuples would cross-route delegates (the tuple-equality scoping in `handle_delegate` can't tell them apart).

**Solution**: Generate a **unique instance ID** at orchestration start, carry it per-pane, and key delegate routing + the per-role clear/respawn decision on that ID (cwd used only to *load* the config; name becomes a pure display label). Durably fixes the worktree class and the same-dir collision.

**Priority**: Medium

Reference: design discussion during PR #151; the canonical-name decoupling shipped there is the band-aid this would replace.

prageethw:
A note on this issue's premise, found while planning vfarcic/dot-agent-deck#421 — **it looks partly stale.**

This issue states that an orchestration has "no unique per-instance ID". PRD #140 appears to have shipped exactly that:

- `mint_orchestration_id()` (`src/agent_pty.rs:383`) mints a per-tab token.
- `TabMembership::Orchestration::orchestration_id` (`src/agent_pty.rs:364`) carries it on every role pane of a tab.
- `OrchestrationIdentity::Instance { id, name }` (`src/state.rs`) is the tokened identity, with `NameCwd` retained only as the legacy fallback for clients predating #140.

`OrchestrationIdentity`'s own doc comment claims the property this issue asks for: *"Equality is the token, so two tabs of the SAME orchestration in the SAME directory are two distinct routing groups. This is what closes issue #140's cross-delivery."* It also notes mixed-variant comparison is never equal, which handles the old-client case.

So the **same-directory collision** described here — the second of the two axes — reads as already addressed for any pane produced by a #140-or-later client.

[...]

PR #469, opened 2026-08-09
docs(prd-468): PRD — split dispatch into placement and spawning

vfarcic:
Adds the PRD for #468.

**@irizzant** — you built the `dispatch` verb and dispatcher mode in #232, so I would particularly value your opinion on this before any of it gets implemented. It changes the shape of the verb you designed, and it touches a boundary you explicitly reasoned about in PRD #220.

## What it proposes

`dispatch` currently does two jobs: it creates a git worktree, and it spawns an agent or orchestration into it. Coupling them makes **placement implicit in the verb** — every dispatch lands in a new tree, because that is the only thing it can do.

The proposal splits them. Agents create the worktree or branch with plain `git`, and `dispatch` takes an explicit **target directory** plus a task. Dispatching into the current directory is then in-place continuation, with no second verb required.

## What motivated it

A single agent analyses an issue, reaches a verdict, and the follow-on work is big enough to want an orchestration. Everything that orchestration needs already exists in the single agent's worktree — the branch, the analysis, any probe tests it wrote. But `dispatch --orchestration` creates a *new* worktree from `main` and strands all of it. The only workable path today is starting the orchestration manually in the existing worktree, which works but leaves the orchestrator with no context, so it re-derives analysis that was already finished.

## The bit I most want you to check

PRD #220 explicitly rules out **mid-flight worktree adoption**, and the reasoning is sound and specific: worker pane cwds are frozen at spawn (`src/agent_pty.rs:736`), the orchestrator's cwd is neither movable nor reported to the daemon, and coordination files are pinned to the recorded cwd via `pane_cwd_map`. Your conclusion was that worktree is **always a pre-spawn decision, never a runtime relocation**.

I believe this proposal preserves that invariant rather than reopening it — the directory is still chosen *before* the spawn, and nothing already running is relocated. A continuation spawns a **new** orchestration into an existing directory after the previous agent has finished. What changes is only who chooses the directory, and when.

**If that reading is wrong, this proposal is wrong**, and you are the best-placed person to say so.

## Known costs, called out in the PRD

- **Placement becomes prompt-enforced instead of code-enforced.** Today it is impossible to put a dispatched worktree on a tmpfs. Afterwards it is only discouraged — and per CLAUDE.md rule 14 that failure is silent and misattributed (a `cc` linker error, or `SIGKILL` on `rustc`, with nothing pointing at the filesystem).
- **Orphan worktrees become routine**, since step 1 can happen without step 2.
- **Config resolution becomes ambiguous.** `spawn.rs:88` deliberately resolves the target from the config seen by `--list-targets`, *not* the worktree's. Those cannot diverge today because `dispatch` creates the tree it spawns into; with an arbitrary target directory they can.
- **Same-directory coordination-file collision** — the same root cause as #156 and the deferral in #140.

Also worth your view: the PRD leaves open whether the CLI should keep a thin convenience wrapper that creates a worktree at the conventional location and dispatches into it. That would preserve today's one-step ergonomics and the rule-14 guarantee, at the cost of the orthogonality the PRD is arguing for. I did not want to pre-empt that decision.

Nothing here is built. This is a proposal for discussion, and I am happy to be talked out of it.

---

PRD: [`prds/468-split-dispatch-placement-spawning.md`](https://github.com/vfarcic/dot-agent-deck/blob/prd-468-split-dispatch-placement/prds/468-split-dispatch-placement-spawning.md) (branch link, live while this PR is open) · [same file on `main`](https://github.com/vfarcic/dot-agent-deck/blob/main/prds/468-split-dispatch-placement-spawning.md) (resolves after merge)

Closes nothing; #468 stays open as the tracking issue.


vfarcic:
@irizzant — flagging this directly rather than relying on the description mention.

You designed the `dispatch` verb in #232, and this PR proposes changing its shape, so your opinion carries more weight here than anyone else's. Nothing is implemented; it is a proposal and I would rather it be argued with than merged quietly.

The single question that decides whether the whole thing is sound: **does splitting placement out of `dispatch` stay on the right side of the pre-spawn boundary you drew in PRD #220?** You ruled out mid-flight worktree adoption for concrete reasons — frozen worker cwds at `src/agent_pty.rs:736`, an orchestrator cwd the daemon is never told about, and coordination files pinned to the recorded cwd. My claim is that choosing the directory earlier, and spawning a *new* orchestration into an existing one, never crosses that line. If you think it does, this proposal does not survive.

Secondary, if you have the appetite: whether the CLI should keep a thin convenience wrapper that creates the worktree at the conventional location, which would preserve today's one-step ergonomics and the CLAUDE.md rule 14 guarantee that placement never lands on a tmpfs. The PRD deliberately leaves that open.

No rush, and a "this is wrong because…" is a more useful reply than an approval.

prageethw (changes requested):
Reviewed at `2a3f8b4` as a specification — read only, no build or test run.

**The core argument is correct and worth pursuing.** `dispatch` genuinely does conflate placement with spawning, the continuation case is genuinely unsupported, and — the question that mattered most — **the #220 pre-spawn invariant genuinely survives.** The invariant #220 protects is that cwd is fixed before `spawn_command`; this proposal moves *who decides* the value, not *when it is bound*. That was the right question to have asked.

I checked the citations rather than taking them, including the one Greptile flagged: `agent_pty.rs:996` (`if let Some(dir) = opts.cwd { cmd.cwd(dir); }`) and `:1071` (`pair.slave.spawn_command(cmd)`) both check out, and the corrected citation genuinely supports the claim — the registry exposes `resize` / `set_agent_label` / `set_agent_type` and no cwd mutator. `spawn.rs:88`, `spawn.rs:117` gated at `:587`, `ui.rs:531`, and `prepare_orchestrator_prompt` already being parameterised on cwd and task are all accurate too.

The Out of Scope reasoning is better than most implementation PRDs manage — particularly rejecting consume-on-read briefs because it *"puts correctness on a cleanup path that a SIGKILL skips, which is precisely how #322's temp roots leak."* That is the correct generalisation of a real incident.

**What the document is missing is the other half of the lifecycle it is splitting.** It specifies creation in detail and never mentions ownership, registration, or removal. That is the theme of the three blockers below, and they compound: F3 makes F4 harmful, and F6's wording is how F1 gets decided by accident.

**For @irizzant's question specifically:** my reading agrees with yours — the pre-spawn boundary holds, since cwd is still bound before `spawn_command` and nothing running is relocated. The risk this proposal actually introduces is not adoption of a live worktree; it is that **nothing owns the directory afterwards.**

---

### F1 — Worktree ownership and tab-close cleanup are unaddressed, and the motivating case makes the default behaviour destructive

`prds/468-split-dispatch-placement-spawning.md:43-44`, with `:85`.

Dispatch does not merely create a worktree today — it **claims it for deletion**. `handle_dispatch` calls `record_worktree(&ctx.worktrees, &paths.worktree_dir, &clone_dir, RemovalPolicy::KeepIfDirty)` (`src/dispatch.rs:331-336`), registering the tree in the daemon-wide `WorktreeRegistry` that the tab-close handler reads and acts on via `remove_worktree`. `dispatch.rs`'s own unit tests guard this under the heading `// --- removal policy (the PRD #120 regression) ---` (`:562`) and assert the policy survives the registry round-trip (`:614`).

**The PRD never mentions registration, removal, `RemovalPolicy`, or the close path anywhere in its 93 lines.** In Scope removes the path derivation and the single-use-name guard and stops there. Both readings available to an implementer are harmful, and the document supports neither over the other:

**Reading A — registration survives.** `dispatch` records whatever directory it is handed, so **closing the tab deletes a directory the deck did not create.** For this PRD's own motivating case that is severe: Success Criterion 2 is "an agent can dispatch an orchestration into *its own* directory", so closing that orchestration's tab runs `remove_worktree` against the *originating agent's worktree* — the branch, the analysis, and the probe tests the Problem Statement exists to preserve. `KeepIfDirty` does not save it: the scenario describes an agent that has committed real work, so the tree is plausibly **clean**, and `KeepIfDirty` removes clean trees by design. Worse, per the PRD's own cited invariant the originating agent's cwd is frozen at spawn — if it is still alive, its cwd becomes a deleted directory.

**Reading B — registration is dropped.** Dispatched worktrees are never cleaned up on close, silently reverting the #120 behaviour those tests protect. M5 says existing tests are "updated rather than duplicated", which would quietly mean deleting the removal-policy assertions without the PRD ever having said cleanup was in scope.

An implementer picks a reading at M1 and either ships a verb that can delete another live agent's working directory, or ships a silent regression of #120 with its guard tests removed as "updates". **Reading A is the path of least resistance, because it is what the code already does.**

*Suggested fix:* state the ownership rule explicitly. The natural one is that `dispatch` only records for removal a directory it created — which after M3 is *none*, making cleanup the agent's responsibility along with creation. If so, say In Scope removes `record_worktree` too, fold it into the orphan-worktree risk (it is then the same risk, not a separate one), and add a milestone for the doc/test changes. Whatever the answer, the in-place-continuation case needs an explicit "never removes the target directory" guarantee.

### F2 — The orphan-worktree mitigation depends on a supplier the PRD deletes

`:85`.

> *Mitigation*: accept it, and keep the `agent/dispatch-*` branch prefix so `git worktree list` still identifies deck-created units mechanically.

The only thing that produces that prefix is the code this PRD removes. `derive_dispatch_paths` builds `slug = format!("dispatch-{clean_name}")` and `branch = format!("agent/{slug}")` (`src/dispatch.rs:196`, `:205`) — and In Scope `:44` plus M3 `:69` delete exactly that derivation. Afterwards the *agent* creates the branch with plain `git`, so the prefix can only be requested by the dispatcher seed, never produced by code.

So the sole stated mitigation for a risk the PRD itself says "becomes routine" is inoperative the moment M3 lands. `git worktree list` then shows whatever branch names agents chose, and "deck-created unit" stops being a mechanically detectable category at all — which also removes the only way to build the cleanup or audit tooling that "accept it" implicitly assumes someone can write later.

*Suggested fix:* either state that the prefix becomes a seed-enforced convention — and accept that orphan detection is then best-effort and unenforceable, a materially different risk posture than the bullet claims — or keep a code-level marker. #425's worktree ownership marker is the existing mechanism for exactly "which of these trees did the deck make".

### F3 — Removing the single-use-name guard removes the only mutual exclusion, and the stated replacement is contradicted by an Open Question

`:87` against `:44` and `:93`.

Today the interlock against two dispatches landing on one slot is **code**: `create_worktree` fails with `dispatch: worktree {} is already claimed by another dispatch` (`src/dispatch.rs:293`) and `dispatch: branch {branch} already exists from an earlier dispatch named '{name}'` (`:309`). In Scope `:44` deletes that guard.

The Risks bullet then offers as mitigation *"**require** the previous agent to have finished before a continuation dispatch"* — while Open Question 3 asks *"Should `dispatch` warn or refuse when the target directory already contains a live agent's coordination files?"*

**These are the same question, answered "required" in one section and "open" in the other.** A requirement with no named enforcer is a hope, and the Risks table presents it as a mitigation already applied while Open Questions says nobody has decided whether anything checks it.

Consequence: the deck moves from code-enforced mutual exclusion to nothing, in a design whose *purpose* is to make dispatching into an already-populated directory the normal case. Two dispatches into one directory produce colliding `work-done-{role}.md` files pinned to the same recorded cwd, and orchestration identity keyed by `(name, cwd)` cannot separate them — the failure the PRD correctly identifies as #156's root cause, now reachable through a supported path rather than only by user error. To answer the concurrency question directly: yes, two dispatches can be placed on the same directory before either spawns, and after M3 nothing detects it.

*Suggested fix:* promote OQ3 to a decision that lands **with** M3 rather than after it, and reword the `:87` mitigation to name the enforcer rather than state a requirement. If the answer is "refuse when coordination files are present", that is the replacement interlock and it belongs in In Scope.

### F4 — The deck's own context file has a fixed name and is overwritten unconditionally

`:47` and `:61`.

In Scope is careful that agent-written briefs go into `.dot-agent-deck/` under a **unique** name. The deck's own context file has no such protection — `orchestrator_context.rs:192,200` writes a fixed `orchestrator-context.md` unconditionally into the target directory. SC3 says this path is "unchanged from #222", which is accurate and is precisely the issue once the target directory is an arbitrary one that may already host an orchestration.

A continuation dispatch into a directory that previously ran an orchestration silently overwrites its context file. Sequentially that is benign and probably intended; combined with F3's missing interlock it is not, because two orchestrations sharing a cwd share one context file and the second's write is invisible to the first. The PRD's care about unique brief names shows the right instinct is already there — it just is not applied to the file the deck writes itself.

### F5 — The seam's interface is unstated: the CLI signature, and the three roles `<name>` currently carries

`:43`, `:67`.

In Scope says only *"`dispatch` accepts an explicit target directory and no longer creates worktrees"* — that is the whole specification of the boundary. It does not say whether the directory is positional or a flag, whether it defaults to the current directory (the in-place case at `:27` needs it to, or needs `.` passed explicitly), or what becomes of `<name>`.

`<name>` is not free to remove. It is documented as *"Short name for the dispatch unit (used for worktree naming)"* (`src/main.rs:129`) but carries **three** roles: the worktree slug and branch (the role M3 deletes); `task_name: format!("dispatch-{name}")` (`src/dispatch.rs:341`), which is what the tab and notifier show; and the pane-id component `sched-dispatch-<name>-<n>-r0` — **which this PRD's own Open Question 1 depends on** when it asks whether the delegate path can resolve such ids.

So M1 cannot be implemented without inventing the signature, and two implementers produce two different CLIs. One who reads "no longer creates worktrees" and drops `<name>` along with the worktree naming silently changes the tab name and the pane id — breaking the very id format OQ1 is tracking.

*Suggested fix:* add the proposed signature to In Scope, even provisionally, and state that `<name>` survives for roles 2 and 3 with its `--help` text corrected — it currently advertises the role being removed.

### F6 — Success Criterion 1 is unsatisfiable as worded

`:59` — *"…with the same end state `dispatch` produces today."*

Today's end state includes the sibling `../<repo>-dispatch-<slug>` path, the `agent/dispatch-<slug>` branch, the registry entry with `RemovalPolicy::KeepIfDirty`, and the single-use-name guard — **every one of which this PRD removes by design.** Read literally the criterion forbids the change it is the success criterion for, and it cannot serve as an acceptance test.

It also masks F1: *"same end state"* is exactly the phrase under which an implementer keeps `record_worktree` without ever consciously deciding to.

*Suggested fix:* scope it to what is meant — "the dispatched agent or orchestration is running in the specified directory with its task delivered, as `dispatch` achieves today" — and let F1's ownership rule state the rest explicitly.

### F7 — Success Criterion 2's operative clause is not observable

`:60` — *"…without re-deriving prior analysis."*

The first half is testable; the second half is LLM behaviour and no test can distinguish it from its opposite. M4 already states the testable version ("the orchestrator works from the existing branch and files"), so the criterion is weaker than the milestone it corresponds to — and SC5 promises an L2 PTY test whose author has no way to assert the criterion as worded. Recommend adopting M4's phrasing.

### F8 — M6 schedules the rule 12 check, but the PRD never records the expected answer

`:72`.

Rule 12's obligation is to *answer* "did this change the TUI↔daemon contract?", and there is a candidate semantic break worth naming now rather than discovering at M6: `SpawnRequest.working_dir` currently always carries a deck-derived worktree path and afterwards carries an arbitrary caller-chosen directory. Same wire, different meaning, across a version boundary — the case rule 12 says needs a `changelog.d/<issue>.breaking.md` fragment rather than a `PROTOCOL_VERSION` bump. Low severity since M6 does schedule the check, but the PRD is the natural place to record it, and the bump policy follows from it.

---

None of this argues against the proposal. It argues that **placement and cleanup are one decision, and the PRD has currently split only the first half.** F1–F3 are the two-phase-operation failure this design needs to answer before anyone builds against the text; requesting changes is about fixing that sequencing before implementation, not about withholding the document from the tree.

vfarcic:
@prageethw CI is green again on `d4e61f9` — all 10 checks pass ([run 31618590185](https://github.com/vfarcic/dot-agent-deck/actions/runs/31618590185)).

The `build-windows` failure was **stale, not a defect in this PR**: it was issue #511 (the `clean_tmp` test handing a Unix-only `/scratch/one` to `--root`, which fails `is_absolute()` on Windows), fixed on main by #512. I updated the branch from main with `gh pr update-branch` — a merge commit, no rebase and no force-push. `build-windows` now passes 1579/1579, including the previously failing `clean_tmp::tests::explicit_roots_replace_the_standard_set`.

**To be explicit about what this does and does not change:** the only delta since your review at `2a3f8b4` is that merge commit. The PRD content is byte-identical, so **F1–F8 are all still outstanding** — this is a CI-only update, not a response to your review.

Re-requesting so it lands back in your queue now that CI is clean, but it's entirely reasonable to leave it parked until the ownership/cleanup findings (F1–F3 especially) are actually answered in the document.

prageethw:
Confirming your read — parking this rather than re-reviewing, so it does not sit in my queue looking like it needs something from me.

The only delta since `2a3f8b4` is the `main` merge commit and the PRD content is byte-identical, so F1–F8 all still stand. A second pass over unchanged prose would produce the same eight findings and cost you a round trip to be told so.

Thanks for saying explicitly that it was a CI-only update. That is the kind of thing that is easy to leave implicit and expensive to guess wrong in either direction.

Back in my queue the moment F1–F3 are answered in the document — those three decide whether the rest is worth re-reading, since ownership and cleanup shape every milestone below them.

prageethw:
Housekeeping, not disengagement — moving this off my review queue and onto yours, since the ball has been in your court since 2026-08-13 and the queue state said otherwise.

Dropped myself as a requested reviewer and assigned you. Nothing about the review changes: it stays `CHANGES_REQUESTED`, F1–F8 all still stand, and the PRD is byte-identical to what I reviewed at `2a3f8b4` (last touched 2026-08-11; `d4e61f9` is just the `main` merge).

**This is self-healing, which is why it is safe to do.** `.github/CODEOWNERS` is pathless (`* @vfarcic @prageethw`) and GitHub omits the author when auto-requesting, so **the next push to this branch re-requests me automatically**. That is exactly the condition I set when I parked it — back in my queue the moment F1–F3 are answered in the document, without either of us having to remember.

The three that gate the rest, restated so they are in one place:

- **F1** — `dispatch` does not merely create a worktree, it claims it for deletion (`record_worktree(… RemovalPolicy::KeepIfDirty)`, read by the tab-close handler). The PRD never mentions registration, removal, `RemovalPolicy` or the close path in 93 lines, so both readings available to an implementer are harmful and the document picks neither.
- **F2** — the orphan mitigation keeps the `agent/dispatch-*` prefix so `git worktree list` can identify deck-created units, but the only thing producing that prefix is `derive_dispatch_paths`, which the same PRD deletes. Afterwards it can only be *requested* by a seed, never *produced* by code.
- **F3** — deleting the single-use-name guard removes the only mutual exclusion (`worktree … already claimed by another dispatch`), and the offered replacement — require the previous agent to have finished — is contradicted by the PRD's own Open Question 3, which asks whether dispatch should warn or refuse at all.

They gate F4–F8 because ownership and cleanup shape every milestone below them; answering these three decides whether the rest is worth re-reading.

No rush from my side — the design question is worth more than the turnaround.

prageethw (approved):
All eight findings from my prior CHANGES_REQUESTED review (F1-F8) are verified addressed in the current diff — the ownership/cleanup section, the orphan-worktree mitigation honesty, the interlock decision, the context-file rationale, the CLI seam signature, and both success-criterion rewordings all map cleanly onto what I asked for. CI is green and no review threads remain unresolved.

PR #473, opened 2026-08-09
docs(governance): route every PR to the other maintainer via a pathless CODEOWNERS

vfarcic:
## What

`main-protected` has required one approving review since 2026-08-09, and GitHub never counts a self-approval — so the *review* half of the rule is genuinely enforced. The **routing** was not. On the day approvals were raised to 1, seven open PRs from the owner had no reviewer requested at all, so the maintainer whose approval they needed had no signal they existed.

- **`.github/CODEOWNERS`** — one pathless rule, `* @vfarcic @prageethw`. GitHub omits the PR author when auto-requesting review from code owners, so this routes every PR to the other maintainer automatically, including PRs opened by an agent (where the gap actually showed).
- **`docs/develop/governance.md`** — a new *How review is requested, and who merges* section, plus a rewrite of the paragraph that used to say there is deliberately no `CODEOWNERS`.
- **`MAINTAINERS.md`** — same correction, and a note to update `CODEOWNERS` in the same change as the maintainer table.
- **`CLAUDE.md`** — the process in rule 8, and in rule 13 why this could not live in the prd-done skill.

## This narrows the earlier "no CODEOWNERS" decision rather than reversing it

What that decision rejected was *per-path* ownership: a path list restates "a maintainer must approve", which the approval count already says, while going stale silently on every rename or split — and stale code-owner paths don't error, they just stop routing. A pathless `*` has no paths to go stale, so that argument doesn't reach it.

It also stays a **router, not a second gate**: `require_code_owner_review` remains `false`, so any maintainer's approval satisfies the one required review exactly as before.

## Process, since three ruleset settings each imply a non-obvious step

- Review is requested via the PR's reviewer **field**. An `@mention` in a comment notifies but never enters the reviewer's *Review requested* queue and isn't what the approval rule counts. Assigning the other maintainer is also wrong — an assignee is whoever drives the PR, normally the author.
- **Request review last.** `dismiss_stale_reviews_on_push: true` voids an approval on any later push, so settling CI and Greptile first is the difference between one round trip and two.
- **The author merges after approval** — nothing constrains who presses the button (`require_last_push_approval` is `false`) and the approver isn't obliged to. But never merge your own *unapproved* PR: for the owner the admin bypass makes that succeed silently rather than fail, which is how the gate decays into ceremony. Flows ending in a merge (`/prd-done`, `/prd-full`) must stop at "green, reviewed, approved".

## Why not the prd-done skill

`.claude/skills/dot-ai-*` is an upstream-synced mirror and a local edit there is provably temporary: `04a3641` narrowed prd-done's review-wait on 2026-06-14, and `e94388d` reverted it byte for byte on 2026-06-19 (`index e5f1b69..b95d270`, the exact inverse). That skill still says "wait for ALL reviews … CodeRabbit"; only CLAUDE.md rule 8 keeps `/prd-full` from hanging. Recorded in rule 13 so the next person doesn't retry it.

## Notes for the reviewer

- **`CODEOWNERS` is read from the base branch**, so this PR can't auto-request its own reviewer — I requested you manually. The next PR after this merges is the real canary.
- **After merge**, worth one command, since malformed entries disable routing without failing anything: `gh api repos/vfarcic/dot-agent-deck/codeowners/errors`
- **No changelog fragment** — governance and maintainer-facing docs only, nothing user-visible.
- `cargo fmt --check` passes. Clippy was **not** run: the diff touches no `.rs`/`.toml` file, and this is a fresh worktree where the gate would mean a full cold build for a guaranteed no-op.
- Docs-only, so `docs/develop/` stays out of `site/sidebars.js` per rule 11.


prageethw (approved):
Reviewed at `84c9c1d`, against `upstream/main`. **Approve.** The thesis holds: the original objection was to *per-path* ownership, a pathless `*` has no paths to go stale, so this genuinely narrows the earlier decision rather than reversing it. Keeping it a router was the right call.

I checked every ruleset claim against the live `main-protected` (id `20587589`) rather than taking them on trust — `required_approving_review_count: 1`, `dismiss_stale_reviews_on_push: true`, `require_last_push_approval: false`, `required_review_thread_resolution: true`, `require_code_owner_review: false`. All accurate. The four files also agree with each other clause by clause; no drift between the CLAUDE.md and `governance.md` near-duplicates.

Nothing below blocks. The first three are one-line edits I'd take before merging; the rest are decisions or follow-ups.

## Worth fixing before merge

**1. There are two stale claims, not one.** Greptile caught `scripts/apply-branch-protection.sh:36–37`. The other is `CONTRIBUTING.md:60`, which still says *"so there is deliberately no `CODEOWNERS`"* — in the bullet that summarises the very document this PR rewrites, so a reader who follows the link lands on text saying the opposite. I swept the tree for `CODEOWNERS`, `code owner`, `nothing to route`; those two are the complete remaining set.

On Greptile's one specifically: it's **cosmetic, not behavioural**. `payload()` hardcodes `"require_code_owner_review": false` at line 156, so `apply` re-asserts exactly the setting this PR relies on — the script can't clobber the routing. Only the prose is wrong. Worth fixing anyway because it sits directly above the `REQUIRED_APPROVALS` knob, read by someone in the act of changing protection. (Watch `usage()` — it's `sed -n '4,22p' "$0"`, so an edit that grows the header past line 22 truncates the help text.)

**2. "A malformed entry disables routing entirely" is an overstatement**, in `.github/CODEOWNERS:17–18`, `governance.md:23` and `CLAUDE.md:34`. GitHub's docs: *"If any line in your CODEOWNERS file contains invalid syntax, that line will be skipped."* Line-scoped, not file-scoped. Indistinguishable today with one rule — but stated as a general fact, in a file that anticipates being edited. Suggest: *"a malformed line is skipped silently — and with a single rule, one skipped line is the whole of the routing"*, which is accurate and if anything more alarming.

**3. The recommended check can't see the edit it's recommended for.** `codeowners/errors` defaults `ref` to the default branch. The bare command is prescribed in three places, each phrased *"after editing it"* — but at that moment your change is on a branch, so it reports on `main` and returns clean for a file you've just broken. That's the same silently-passing check the PR is written to guard against. Suggest `?ref=<branch>` pre-merge, bare form after merge.

Also worth stating what a good answer looks like, since none of the three places does. Measured: this repo → `404`; `cli/cli` → `{"errors":[]}`. So **after merge a 404 is itself the alarm**, not an absence of news — it means GitHub found no CODEOWNERS at that ref. Useful property: `gh` exits non-zero on the 404, so a scripted check fails loudly.

## Worth a decision

**4. Renovate turns the router into noise.** `renovate[bot]` isn't an owner, so both maintainers get requested on every Renovate PR — and Renovate's `pull_request` bypass plus five `automerge` groups in `renovate.json` means those merge on green CI without anyone acting on the request. Measured: **24 of the last 60 merged PRs (40%) were Renovate's.**

The predictable steady state is a stream of review requests that resolve themselves, which is exactly the habituation that teaches you to ignore the queue — the decay this PR exists to prevent, relocated from "no signal" to "signal nobody reads". This is the one finding that can undo the PR's own goal.

I don't have a clean fix, and would rather say so than hand over one that doesn't work: CODEOWNERS has no author-based exclusion, and Renovate's `reviewers`/`ignoreReviewers` govern reviewers *Renovate* adds, not GitHub's code-owner auto-request. The alternative — a trailing ownerless rule over the paths Renovate touches — would also un-route *human* PRs to `.github/workflows/`, which I wouldn't trade. My suggestion is to accept it and write one sentence in `governance.md`, so it lands as a priced cost rather than a surprise.

**5. The router and "request review last" pull against each other.** Routing fires when the PR is *opened*, so on a non-draft PR the reviewer is requested first — the precise thing the next sentence says not to do. With `dismiss_stale_reviews_on_push: true`, a reviewer who responds promptly gets their approval voided by the author's next push. Under the manual regime "request review last" was enforceable by not typing `--reviewer` yet; automatic routing removes that lever and the doc doesn't say what replaces it.

Every piece of the answer is already in the PR: **open as draft → settle CI and Greptile → `gh pr ready`** makes `gh pr ready` the moment routing fires, so "request review last" becomes mechanical instead of aspirational. But `CLAUDE.md:34` currently says *"do not open a PR with `--draft` unless you mean to"*, which points the other way. I'd invert it.

**6. "The maintainer who did not open it" is imprecise for the case you most want covered.** On a PR from anyone who isn't an owner — outside contributor, Renovate, or an agent under a separate identity — neither owner is the author, so GitHub requests **both**. The PR body names agent-opened PRs as the motivating gap, so this isn't hypothetical. Suggest: *"to the other one on a maintainer's PR, to both on anyone else's."*

## Follow-ups, not objections

**7. The one silent failure not listed.** An owner losing write access — or being renamed or deleted — invalidates the entry and stops routing, with **no edit to CODEOWNERS at all**. So the `codeowners/errors` check, bound to *"after any edit"*, never fires for it. The check is right; its trigger is the wrong event. One clause in the CODEOWNERS comment closes it: *"both entries must retain write access — GitHub silently drops an owner who doesn't have it."* (Unverified: GitHub documents that endpoint as listing *syntax* errors, so I wouldn't assume it catches an ineligible owner.)

**8. The router-not-gate invariant is unobservable.** `apply-branch-protection.sh:201` prints `.rules[] | .type` — rule types only, never the `pull_request` parameters. So `require_code_owner_review`, the flag the whole "router, not a gate" claim rests on, is invisible to the repo's own inspection tool. A UI toggle to `true` would be silently reverted by the next `apply` rather than surfaced.

**9. Nothing says an approval has to be a *human* act.** GitHub enforces "not the same account", not "a second person looked". This repo runs agents under maintainer identities, so an agent on one maintainer's token can satisfy the sole required review on the other's PR. The new text closes the adjacent hole by name — *"never merge your own unapproved PR"* — but is silent on the approve step, which is the one that actually stands in for review.

**10. The offboarding path is one clause where the grant path is a sequenced procedure.** With exactly two maintainers, losing one doesn't degrade the gate gracefully — it makes `main` mergeable only by admin bypass. Worth stating alongside the revoke instruction.

**11. Flagging so nobody later "disproves" it:** *"GitHub omits the PR author when auto-requesting from code owners"* carries the entire mechanism and appears in all four files, but I couldn't find it stated in GitHub's docs — not on About code owners, not in the REST reference. It's real and structurally forced (the review-request API won't accept the author at all), so I'm satisfied it holds. Noting it only so a future reader who checks the docs, finds nothing, and concludes the PR was wrong has this instead.

**12. The existing backlog stays unrouted.** CODEOWNERS only routes PRs opened after it lands. Four open PRs currently have no reviewer requested: #469 and #464 (yours), #427 and #419 (mine). `governance.md:36` correctly covers this case with the manual commands, but nothing assigns the cleanup — and an unassigned cleanup is the failure mode this PR opens by describing. Four `gh pr edit <n> --add-reviewer` calls in the same sitting as the merge.

## Conventions

Rule 10 (no hard-wrapped prose) — clean; the wrapped lines are in `.github/CODEOWNERS`, a config file rule 10 doesn't reach. Rule 11 — `governance.md` stays out of `site/sidebars.js` and `docusaurus.config.js:41` still excludes `develop/**`. No changelog fragment is the right call. The disclosed clippy skip is fine; only observation is that rule 2 reads unconditionally, so if a docs-only carve-out is meant to be standing it belongs in rule 2 rather than re-argued per PR.

Nice touch, probably accidental: `*` is recursive in CODEOWNERS, so the file owns itself, which is exactly GitHub's own hardening advice.

PR #497, opened 2026-08-11
fix: bound client-side frame read by MAX_FRAME_LEN (#478)

### LINKED ISSUE #478: Client-side frame read allocates from an unbounded u32 length — `MAX_FRAME_LEN` is enforced in `read_frame` but not in `send_daemon_request_blocking_with_timeout`
`read_frame` refuses a frame longer than `MAX_FRAME_LEN` (16 MiB) specifically so a forged length prefix cannot make us allocate gigabytes. The blocking client helper on the TUI side reads the same protocol and applies no such bound — it allocates directly from an attacker-influenced `u32`.

Found during a code audit of an unrelated change that made this call site newly load-bearing on an interactive keypath.

## The asymmetry

`src/daemon_protocol.rs` — guarded, with the rationale stated in the comment above `MAX_FRAME_LEN`:

```rust
/// … or buggy peer trying to allocate gigabytes off a forged length prefix.
/// 16 MiB is well above any reasonable PTY chunk or scrollback snapshot.
const MAX_FRAME_LEN: usize = 16 * 1024 * 1024;

let len = u32::from_be_bytes([header[1], header[2], header[3], header[4]]) as usize;
if len > MAX_FRAME_LEN {
    return Err(io::Error::new(
        io::ErrorKind::InvalidData,
        format!("frame length {len} exceeds {MAX_FRAME_LEN}"),
    ));
}
let mut payload = vec![0u8; len];
```

`src/ui.rs`, `send_daemon_request_blocking_with_timeout` — unguarded:

```rust
let len = u32::from_be_bytes([
    resp_header[1], resp_header[2], resp_header[3], resp_header[4],
]) as usize;
let mut body = vec![0u8; len];
stream.read_exact(&mut body)?;
```

`u32::MAX` is ~4 GiB. The allocation happens *before* `read_exact`, so the peer does not need to send the bytes — a 5-byte header is enough to trigger it.

## Impact

Bounded by the threat model: the attach socket is created `0o600`, so this needs a same-uid peer or a compromised/buggy daemon rather than a remote attacker. That is why I would not call it critical.

What makes it worth fixing anyway is that it is a **guard the codebase already decided it wanted**, applied on one side of the protocol and not the other. The reasoning in `MAX_FRAME_LEN`'s own comment — "a buggy peer" — covers the unguarded path just as well as the guarded one. A malfunctioning daemon is squarely inside the case the constant was written for.

This runs synchronously on the `Ctrl+n` new-pane keypath, so the failure mode is the TUI ballooning or being OOM-killed on a keystroke.

## Second, smaller point

`stream.set_timeouts(timeout)` sets a **per-syscall** timeout, not a total deadline. A peer that drips one byte per interval keeps `read_exact` alive indefinitely while every individual read stays inside the timeout. Call sites documenting this helper as bounded ("a wedged daemon can't freeze form-open for more than a quarter second") are accurate for a *silent* peer and not for a *slow* one.

## Suggested fix

Apply the same `MAX_FRAME_LEN` check on the client read before allocating. It is a handful of lines and reuses the existing constant. The slow-drip point is a separate, larger change (a total deadline rather than per-syscall timeouts) and may be worth its own issue — happy to split it out if you prefer.

I'm glad to open a PR for the bound if that's useful.

vfarcic:
Closes #478.

`read_frame` refuses a frame over 16 MiB precisely so a forged or corrupted length prefix cannot make us allocate gigabytes. The TUI's synchronous one-shot client read the same 5-byte header and sized its body buffer straight off the `u32` with no such check. The allocation happened *before* `read_exact`, so a 5-byte header claiming ~4 GiB was enough — the peer never had to send a byte of payload. This runs synchronously on the `Ctrl+n` new-pane key path (via `live_orchestration_cwds`), so the failure mode was the TUI ballooning or being OOM-killed on a keystroke.

Threat model is as @prageethw scoped it in the issue: the attach socket is `0o600`, so this needs a same-uid peer or a buggy/compromised daemon rather than a remote attacker. It is worth fixing because it is a guard the codebase already decided it wanted, applied on one side of the protocol and not the other — and "a buggy peer", the case `MAX_FRAME_LEN`'s own comment names, covers the unguarded path exactly.

## What changed

- **`src/daemon_protocol.rs`** — `MAX_FRAME_LEN` widened from private to `pub(crate)`, with a note that the bound belongs to the *protocol*, not to `read_frame`. The 16 MiB literal is still spelled exactly once.
- **`src/ui.rs`** — the response decode is split out of `send_daemon_request_blocking_with_timeout` into `read_resp_frame_blocking`, which checks the length against `MAX_FRAME_LEN` before `vec![0u8; len]` and returns the same `io::ErrorKind::InvalidData` in the same message shape (`frame length {len} exceeds {MAX_FRAME_LEN}`) as `read_frame`. Splitting it out is what makes the framing testable — the send path itself needs a live socket at `config::attach_socket_path()`.
- **`changelog.d/478.bugfix.md`**.

Apart from the new bound, `read_resp_frame_blocking` is the original code moved verbatim, so every frame within the cap decodes exactly as before.

## Other client-side readers

`src/ui.rs:761` was the only unbounded reader in production. Every other consumer — `daemon_client.rs`, `embedded_pane.rs` — goes through `read_frame`, which was already bounded. `tests/common/mod.rs` has two near-identical copies, but they read from a daemon the test itself spawned, so there is no untrusted peer, and `pub(crate)` deliberately does not reach the integration-test crate. Left alone; happy to widen if you'd rather they share the constant.

## Surfaced as a failed request, not a panic

All four callers already treat `Err` as a failed request: `live_schedule_names` returns `HashSet::new()`, `live_orchestration_cwds` uses `let Ok(..) else { return Vec::new() }` (fails open — the form opens with no hint), `ScheduleRunNow` shows `Run-now failed: {e}`, and the reload-after-delete discards with `let _`. An over-long frame therefore degrades rather than taking the TUI down.

**Not a break — no `PROTOCOL_VERSION` bump, and no `.breaking.md` fragment.** Checked rather than assumed:

- **No wire shape change.** Same 5-byte header (1 kind + 4 BE length), same payload, no field added, removed, or retyped.
- **No semantic break behind a stable wire.** No field changes meaning; this only rejects input.
- **The newly-rejected input is unreachable from any conformant daemon.** The sole daemon-side RESP writer is `daemon_protocol.rs:1814`, which goes through `write_frame` — and `write_frame` has always returned `InvalidInput` for a payload over the same `MAX_FRAME_LEN`. No released daemon can emit a frame this client now refuses.

Old daemon ↔ new TUI and new daemon ↔ old TUI therefore behave exactly as before, so the manual previous-release-daemon run does not apply here. If a reviewer disagrees with that reading, the run is the next step and I'll do it.

## Gates

- `cargo test-fast` — **1661 passed, 0 failed**.
- `cargo test-e2e` — 3698 run, 3694 passed, **4 failed**. All four investigated and none is from this change:
  - `chain_smoke_pi_002_worker_receives_delegate_and_signals_work_done` — flake; **passed on rerun**.
  - `focus_007_lock_governed_focus_contract_on_real_binary`, `idle_worker_011_silent_worker_prompt_is_visible_in_attached_tui`, `card_stats_005_real_agent_card_narrows_without_restructuring` — fail consistently, and **fail identically on the pre-change baseline `a33ff21`**, which I checked out and ran to confirm. They are pre-existing on `main` and all three are TUI focus/render assertions that never touch the sync daemon-client decode path. Flagging them here since they are not this PR's to fix.

## Tests

Four unit tests next to the code, mirroring `daemon_protocol::tests::frame_rejects_oversize`:

- `resp_frame_rejects_oversize_before_allocating` — feeds a **header-only** 5-byte frame claiming `MAX_FRAME_LEN + 1` and `u32::MAX`, asserts `InvalidData`, and asserts the cursor never advanced past byte 5, i.e. the body read was never attempted and no buffer was ever sized from the forged length.
- `resp_frame_accepts_exactly_max_frame_len` — pins `>` vs `>=`: an at-limit frame must reach the body read (failing as `UnexpectedEof` here) rather than being rejected as over-long.
- `resp_frame_decodes_well_formed_response` and `resp_frame_rejects_wrong_kind` — the happy path and the pre-existing kind check still hold.

I verified the oversize test genuinely gates the fix by removing the bound and watching it fail (`UnexpectedEof` instead of `InvalidData`) before restoring it.

No TUI test under rule 4: the guard has no observable effect in normal operation — it is reachable only from a malfunctioning peer — so the framing seam is the right level. `read_resp_frame_blocking` and its callers are private, so an integration test could not reach them anyway.

## Not addressed

The issue's **second point** — `set_timeouts` being a per-syscall timeout rather than a total deadline, so a slow-drip peer keeps `read_exact` alive indefinitely while every individual read stays inside the timeout — is untouched here, as @prageethw suggested. It still applies after this change, including to the `DAEMON_HINT_TIMEOUT` call site whose doc comment describes it as bounded ("can't freeze form-open for more than a quarter second"), which remains accurate for a *silent* peer and not for a *slow* one. Worth its own issue.

### DIFF
--- a/changelog.d/478.bugfix.md
+++ b/changelog.d/478.bugfix.md
@@ -1,0 +1,7 @@
+## Bounded Frame Length on the TUI's Synchronous Daemon Client
+
+The TUI's one-shot daemon queries now reject an over-long response frame instead of allocating from the length prefix on the wire. `read_frame` has always refused a frame larger than 16 MiB precisely so a forged or corrupted length cannot make the process allocate gigabytes; the synchronous client the TUI uses for its blocking queries read the same 5-byte header without that check and sized its buffer straight off the `u32`. Since the allocation happened before the body was read, a 5-byte header claiming 4 GiB was enough — no payload required.
+
+This ran on the `Ctrl+n` new-pane key path, where the deck asks the daemon which directories already host a live orchestration, so the visible failure was the TUI ballooning or being killed on a keystroke. Both sides now read the same bound and return the same `InvalidData` error, and every caller already treated a failed request as best-effort, so an over-long frame degrades to "no hint" or a `Run-now failed: …` status line rather than taking the TUI down.
+
+The wire format is unchanged. A daemon can never have sent a frame this affects — the writer has always enforced the same cap — so builds across versions interoperate exactly as before.
--- a/src/daemon_protocol.rs
+++ b/src/daemon_protocol.rs
@@ -197,7 +197,13 @@ pub const PROTOCOL_VERSION: u32 = 6;
 /// Hard cap on a single frame's payload length. Defends against a malicious
 /// or buggy peer trying to allocate gigabytes off a forged length prefix.
 /// 16 MiB is well above any reasonable PTY chunk or scrollback snapshot.
-const MAX_FRAME_LEN: usize = 16 * 1024 * 1024;
+///
+/// `pub(crate)` rather than private (issue #478) because the bound belongs to
+/// the *protocol*, not to [`read_frame`]: the TUI's synchronous one-shot client
+/// (`ui::send_daemon_request_blocking_with_timeout`) decodes the same 5-byte
+/// header without the async reader, and used to allocate straight off the u32.
+/// Both sides now read the one constant — do not re-spell the 16 MiB literal.
+pub(crate) const MAX_FRAME_LEN: usize = 16 * 1024 * 1024;
 
 /// Bounded timeout for a single STREAM_OUT/STREAM_END write to a client. If
 /// a client stops draining its socket, the OS send buffer fills and our
--- a/src/ui.rs
+++ b/src/ui.rs
@@ -735,8 +735,8 @@ fn send_daemon_request_blocking_with_timeout(
     req: &crate::daemon_protocol::AttachRequest,
     timeout: std::time::Duration,
 ) -> std::io::Result<crate::daemon_protocol::AttachResponse> {
-    use crate::daemon_protocol::{KIND_REQ, KIND_RESP};
-    use std::io::{Read, Write};
+    use crate::daemon_protocol::KIND_REQ;
+    use std::io::Write;
 
     let path = config::attach_socket_path();
     let mut stream = crate::platform::ipc::IpcClient::connect(&path)?;
@@ -750,8 +750,35 @@ fn send_daemon_request_blocking_with_timeout(
     stream.write_all(&payload)?;
     stream.flush()?;
 
+    read_resp_frame_blocking(&mut stream)
+}
+
+/// Read one `KIND_RESP` frame off a synchronous reader and decode its body.
+///
+/// Split out of [`send_daemon_request_blocking_with_timeout`] so the framing
+/// can be exercised against a plain in-memory reader (issue #478) — the send
+/// path itself needs a live socket at [`config::attach_socket_path`].
+///
+/// The length prefix is checked against
+/// [`crate::daemon_protocol::MAX_FRAME_LEN`] BEFORE the body is allocated,
+/// mirroring [`crate::daemon_protocol::read_frame`] — same bound, same
+/// [`std::io::ErrorKind::InvalidData`], same message shape. Without it a
+/// 5-byte header claiming `u32::MAX` made this allocate ~4 GiB without the
+/// peer sending a single body byte, and this runs SYNCHRONOUSLY on the
+/// `Ctrl+n` new-pane key path (via [`live_orchestration_cwds`]), so the
+/// failure mode was the TUI ballooning or being OOM-killed on a keystroke.
+/// The socket is `0o600`, so the reachable case is a buggy or compromised
+/// same-uid daemon — exactly the "buggy peer" `MAX_FRAME_LEN` was written for.
+/// Every caller already treats `Err` as a failed request (best-effort hint,
+/// or a `Run-now failed: …` status line), so an over-long frame degrades
+/// instead of panicking.
+fn read_resp_frame_blocking<R: std::io::Read>(
+    r: &mut R,
+) -> std::io::Result<crate::daemon_protocol::AttachResponse> {
+    use crate::daemon_protocol::{KIND_RESP, MAX_FRAME_LEN};
+
     let mut resp_header = [0u8; 5];
-    stream.read_exact(&mut resp_header)?;
+    r.read_exact(&mut resp_header)?;
     if resp_header[0] != KIND_RESP {
         return Err(std::io::Error::other(format!(
             "expected RESP frame, got kind 0x{:02x}",
@@ -764,8 +791,14 @@ fn send_daemon_request_blocking_with_timeout(
         resp_header[3],
         resp_header[4],
     ]) as usize;
+    if len > MAX_FRAME_LEN {
+        return Err(std::io::Error::new(
+            std::io::ErrorKind::InvalidData,
+            format!("frame length {len} exceeds {MAX_FRAME_LEN}"),
+        ));
+    }
     let mut body = vec![0u8; len];
-    stream.read_exact(&mut body)?;
+    r.read_exact(&mut body)?;
     serde_json::from_slice(&body).map_err(std::io::Error::other)
 }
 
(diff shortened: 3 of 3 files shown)


prageethw (approved):
Reviewed at `b64650c` — code read only, no local build or test run (CI is green here and I did not re-run it locally). I read this twice, once as ordinary review and once as a security pass; both landed on approve.

The fix is correct and the hard part is right: **the bound strictly precedes the allocation.** `len > MAX_FRAME_LEN` at `src/ui.rs:794` comes before the only `len`-sized allocation at `:800`, with no `with_capacity` or `reserve` in between — which is the whole vulnerability, and the ordering is what closes it. Some specifics I verified rather than assumed:

- **Symmetry holds exactly.** `len > MAX_FRAME_LEN` is the identical predicate at `daemon_protocol.rs:244` (`read_frame`), `:273` (`write_frame`) and the new `ui.rs:794` — same comparison, same `InvalidData`, same message. No `>` / `>=` drift between the two ends, and the at-limit case is pinned by a dedicated test.
- **The fix is complete.** Only two length-prefix readers exist in `src/` — `daemon_protocol.rs:243`, already bounded, and `ui.rs:788`, bounded here. There is no third.
- **No stream desync.** The connection is one-shot and `stream` is a local dropped when `?` propagates, so the socket is torn down rather than optimistically resynchronised mid-frame.
- **Integer handling is sound on every supported target.** The length is `u32`, so `as usize` is a *widening* conversion — the 32-bit truncation hole does not arise. Explicit big-endian both directions, and no arithmetic before the check.
- **No remote surface.** `connect` is an `ssh -t` exec wrapper and the laptop-side socket bridge was deleted in M2.7, so this is same-host only: a `0o600` uid-qualified Unix socket, or a Windows named pipe with an owner-SID check. The threat model in the PR description is accurate.
- **No new rejection of legitimate traffic.** `KIND_RESP` has exactly one writer, `write_resp`, which goes through the already-capped `write_frame` — so no released daemon can emit a frame this client now refuses. That is also why **no `PROTOCOL_VERSION` bump and no `.breaking.md`** is the right call, which I checked explicitly since a new client-side rejection is the shape that usually does need one.
- **No panics on the length path** — no `unwrap` / `expect` / `try_into`, indexing is on a fixed `[u8; 5]`, and `serde_json`'s recursion limit is at its default (the crate does not enable `unbounded_depth`).
- **All four callers degrade cleanly** — `HashSet::new()`, a fail-open `Vec::new()`, a `Run-now failed: {e}` status line, and a deliberate `let _`. No panic and no frozen dashboard.

`resp_frame_rejects_oversize_before_allocating` asserting `cursor.position() == 5` after rejection is the right test: it proves the body read was never *attempted*, which is the correct observable proxy for "nothing was allocated from the forged length" — much better than only asserting that an error came back.

Correcting my own initial read: the changed-file list suggests no tests, but four unit tests ship inline in `src/ui.rs`'s `mod tests`, covering over-limit, at-limit, happy path and wrong-kind. I have no missing test to name.

---

### 1. The deferred half of #478 will lose its only record when this merges

`src/ui.rs:801`.

Issue #478 has a second point — per-syscall timeout versus a total deadline — which this PR deliberately defers and discloses. That is a reasonable scope call. The problem is bookkeeping: there is no follow-up issue for it anywhere upstream, and `Closes #478` will delete the only record that it exists.

The residual is real. `set_timeouts` is per-syscall and re-arms on every byte, so a peer that declares a *legal* length and then writes one byte per 200 ms holds the synchronous `Ctrl+n` key path open indefinitely — at the 16 MiB cap that is roughly 38 days. Worth noting that the cap gives a slow drip *more* legal runway, not less, so this bound and that hole interact rather than being independent.

**The fix pattern already exists in this repo**: `src/hook.rs:638` `read_reply_line` implements exactly the total-operation deadline with per-read re-arming. So the follow-up is a port, not a design problem.

Ask: file the follow-up and link it before merging, so the deferral stays tracked.

### 2. Minor

- **`src/ui.rs:800-801`** — `read_frame` guards its body read with `if len > 0` and this one does not. Provably harmless (`read_exact` on an empty buffer is a no-op) and carried over verbatim from the pre-change code — raised only because the PR positions the two as mirrors. Fine to leave.
- **The request *write* half of the same function** still hand-rolls its header with no bound, so the function is now asymmetric. I could not construct a reachable scenario, so treat this as an observation rather than a finding.
- **The doc comment and changelog overstate the failure mode.** They describe the header-only attack as balloon-then-OOM-kill, but `calloc` maps 4 GiB lazily, so under strict overcommit the real immediate failure is a `handle_alloc_error` **abort**. Documentation precision only — the fix is correct either way, and an abort is just as much a denial of service.

### On your open question about the `tests/common/mod.rs` duplicates

Leave them as they are, and I'd recommend explicitly **against** widening `MAX_FRAME_LEN` for the harness copies. Those read from a test-spawned daemon, and raising the constant's visibility to `pub` to satisfy a test would export a protocol internal to the whole crate's consumers. `pub(crate)` is the right visibility. Both my passes reached this independently.

Issue #436, opened 2026-08-08
cargo clippy never lints the xtask/* workspace members, and one lint is already sitting there


vfarcic:
Split out from #407.

Rule 2's clippy gate — even after #407 widens it to `cargo clippy --all-targets --features e2e -- -D warnings` — still only covers the **root package**. Cargo's default target selection in a workspace with a root package is that package alone, so `xtask/docs`, `xtask/linkage-check`, and `xtask/spec` are never linted by any gate. CI's `build` job (`.github/workflows/ci.yml`) inherits the same blind spot.

`xtask/spec` and `xtask-docs` are path dev-dependencies, so they get *built* as dependencies — but building is not linting, and `xtask/linkage-check` is not even built.

## This is not hypothetical — there is a live warning on `main`

```
$ cargo clippy --workspace --all-targets --features e2e -- -D warnings
error: consider using `sort_by_key`
   --> xtask/linkage-check/src/clean_tmp.rs:210:5
    |
210 |     out.sort_by(|a, b| b.bytes.cmp(&a.bytes));
    |     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    |
    = note: `-D clippy::unnecessary-sort-by` implied by `-D warnings`
help: try
    |
210 -     out.sort_by(|a, b| b.bytes.cmp(&a.bytes));
210 +     out.sort_by_key(|b| std::cmp::Reverse(b.bytes));
    |
```

Measured on `main` at 6b06069. This is why the change is separate from #407: adding `--workspace` there would have shipped a gate that fails on the first commit, so #407 deliberately stops at the root package and points here.

## Suggested fix

1. Fix `xtask/linkage-check/src/clean_tmp.rs:210` (the suggested `sort_by_key` + `Reverse` rewrite is a one-liner).
2. Re-run `cargo clippy --workspace --all-targets --features e2e -- -D warnings` and clear anything else it surfaces across the three `xtask` members.
3. Add `--workspace` to rule 2's command in `CLAUDE.md` and to the Linux `build` job's clippy step in `.github/workflows/ci.yml`, keeping them identical.
4. Leave `build-windows` / `build-macos` on bare `cargo clippy -- -D warnings`, for the reason already documented in those jobs' comments.

Step 3 should follow step 2 in the same PR, so the gate never lands red.

PR #498, opened 2026-08-11
ci: select the whole workspace in the test gates (#489, #470)

### LINKED ISSUE #470: cargo test-fast omits --workspace, so no gate ever runs the xtask crates' tests
## Symptom

`cargo test-fast` is an alias for `cargo nextest run` — **without `--workspace`**. Cargo's default target selection is the root package alone, so the `xtask/*` members' tests are executed by **no gate anywhere**:

- not by `cargo test-fast` (the per-task gate, rule 5)
- not by `cargo test-e2e` (`cargo nextest run --features e2e`, same default selection)
- not by CI, which runs the same aliases

`cargo clippy --workspace --all-targets --features e2e` *compiles* them, so a type error or lint is caught. But compiling is not running: assertions never execute.

## How it surfaced

While implementing #322, two new tests were added to `xtask/linkage-check/src/clean_tmp.rs`. They pass — but only when run explicitly:

```
cargo nextest run -p xtask-linkage-check
39 tests run: 39 passed, 0 skipped
```

Nothing in the normal workflow runs that command. The tests would have shipped green-by-default and silently stopped protecting anything the moment they broke.

## Relationship to #436

This is the exact gap `--workspace` closed for **clippy** in #436, which found an `unnecessary_sort_by` sitting in `xtask/linkage-check/src/clean_tmp.rs` on `main` because that crate was linted by no gate. CLAUDE.md rule 2 records the reasoning and notes the widening was free warm (0.20s either way).

The test gate still has the hole. Same crates, same cause, same fix shape.

## Proposed fix

Add `--workspace` to the `test-fast` alias (and consider `test-e2e` for symmetry) in `.cargo/config.toml`, then update CLAUDE.md rule 5.

Worth checking before landing:

- whether any xtask test is slow, needs network, or is otherwise unsuited to the fast tier — the whole point of the tier is that it is cheap and deterministic;
- the warm-cache cost of the widened alias, so rule 5 can record it the way rule 2 records clippy's.

## Why it was not fixed in #322's branch

#322 broadens a destructive command (`clean-e2e-tmp`) and already had to touch `clean_tmp.rs`, which **#461 is concurrently rewriting**. Changing a global cargo alias on that branch was judged scope creep on a file with an active conflict; the reviewer agreed. Filing separately so it is not lost.

Discovered while gating #322.

### LINKED ISSUE #489: xtask unit tests run in no gate: `cargo nextest run` selects the root package only
The sibling of #436, for tests rather than lints.

## What's wrong

`cargo test-fast` is `cargo nextest run`, and CI's `build`, `build-macos` and `build-windows` jobs all run bare `cargo nextest run`. The workspace root is itself a package and there is no `default-members`, so cargo's default target selection is **the root package alone**. Every test in `xtask/linkage-check`, `xtask/docs` and `xtask/spec` therefore executes in **no gate anywhere**.

Measured on `agent/dispatch-reaper-461`:

```
$ cargo nextest list            | grep -c clean_tmp     # what test-fast and CI run
0
$ cargo nextest list --workspace | grep -c clean_tmp
25
$ cargo nextest run --workspace
1729 tests run: 1729 passed        # vs 1657 without --workspace
```

This is exactly the mechanism #436 fixed for clippy ("Cargo's default target selection is the root package alone, so `xtask/docs`, `xtask/linkage-check` and `xtask/spec` were linted by no gate anywhere") — `--workspace` was added to the clippy command but never to the test command.

## Why it matters concretely

`xtask/linkage-check/src/clean_tmp.rs` (the `clean-e2e-tmp` reaper, a **deletion** tool) carries `source_has_no_pid_recycling_machinery`, whose own doc comment calls it "**The** guard against the recycled-PID branch coming back … and the only mechanism that can be one". That branch was built twice and both times deleted live e2e roots. The guard is a runtime assertion, so today it **never executes in CI** — restore the deletion-unsafe code verbatim and every gate stays green.

The compile-time half survives: `cargo clippy --workspace --all-targets` does *build* those tests, which is what still enforces `owner_of`'s signature. Only the runtime assertions are inert.

## Fix

Add `--workspace` to the test command in `.cargo/config.toml`'s `test-fast` / `test-e2e` aliases and to the CI jobs.

Worth doing as its own PR rather than folded into a feature branch: flipping the selection may surface pre-existing failures in `xtask/docs` or `xtask/spec` that have never run, and those should be triaged on their own rather than turning an unrelated PR red.

Found while verifying #467.

vfarcic:
Closes #489. Closes #470 (duplicate of #489, filed from a different verification).

## What was wrong

The workspace root is itself a package and there is no `default-members`, so cargo's default target selection is the **root package alone**. Every test in `xtask/linkage-check` and `xtask/docs` executed in **no gate anywhere** — not `cargo test-fast`, not `cargo test-e2e`, and not CI, which runs the same commands.

```
cargo nextest run              # 1657 tests   <- what every gate ran
cargo nextest run --workspace  # 1705 tests
```

This is the test-side sibling of #436, which added `--workspace` to clippy. The reason the gap outlived that fix is that **clippy's half already covered the compile-time side**: those tests are *built*, so a type error or a lint in them still failed. Only their **runtime assertions** were inert — a test could be gutted to `assert!(true)` and every gate stayed green. That matters most for `xtask/linkage-check/src/clean_tmp.rs`, the `clean-e2e-tmp` reaper: it is a **deletion** tool whose safety properties (owned-prefix matching, opt-in untagged reaping, never following a symlink out of the temp root) are runtime assertions and nothing else.

## What changed

`--workspace` added everywhere the command is spelled out, so the copies cannot drift apart:

| Where | Why it is in scope |
|---|---|
| `.cargo/config.toml` — `test-fast`, `test-e2e` | the rule 5 gates themselves |
| `.github/workflows/ci.yml` — `build`, `build-windows`, `build-macos` | CI must match the aliases |
| `bacon.toml` — `test-fast`, `test-e2e` jobs | a watch loop narrower than the gate is worse than none |
| `.claude/skills/verify-pr/checks.sh` — `test-fast`, `e2e` steps | a reviewer's gate must not be narrower than the author's |

Plus CLAUDE.md rule 5, recording the mechanism and the measured cost the way rule 2 already does for clippy, and a `misc` changelog fragment.

### On `build-windows` / `build-macos`

The task flagged that rule 2 notes those two jobs deliberately diverge from Linux. **They get the flag anyway, and the divergence does not apply here.** That divergence is caused by `--all-targets --features e2e` — the L2 tier is Unix-only and reports dozens of E0425s against the Windows target — and `--workspace` is no part of that mechanism. Meanwhile the aliases live in `.cargo/config.toml` and apply on **every** platform, so a macOS or Windows contributor typing `cargo test-fast` now runs the xtask tests; if their CI job did not, a platform-specific break in them would be invisible on PRs and surface only on that contributor's machine — precisely the failure mode `build-macos` was added to close. The one xtask test with a platform dependency (`symlinks_named_like_owned_dirs_are_skipped`) is already `#[cfg(unix)]`.

## Are the newly-gated tests suited to the fast tier?

Checked explicitly, since the tier's whole point is that it is cheap and deterministic. **Yes, all 48.** Every one is in-memory or `tempfile::tempdir()`-based; none shells out to git, reads a repo file, touches the network, or sleeps. (`list_tests.rs` *does* have `git show` / merge-base helpers, but they live in the production `run()` path, not in any `#[test]`.) Nothing is excluded.

`xtask/spec` is a workspace member too but currently has **no** tests, so the 48 come from the other two crates.

**Cost is free warm**, matching rule 2's finding for clippy:

- the 48 tests run in **~0.07s** in aggregate; every one is under 40ms
- full tier, back to back: **26.4–26.9s** root-only vs **25.3–26.7s** with `--workspace` — below run-to-run noise, because the tier's wall clock is set by its slowest single test (~20s), not by its count
- warm target selection: ~0.9–1.5s either way

## Gates

| Gate | Result |
|---|---|
| `cargo clippy --workspace --all-targets --features e2e -- -D warnings` | clean |
| `cargo test-fast` (now `--workspace`) | **1705/1705 passed** |
| `cargo xtask linkage-check` | ok (497 catalog ids, 424 annotations, 7 rules) |
| `cargo test-e2e` (now `--workspace`) | 3742 selected; see below |

**No xtask test failed.** The issue anticipated that flipping the selection might surface pre-existing failures in the never-run crates — it did not. All 48 passed on the first run.

## Pre-existing e2e failures, unrelated to this PR

The first `cargo test-e2e` run reported 6 failures + 1 timeout. Triaged rather than assumed:

**4 were environmental.** A second full e2e suite was running concurrently in another worktree on the same machine, swap was exhausted (7G/7G) and `/tmp` — a tmpfs — was 64% full with 2.1 GB of stale `dad-tests-*` roots. That is CLAUDE.md rule 14's failure mode exactly, and it showed up as an `aws-lc-sys` gcc failure inside `version_001`'s nested build. After `cargo xtask clean-e2e-tmp --apply` and on a quiet machine, `devin_live_001`, `new_pane_016`, `version_001` and `chain_smoke_pi_002` all pass.

**3 fail on `main` too, with or without this change:**

- `focus_007_lock_governed_focus_contract_on_real_binary` — `tests/e2e_orchestration_focus.rs:212` — injecting `WaitingForInput` for the non-orchestrator `alpha` role never steers focus onto its expanded pane box
- `idle_worker_011_silent_worker_prompt_is_visible_in_attached_tui` — `tests/e2e_idle_worker_detector.rs:191`
- `card_stats_005_real_agent_card_narrows_without_restructuring`

Demonstrated by stashing this diff and rerunning the same seven tests on a pristine tree: **identical outcome, same 3 fail and same 4 pass.** That is the expected result — this diff touches no Rust source, no test source, and nothing that affects how the root package compiles; `--workspace` only widens *package selection*, so the root package's e2e binaries are byte-identical. These are broken on `main` and want their own issue; deliberately not fixed here, since folding unrelated e2e repairs into a gate-selection change is what #489 explicitly asked to avoid.

## Contract check (rule 12)

Not applicable: no daemon, protocol, orchestration or hook code is touched. No `PROTOCOL_VERSION` bump, no `.breaking.md` fragment — the changelog fragment is `misc`.


prageethw (approved):
Reviewed at `fbfb12f` — code read only, no local build or test run (CI is green here and I did not re-run it locally).

This is a clean fix for a gate that was passing while testing less than it claimed, and I checked the load-bearing claims rather than taking them:

- **`--workspace` genuinely reaches every member.** `Cargo.toml` declares `members = [".", "xtask/docs", "xtask/linkage-check", "xtask/spec"]` with **no** `default-members` and **no** `[workspace] exclude`, and `find . -name Cargo.toml` (excluding `target/`) returns exactly those four manifests — so no crate lives outside the workspace and nothing escapes the selector. The causal claim in the PR body is right: root is itself a package, so with no `default-members`, cargo's default selection was the root package alone.
- **Every spelled-out surface is consistent.** Grepping the tree for `nextest run` finds nine literal spellings; this PR updates seven of them, and the two it leaves are filtered single-test reruns, so neither is a gate. Everything else in the repo goes through the aliases (`CONTRIBUTING.md`, `reproduce-first/SKILL.md`, `.dot-agent-deck.toml`, `Cargo.toml`'s comments) and inherits the fix for free. `Taskfile.yml` has no test entrypoint, so there is no missed surface there.
- **Feature flags travel with the selection.** `e2e = []` is declared only on the root package; the three `xtask` manifests declare no `[features]`, which is fine — a bare `--features e2e` alongside `--workspace` applies to the selected packages that define it and errors only if none do. Both `test-e2e` spellings, `checks.sh:245`, and the two clippy gates (`ci.yml:170`, `checks.sh:161`) all carry `--all-targets --features e2e` where they should.
- **The `build-windows` / `build-macos` divergence is correctly left alone.** That divergence is caused by `--all-targets --features e2e` against a Unix-only L2 tier; `--workspace` is no part of that mechanism, and the new comments explain it accurately.
- **Nothing runs twice.** `--workspace` replaces the default package selection rather than adding to it.
- **`CLAUDE.md` rule 5 now matches the config byte-for-byte**, and the added paragraphs (mechanism, the compile-time/runtime split that let the gap survive #436, the measured cost) all match the diff.
- **The newly-gated tests are fit for the fast tier.** I spot-checked rather than assuming: the `Command::new("git")` calls and the `current_dir()` walks all live in production `run()` paths outside the `#[cfg(test)]` blocks, and the one platform-dependent test is already `#[cfg(unix)]`. Settled empirically anyway — `build`, `build-windows` and `build-macos` are all green on this head, so the widened selection really did execute those tests on all three platforms.

Both issues are fully covered (#489's aliases and CI jobs; #470's rule 5 update, fitness check and recorded cost), and the PR goes further with `bacon.toml` and `checks.sh`. Nothing below is blocking.

---

### 1. `bacon.toml`'s `clippy` job is still narrower than the gate it sits next to

`bacon.toml:17`.

The new comment three lines above states the principle — *"a watch loop that runs a narrower tier than the per-task gate is worse than no watch loop"* — and then `[jobs.clippy]` is still `["cargo", "clippy", "--", "-D", "warnings"]`. That is narrower than rule 2's required command in exactly the two ways this project has already been bitten by: no `--workspace` (#436's gap, this PR's direct sibling) and no `--all-targets --features e2e` (#407's gap, which compiled 64 e2e-gated test files to nothing and reported clean).

Failure scenario: a developer leaves the watch loop on the clippy job while editing `xtask/linkage-check/src/clean_tmp.rs` and introduces a lint there. Bacon reports clean all afternoon, because the bare command never selects that package. The rule 2 pre-commit gate or CI's `build` job then fails on a lint the watch loop was structurally incapable of seeing — the identical shape to the `unnecessary_sort_by` that sat in that exact file and is cited in rule 2.

Pre-existing rather than introduced here, so filing it as a follow-up is equally defensible — but this PR is editing this file specifically to close the "watch loop narrower than the gate" hole, and it closes the test half while leaving the lint half open under its own new comment. One-line change in a file already open in the diff.

### 2. Nothing prevents the selector from drifting back out of sync

Conceptually `.cargo/config.toml:19-20` as the source of truth.

After this PR the selector is duplicated across nine literal spellings in four files, kept in lockstep only by the comments the PR adds. There is no automated check that they agree — and the project has now had this exact drift twice, since #436 fixed clippy's selection and the identical test-side gap survived unnoticed until #489/#470. The failure mode is silent by construction: a gate that selects less still reports green.

Failure scenario: a future PR edits `ci.yml`'s `build` step — to add `--no-fail-fast`, or to split the job — and drops `--workspace` in the rewrite. CI stays green, the aliases still carry the flag locally, and the xtask crates quietly lose CI coverage again, detectable only by someone re-measuring the selected-test count, which is how #489 was found in the first place.

A `linkage-check` rule asserting that every `nextest run` spelling under `.cargo/`, `.github/workflows/`, `bacon.toml` and `.claude/skills/` carries `--workspace` would make the invariant enforceable rather than commented. Follow-up, explicitly not a reason to hold this PR.

### 3. Two stale copies outside the diff

Neither is in the diff, so they are here rather than inline:

- `.claude/skills/verify-pr/SKILL.md:124` and `:171` still spell `cargo nextest run --features e2e` without `--workspace`, while `checks.sh` in the same skill directory was updated. The practical impact is weak — both are filtered reruns of a named root-package test, so the selector is immaterial to their intended use. The one concrete miss is a reviewer following line 171 verbatim to rerun a failing **xtask** test per rule 6: the filter selects nothing and nextest reports zero tests rather than the failure they were chasing. The stronger argument is simply that these are the surviving copies that make the PR body's *"so the copies cannot drift apart"* not quite true.
- `flake.nix:235` still says *"`cargo test-fast` is an alias for `cargo nextest run`"*. The comment's actual point (the shell needs `cargo-nextest`) is unaffected and still correct; it is just one more place a reader learns the wrong expansion, which is the class of drift this PR exists to eliminate. Purely cosmetic.

---

On the numbers: "1657 → 1705 tests" is not independently verifiable without running the suite, which I did not — but it is internally consistent with #489's own 1657 → 1729 measurement on a different branch, and the direction and mechanism are confirmed by the manifest inspection above. Rule 12 is correctly marked not applicable (no daemon, protocol, orchestration or hook code touched) and a `misc` fragment is the right classification.

Issue #492, opened 2026-08-10
shell-activity: move the `ps` sample off the Tokio worker and bound it with a timeout


vfarcic:
## Problem

`process_table()` (`src/platform/proc/unix.rs:233-237`) runs a **synchronous** `std::process::Command::output()` with no timeout. It is called from `AgentPtyRegistry::shell_foreground_busy_snapshot`, which the daemon's shell-activity monitor invokes directly from an async task every 500ms.

So it is not merely untimed — it occupies a **Tokio worker thread** for the duration of the subprocess. A `ps` stuck in D-state (NFS, a wedged filesystem) stalls hook ingestion, client requests and shutdown, not just this signal. The 500ms cadence makes it a recurring exposure rather than a one-off.

Raised by Greptile (P2) on #390 and disclosed in that PR's description rather than fixed there, so the fix could be proved once against both trees rather than landing unproved in one.

## Fix shape

Two halves, worth doing together:

1. A `spawn_blocking` boundary around the snapshot call, so the subprocess stops occupying a worker.
2. A genuine timeout — async `Command` plus `tokio::time::timeout`.

## The one non-obvious decision

A timed-out sample must mean **`None` / "no opinion"**, never `Some(false)`. A wedged `ps` must not silently flip busy panes to idle. Per Greptile's carry-forward note, that reasoning belongs in a comment **at the call site**, not only in an issue or PR description — it is not recoverable from the code alone.

The `Option<Vec<_>>` return type added in #390 (`b84135d`) already makes this expressible: the monitor tick skips entirely on `None`, leaving `last_known` intact and emitting nothing.

## Context

- Introduced by #390 (PRD #386), which closed #386.
- Related: the poll cadence itself is earmarked for measurement in PRD #386 M5.

vfarcic:
Duplicate of #429, which predates this by two days (filed 2026-08-08) and already covers the same defect more completely — including the `None` vs `Some(false)` decision and Greptile's note that the reasoning belongs in a comment at the call site.

Closing in favour of #429. My mistake for not searching the tracker before filing.

PR #500, opened 2026-08-11
fix(shell-activity): skip the `ps` sample with no panes, and never let a wedged `ps` block a worker or read as Idle (#493, #429)

### LINKED ISSUE #429: shell-activity: the `ps` process-table sample blocks a Tokio worker and runs without a timeout
Follow-up tracking for a defect disclosed rather than fixed in #390 (PRD #386). Found by Greptile (P2) and confirmed in detail by @prageethw, who deliberately deferred the fix to keep the fork and upstream trees running identical code — the fix is proved on neither tree today, and Greptile accepted that reasoning. Filing here because the fork's tracking does not travel with the curated port, so without this issue upstream would carry the defect untracked.

## The defect

`crate::platform::proc::process_table()` (`src/platform/proc/unix.rs`) runs a synchronous `std::process::Command::output()`. It is called from `AgentPtyRegistry::shell_foreground_busy_snapshot`, which `run_shell_activity_monitor` calls directly from an async task every 500 ms.

Two separate problems:

1. **No timeout.** A `ps` that never returns hangs the call indefinitely.
2. **It occupies a Tokio worker thread**, which is the more serious half. A `ps` wedged in D-state (NFS, a stuck filesystem) stalls hook ingestion, client requests and daemon shutdown — not just this signal. The 500 ms cadence makes it a recurring exposure rather than a one-off.

## Fix shape

Two parts, per the discussion on #390:

- A `spawn_blocking` boundary around the snapshot call — the small, obvious first step.
- A genuine timeout: async `Command` plus `tokio::time::timeout`.

## The one non-obvious decision

**A timed-out sample must yield `None` ("no opinion"), never `Some(false)`.** `descendant_shell_activity` already draws this distinction deliberately, and callers are documented to treat `None` as "leave the pane's status alone." Collapsing a timeout to `Some(false)` would let a wedged `ps` silently flip every busy pane to `Idle` — reintroducing the exact stale-`Idle` bug PRD #386 exists to fix.

Per Greptile's parting note on #390, put that reasoning in a comment **at the call site**, not only in a PR description — it is not recoverable from the code alone.

## Related

- #390 — the PR that disclosed this
- #386 — the descendant-scan PRD
- Separate concern, same monitor: the poll's measured CPU cost (~9% of a core), discussed on #390.

### LINKED ISSUE #493: shell-activity: skip the `ps` sample entirely when no live panes exist
## Problem

The shell-activity monitor samples the whole process table on every tick **before** it knows whether there is anything to classify.

`AgentPtyRegistry::shell_foreground_busy_snapshot` (`src/agent_pty.rs:4116`) calls `process_table()` as its first statement, and only *then* takes the registry lock and filters for live panes:

```rust
let table = crate::platform::proc::process_table()?;   // <-- forks `ps -A` unconditionally
let inner = self.inner.lock().unwrap();
Some(inner.agents.values().filter(|agent| !agent.exited...
```

`run_shell_activity_monitor` (`src/daemon.rs`) calls it every 500ms with no guard. So a daemon with **zero panes** still forks `ps -A` twice per second, and each sample additionally issues a `getsid(2)` per process row — hundreds on a typical dev box.

## Measured

Verified while reviewing #390, not inferred: a daemon started with no panes at all, sampled at 20Hz for 5s, caught a `ps` child in 5 of 100 samples — consistent with the 2Hz poll given how short-lived each `ps -A` is.

This is not bounded by the idle-shutdown path either. The daemon idle-shuts-down only with no clients **and** no agents, so a TUI attached with no agent panes open polls forever at 2Hz for zero benefit.

## Fix

Return early when there are no live panes, before sampling. Something like: take the lock, collect the candidate panes, drop the lock, and only sample the table if that set is non-empty. Note the ordering matters — the current code deliberately samples *outside* the lock so the subprocess never runs while holding it, and any fix should preserve that.

## Context

- Introduced by #390 (PRD #386), which closed #386.
- Fits naturally with PRD #386's **M5**, which already earmarks the poll cadence and the Route A/Route B (`ps` vs native enumeration) cost question for measurement.
- Related: #492 (the same call being synchronous and untimed on a Tokio worker).

vfarcic:
Fixes #493. Fixes #429.

Two defects in the same 2 Hz poll — `AgentPtyRegistry::shell_foreground_busy_snapshot` and its caller `run_shell_activity_monitor` — both disclosed rather than fixed by #390, and fixed here as one unit because they touch the same call. (#492 was closed as a duplicate of #429.)

## #493 — the sample happened before we knew there was anything to classify

`process_table()` was the **first statement** of the snapshot, so the whole process table was sampled and only *then* filtered against live panes. A daemon with **zero panes** therefore forked `ps -A` twice a second — each sample also issuing a `getsid(2)` per process row — to classify nobody. Nothing bounded it: idle shutdown needs no clients **and** no agents, so a TUI attached with no panes open polled forever for no possible benefit.

The snapshot is now two pieces:

- `shell_activity_candidates()` — the **lock half**. Takes the registry lock, resolves each live pane's `pane_id`, `shell_pid` and per-agent argv shapes into an owned `ShellActivityCandidate`, and returns. This is where PRD #386's Open Question 2 selection now lives, because a pane's agent kind is only visible under the lock.
- `classify_shell_activity()` — the **pure half**. No lock, no I/O, just the descendant walk over a table the caller sampled.

The sample sits between them and is skipped entirely when the candidate set is empty.

**The ordering property is preserved.** The original sample-first ordering existed so no fork/exec ever ran while holding a lock the TUI-facing paths also take. That still holds — but now because the lock is *dropped* before sampling, not because the sample comes first. This is why `ShellActivityCandidate` carries owned data (including `shell_pid`) rather than borrowing from the registry; it is the mechanism, not incidental.

An empty registry answers `Some(vec![])`, **not** `None`. `None` means "the sample failed", and a caller seeing it would skip the tick and never clear its edge-detection map, so a reused pane id would inherit a stale reading.

## #429 — the sample blocked a Tokio worker and had no timeout

The daemon now samples through `process_table_async()` (tokio `Command`, `kill_on_drop(true)`) under a 2-second `tokio::time::timeout`. `process_table()` stays for synchronous callers and is documented as never-from-async.

**Why an async child rather than `spawn_blocking`.** The issue lists `spawn_blocking` as the small first step and the async `Command` + timeout as the genuine fix; the second subsumes the first, so only it is here. `spawn_blocking` merely relocates the stall to the blocking pool — and worse, `tokio::time::timeout` around a `spawn_blocking` handle **does not cancel the thread**, so a permanently-wedged `ps` at 2 Hz would leak one pool thread per tick up to the 512-thread cap. Awaiting an async child frees the worker *and* makes the timeout real: dropping the future kills the child.

One residual, documented rather than engineered around: `Command::spawn` still does the fork/exec on the calling thread, and glibc's `posix_spawn` uses `CLONE_VFORK`, so a `/bin/ps` binary on a wedged mount could block there. The realistic wedge in the issue is `ps` reading `/proc`, which is fully covered.

### The load-bearing decision: a timed-out sample yields `None`, never `Some(false)`

Both the failed-sample and timed-out arms mean **"no opinion"**. A wedged `ps` is a statement about `ps`, not about the panes — they are exactly as busy as they were a moment ago. Collapsing a timeout to `Some(false)` would synthesize a `ShellIdle` for **every live pane** and silently flip them all to `Idle`: the precise stale-`Idle` bug PRD #386 exists to fix, reintroduced with a new trigger and nothing in the log to find it by. `descendant_shell_activity` already draws this distinction deliberately and its callers are documented to treat `None` as "leave the pane's status alone".

Per the issue, **that reasoning is a comment at the call site** in `run_shell_activity_monitor`, not only here — it is not recoverable from the code. Both arms `continue`, leaving `last_known` untouched (clearing it would make every pane look new next tick and re-emit a spurious edge for each one); a timeout logs a `warn!`.

## Measured cost (the number PRD #386 M5 wants)

Release build, idle 16-core Linux box, ~620 processes, 40 samples of the real `process_table()`:

| | per sample | duty cycle at 2 Hz |
|---|---|---|
| Wall time | ~49 ms | **~10% of one core** |
| Our own CPU (`getsid` loop + parse) | ~1.4 ms | ~0.3% of one core |

This independently reproduces the ~9% figure discussed on #390. The gap between the rows is the finding: **~97% of the cost is the `ps` child and the wait on it**, not anything this process computes — so it was also ~49 ms of *worker-thread occupancy* every 500 ms, even with zero panes. That is the same number for both issues: it is what #493 was spending on nothing, and what #429 was blocking a worker for. It also frames M5's Route A vs. Route B question: native enumeration would target the 49 ms, not the 1.4 ms.

Recorded in `docs/develop/shell-activity-signal.md` rather than only here.

## Rule 12 — `PROTOCOL_VERSION` vs. `changelog.d/<issue>.breaking.md`

**Neither. `PROTOCOL_VERSION` stays at 7, and no `.breaking.md` fragment.**

- **No structural wire change.** `src/daemon_protocol.rs`, `src/event.rs` and `src/state.rs` are not in the diff at all. No frame, field, handler signature or serde surface is touched — the only `EventType`/`AgentEvent` lines in the diff are inside new test code seeding `AppState`.
- **No semantic break behind a stable wire.** The set and meaning of emitted `ShellBusy`/`ShellIdle` events is unchanged. The new timeout arm emits nothing, but before this change there was *no timeout at all* — a wedged `ps` hung the task forever and likewise emitted nothing. With zero panes nothing was emitted before or after either. There is no pairing of an old and a new build that observes a different event stream for the same machine state.
- So per `docs/develop/versioning.md` this is feature/bugfix → **patch**: two `.bugfix.md` fragments, one per issue, so each gets its own link in the release notes.

**On the cross-version manual test:** the meaningful baseline for *this* delta is `origin/main` (protocol 7), not the last release — v0.35.10 predates #390 and sits at protocol 6, so any mismatch against it is #390's already-documented break, not this PR's. Since the diff contains no protocol-facing production code (evidence above), the check here is static/handshake-level rather than a full interactive delegate + hooks click-through against a separately-built old daemon. Flagging that explicitly rather than silently calling it done.

## Tests (rule 4)

Two in-crate `#[tokio::test]`s on the real monitor, with the process-table sample injected via a new `run_shell_activity_monitor_with`. The deadline is applied *around* the injected sampler, so a test sampler that never completes exercises the real timeout path rather than a stub.

- `shell_activity_monitor_never_samples_the_process_table_with_no_live_panes` — empty registry, counting sampler, several ticks: the sampler must never be called.
- `shell_activity_monitor_leaves_statuses_alone_when_the_sample_times_out` — real `/bin/sh` pane seeded to `Working`, sampler is `std::future::pending()` (the wedged `ps`): status must stay `Working` and nothing may be broadcast.

Plus a unit test pinning the empty-registry `Some(vec![])` contract at the synchronous seam.

**Both monitor tests were verified to go red against their defect**, by temporarily reintroducing each:

- sampling unconditionally → `sampler called: left: 4, right: 0`
- collapsing the timeout to "every pane idle" → `status: left: Idle, right: Working`

L1/in-crate is the right tier: both are decisions inside the daemon's poll task with no rendered surface, and the injected sampler makes them deterministic rather than dependent on a wedged filesystem or an empty machine. The existing real-PTY and real-agent coverage (`status/shell-activity/004`–`007`) is unchanged and still exercises the production path end to end.

## Gates

`cargo fmt --check`, `cargo clippy --workspace --all-targets --features e2e -- -D warnings`, `cargo test-fast` (1678 passed), `cargo test-e2e`.

### e2e result

`cargo test-e2e`: **3738 passed, 6 failed** out of 3744, all six in the real-agent/real-binary tier. On individual rerun (rule 6), `shell_activity_006`, `devin_live_001` and `dispatch_013` **pass** — `006` failed its own declared prompt-adherence precondition (no `ToolStart`/`Bash` event ever arrived, so the model never made the call).

The remaining three — `focus_007`, `idle_worker_011`, `card_stats_005` — fail **identically on unmodified `origin/main`** (checked out and rerun to confirm: same three, same assertions, same timings). They are pre-existing on `main`, not introduced here.

Most relevant to this PR: **`shell_activity_005` (the rot canary) and `007` (no-false-positive-at-idle) both passed on the first full run.** Those are the two tests that prove the signal still fires — and still doesn't over-fire — for a real interactive agent, which is the coverage that matters for a change to how the sample is taken.

---

## Review rounds (Greptile, all three addressed)

Greptile found a chain of three related defects in the timeout, each one a narrowing of the previous. All are fixed with tests that were individually verified to go red against their defect. Worth reading as one story, because the guards look redundant until you see why they are not.

**Round 1 — `unix.rs:303`, timeouts accumulate wedged children** (`308d697`). The deadline as first written bounded the `ps` *child*: on expiry the future was dropped and the next tick started a fresh sample. But a D-state process ignores the `SIGKILL` that `kill_on_drop` sends until it comes unstuck, so the abandoned `ps` stays on the process table and the retry adds another every 2.5s (~24/minute) — turning the stall this PR targets into a pid leak. The deadline now bounds the **wait**: an overrunning future is retained and re-awaited, so **at most one `ps` child exists at a time**.

**Round 2 — `daemon.rs:1198`, a stale table across a pane-less gap** (`7871328`). Retention is what makes a *late* answer possible, and that needed a bound I had not added. A wedge outlasting every pane and then recovering would classify today's pids against a table from before they existed. The sample now carries its start instant and an answer older than `MAX_TABLE_AGE` (3s) is discarded.

**Round 3 — `daemon.rs:1217`, pid reuse inside the age limit** (`d1e06c8`). A bound is not an identity check: a pane can be replaced *within* the 3s window. The sample now also carries the candidate set as of its start, and on the resumed path only panes whose `(pane_id, shell_pid)` pair is unchanged are classified.

**Why the three do not conflict.** Retention refuses to abandon an *un-answered* sample (that is the only thing that can accumulate). The other two refuse to *believe* an answered one that has aged out or is being asked about a pane it never saw — and discarding an answered sample is free, because its child is already finished. Round 3's filter is scoped to the resumed path, where `at_start != candidates` is even possible; a sample started on the current tick has an identical set by construction, so the normal path pays nothing.

**A note on one test, since it nearly shipped green for the wrong reason.** The round-3 test first forced its pane replacement with `close_agent`, and passed *without* the fix. Instrumenting the branch showed `close_agent`'s SIGTERM grace window (`/bin/sh` ignores SIGTERM) blocking the current-thread runtime long enough that the sample landed 3.4s old — so round 2's freshness guard swallowed it and the assertion held for an unrelated reason. The committed version adds the second pane without closing the first and *also* asserts the unchanged pane reaches `Working`, which proves the late answer was accepted rather than aged out. That rationale is in the test's doc comment.

Final gates on `d1e06c8`: `cargo fmt --check`, `cargo clippy --workspace --all-targets --features e2e -- -D warnings`, `cargo test-fast` **1678 passed**, `shell_activity` e2e 13/14 with `006` failing only its own declared prompt-adherence precondition and passing on rerun against this commit.


vfarcic:
Thanks — the P1 is valid and is fixed in 308d697.

**Timeouts accumulate wedged children** (`src/platform/proc/unix.rs:303`) — correct, and it was a gap this PR introduced rather than one it inherited. The deadline as written bounded the *child*, not just the wait: on expiry the future was dropped and the next tick started a fresh sample. A process in uninterruptible sleep does not act on the `SIGKILL` that `kill_on_drop` sends until it leaves D-state, so the abandoned `ps` stays on the process table and the retry adds another every 2.5s — roughly 24 a minute, turning the stalled signal this PR set out to fix into a pid leak.

The fix keeps the deadline as a bound on the **wait** and not on the child: an overrunning future is retained and re-awaited on the following tick, so **at most one `ps` child exists at a time** however long the filesystem stays wedged. A sample that answers "failed" is finished and still dropped, so the next tick starts fresh. The overrun now logs once per wedged sample rather than once per tick.

One consequence worth naming explicitly: a late answer is classified against the tick that *receives* it. Candidates are re-resolved every tick so they are current; only the table can be as old as the sample. That is the right trade — a late answer about a live pane beats no answer — and it is documented at `inflight`'s declaration.

The existing timeout test now also counts samples *started* and requires exactly one. Verified red against the abandon-and-retry shape (3 samples instead of 1), so the invariant is pinned rather than asserted. `docs/develop/shell-activity-signal.md` and the changelog fragment both carry the wait-not-the-child distinction, since the natural-looking version is the wrong one.

Re-ran `cargo test-fast` (1676 passed) and the full `shell_activity` e2e set (12 passed, including all three real-agent tests `005`/`006`/`007`).

vfarcic:
Valid again, and a good catch on the interaction — round 1's retention is precisely what makes a *late* answer possible, so it needed a bound of its own that I did not add. Fixed in 7871328.

**Stale table across a pane-less gap** (`src/daemon.rs:1177`) — confirmed. A wedge that outlasts every pane and then recovers would classify today's pids against a table taken before they existed. Most of the time that is harmless, since `descendant_shell_activity` returns `None` for a pid the table lacks and the pane is simply skipped; the harmful case is exactly the one you named, pid reuse — and it is worse than it first looks, because `last_known` has no entry for a newly-opened pane, so `changed` is true and the wrong reading emits an event on its very first tick rather than being swallowed as a non-transition.

The in-flight sample now carries its start instant, and an answer older than `MAX_TABLE_AGE` (3s) is discarded instead of classified. The two rules do not conflict, which is the part worth stating: discarding an **answered** sample is free and cannot accumulate, because that child is already finished — only abandoning an **un-answered** one leaves an undead `ps`. So round 1 keeps waiting, and round 2 declines to believe a late answer.

I kept retention unconditional, including on a no-candidate tick, rather than dropping the future there. Dropping looks tidier and would also close this hole, but it reopens round 1's accumulation path through pane churn (close to zero, reopen, second `ps` joins the first undead one), and #493 already guarantees a paneless daemon starts no sample at all. The staleness bound closes the hole without that trade. The residual is one retained child — the same one we would be waiting on anyway.

3s is ~60x a healthy sample (~49ms) and well clear of a heavily loaded one, so it never trips in normal operation.

New test hands the monitor a sampler that answers past the bound with a table claiming the pane *is* busy; the session must stay `Idle`. Verified red without the check (`Working` instead of `Idle`). `cargo test-fast` 1677 passed; `shell_activity` e2e 13 passed, including real-agent `005`/`006`/`007`.

vfarcic:
Valid — this is the residual my round-2 bound left, and you are right that a bound is not an identity check. Fixed in d1e06c8.

**Stale table within the age limit under pid reuse** (`src/daemon.rs:1217`) — the in-flight sample now carries the candidate set as it was when the sample **started**, and on the resumed path only panes whose `(pane_id, shell_pid)` pair is unchanged get classified. A same-slot respawn keeps the pane id but takes a new pid; a fresh pane brings a new pane id — either way the pair differs, and the pane waits for the next sample. That is the honest answer rather than a conservative one: the table predates the pane and genuinely cannot describe it.

Scoped to `was_resumed`. A sample started on the current tick has `at_start == candidates` by construction, so the filter would be a no-op and the common path allocates nothing.

**One thing worth flagging about the test, because it caught me.** The first version forced the replacement by closing pane A — and it passed *without* the fix. Instrumenting the branch showed why: `close_agent`'s SIGTERM grace window (`/bin/sh` ignores SIGTERM) blocked the current-thread runtime long enough that the sample landed 3.4s old, so round 2's **freshness** guard swallowed the answer and pane B stayed `Idle` for a reason having nothing to do with identity matching. Green for the wrong reason.

The committed test therefore adds pane B *without* closing A, and asserts that A reaches `Working` too. That assertion is the point: A being promoted proves the late answer was **accepted** rather than aged out, so B staying `Idle` can only be the identity filter. Verified red without the guard (B goes `Working`). The rationale is in the test's doc comment so the next person does not re-introduce the weaker version.

Gates: `cargo fmt --check`, `cargo clippy --workspace --all-targets --features e2e -- -D warnings`, `cargo test-fast` (1678 passed). `shell_activity` e2e: 13/14, with `006` failing its own declared prompt-adherence precondition ("roughly 2 runs in 3" per the test's own note) and **passing on rerun** against this exact commit — `005` and `007`, the rot canary and the no-false-positive check, pass consistently.

prageethw (approved):
Reviewed at `d1e06c8` — code read only, no local build or test run (CI is green here and I did not re-run it locally). I read this from two angles, ordinary review and a security/robustness pass; both landed on approve.

Both issues are genuinely fixed rather than relocated, and the parts that were hardest to get right are right:

- **The wait is bounded on the whole operation, not on one read.** `timeout(SAMPLE_TIMEOUT, pending.as_mut())` wraps the entire `output()` future — spawn, both pipe drains, and the child exit — so there is no "bounded read of an unbounded stream" hole. `output()` drains stdout and stderr concurrently with the wait, so the discarded-stderr pipe cannot deadlock the drain either; the comment saying so is accurate.
- **Failure degrades to an explicit absence, not to a plausible value.** `Ok(None)` and `Err(_elapsed)` both leave `last_known` untouched and broadcast nothing, `Some(false)` is never synthesized from a failure anywhere on the path, and `classify_shell_activity` drops a `None` rather than folding it to `false`. That is the load-bearing #429 decision and it holds in the code, not just in the comment.
- **Retaining the in-flight future rather than dropping it is the right call for a D-state child**, and the reasoning in the comment is correct.
- **Greptile's three inline P1s are genuinely closed** by `308d697` / `7871328` / `d1e06c8`, not papered over — I re-derived all three against the current code.

One correction to my own initial read, since I had it wrong: the changed-file list shows no test files, but the PR **does** add five tests — four `#[tokio::test]`s against the real monitor through the new `run_shell_activity_monitor_with` injection seam, plus a unit test pinning the `Some(vec![])` contract. `tests/CATALOG.md` covers `tests/`, not in-crate `#[cfg(test)]` modules, so nothing is missed there. Test coverage is not a finding against this PR; F6 below is the one real gap.

Nothing below is blocking. **F1 is the one I would file as a follow-up issue** — it is the only finding that can silence the signal permanently.

---

### F1 — `MAX_TABLE_AGE` and `SAMPLE_TIMEOUT` leave a band where the signal dies permanently

`src/daemon.rs:1172`, interacting with `:1121` and the discard arm at `:1230-1238`.

With `POLL_INTERVAL` 500 ms, `SAMPLE_TIMEOUT` 2 s and `MAX_TABLE_AGE` 3 s, a sample taking `T` resolves as: `T ≤ 2` accepted on its own tick, `2 < T ≤ 3` accepted on the resumed tick, and **`T > 3` discarded, always, with no recovery path**. The discard drops `inflight`, the next tick starts a fresh sample that takes the same `T`, and so on. So the retention machinery only ever yields a usable answer inside a one-second band, and a `ps` consistently slower than 3 s means the signal never produces another reading.

Failure scenario: a box with a very large process table, or a slow-but-not-wedged network filesystem, where `ps -A -w -w -o pid=,ppid=,tty=,args=` consistently takes ~3.5 s. t=0.5 sample starts → t=2.5 timeout, retained, one `warn!` → t=3.0 resumed tick awaits → t=4.0 answers, `age = 3.5 s > 3 s` → discarded → t=4.5 fresh sample → repeat forever. Every pane running a long command gets knocked to `Idle` by the `Stop` hook and is never re-promoted — the same end state the mechanism exists to prevent, reached by the monitor never speaking rather than by it speaking wrongly.

Second-order: the discard `warn!` at `:1231` is **not** rate-limited, unlike the overrun `warn!` which is gated on `inflight_reported` at `:1307`. In that scenario it is one line every ~3.5 s, forever.

Suggested shape: derive `MAX_TABLE_AGE` from `SAMPLE_TIMEOUT` (e.g. `2 ×`) so the two cannot drift into a dead band, and gate the discard log the same way as the overrun one.

### F2 — `last_known.retain` conflates "pane went away" with "pane was unobservable this tick"

`src/daemon.rs:1331`, with `classify_shell_activity`'s `filter_map` in `src/agent_pty.rs:4223`.

`seen` is built from the *classified snapshot*, not the *candidate set*. A pane that is live but unobservable in one sample is therefore evicted from `last_known`, and on the next good sample it looks brand new and re-emits an edge. That is exactly the spurious-edge failure the timeout arm's own comment at `:1294-1298` refuses to cause — the guard exists at whole-sample granularity and is absent at per-pane granularity.

Failure scenario: pane P is busy, `last_known[P] == true`, card reads `Working`. On one tick `getsid_or_negative` returns `-1` because the pid was reaped between `ps` writing the row and the `getsid` call, so `descendant_shell_activity` returns `None` and P drops out of the snapshot. `retain` evicts it. Next tick P classifies busy again, `insert` returns `None`, `changed == true`, and a fresh `ShellBusy` is broadcast for a pane whose state never changed.

Impact today is mild — `apply_event`'s `ShellBusy` arm only promotes `Idle`/`Unknown`, so it is usually a no-op — and it is pre-existing. Flagging it because #500 is where per-pane versus per-sample granularity is being reasoned about. Fix: `retain` against the candidate set keyed on `(pane_id, shell_pid)`; keying on `pane_id` alone would regress the replaced-pane case.

### F3 — The capture bounds time and freshness, but not size

`src/platform/proc/unix.rs:223` (`PS_TABLE_ARGS`), `:296-305` (`process_table_async`), `src/platform/proc/scan.rs:238-255` (`parse_ps_table`).

Nothing caps the table or an individual row. `-w -w` deliberately disables `ps` truncation (correct for the matching logic), `output()` reads stdout to EOF into a `Vec<u8>`, and `parse_ps_table` then allocates one owned `String` per row. So the input size is set by whatever processes exist on the machine, and a process's argv is controlled by whoever spawned it.

Failure scenario: any local unprivileged user spawns a few processes with multi-hundred-KB argv (Linux allows up to `MAX_ARG_STRLEN`, 128 KB per string). With one pane open the daemon re-reads and re-allocates that payload every 500 ms — the raw `Vec<u8>`, a `String` copy if the bytes are not valid UTF-8, and one owned `String` per row — driven entirely by processes the deck has nothing to do with, on a repeating timer with no backoff. `ShellToolShape::matches` then runs a substring search over each oversized row.

**This is pre-existing and out of the declared scope of #493/#429 — I am not asking for it here.** Noting it because it is the one robustness item the change leaves fully open, and because #500 actually makes the follow-up *easier* by creating `PS_TABLE_ARGS` as the single shared definition both samplers use. If it is picked up: bound the read (spawn with a piped stdout, read at most N MB, discard the remainder) rather than truncating individual rows — `matches` is a substring test and the interesting needles can sit arbitrarily far into a line, so per-row truncation would convert a resource issue into a silent false negative, which is the failure mode `docs/develop/shell-activity-signal.md` explicitly says to fear.

Also on the async path specifically: tokio's `Command::output()` unconditionally forces both stdout and stderr to pipes and cannot be talked out of it, so the async path buffers stderr in memory where the sync path discards it at the fd level with `.stderr(Stdio::null())`. The PR's comment notes the pipe-forcing, so this is understood rather than missed; marginal in practice since `ps` is not chatty on stderr.

### F4 — A retained sample is never polled *or* dropped on a paneless tick, so a `ps` that finishes there is never reaped

`src/daemon.rs:1149`, the `candidates.is_empty()` arm.

By explicit design that branch neither polls nor drops `inflight`. That is right for a child still wedged in D-state, but not for one that *finishes* while the daemon is paneless: nothing awaits it and nothing drops it, so it is never reaped. This is a real gap in tokio's model rather than a theoretical one — `Reaper`'s `Drop` is what pushes an un-exited child onto the orphan queue, and the SIGCHLD-driven reaper only ever reaps queue members, so a `Child` owned by a future that is neither polled nor dropped is in neither place and stays `<defunct>`.

Failure scenario: one pane open, a sample overruns and is retained; the user closes the last pane; every subsequent tick takes the empty arm and never touches `inflight`; the `ps` unwedges and exits and becomes a zombie for as long as the daemon stays paneless — which for an attached-but-empty dashboard is exactly the long-lived state #493 is about.

Bounded at exactly one zombie, so minor. The comment's "the same child we would be waiting on anyway" is true for a wedged child but understates this case. A zero-duration poll on that branch (discarding the answer, since with no panes there is nothing to classify) would close it while preserving #493's invariant — the invariant is "never *fork* `ps` with no panes", and finishing an already-running child forks nothing.

### F5 — The deadline restarts on each resumed tick

`src/daemon.rs:1225`.

A sample already past `MAX_TABLE_AGE` is still awaited a further 2 s before being thrown away, delaying the first reading after a paneless gap by ~2.4 s.

### F6 — The failed-sample arm has no test with panes present

`src/daemon.rs:1305` (`Ok(None)`).

Given the injection seam this PR already added, this is roughly ten lines. Without it, a later refactor folding this arm into the empty-snapshot path would regress silently — and that regression is precisely a failure reading as `Idle`.

### F7 — Minor

- `src/daemon.rs` — `inflight_reported` is reset only when a *new* sample starts, so a permanently wedged `ps` logs exactly once and then goes quiet forever. An NFS mount that hangs at 09:00 produces one line at 09:00:02, and someone investigating a frozen badge at 15:00 finds nothing recent. A decaying re-log (60 s, or a doubling backoff) with the elapsed wedge duration would fix it. Relevant because the doc frames the failure mode to watch for as silence rather than noise.
- `src/platform/proc/unix.rs` — `ps` is resolved through `PATH` on both sampler paths, twice a second, for the daemon's life. The argument vector is a compile-time constant with no untrusted input, so this is hardening rather than a live issue and it is pre-existing — but since the invocation is fixed anyway, pinning `/bin/ps` then `/usr/bin/ps` before the bare name removes the environment from the equation.
- `src/platform/proc/windows.rs:613` — parity is semantically fine (`process_table_async` returns `None` without awaiting, so the timeout is a correct no-op and the caller needs no `cfg` branch), but Windows now pays the registry lock plus a candidate clone every 500 ms for a sampler that can never answer.
- The four new monitor tests sleep on real wall-clock time (~16.5 s total, ~5.6 s wall-clock under nextest's concurrency) rather than using `tokio::time::pause()`. There is a defensible reason — three spawn a real `/bin/sh` pane and exercise the real deadline, which `pause()` would stub out — so this is an observation rather than a finding.

---

### Checked and clean

- **Reaping on every path except F4**: answered sample → `wait_with_output`; too-old → `continue` drops the local `pending` (already `take()`n) → `kill_on_drop` → orphan queue; failed sample → same; daemon shutdown → task abort drops the future → same.
- **The no-panes skip does not skip cleanup.** It falls through to `last_known.retain` against an empty `seen`, which clears the map — *more* cleanup than the failure paths perform. `Some(vec![])` vs `None` is preserved at the synchronous seam.
- **No lock is held across the fork/exec** — `shell_activity_candidates` returns fully owned values, so the registry mutex is released before `sample()` runs.
- **The stale-table guards are correctly conservative.** `MAX_TABLE_AGE` plus the `(pane_id, shell_pid)` identity filter means a late answer applies only to panes whose identity is provably unchanged. Panes excluded by that filter drop out of `seen` and hence `last_known`, but any pane that can be excluded either post-dates the sample or has changed identity, so no spurious edge results.
- **Duration arithmetic** — all three constants are non-zero compile-time values, no timeout is computed, `started` is a monotonic `tokio::time::Instant` with only `elapsed()` taken from it.
- **No new `unsafe`**; the only `unsafe` in the touched files is the pre-existing `libc::getsid`, correctly documented as a pure kernel lookup with no I/O wait, which is what justifies leaving it on the caller's thread.
- **No untrusted data reaches the log** — both `warn!` calls carry only counts and durations, so attacker-controlled process names cannot inject into it.
- **Argv-veto spoofing considered and dismissed**: `ShellToolShape::matches` is a substring test over controllable argv, but the shape is only a veto, and to influence a reading the process must already be a transitive descendant of that pane's PTY child in a different POSIX session — i.e. already inside the agent's own tree, which could produce a `Working` badge legitimately.

Issue #322, opened 2026-08-01
E2E harness leaves stale temp dirs; a full gate run can false-red on tmpfs exhaustion


vfarcic:
## Symptom

A full `cargo test-e2e` run can fail en masse with errors that have nothing to do with the code under test. Observed while gating PR #310:

- 103 of 2742 tests failed; **88 failure bodies contained `Disk quota exceeded (os error 122)`** verbatim.
- The remainder were downstream of the same exhaustion — `git init` failing inside a temp dir, a Pi extension failing to write, daemons never booting.
- After reclaiming disk space and changing *nothing else*, the identical suite passed **2742/2742**.

The diagnosis cost a full round trip. A green branch produced a 103-test red wall whose most prominent failures (`the REAL interactive OpenCode worker never became visibly input-ready`, `Pi pane's card never showed Thinking`) read like genuine product regressions.

## Root cause

`/tmp` here is a **14 GB tmpfs** — RAM-backed — and the e2e harness does not reliably clean up after itself. At the time of failure:

| | |
|---|---|
| `.tmp*` dirs under `/tmp` | **204** |
| Total size | **2.9 GB** |
| Older than 6 h | **140** |
| Modified in the last hour | 2 |

Plus multi-hundred-MB leftovers from earlier runs (`dot-agent-deck-main-prd241` 1.5 G, `dot-agent-deck-win-check` 830 M, `dad-v0350-target` 465 M). Reclaiming the stale set freed **6.1 GB** and took `/tmp` from 80% to 42%.

Because the filesystem is RAM-backed, every abandoned temp dir is permanently held memory until a human notices. Failed runs are the worst offenders — a run that dies mid-test never reaches its cleanup path, so the failure mode is self-amplifying: the more the suite fails, the more likely it is to fail again.

## Proposed fix

Any of these would help; the first two are the substantive ones.

1. **Clean up temp dirs even on failure.** Whatever creates `.tmp*` should remove it on the panic/abort path, not only on success.
2. **Pre-flight check.** Before the suite runs, assert a minimum free space on the temp filesystem and fail fast with an explicit message (`e2e needs ~N GB free in /tmp; found X`) instead of producing dozens of misleading assertion failures.
3. **A prune helper** (`cargo xtask clean-e2e-tmp` or similar) so reclaiming is one obvious command rather than hand-rolled `find`.
4. Consider whether the harness should use a dedicated directory it fully owns, rather than sharing `/tmp` with other tooling and parallel runs.

[...]

PR #504, opened 2026-08-11
ci+docs: fail fast on a missing RELEASE_TOKEN, and record the required-checks + auto-merge switch

vfarcic:
Three things that have to land together: the workflow hardening, the rule 8 / governance.md text that described the world before it, and the protection script that reconstructs the ruleset both of them describe.

## 1. Fail fast on a missing `RELEASE_TOKEN` (`bc11bc6`, hardened in `5817bd7`)

`release.yml` and `docs-publish.yml` both push straight to `main` and checked out with `token: ${{ secrets.RELEASE_TOKEN || github.token }}`. That fallback was correct when written, and its own comment names the condition it depended on: it keeps the workflow *"behaving exactly as it does today while the secret is unset, so this change is safe to merge before `main` is protected."* **Both halves flipped on 2026-08-08** — `main` became ruleset-protected and the secret was set the same day.

All it can do now is degrade an **unset or empty** secret to `github-actions[bot]`, which is deliberately **not** a bypass actor (`GITHUB_TOKEN` cannot be one on a user-owned repo). The run then fails *late*:

| Workflow | Fails at | State left behind |
|---|---|---|
| `release.yml` | "Commit changelog updates" | Tag cut, changelog written — the v0.35.6 `GH006` mode |
| `docs-publish.yml` | "Commit chart changes" | Image in GHCR, `main`'s chart pointer stale — how v0.35.9 failed |

Each job now verifies the secret in its **first** step, with `set -euo pipefail` like every neighbouring block. `docs-publish.yml`'s guard also names the secret-forwarding requirement on `release.yml`'s `docs` job, since a reusable workflow invoked via `uses:` gets no secrets unless the caller passes them — the exact v0.35.9 cause, fixed in `428acef`.

**Scope of the guard, stated precisely** (corrected after review — the first version of this PR overclaimed it): it tests `-z`, so it catches *unset or empty*. A token that is present but **expired, revoked or under-scoped passes it and still dies at `git push origin main`** in the same half-done state. The old fallback did not cover that case either — a GitHub `||` returns the left operand whenever it is non-empty, so an expired PAT was passed straight through to `actions/checkout`; the two behaviours differ only for an unset or empty secret. Making the guard *probe* the token was considered and rejected: an extra API call and a new failure mode on the release path, to cover a case that has never occurred here. The `docs-publish.yml` canary is what actually validates a token, so it should be re-run after every PAT rotation — now recorded in both docs.

## 2. Record the settings change (`ca00958`, corrected in `329a14b`)

The repository settings changed on 2026-08-11, independently of this PR: the ruleset gained `required_status_checks` with `build`, `build-macos`, `build-windows`, `security`, and `allow_auto_merge` became `true`. Rule 8 and `docs/develop/governance.md` still described the old world — including documenting the `|| github.token` fallback as a hazard to live with, which is why this text belongs on *this* branch rather than its own.

Stale claims removed: rule 8's "**`main` has no required status checks at all**" and its "treat re-adding them as unverified rather than safe"; the ruleset described as three rules; "the author merges after approval" as the only route; governance.md's "Auto-merge is currently disabled on this repository" and its rollout note that step 6 "is still open".

Decisions now recorded, which is the part worth reviewing:

- **Greptile stays deliberately unrequired.** Rule 8 already noted it is *technically* safe to require. It is excluded anyway because an approval is the judgment call and waiving a finding is a legitimate approval — requiring the reviewer would turn advice into a veto. Only objective jobs are required.
- **Arming auto-merge is not a bypass.** It waits for the same approval a manual merge needs, so it is safe to arm before review lands — **but only while the ruleset is in force**, since `allow_auto_merge` is a repository setting the ruleset does not own. Emergency override therefore disarms before it lifts (§5 below).
- **`/prd-done` and `/prd-full` may now arm auto-merge and hand off**, where before they had to stop at "green, reviewed, approved". Merging directly is still forbidden.
- **The write-vs-admin asymmetry is accepted, not resolved.** The four checks bind @prageethw and stay advisory for the admin owner. That is tolerable only because they are objective — a red `build-windows` is not a matter of taste.

## 3. Make `apply` restore the real ruleset (`f7bb367`)

@prageethw's blocker, and it was outside the original three files, which was the point: this PR declared the rollout complete while the only tool that reconstructs the ruleset still emitted its pre-rollout shape. `payload()` emitted three rules, `required_status_checks` appeared nowhere in the script, and `cmd_apply` ships the payload with `gh api --method PUT` — a full replacement. So `apply` **deleted** the four required checks and reported success, which is exactly what governance.md's Emergency override told you to run, captioned *"put it back"*.

`REQUIRED_CHECKS` and `STRICT_REQUIRED_CHECKS` are overridable like every other tunable; `REQUIRED_CHECKS=` omits the rule for a fork whose CI produces different job names. `status` now prints the contexts rather than just the rule type, so it can serve as the pre-tag check that the gate and the admin bypass are both intact.

## 4. The skipped-checks question, settled

`required_status_checks` plus `ci.yml`'s path-filtered `changes` gate looks like it could make docs-only PRs permanently unmergeable — the "a required check that never reported blocks the PR forever" trap this PR records for #416, but reachable by anyone. **It does not, on two independent grounds**, and the docs previously said nothing either way:

- GitHub treats a job skipped by a conditional as **passing**: *"A job that is skipped will report its status as 'Success'. It will not prevent a pull request from merging, even if it is a required check."* The #416 trap is a check that never reported **at all** — no check run exists to evaluate — which is a different mechanism.
- The `changes` gate is scoped **by author** to `renovate[bot]`, so a human PR gets the full matrix whatever its paths. Measured: #504 (workflows + docs), #499 (one skill file), #469 and #464 (one PRD file each) all ran `build`, `build-macos`, `build-windows` and `security`.

Both are now in governance.md step 6, with the note that the author scope is defence in depth rather than the load-bearing half — worth knowing before the `changes` gate is ever widened beyond Renovate.

## 5. Review items also addressed

- **F3 settled, not merely restated.** #510 was authored *and* merged by `app/renovate` with `reviewDecision: REVIEW_REQUIRED` and all four required checks `SKIPPED`. Renovate's bypass covers checks as well as approval, so `main` has a zero-human-approval, zero-green-check lane — stated as a security property, accepted by design with `ci.yml`'s push-to-`main` trigger as the net, closable with `RENOVATE_BYPASS=false`. The claim that #503 was the live test is dropped: it touches `Cargo.lock`, so it gets the full matrix and structurally cannot exercise the skip path.
- **F4.** The bypass actors *did* survive the full `PUT` — `gh api repos/vfarcic/dot-agent-deck/rulesets/20587589` still lists `RepositoryRole` id 5 `always` and `Integration` id 2740 `pull_request` (admin-only, which is why review could not check it). But no direct push to `main` has happened since the checks returned at 2026-08-11T15:00:51Z, so "closed on both sides" is now "in place, not exercised", with a recommendation to run `status` before the next tag.
- **F5.** Emergency override now documents the way back for both switches, in order: disarm armed PRs and disable `allow_auto_merge`, *then* delete the ruleset; restore the ruleset, `status` to confirm, *then* re-enable arming.
- **F6 deliberately deferred to #514.** `secrets: inherit` does forward HOMEBREW_TAP_TOKEN and SCOOP_BUCKET_TOKEN into a job that reads neither, and the new comments no longer endorse it. Not narrowed inline because the callee is also `workflow_dispatch`-able and that path has to be canaried first — guessing on the release path is how both prior failures happened.
- **Minor:** `set -euo pipefail` in both guard blocks; `f1204ec`'s committer no longer described as its pusher; the script's stale "there is no CODEOWNERS file" comment replaced with the pathless router that is actually there.
- **Forks.** New governance.md section: removing the fallback breaks releases on any fork with no `RELEASE_TOKEN` (`prageethw/dot-agent-deck` shipped five releases in six days), with the one-PAT fix and the note that a fork's secret list is not readable from upstream.
- **#515** filed for the pre-existing `sudo wget` install of `yq` with no checksum verification in a job holding the admin PAT.

## Verification

YAML re-parsed for both workflows; guard confirmed as step 1 with the checkout at step 2 and no `${{ }}` in either guard body. `payload()` diffed against the live ruleset: the `required_status_checks` rule is byte-identical, and the full payload matches except the server-side `required_reviewers: []` read default on the pre-existing `pull_request` rule. Read-only `status` run; **`apply` and `delete` were deliberately never run** against the repository. shellcheck clean at `--severity=style`. `cargo fmt --check`, `cargo clippy --workspace --all-targets --features e2e -- -D warnings` and `cargo xtask linkage-check` all green.

`RELEASE_TOKEN` is present (rotated 2026-08-08), so the workflow commit changes nothing about today's behaviour — it changes what happens the day the secret is missing.


prageethw (changes requested):
Reviewed at `ca00958` — code read only, no build, no test, and no writing `gh` call (nothing was mutated, no release or publish triggered, no secret value read).

**The workflow half is correct and would pass on its own.** I am requesting changes on one thing, and it is not in the diff — which is exactly why it is worth raising.

The guard itself is right, and right in the ways that are easy to get wrong:

- **Safe form.** `env:`-bound with `[ -z "${RELEASE_TOKEN}" ]`, not `${{ secrets.RELEASE_TOKEN }}` interpolated into the `run:` body — so the value never enters the rendered script text. No `${{ }}` appears in any new `run:` block at all.
- **First step, before `actions/checkout`**, and in `docs-publish.yml` ahead of the GHCR push — so no repo content is resident while the secret is in scope, and the v0.35.9 half-state is unreachable via a missing secret.
- **Fails closed and early** in both workflows, which do not diverge.
- **No privilege widening** — step-scoped `env:`, `permissions:` untouched.
- **No fork-PR hazard**: neither workflow has a `pull_request` trigger, so the guard cannot turn external contributions red for a reason the contributor cannot fix.
- Coverage is complete — every job that needs the token has one.

I also checked the documentation half against the live repo rather than reading it for plausibility, and **every checkable claim holds**: the ruleset's four rules, the four contexts, `strict_required_status_checks_policy: false`, the `pull_request` parameters, `allow_auto_merge: true`, #416 still stuck as a fork PR, the four job names in `ci.yml` (real job ids with no `name:` override, so the check names will match), and exactly two direct pushes to `main`. Auto-merge arming is correctly characterised as not a bypass.

---

## The blocker — `scripts/apply-branch-protection.sh` is now a silent revert device

The script is **outside the four files this PR touches**, which is precisely the problem: the PR declares the required-checks rollout complete and re-frames `governance.md` around it, while leaving the only tool that reconstructs the ruleset at its pre-rollout shape.

I verified each link rather than inferring it:

- The **live** `main-protected` ruleset has four rules, including `required_status_checks` with contexts `build`, `build-macos`, `build-windows`, `security`.
- `payload()` emits **three** — `deletion`, `non_fast_forward`, `pull_request`. The string `required_status_checks` appears **nowhere in the script**.
- `cmd_apply` ships it with `gh api --method PUT`, a full replacement.

So running `apply` deletes the four checks. And `governance.md:137-140` documents the Emergency override as exactly that, captioned *"put it back"*:

```bash
scripts/apply-branch-protection.sh delete   # remove the gate
scripts/apply-branch-protection.sh apply    # put it back
```

It does not put it back. It reinstates a weaker ruleset with the checks silently gone, and nothing reports a loss — the command succeeds. `governance.md:69` compounds it by re-framing these steps as the onboarding procedure to re-run.

This is the exact partial-payload hazard **your own new text at `governance.md:97` warns about**, one rule over. In rule-16 terms: the new text makes a reader consume "a full `PUT` of every rule", and the only named supplier does not supply it.

Cheap to fix — add the rule to `payload()` — and I would rather it land with this PR than as a follow-up, because the window where the docs say "put it back" and the script quietly does not is the window where someone runs it in an emergency.

---

## Other findings — none blocking

**F2 — the guard tests `-z` only, but three places claim it covers an expired secret.** `governance.md:61`, the new `CLAUDE.md` rule 8 paragraph, and the PR body all say so. A GitHub `||` returns the left operand when non-empty, so the old fallback never degraded an *expired* PAT either — the claim is wrong about both the old behaviour and the new coverage. A present-but-under-scoped or expired token still passes the guard and dies later at `git push origin main`, with the tag already cut and the changelog already written. Either narrow the claim to "unset or empty", or make the guard probe the token.

**F3 — the new `UNVERIFIED` bullet contradicts the section it links to, and is now falsified.** `governance.md:100` calls the Renovate-bypasses-required-checks question open; unchanged `governance.md:115` asserts it as fact. **PR #510 settles it**: authored *and* merged by `app/renovate` at 2026-08-11T16:21:23Z — 81 minutes after the checks went up — with `reviewDecision: REVIEW_REQUIRED`. The head commit here predates that by ~72 minutes, so the bullet was accurate when written and is merely stale.

Worth stating plainly because it is security-relevant rather than editorial: that is a confirmed **zero-human-approval merge lane into `main`**. Accepted by design, with `ci.yml`'s push-to-`main` trigger as the net — but the docs should now say so as fact.

The bullet also names #503 as its live test, which cannot work: #503 touches `Cargo.lock`, so it gets the full matrix. The devbox-only and flake-only automerge groups skip all four required jobs via `ci.yml`'s `changes` gate, so they hit the same "never reported" trap recorded for #416 — a second mechanism the note does not mention and #503 structurally cannot exercise.

**F4 — the guard proves presence, not bypass capability.** No direct push to `main` has happened since `required_status_checks` returned (checks added 2026-08-11T15:00:51Z; last release v0.35.10 at 2026-08-10T17:16Z). Since the v0.35.6 failure was specifically a required-status-checks rejection on a direct push, the "closed on both sides" claim is untested. I could not confirm the Admin bypass actor survived the 2026-08-11 full `PUT` — `bypass_actors` is admin-only and my access is `push` — so this is **unverified, not broken**. Running `scripts/apply-branch-protection.sh status` before the next tag would settle it, once the script is fixed per the blocker.

**F5 — no way back documented for either switch.** `allow_auto_merge` is a *repo setting* that survives a ruleset `delete`, so armed PRs land unreviewed the moment Emergency override runs — the outcome the page's own argument for arming early says cannot happen.

**F6 (pre-existing) — `secrets: inherit`** forwards `HOMEBREW_TAP_TOKEN` and `SCOOP_BUCKET_TOKEN`, cross-repo write credentials, into a docs-publishing job that needs neither. Not changed here, but two new comment blocks now instruct readers to preserve it, which entrenches it. A narrow explicit map would be better.

**Minor:**
- `CLAUDE.md:42` says `f1204ec` was "pushed to `main` by `github-actions[bot]` … same identity", conflating the *committer* (set via `git config user.name`) with the *pusher* (the PAT). As written it implies `github-actions[bot]` can push to protected `main` — the opposite of the rule's own point.
- The guard blocks omit `set -euo pipefail`, unlike every neighbouring `run:` in the same files. Harmless today.
- (pre-existing) `docs-publish.yml:117` installs `yq` by `sudo wget` with no checksum verification, then executes it in a job holding the admin PAT.
- (pre-existing) `scripts/apply-branch-protection.sh:36` claims "there is no CODEOWNERS file"; `.github/CODEOWNERS` exists.

vfarcic:
Thanks @prageethw — the blocker was real and the framing was right: the PR declared the rollout complete while leaving the only tool that reconstructs the ruleset at its pre-rollout shape. All of it is addressed in `f7bb367`, `5817bd7` and `329a14b`. Point by point.

## BLOCKER — `scripts/apply-branch-protection.sh` reinstated a weaker ruleset

**Fixed** in `f7bb367`. Confirmed exactly as you described: three rules emitted, `required_status_checks` nowhere in the file, shipped with `gh api --method PUT`.

`payload()` now emits the rule, with the live shape (`strict_required_status_checks_policy: false`, `do_not_enforce_on_create: false`, the four contexts). Overridable like the other tunables via `REQUIRED_CHECKS` / `STRICT_REQUIRED_CHECKS`, with one deliberate deviation worth flagging for review: `REQUIRED_CHECKS` uses `-` rather than `:-`, because an explicitly empty value has to mean "omit the rule" — which is what a fork whose CI produces different job names needs — and `:-` would silently restore the default instead. Unset still means the four defaults. `STRICT_REQUIRED_CHECKS` is validated to true/false, and contexts containing `"` or `\` are rejected rather than interpolated, matching the existing `RENOVATE_APP_ID` treatment. `set -f` around the word split so a stray `*` cannot glob into filenames.

I also made `status` print the *contexts*, not just the rule type — "the rule is present" was not enough to confirm the gate is intact, and it is what makes your F4 recommendation actionable:

```
== rules in main-protected ==
deletion
non_fast_forward
pull_request
required_status_checks: build, build-macos, build-windows, security

== bypass actors ==
RepositoryRole id=5 mode=always
Integration id=2740 mode=pull_request
```

Header gains a third warning: `apply` is a full `PUT`, so anything configured in the UI and absent from the script is deleted. `usage()`'s `sed` range follows.

**Verification, without mutating anything.** `payload()`'s output diffed against `gh api repos/vfarcic/dot-agent-deck/rulesets/20587589`: the `required_status_checks` rule is byte-identical to live, and the whole payload matches except a server-side `required_reviewers: []` read default on the pre-existing `pull_request` rule — that ruleset was created by this script, which has never emitted the field, so a `PUT` without it reproduces the same state. **`apply` and `delete` were never run.** shellcheck clean at `--severity=style`.

## Ahead of your findings — the skipped-checks question your F3 opened

Your note that the devbox-only and flake-only groups skip all four required jobs raised a worse possibility than the blocker: if `SKIPPED` does not satisfy a required check, then pairing required checks with a path-filtered CI had made a whole class of PRs permanently unmergeable — the #416 trap reachable by anyone, including this PR. So I checked it before anything else.

**It does not hold, on two independent grounds:**

1. **A job skipped by a conditional counts as passing.** GitHub is explicit: *"A job that is skipped will report its status as 'Success'. It will not prevent a pull request from merging, even if it is a required check."* The #416 trap is a check that never reported **at all** — no check run exists to evaluate. A conditionally-skipped job produces one, with conclusion `skipped`. Different mechanism, and the distinction is exactly what makes #416 stuck and #510 mergeable.
2. **`ci.yml`'s `changes` gate is scoped by *author*, not only by path** — `if [ "${PR_AUTHOR:-}" != "renovate[bot]" ]` → full CI. So a human PR touching only Markdown, only `prds/` or only devbox files gets the full matrix regardless. Measured on the four most likely candidates: #504 (workflows + docs), #499 (one `.claude/skills` file), #469 and #464 (one `prds/*.md` each) — all four ran `build`, `build-macos`, `build-windows` and `security`, none skipped.

So nothing to fix, but the docs said nothing either way, which is why the question was live at all. Both facts are now in governance.md step 6 and rule 8, with the note that the author scope is *defence in depth* rather than the load-bearing half — worth knowing before the `changes` gate is ever widened beyond Renovate.

(Unrelated, but surfaced by that sweep and worth your attention independently: #469 has a red `build-windows` and #464 has red `build-macos` **and** `build-windows`, both on PRs that only add a `prds/*.md`. Those are now genuinely blocking rather than advisory.)

## F2 — the guard tests `-z` only

**Fixed by narrowing, in all three places** (governance.md, rule 8, and the PR body). You are right on both halves, including that the old fallback never degraded an expired PAT either — a GitHub `||` returns the left operand whenever it is non-empty, so the two behaviours differ *only* for an unset or empty secret. Both pages and both workflow comments now say plainly that a present-but-expired, revoked or under-scoped token passes the guard and still dies at `git push origin main`, with the tag already cut or the image already in GHCR.

I chose narrowing over probing deliberately: a probe buys coverage of a case that has never occurred here, at the cost of an extra network dependency and a new failure mode on the release path. What it is replaced with is a pointer to the thing that actually validates a token — step 5's `docs-publish.yml` canary — plus an explicit instruction to re-run it after every PAT rotation, which was the real gap.

## F3 — the UNVERIFIED bullet

**Replaced with the measurement**, and thank you for finding #510; you are also right that it contradicted its own page one section over. Confirmed independently: authored *and* merged by `app/renovate` at 2026-08-11T16:21:23Z, `reviewDecision: REVIEW_REQUIRED`, `devbox.json` + `devbox.lock` only, and all four required checks `SKIPPED` rather than passed.

That last detail is the sharper half, so it is recorded: Renovate did not merely bypass the approval, it merged with required checks that never succeeded. Both pages now state it as a **zero-human-approval, zero-green-check lane onto `main`** — security-relevant rather than editorial, as you put it — accepted by design with `ci.yml`'s push-to-`main` trigger as the net, and with the observation that on those two paths the post-merge run is the *first* time any Rust job sees the change. `RENOVATE_BYPASS=false` closes it.

The #503 claim is dropped for the reason you give: `Cargo.lock` means the full matrix, so it structurally cannot exercise the skip path.

## F4 — presence vs bypass capability

**Recorded, and half of it resolved in your favour.** The bypass actors *did* survive the full `PUT`: `gh api repos/vfarcic/dot-agent-deck/rulesets/20587589` still reports `actor_id: 5` (`RepositoryRole`, `always`) and `actor_id: 2740` (`Integration`, `pull_request`). You could not check this because `bypass_actors` is admin-only and your access is `push` — noted in the docs so the next `write` maintainer asks rather than assumes, and `status` now prints it.

Your other half stands and is now written down: no direct push to `main` since the checks returned at 2026-08-11T15:00:51Z (v0.35.10 published the day before at 17:21Z), and since v0.35.6 died specifically on a required-checks rejection of a check-run-less direct push, that leg is untested. "Closed on both sides" is now "in place, not exercised", with your recommendation adopted — run `status` before the next tag, and prefer the canary over a real release as the first thing to exercise it.

## F5 — no way back documented

**Fixed, and this was the sharpest of the non-blocking findings** — the procedure did the opposite of what the page's own argument claimed. Emergency override now handles both switches with the ordering explicit: list and disarm armed PRs, disable `allow_auto_merge`, *then* delete the ruleset; and on the way back, restore the ruleset, run `status` to confirm all four rules and both bypass actors, *then* re-enable arming — so there is never a window with auto-merge available on an unprotected `main`. The auto-merge bullet now carries the caveat that "arming is not a bypass" holds only while the ruleset is in force. Every `gh` invocation in the new block was run or `--help`-checked rather than written from memory.

## F6 — `secrets: inherit`

**Comments fixed; the narrowing deliberately deferred to #514.** Your read is right: it forwards HOMEBREW_TAP_TOKEN and SCOOP_BUCKET_TOKEN into a job that reads neither, and my new comment blocks had started instructing readers to preserve it, which entrenched a thing worth removing. They now say it is broader than needed and point at the issue. I also refreshed the `docs` job comment, which still claimed the checkout "silently falls back to `github.token`" — stale as of `bc11bc6`.

I did verify the prerequisite: `docs-publish.yml` references only `secrets.RELEASE_TOKEN` and `secrets.GITHUB_TOKEN`, and the latter is provided automatically to a called workflow. What I could not verify is the last step. The narrow map needs a matching `on.workflow_call.secrets` declaration in the callee, and `docs-publish.yml` is also `workflow_dispatch`-able; that path is expected to keep reading the repository secret, but GitHub's docs do not state it and confirming it needs a real run — which pushes a chart bump to `main` and publishes an image. Getting it wrong fails the next release's docs job. Guessing on this path is how v0.35.6 and v0.35.9 both went, so #514 carries the two edits plus the canary that validates them, rather than a drive-by on a review round. Say the word if you would rather it landed here.

## Minor

- **`set -euo pipefail`** added to both guard blocks. ✅
- **`f1204ec` committer vs pusher** — fixed. The committer is `github-actions[bot]` because the step sets `git config user.name`; the *pusher*, and so the identity the ruleset evaluated, is the `RELEASE_TOKEN` PAT's admin account. The passage now says so explicitly, and the evidence it exists for — that only the invocation differed — is preserved. ✅
- **`apply-branch-protection.sh:36`'s "there is no CODEOWNERS file"** — fixed; it now describes the pathless router that is there, and keeps the original reasoning where it still applies, to *per-path* ownership. ✅
- **`yq` by `sudo wget` with no checksum, in a job holding the admin PAT** — filed as **#515** rather than fixed here, to keep this diff to the change under review. It lays out the options (checksum verification, dropping `yq` for the three trivial edits, a SHA-pinned action, or devbox) and notes the blast radius you identified: it runs after the checkout and two steps before the push to `main`.

## Downstream note — forks

**Recorded, in a new governance.md "Forks" section.** Independently confirmed the part I could see: `prageethw/dot-agent-deck` shipped five releases in the six days to 2026-08-11 (v0.35.8 → v0.37.1). The secret list I could *not* see — `gh secret list --repo prageethw/dot-agent-deck` returns `HTTP 403` from upstream — so that is attributed to your review and flagged as fork-side to confirm, which is itself worth documenting.

The section gives the fix a fork operator needs (one PAT with **Contents: read and write**; on an unprotected fork `main` it needs no admin rights and no bypass, since the token only has to push), notes the alternative of a local one-line patch, and adds the mirror-image hazard for the script: `REQUIRED_CHECKS=` is what a fork should pass, because requiring a context its CI never produces is the #416 trap self-inflicted.

## Gates

`cargo fmt --check`, `cargo clippy --workspace --all-targets --features e2e -- -D warnings`, and `cargo xtask linkage-check` all green. Both workflow files re-parsed as YAML, guard confirmed still step 1 with the checkout at step 2 and no `${{ }}` in either guard body. shellcheck clean on the script. Markdown prose kept to single lines per rule 10.

Over to you — **not merging**, per rule 8 this needs your approval. Re-requesting review now that the pushes are done, since `dismiss_stale_reviews_on_push` would have voided an earlier one.

vfarcic:
## CI status: green except `build-windows`, which is `main`, not this PR

Everything settled on `edbcb3f`. `build`, `build-macos`, `security`, `nix`, `devbox`, `docs-build`, `windows-cross-check`, `changes` and `label` all pass. **Greptile: 5/5, no inline comments, no findings** (read per rule 8 — there was nothing to respond to).

`build-windows` fails, and it is not caused by anything here. The failing test is `xtask-linkage-check::clean_tmp::tests::explicit_roots_replace_the_standard_set`, at `cargo nextest run --workspace`. Three independent confirmations:

- **The test does not exist on this branch.** `git grep` finds it only on `main` — it arrives through the pull request's merge ref, from `3d1b6f4` (#322, merged as #472). This PR contains no Rust at all.
- **The identical test fails on #469 and #464**, which add a single `prds/*.md` each. Same test name, same job.
- **#512 is already open and fixes exactly that file** — *"fix(xtask): use a platform-absolute `--root` path in the clean_tmp test"*, touching only `xtask/linkage-check/src/clean_tmp.rs`.

So `main` is red on Windows and every open pull request inherits it. **#512 is the one that unblocks the queue** — this PR, #469 and #464 all go green behind it, and none of them can merge before it now that `build-windows` is required.

That is worth more than a comment, because it is a direct consequence of the switch this PR documents, so `2a528d7` records it as a fourth thing step 6 surfaced. Before required checks, `ci.yml`'s post-merge run on `main` made a broken `main` a *notification* while pull requests kept merging; now the same break stops all of them at once. The gate working as designed, but with a cost worth naming — and the operational rule that falls out of it: when several unrelated pull requests go red on the same required job, suspect `main` before suspecting the pull requests.

Nothing to do on this PR for it. Flagging so the red check is not read as a defect here, and so #512 gets priority.

vfarcic:
## CI is green on `c7b059f`

All checks pass, including `build-windows` — the only red job on the previous head. All four required contexts are green: `build`, `build-macos`, `build-windows`, `security` ([run](https://github.com/vfarcic/dot-agent-deck/actions/runs/31618032039)).

The `build-windows` failure was the stale `main` breakage, exactly as flagged above, and nothing in this diff. It was `xtask-linkage-check::clean_tmp::tests::explicit_roots_replace_the_standard_set` panicking at `xtask/linkage-check/src/clean_tmp.rs:1386` with `--root needs an absolute path, got "/scratch/one"` — #511, fixed by #512 (merged as `90918c06`, touching only `xtask/linkage-check/src/clean_tmp.rs`). This branch was based on pre-fix `main`, so it kept inheriting it; `90918c06` was confirmed not to be an ancestor of the old head `2a528d7`.

Updated via `gh pr update-branch`, so it is a **merge** of `main` into the branch — no rebase and no force-push, and `2a528d7` is still an ancestor. The branch is now current with `main` at `3496e7a`.

Greptile re-reviewed the new head and passed, again with no inline comments.

**What is still open:** the review decision is `CHANGES_REQUESTED` from the 2026-08-11 review at `ca00958`, and one approval is required. There are no unresolved review threads — that review came through as a top-level body with zero inline comments, so `required_review_thread_resolution` is satisfied. Green CI does not make this mergeable, and it is not being merged.

@prageethw — the red check that was sitting on this is gone and it is not a defect here, so this is ready for another look whenever you have a moment. My responses to the blocker and F2–F6 are in the comment above.

prageethw (changes requested):
The blocker is fixed. I verified `payload()` against the live ruleset (`20587589`) rather than against the description, and every rule the read-only endpoints expose is now modelled:

| Surface | Live ruleset | `payload()` | Verdict |
|---|---|---|---|
| `deletion` | present | present | matches |
| `non_fast_forward` | present | present | matches |
| `pull_request` | approvals 1, stale-review dismissal, thread resolution, all three merge methods | same effective settings (omits the empty server-default `required_reviewers: []`) | matches effectively |
| `required_status_checks` | `build`, `build-macos`, `build-windows`, `security`; strict false; do-not-enforce-on-create false | identical by default | matches |
| bypass actors | API redacts them to this credential | Admin role 5 / `always`, Renovate integration 2740 / `pull_request` | not independently verifiable here |

One caveat on method, stated so you can discount it appropriately: the ruleset API returns `bypass_actors: null` to my write-level credential, so I could not re-read the live Admin/Renovate entries independently. Everything else was compared directly against the live rule.

## Blocking — the `-` versus `:-` deviation you flagged for review

You asked for a look at it, and I think it is more than a style call.

`REQUIRED_CHECKS=""` makes `set -- $REQUIRED_CHECKS` produce zero words (`scripts/apply-branch-protection.sh:158`), `required_status_checks_rule_json` returns no JSON, and because this ships `gh api --method PUT`, the whole ruleset is **replaced** with three rules. All four required checks come off a live repository — with no error, no warning, and no acknowledgement.

That is the same class of defect as the original blocker: a tool that reconstructs branch protection while silently emitting something weaker than what it replaces. The difference is the trigger. The first version needed someone to run a stale script; this version needs an empty environment variable, which is a far more ordinary way to arrive at the same outcome — a CI step with an unset variable, a sourced env file, a typo in an export.

`docs/develop/governance.md` presents `REQUIRED_CHECKS=` as a routine fork option, and that framing is what makes an accidental empty value read as supported rather than dangerous.

I am not asking you to collapse unset-versus-empty — `-` was a deliberate choice and the intent behind it is sound. What it needs is for the dangerous branch to be **affirmative**:

- gate the empty case behind something like `ALLOW_NO_REQUIRED_CHECKS=true`, and error when `REQUIRED_CHECKS` is empty without it; or
- make empty an error outright, and give forks a separately named no-checks mode.

Either way the four defaults stay when the variable is unset, and a fork that genuinely wants no required checks has to say so in a way that reads as a decision rather than an accident.

## Everything else passes

**Guard ordering.** Both token guards are step 1 — before checkout, tags, commits, publication or pushes. A guard that ran after the first mutating step would be worse than none, and neither does.

**Credential coverage after removing the `github.token` fallback.** `release.yml` checks out with `secrets.RELEASE_TOKEN` and its reusable `docs` call keeps `secrets: inherit`, so the callee still receives it; `docs-publish.yml` has it before both the GHCR push and the `git push origin main`. Both of the direct-to-`main` pushes rule 8 records still have a credential.

**No leak.** Both guards bind the secret through `env`, test only emptiness, and print no value. The stated limitation — that a non-empty but expired, revoked or under-scoped token still passes the guard — is accurate and honestly scoped.

**Status output is honest.** `cmd_status` reconstructs the printed list from `.parameters.required_status_checks[].context` rather than echoing a hardcoded set, so it surfaces drift instead of hiding it. That is the right call and worth keeping.

**Docs** match the shipped behaviour throughout, apart from the `REQUIRED_CHECKS=` framing above.

Requesting changes for the empty-value footgun only.

vfarcic:
Thanks @prageethw — the empty-value footgun is fixed in `b55d811`, and you were right that it is more than a style call. I reproduced it before changing anything: `REQUIRED_CHECKS=""` → `set -- $REQUIRED_CHECKS` yields zero words → `required_status_checks_rule_json` returns no JSON → `payload()` emits three rules → `apply`'s `PUT` replaces the ruleset. Exit 0, no warning, and the PR page looks identical afterwards.

## The fix — the dangerous branch is now affirmative

I took your first option. `-` rather than `:-` stays, so unset still means the four defaults and the fork escape hatch still exists; what changed is that reaching it takes a second, plainly named variable:

```bash
REQUIRED_CHECKS= ALLOW_NO_REQUIRED_CHECKS=true scripts/apply-branch-protection.sh apply
```

An empty `REQUIRED_CHECKS` on its own is now a hard error that names both alternatives (set your own job names, or acknowledge). Three details worth flagging:

- **The flag never causes a removal.** With `REQUIRED_CHECKS` non-empty it is inert and the contexts are emitted exactly as before. It only permits what an empty value already asked for.
- **It is validated to `true`/`false`.** A typo like `ALLOW_NO_REQUIRED_CHECKS=yes` would otherwise fail closed into the empty-value refusal, whose message then blames `REQUIRED_CHECKS` for a mistake made in the other variable.
- **The test is on the word count after splitting**, not `[ -z "$REQUIRED_CHECKS" ]`, so `REQUIRED_CHECKS="   "` lands on the same decision instead of slipping through. That is the same accident with a stray space in it.

## Two related defects the verification turned up

Both are the same class you named — a tool that silently emits something weaker than what it replaces — and neither was visible from reading the diff.

**1. `payload()`'s validation was inert in a command substitution.** It relied on `set -e` to abort at `checks_rule="$(required_status_checks_rule_json)"`. Bash ignores errexit inside a command substitution used as an assignment's RHS — *"if a shell function executes in a context where `-e` is being ignored, none of the commands executed within the function body will be affected by the `-e` setting"* — so a caller written as `body="$(payload)"` prints the refusal, leaves `checks_rule` empty, and emits a ruleset with the rule missing. Exit 0. I hit this by writing exactly that caller, which is a fair warning about how easy it is to reintroduce. Measured:

| call form | inner refusal propagates? |
|---|---|
| `payload \| gh api …` (as shipped) | yes |
| `payload` directly | yes |
| `body="$(payload)"` | **no — emits a 3-rule body, exit 0** |
| `if ! body="$(payload)"` | **no — `payload` itself returns 0** |

`payload()` now uses explicit `|| return 1`, which propagates from all four. The script's existing comment claiming errexit propagation was true only at one level of nesting; the note now says what actually does the work, so the next person does not re-learn it the way I did.

**2. `cmd_apply` piped `payload` into `gh`.** On a refusal that still started `gh api --input -` and handed it empty stdin, leaving "is anything sent" up to how `gh` treats an empty body. It now renders the payload into a variable first, so a refusal transmits nothing at all.

## Docs

Your framing point is the one I think mattered most — `REQUIRED_CHECKS=` presented as a routine fork option is what makes an accidental empty value read as supported. Step 6 and the Forks section now say the omission takes two variables and state why: on a repository that *does* have checks, those same two keystrokes remove all four, and an unset variable is a far more ordinary way to arrive there than being a fork is. `CLAUDE.md` rule 8 matches.

## Verification

Ten variable combinations against a **stubbed `gh`** that treats any write as a test failure — only the defaults and the acknowledged pair reach a write:

```
SENT  unset defaults              -> [deletion, non_fast_forward, pull_request, required_status_checks]
BLOCK REQUIRED_CHECKS=            -> error: REQUIRED_CHECKS is empty, which emits a ruleset with NO required status checks…
BLOCK REQUIRED_CHECKS="   "       -> same refusal
SENT  REQUIRED_CHECKS= + ack=true -> [deletion, non_fast_forward, pull_request]
SENT  custom names + ack=true     -> [… , required_status_checks]   (ack inert)
BLOCK ack=yes / ack=false / RENOVATE_APP_ID=abc / STRICT=maybe / quote-in-context
```

`payload()`'s output diffed against live ruleset `20587589`: `required_status_checks` and `bypass_actors` are identical, with the same single `required_reviewers: []` server read-default on `pull_request` you noted. **`apply` and `delete` were never run** — the only live calls were read-only, and `status` still reports all four contexts and both bypass actors:

```
required_status_checks: build, build-macos, build-windows, security
RepositoryRole id=5 mode=always
Integration id=2740 mode=pull_request
```

On your F4 caveat: `bypass_actors` came back populated to this credential, so the admin/Renovate entries above are a direct read rather than an inference — that is the half you could not see from a `write` token.

## Gates

`cargo fmt --check`, `cargo clippy --workspace --all-targets --features e2e -- -D warnings` and `cargo xtask linkage-check` green; shellcheck clean at `--severity=style`; both workflow files re-parsed as YAML, with the guard confirmed still step 1, checkout step 2 on the bare `secrets.RELEASE_TOKEN`, no `${{ }}` in either guard body, and `secrets: inherit` intact on the `docs` job. All 11 checks pass on `b55d811`, including the four required. Greptile re-reviewed at 5/5 with no inline comments.

Nothing merged and nothing approved — over to you.

prageethw (approved):
Both blockers (the missing required_status_checks in payload(), and the empty-REQUIRED_CHECKS silent-omission footgun) are fixed exactly as proposed. The overclaimed guard-coverage language was narrowed rather than left standing, the stale Renovate-bypass claim was replaced with confirmed measurement against #510, and the untested 'closed on both sides' claim was honestly downgraded. No unresolved threads, CI fully green including all required checks.

PR #512, opened 2026-08-11
fix(xtask): use a platform-absolute `--root` path in the clean_tmp test

### LINKED ISSUE #511: build-windows is red on main: clean_tmp test passes a Unix-only absolute path to `--root`
Found while verifying #464, whose `build-windows` job went red on a **markdown-only** diff. The failure is inherited from `main`, not caused by that PR.

## The failure

```
FAIL xtask-linkage-check::bin/xtask-linkage-check clean_tmp::tests::explicit_roots_replace_the_standard_set
thread '...' panicked at xtask\linkage-check\src\clean_tmp.rs:1386:
  parse: "--root needs an absolute path, got \"/scratch/one\""
Summary 1579 tests run: 1578 passed, 1 failed, 0 skipped
```

## Cause

`parse_args` validates `--root` with `Path::is_absolute()` (`xtask/linkage-check/src/clean_tmp.rs:637`), which is **platform-specific**: a leading `/` is absolute on Unix but *not* on Windows, where an absolute path needs a drive prefix (`C:\...`). The test at `:1379` hardcodes `/scratch/one` and `/scratch/two`, so on Windows its own fixture fails the validation the surrounding code is exercising.

**Production behaviour is correct** — a Windows user passing `--root C:\scratch\one` works, and rejecting `/scratch/one` there is right. This is a test-fixture bug only.

## Why it is isolated

Exactly one test feeds absolute paths through `parse_args`. The other two call sites are fine on every platform: `:1393` passes `&[]`, and `:1439` passes a deliberately-relative `"scratch"` asserting `is_err()`. That is why precisely 1 of 1579 tests failed rather than a cluster.

## Timeline

| When | What |
|---|---|
| 2026-08-11T16:09:52Z | #472 merges, adding this test (`28d392b`, "move harness temp roots off the RAM-backed tmpfs") |
| 2026-08-11T16:09:55Z | #472's own `main` CI run is **cancelled**, superseded by a following push — so the red was never seen at merge time |
| 2026-08-11T16:21:25Z | `main` at `3cf4b68` runs and **fails** |

`main` is red now, and every open PR inherits it on its merge commit.

## Why it slipped through

This is the gap CLAUDE.md rule 2 describes. `xtask/linkage-check` was linted by no gate at all until #436 added `--workspace`, and `build-windows` deliberately stays on bare `cargo clippy -- -D warnings` (the L2 tier is Unix-only). So this member gets the least cross-platform scrutiny in the repo, while `cargo nextest run --workspace` on Windows does execute its tests. A cancelled merge run then removed the last chance to catch it.

## Fix

Make the two literals platform-appropriate — `#[cfg(windows)]` drive-prefixed, `#[cfg(not(windows))]` as today. Nothing touches the filesystem; the paths only need to satisfy the absoluteness check.

Note `xtask/linkage-check/src/clean_tmp.rs` is also modified by open PR #467 (PID-liveness reaping, #461), though not this test — whichever lands second may need a trivial rebase.

vfarcic:
Fixes #511. Unblocks `build-windows`, which is currently **red on `main`** and inherited by every open PR on its merge commit.

## The bug

`parse_args` validates `--root` with `Path::is_absolute()` (`clean_tmp.rs:637`), which is **platform-specific**: a leading `/` is absolute on Unix but *not* on Windows, where an absolute path needs a drive prefix. The test hardcoded `/scratch/one`, so on Windows its own fixture failed the very validation the surrounding code exercises:

```
FAIL clean_tmp::tests::explicit_roots_replace_the_standard_set
  parse: "--root needs an absolute path, got \"/scratch/one\""
Summary 1579 tests run: 1578 passed, 1 failed
```

**Production behaviour was never wrong.** Rejecting `/scratch/one` on Windows is correct, and `--root C:\scratch\one` works. This is a test-fixture bug only.

## The fix

The two literals become `#[cfg(windows)]` / `#[cfg(not(windows))]` consts. Nothing touches the filesystem — the paths only need to satisfy the absoluteness check, so the directories need not exist.

Scoped to one test deliberately: it is the only call site that feeds *absolute* paths through `parse_args`. The other two are correct on every platform — one passes `&[]`, the other a deliberately-relative `"scratch"` asserting `is_err()`. That is why exactly 1 of 1579 failed rather than a cluster.

## Verification

| Check | Result |
|---|---|
| `cargo test-fast` | **2523 passed, 0 skipped** |
| `cargo test-fast explicit_roots_replace_the_standard_set` (rule 6, isolated) | pass |
| `cargo xtask linkage-check` | ok |
| `windows-cross-check.sh -p xtask-linkage-check` | compiles for `x86_64-pc-windows-msvc` |

**One limitation worth stating plainly:** the last row is a *type-check*, not a test run, so it proves the `cfg(windows)` branch compiles — not that the test passes on real Windows. Only this PR's `build-windows` job can show that. Note also that `windows-cross-check.sh` runs `cargo check --tests` **without `--workspace`**, so it does not cover `xtask/*` by default; I passed `-p xtask-linkage-check` explicitly. That is arguably a gap in the script worth a follow-up, given CLAUDE.md rule 2 widened the *clippy* gate to `--workspace` for exactly this reason.

## How it reached `main`

Introduced by `28d392b` (#472, "move harness temp roots off the RAM-backed tmpfs"), merged 16:09:52Z. Its own `main` CI run was **cancelled** three seconds later by a following push, so the red was never seen at merge time; `main` then went red at `3cf4b68`.

This is also the gap CLAUDE.md rule 2 describes: `xtask/linkage-check` was linted by no gate until #436 added `--workspace`, and `build-windows` deliberately stays on bare `cargo clippy`, so this member gets the least cross-platform scrutiny in the repo — while `cargo nextest run --workspace` on Windows does execute its tests.

## Notes

- **No changelog fragment.** Test-fixture only, zero user-visible effect, so it has no release-note story. Happy to add a `misc` fragment if you would rather every PR carry one.
- `clean_tmp.rs` is also modified by open #467 (PID-liveness reaping, #461), though not this test — whichever lands second may need a trivial rebase.

### DIFF
--- a/xtask/linkage-check/src/clean_tmp.rs
+++ b/xtask/linkage-check/src/clean_tmp.rs
@@ -1372,6 +1372,21 @@ mod tests {
         }
     }
 
+    /// Two absolute paths that satisfy [`parse_args`]'s `is_absolute()` check on
+    /// the host platform. `is_absolute()` is platform-specific — a leading `/`
+    /// is absolute on Unix but NOT on Windows, which requires a drive prefix —
+    /// so a Unix-only literal here fails the very validation the test is
+    /// exercising, on Windows only (issue #511). Nothing touches the
+    /// filesystem; these directories need not exist.
+    #[cfg(windows)]
+    const SCRATCH_ONE: &str = r"C:\scratch\one";
+    #[cfg(windows)]
+    const SCRATCH_TWO: &str = r"C:\scratch\two";
+    #[cfg(not(windows))]
+    const SCRATCH_ONE: &str = "/scratch/one";
+    #[cfg(not(windows))]
+    const SCRATCH_TWO: &str = "/scratch/two";
+
     /// `--root` replaces the standard set rather than adding to it, so a
     /// deliberate scan of one directory cannot quietly also delete from
     /// `/var/tmp` or the system temp dir.
@@ -1379,15 +1394,15 @@ mod tests {
     fn explicit_roots_replace_the_standard_set() {
         let (_opts, roots) = parse_args(&[
             "--root".to_string(),
-            "/scratch/one".to_string(),
+            SCRATCH_ONE.to_string(),
             "--root".to_string(),
-            "/scratch/two".to_string(),
+            SCRATCH_TWO.to_string(),
         ])
         .expect("parse")
         .expect("options");
         assert_eq!(
             roots,
-            vec![PathBuf::from("/scratch/one"), PathBuf::from("/scratch/two")],
+            vec![PathBuf::from(SCRATCH_ONE), PathBuf::from(SCRATCH_TWO)],
         );
         assert!(
             parse_args(&[])


prageethw (approved):
Reviewed at `462dedc` — code read only, no build and no test run.

**Approving, and this is the one I would land first.** All four things I wanted to check hold up:

- **The fix is correct on both platforms.** `parse_args` does only `PathBuf::from` plus `is_absolute()` with no filesystem access, so the "these directories need not exist" claim is accurate, and both literals satisfy `is_absolute()` on their own platform.
- **It fixes the whole failure.** `explicit_roots_replace_the_standard_set` is the **only** test in the module that feeds absolute paths through `parse_args` — every other path literal is either `#[cfg(unix)]`-gated or never reaches `is_absolute()`. So no sibling test is left red on Windows, which the green `build-windows` on this PR confirms independently.
- **The load-bearing production claim is right.** Rejecting `/scratch/one` on Windows *is* correct behaviour, and it is corroborated elsewhere in the tree: `tests/common/mod.rs:7455-7462` already refuses drive-less rooted paths for the same stated reason. So this is a test-fixture bug only, exactly as the description says.
- Scope and title match the change.

### One suggestion

**The production claim is now asserted only in a commit message.** `a_relative_root_is_refused` tests `"scratch"`, which is relative on every platform, so nothing pins the Windows-specific *rooted-but-drive-less* shape that this PR is entirely about. If someone later relaxes the check from `is_absolute()` to `has_root()` — a plausible "fix" for a Windows path complaint — no test on any platform would catch it, and the behaviour this PR just confirmed as correct would silently invert.

A `#[cfg(windows)]` assertion that `--root /scratch/one` is refused would cost two lines and pin it where the next person will look.

### On the overlap with #467

**These two PRs fix the same defect.** #467's commit `96473894` makes the same change to the same test, differing only in shape — `const SCRATCH_ROOTS: [&str; 2]` there, `SCRATCH_ONE`/`SCRATCH_TWO` here — with identical values and the same `#[cfg(windows)]` split.

**My recommendation is to land this one first**, and I want to be explicit that this is *not* the "close one as a duplicate" situation:

- This is the minimal unblock — one file, +18/-3, test-only, all 11 checks green — for a `main` whose `build-windows` redness is currently inherited by every open PR on its merge commit.
- #467 is a ~1400-line feature PR with review items still open, so gating a red-`main` fix behind it costs days.
- #467's extra scope — the `EFBIG` fix for `exabyte_sparse_files_saturate_instead_of_aborting_the_sweep` on Linux and macOS — is genuinely independent and does not need to travel with the Windows fix.

After this lands, #467 should take `main`'s side of the conflicting hunk (delete `SCRATCH_ROOTS`, use `SCRATCH_ONE`/`SCRATCH_TWO`) and drop the now-stale paragraph 2 of `96473894`'s commit message in the same pass.

PR #539, opened 2026-08-13
feat(orchestration): suggest and enforce unique names on the new-pane form

### LINKED ISSUE #538: The new-pane form's Name pre-fills with the directory basename and enforces no uniqueness, so two orchestrations in one repo are indistinguishable
The new-pane form's **Name** field pre-fills with the picked directory's basename (`src/ui.rs`, `DirPickerIntent::NewPane`) and nothing checks uniqueness. A user who never touches the field still submits a non-empty name — **the same one every other orchestration in that directory submits.**

So opening two orchestrations in one repo gives both the identical name. There is no answer to *"which orchestration is this?"* that survives a restart:

- `orchestration_id` is unique but **minted fresh on every tab open** (`mint_orchestration_id`), so a reopened tab does not recognise itself.
- `name` + `orchestration_cwd` are identical for two tabs of one orchestration in one directory — which is the normal case when working on one repo.

## Why this is worth fixing on its own

#220's own problem statement notes that the friction of getting an orchestration into a worktree *"pushes users toward same-directory concurrency instead"*, and that **the #140 reporter hit the concurrency bug precisely because of it**. Same-directory concurrency is therefore a normal state, not an edge case — and in that state the deck cannot tell its own orchestrations apart, in the tab title or anywhere downstream.

This is **not** #220/M1.1, which is a naming/collision scheme for the `dispatch` verb's git worktree and branch names (and is already done). Two different naming concerns that share a word. Filing separately so neither gets closed by the other.

## Proposed shape

Make the typed name a real identity: **required**, **unique among live orchestrations**, and **suggested** so accepting it costs one keystroke — the next free `<foldername>-orchestrator-N`.

The uniqueness check needs no new machinery and **no extra daemon round-trip**: `live_orchestration_cwds()` is already called at form-open whenever the project offers an orchestration, and `TabMembership::Orchestration` already carries `display_title: Option<String>` over the same `ListAgents` round trip — it is simply discarded today. Verified against `main` @ `3496e7ab`: collecting the titles alongside the cwds is self-contained to `src/ui.rs` and touches neither `daemon_protocol.rs` nor `agent_pty.rs`.

The refusal renders on PRD #140's existing guard seam, so it is a form-validation change rather than a new surface.

## Note on `display_title`

This promotes `display_title` from decoration to identity. Today it is documented as title-only (`src/agent_pty.rs`: *"is title-only and never feeds delegate/role lookups"*; `src/ui.rs`: *"name to the tab TITLE only"*). That contract change is the substantive part of the review, and PRD #107's title-vs-identity split has to be preserved rather than collapsed.

## Provenance

Implemented and running in a fork; offering it here because the form and `display_title` are upstream code and the fix belongs where the code lives. Happy to open the PR — the port is already done and gate-clean — once there is agreement this is wanted and that the `display_title` promotion is acceptable.

prageethw:
Closes #538

## What changed

The new-pane form's Name field is pre-filled with the directory basename. It enforced no uniqueness, so two orchestrations spawned from the same repo were indistinguishable — in the tab bar, and in anything reading `display_title` over the wire. I think this might not be important to upstream, as the plan is for humans to create worktrees and point them to work on it from the get-go, just sending it here in case it changed; otherwise, please close this.

When an orchestration is selected, the Name field now suggests the next free `<foldername>-orchestrator-N` name instead of the bare basename, and submit is refused outright — not merely disabled — when the typed name matches a name a live orchestration already holds. This rides the existing same-cwd-warning guard seam, which now drops `[Submit]` entirely for a colliding name rather than only disabling it.

## The actual review: `display_title` moves from decoration to identity

This is the substantive change, and it's worth weighing on its own rather than as a side effect of the uniqueness feature. `display_title` is currently documented in two places (`src/agent_pty.rs`, `src/ui.rs`) as presentation-only — cosmetic, title-only, with no bearing on identity or behavior. This PR uses it as the uniqueness universe for the collision check: a submit is refused when the typed name equals another live orchestration's `display_title` (falling back to its canonical name when unset). That's a real contract change — `display_title` now has a behavioral consequence, not just a rendering one — and it should be reviewed as one. If that split between title and identity matters for other reasons in this codebase, this PR is the place to push back on collapsing it.

## No added daemon round-trip

The uniqueness check does not add a call to the daemon. `live_orchestration_cwds()` is folded into `live_orchestration_cwds_and_titles()`, so the check rides the same single `ListAgents` call the same-cwd warning already pays for at form-open time. `TabMembership::Orchestration` already carries `display_title` over that same round trip; this PR is the first thing to actually read it. That's what makes the change cheap on the common path, and it's probably the first thing you'll want to confirm rather than take on faith.

## Test coverage added beyond what this commit's history motivated

I added two unit tests — `orchestration/identity/002` and `orchestration/identity/003` — covering the suggestion and the refusal directly, plus updated two pre-existing tests that asserted on the old bare-basename pre-fill so they drive the field the way a user now would (backspacing over the suggestion before typing a custom name).

I'm calling this out because it wasn't a given: the commit this PR is drawn from shipped `suggest_orchestration_name` / `name_collision` and the whole refusal UI with zero test coverage, and I traced that state back through its parent commit to confirm it wasn't accidental — the tests genuinely weren't written. Offering an untested behavior change here seemed like the wrong call, so I added coverage rather than porting the gap as-is.

## What I deliberately left out

- `tests/e2e_orchestration_worktree.rs` — this file only exists in the environment this PR is drawn from because of an unrelated, environment-specific commit; there's no equivalent here, and porting it would have introduced a file with no reason to exist against this history.
- An L2 PTY test (`orchestration/identity/004`) that exercised this feature end-to-end — it lived outside the three files in this diff, needed a fixture I could not verify exists in this repo, and its catalog text referenced entries specific to the environment it was written in. Rather than guess at porting it, I'm flagging the gap here: end-to-end PTY coverage for this feature doesn't exist in this PR, and if that matters for merge I'd rather add it properly than paper over it.

## Files changed

- `src/ui.rs` — the suggestion/collision logic, the guard-seam change, and the two new + two updated unit tests
- `tests/CATALOG.md` — catalog entries for the new tests
- `changelog.d/538.feature.md` — changelog fragment

### DIFF
--- a/changelog.d/538.feature.md
+++ b/changelog.d/538.feature.md
@@ -2,6 +2,6 @@
 
 The new-pane form's Name field pre-filled with the directory basename and enforced no uniqueness, so two orchestrations spawned from the same repo were indistinguishable in the tab bar and in any tooling that reads `display_title`.
 
-When an orchestration is selected, the Name field now suggests the next free `<foldername>-orchestrator-N` name instead of the bare basename, and submit is refused outright — not merely disabled — when the typed name matches a name a live orchestration already holds. This rides the existing PRD #140 same-cwd-warning guard seam, which now drops `[Submit]` entirely for a colliding name.
+When an orchestration is selected, the Name field now suggests the next free `<foldername>-orchestrator-N` name instead of the bare basename, and submit is refused outright — not merely disabled — when the resulting title matches a name a live orchestration already holds. This rides the existing PRD #140 same-cwd-warning guard seam, which now drops `[Submit]` entirely for a colliding name.
 
 The uniqueness check adds no daemon round-trip: `live_orchestration_cwds()` is folded into `live_orchestration_cwds_and_titles()`, so the check rides the same single `ListAgents` call the same-cwd warning already pays for at form-open time. The uniqueness universe is each live orchestration's `display_title` (falling back to its canonical name), which already travels over the wire on `TabMembership::Orchestration` — no protocol change.
--- a/changelog.d/538.feature.md
+++ b/changelog.d/538.feature.md
@@ -1,0 +1,7 @@
+## The new-pane form suggests a unique orchestration name, and refuses a colliding one
+
+The new-pane form's Name field pre-filled with the directory basename and enforced no uniqueness, so two orchestrations spawned from the same repo were indistinguishable in the tab bar and in any tooling that reads `display_title`.
+
+When an orchestration is selected, the Name field now suggests the next free `<foldername>-orchestrator-N` name instead of the bare basename, and submit is refused outright — not merely disabled — when the typed name matches a name a live orchestration already holds. This rides the existing PRD #140 same-cwd-warning guard seam, which now drops `[Submit]` entirely for a colliding name.
+
+The uniqueness check adds no daemon round-trip: `live_orchestration_cwds()` is folded into `live_orchestration_cwds_and_titles()`, so the check rides the same single `ListAgents` call the same-cwd warning already pays for at form-open time. The uniqueness universe is each live orchestration's `display_title` (falling back to its canonical name), which already travels over the wire on `TabMembership::Orchestration` — no protocol change.
--- a/src/ui.rs
+++ b/src/ui.rs
@@ -1058,6 +1058,17 @@ struct NewPaneFormState {
     /// filesystem). Empty (the `new` default) means "nothing known live" — the
     /// suggestion starts at `-orchestrator-1` and nothing is ever refused.
     live_orchestration_names: Vec<String>,
+    /// Whether the user has edited the Name field by hand since the form
+    /// opened — set by the two text-edit key arms in
+    /// `handle_new_pane_form_key` (`FormField::Name` only), never by the
+    /// basename pre-fill `transition_after_dir_pick` applies. While `false`,
+    /// [`Self::suggest_name_if_orchestration_selected`] is free to overwrite
+    /// `name` with the generated suggestion; once `true`, that fn leaves the
+    /// field alone — a generated default may replace a generated default,
+    /// never a human edit. Guards re-clicking the selected chip, arrowing
+    /// between orchestrations, and arrowing off and back onto one, all of
+    /// which land on the same call.
+    name_touched: bool,
 }
 
 // ---------------------------------------------------------------------------
@@ -1153,6 +1164,9 @@ impl NewPaneFormState {
             // known live, so the suggestion starts at `-orchestrator-1` and
             // nothing is ever refused.
             live_orchestration_names: Vec::new(),
+            // A freshly opened form has no human edit yet — the basename
+            // pre-fill below is not one.
+            name_touched: false,
         }
     }
 
@@ -1211,24 +1225,51 @@ impl NewPaneFormState {
     /// the selection LANDS on an orchestration — called from every path that
     /// can change `selection_index` (arrow keys, click). A no-op when the
     /// current selection isn't an orchestration (a plain mode/card/authoring
-    /// option keeps whatever the user already typed).
+    /// option keeps whatever the user already typed), and a no-op once
+    /// [`Self::name_touched`] is set — a generated default may replace a
+    /// generated default, never a human edit. Without this guard, re-clicking
+    /// the already-selected chip, arrowing between two orchestrations, or
+    /// arrowing off one and back all silently clobber whatever the user typed.
     fn suggest_name_if_orchestration_selected(&mut self) {
+        if self.name_touched {
+            return;
+        }
         if self.selected_orchestration().is_some() {
             self.name = self.suggest_orchestration_name();
         }
     }
 
-    /// Whether the typed Name matches a name a live orchestration already
+    /// The title this submission will ACTUALLY take: the typed Name when it
+    /// is non-empty, otherwise the canonical fallback
+    /// `open_orchestration_tab` applies
+    /// ([`crate::project_config::resolve_orchestration_name`],
+    /// `src/tab.rs:846-849`). `None` when no orchestration is selected — a
+    /// plain mode/card/authoring option carries no identity uniqueness
+    /// constraint. [`Self::name_collision`] must compare THIS and not the raw
+    /// field — an empty field is not "no title", it is the canonical title,
+    /// and comparing `""` against the live titles can never match because
+    /// `live_orchestration_cwds_and_titles` filters empties out
+    /// (`src/ui.rs:944`). Uses `is_empty()`, mirroring `src/tab.rs:846`'s
+    /// predicate byte for byte — trimming here would let a whitespace-only
+    /// name resolve differently in the guard than in the tab.
+    fn resolved_title(&self) -> Option<String> {
+        let orch = self.selected_orchestration()?;
+        Some(if self.name.is_empty() {
+            crate::project_config::resolve_orchestration_name(&orch.name, &self.dir)
+        } else {
+            self.name.clone()
+        })
+    }
+
+    /// Whether the title this submission will actually take (see
+    /// [`Self::resolved_title`]) matches a name a live orchestration already
     /// holds. Only meaningful when an orchestration is selected — a plain
     /// mode/card/authoring option carries no identity uniqueness constraint.
     /// Drives the blocking refusal at submit and the `[Submit]`-button-gone
     /// render on the guard seam.
     fn name_collision(&self) -> bool {
-        self.selected_orchestration().is_some()
-            && self
-                .live_orchestration_names
-                .iter()
-                .any(|l| l == &self.name)
+        self.resolved_title()
+            .is_some_and(|t| self.live_orchestration_names.iter().any(|l| l == &t))
     }
 
     /// PRD #140 M4.0: whether the form should render
@@ -1301,6 +1342,9 @@ impl NewPaneFormState {
             // The locked schedule form can't select an orchestration either,
             // so there is never a name to suggest or collide.
             live_orchestration_names: Vec::new(),
+            // The Name field is hidden and fixed to `SCHEDULE_MODE_NAME` on
+            // this form, so there is nothing to touch.
+            name_touched: false,
         };
         // Lock the selection onto the built-in schedule option (index 1 with no
         // modes/orchestrations) so the existing schedule spawn branch fires.
@@ -6633,7 +6677,11 @@ fn handle_new_pane_form_key(key: KeyEvent, ui: &mut UiState) -> Action {
         },
         KeyCode::Backspace if matches!(form.focused, FormField::Name | FormField::Command) => {
             let field = match form.focused {
-                FormField::Name => &mut form.name,
+                FormField::Name => {
+                    // A human edit — the suggestion must never overwrite it.
+                    form.name_touched = true;
+                    &mut form.name
+                }
                 FormField::Command => &mut form.command,
                 FormField::Mode | FormField::Agent => unreachable!(),
             };
@@ -6641,7 +6689,11 @@ fn handle_new_pane_form_key(key: KeyEvent, ui: &mut UiState) -> Action {
         }
         KeyCode::Char(c) if matches!(form.focused, FormField::Name | FormField::Command) => {
             let field = match form.focused {
-                FormField::Name => &mut form.name,
+                FormField::Name => {
+                    // A human edit — the suggestion must never overwrite it.
+                    form.name_touched = true;
+                    &mut form.name
+                }
                 FormField::Command => &mut form.command,
                 FormField::Mode | FormField::Agent => unreachable!(),
             };
@@ -27984,13 +28036,22 @@ mod tests {
              suggest N=2, skipping the taken slot"
         );
 
-        // Simulate a stale/resubmitted taken name still sitting in the field
-        // (e.g. the user typed over the suggestion with a value someone
-        // else's live orchestration already holds).
-        ui.new_pane_form.as_mut().unwrap().name = "myproj-orchestrator-1".to_string();
-
         let enter = KeyEvent::new(KeyCode::Enter, KeyModifiers::NONE);
         handle_new_pane_form_key(enter, &mut ui); // Mode -> Name
+
+        // Simulate a stale/resubmitted taken name still sitting in the field
+        // (e.g. the user typed over the suggestion with a value someone
+        // else's live orchestration already holds) — reached through the
+        // real key handler, not by direct field assignment, so the test pins
+        // the user-facing path rather than a shortcut around it.
+        let backspace = KeyEvent::new(KeyCode::Backspace, KeyModifiers::NONE);
+        for _ in 0.."myproj-orchestrator-2".len() {
+            handle_new_pane_form_key(backspace, &mut ui);
+        }
+        for c in "myproj-orchestrator-1".chars() {
+            handle_new_pane_form_key(KeyEvent::new(KeyCode::Char(c), KeyModifiers::NONE), &mut ui);
+        }
+
         let result = handle_new_pane_form_key(enter, &mut ui); // attempted submit
 
         assert!(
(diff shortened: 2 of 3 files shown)


prageethw:
Note on the framing, not on the change — **this is a wording nit and explicitly not a reason to hold the PR.**

Checked first, because it decides whether there is anything to say at all: `git grep -n 'name_collision\|suggest_orchestration_name\|live_orchestration_names' upstream/main -- src/` returns **nothing**. There is no suggestion, no collision check and no refusal on `main` today. So this PR is a strict improvement — it takes the failure case from *"two orchestrations in one repo are always indistinguishable"* down to *"indistinguishable only when two new-pane forms are open at once."*

## The one thing worth adjusting

The body says submit is *"refused outright — not merely disabled"*, and the changelog reads as enforcement. That is true for the case the check can see, and it is worth being precise about which case that is, because `name_collision()` compares against a list captured **once, at form-open**, from a single `ListAgents` round-trip, and never refreshed. Two forms open simultaneously therefore each hold a snapshot that does not contain the other's name: both are suggested the same next free `N`, and **neither submit is refused**. Nothing at the marker-write layer enforces owner uniqueness either, so the form gate is the only enforcement point and it is advisory against a genuinely concurrent second form.

I am flagging it because I have watched this exact sentence cause a problem before. A prior implementation of this same feature on my fork shipped a PRD and changelog claiming a guarantee the code did not provide; a reviewer and an auditor caught it independently, and it needed a follow-up issue purely to record what the guard could not do. The code was fine both times — the prose was what misled.

The reason it matters more than pedantry: a partial guard *described* as enforcement can be worse than the same code described accurately, because it invites trust in precisely the case it cannot see. This is not theoretical for this repo — a session earlier today ran two concurrent orchestrations against the same checkout, which is the exact window the snapshot misses.

## Suggested

One clause in the changelog fragment and the body, something like:

> The check compares against the orchestrations live when the form was opened, so it is advisory against a second form opened concurrently; it makes the common single-form case correct and makes identities distinguishable once they exist.

And, if it seems worth tracking rather than just documenting, a follow-up issue for closing the concurrent window — the natural fix is re-checking at submit rather than trusting the form-open snapshot, or enforcing at the marker-write layer where uniqueness could actually be authoritative.

## Separately

Raising the `display_title` contract change yourself was the right call, and it is the part I would most want a second opinion on. Moving it from documented presentation-only to a uniqueness universe with behavioural consequence is a real change to what that field means, and it is the kind of thing that is easy to notice only later, from the other side of the wire.

prageethw:
## Both P1s answered at `25681933` — one fixed, one argued, plus a third defect neither of us had spotted

All 10 checks green on the new head, Greptile re-reviewed with no new findings. The two inline threads got marked resolved, but neither had a reply on it, so the substance is here rather than buried in a collapsed thread.

---

### P1 "Chip re-selection overwrites edited names" — **fixed, and it was broader than reported**

Confirmed, and the mechanism is exactly as described: `Action::FormSelectMode` called `suggest_name_if_orchestration_selected()` unconditionally, and that function overwrote `self.name` unconditionally.

Reviewing it turned up that the clobber fires on **three** paths, not one:

1. clicking the already-selected chip — the reported case;
2. arrowing from orchestration A to orchestration B (`select_next_mode` / `select_previous_mode`);
3. arrowing off an orchestration onto a mode and back.

Cases 2 and 3 are genuine selection *changes*, so the obvious fix — `if idx != self.selection_index` — closes only case 1 and leaves the other two. It would also have changed the pre-existing behaviour of re-clicking a chip, which today re-focuses `FormField::Mode`, a real and wanted effect.

Fixed instead with a `name_touched` dirty flag set in the `Backspace` and `Char(c)` arms of `handle_new_pane_form_key` (Name field only), with an early return in `suggest_name_if_orchestration_selected`. That closes all three, and it does **not** disturb the behaviour the PR is built around: the Name field's initial pre-fill comes from `transition_after_dir_pick`, which is not user input, so `name_touched` is still `false` when the orchestration is first selected and the `myproj` → `myproj-orchestrator-1` replacement still happens.

Pinned by `orchestration/identity/004`, which types the name through the **real key handler** and asserts survival across both the same-index re-click and the arrow-away-and-back — the case a selection-changed guard would miss.

---

### The third defect: an empty Name field bypassed the collision guard entirely

Not raised in review; found while working the P1 above, and worth stating plainly because it defeated the feature this PR is named for.

`name_collision` compared the **raw** Name field, but the tab title is a **resolved** value. When the field is empty those are different values and the guard was checking one the spawn never uses:

- `live_orchestration_cwds_and_titles` filters empty titles out, so `live_orchestration_names` can never contain `""` — an empty field returned `false` **unconditionally**.
- Neither submit door had an emptiness check, so the form rendered a perfectly ordinary submittable `[Submit]` with no warning.
- `req.name == ""` → `display_title = None` → the title falls back to `resolve_orchestration_name`, a pure function of config + directory.

So clearing the field twice produced two identically-titled tabs. **One user, two `Ctrl+n` passes, no concurrency and no second client.**

The crisp form is the asymmetry: **typing** the canonical name was correctly refused; **clearing** the field — which resolves to that same name — was not. The guard was inconsistent with itself.

Fixed by comparing the resolved title rather than the raw field, via a `resolved_title()` helper. Deliberately *not* by refusing an empty submission and *not* by silently substituting the suggestion — an empty field has a long-standing meaning ("use the canonical name") and both alternatives would change behaviour beyond this PR's scope. It uses `is_empty()` rather than `trim().is_empty()` so it mirrors `src/tab.rs`'s predicate byte for byte; trimming would reopen the same divergence one character to the left. Every non-empty name takes the unchanged path, so no existing assertion moved.

Pinned by `orchestration/identity/005`, deliberately its own catalog id rather than folded into `004` — they are different defects (`004` is the write side, `005` the read side), and one scenario per id is what stops either half being silently dropped in a later refactor while the id stays green.

---

### P1 "Stale snapshot permits duplicate titles" — **the race is real; I'm arguing it doesn't belong on this seam**

Conceded without reservation: two attached clients can each open a form, each capture a `ListAgents` snapshot, and each be told a title is free. The snapshot is taken once at open and never refreshed, there is no authoritative check anywhere on the spawn path, and both tabs get indistinguishable titles. The description is accurate.

**Not conceded: that this PR is the place to fix it.** Four reasons, each checked against the code before writing it here.

**1. The comparison is against `main`, not against perfection.** `git grep -n 'name_collision\|suggest_orchestration_name\|live_orchestration_names'` on `main` returns **zero** hits for all three symbols — there is no uniqueness check today at any layer. On `main` two orchestrations from one repo are *always* indistinguishable, because the Name field pre-fills with the bare directory basename and both spawns take it. This PR narrows that from "always" to "only when two forms are open concurrently against the same title". Blocking on the residual case restores a strictly worse state.

**2. No form-time check can close it, so the fix does not belong on this seam at all.** The objection isn't that this snapshot is refreshed too rarely — it's that a snapshot is not an authority. Refreshing at submit time shrinks the window to the round-trip latency and leaves the identical bug. Worth noting the one-shot-snapshot pattern is **pre-existing**, not introduced here: `live_orchestration_cwds` has 17 hits on `main` and PRD #140's same-cwd warning already renders a verdict from exactly this snapshot. This PR folds a second consumer onto the same call.

**3. It crosses a compatibility boundary.** The wire *shape* wouldn't move — `AttachResponse` already carries `ok`/`error`, and `StartAgent` already carries everything the check needs — so no `PROTOCOL_VERSION` bump. But `StartAgent`'s *meaning* would change: a request that has always been accepted could now be refused. An older TUI against a newer daemon surfaces that as a bare pane-spawn error with no form to return to; a newer TUI against an older daemon silently gets no enforcement while appearing to have it. Same-wire/different-meaning, so it wants a `.breaking.md` and a two-direction cross-version check — not a rider on a form-behaviour change.

**4. There's a design decision here a form-time patch would pre-empt.** An orchestration tab is **N separate `StartAgent` calls**, one per role. A naive per-spawn refusal either rejects role 0 and leaves a half-built tab, or rejects roles 1..N of a tab whose own role 0 just registered the title. It has to be keyed on the per-tab `orchestration_id` token so a tab never collides with itself — and then someone has to decide what a genuine collision does (refuse and surface it to the still-open form? auto-suffix? claim-and-hold?). None of that is settled, and settling it inside a form PR is how it gets settled badly.

The same conclusion was reached independently downstream ([prageethw/dot-agent-deck#201](https://github.com/prageethw/dot-agent-deck/issues/201)): form-time uniqueness is inherently **advisory**, and a durable answer belongs in the product, arbitrated by the daemon.

**Filed as #555** rather than left implicit, with the full design scope — where the check goes, the `orchestration_id` keying, refuse-over-auto-suffix and why, the stale-registry-entry case that would otherwise make a name permanently unclaimable, and the compatibility note.

**The deferral is only honest because the cheap half is closed here.** That's why the empty-name defect above is in this PR rather than bundled into the same follow-up: a PR titled "enforce unique names" that defers the hard half *and* ships a guard defeated by clearing one field wouldn't be enforcing anything a user couldn't trivially step around.

---

### One more thing, on test coverage

`render_new_pane_orchestration_name_collision_to_buffer` was added as a `pub fn` and called from **nowhere**. Because it's `pub` in a lib target, `dead_code` never fires, so no gate caught it — and the refusal's most user-visible behaviour, `[Submit]` **disappearing** from the action row rather than being dimmed, was pinned by nothing. A refactor restoring a dimmed-but-present `[Submit]` would have passed every gate.

`identity/005` now asserts through that seam. Also fixed `identity_003`, which set the field by direct assignment while its comment claimed it simulated typing — it skipped the very handler the `name_touched` fix hooks, so it couldn't have caught a regression in how the state is *reached*.

And the changelog fragment said submit is refused "when the **typed** name matches" — precisely true of the old code and precisely the empty-name bug. Reworded to "the resulting title".

vfarcic:
## Answering the question #538 asked, plus one required change

You asked in #538 for agreement before building this, and opened the PR saying you'd take a close if the answer was no. Sorry that sat — answering it properly now.

**Keep it.** I would never run two orchestrations in one directory, and I don't think users should either — a worktree per line of work is the model, and #220 exists to make that one step instead of a chore. But same-directory concurrency clearly happens (it's how the #140 reporter got there), and a guard rail for a workflow I'd steer people away from is still worth carrying. Three files, no dependencies, no protocol change — it costs nothing to keep.

I'm also fine with the `display_title` promotion you flagged as the substantive part. Routing never consults it — delegate/work-done key on the per-tab instance token, and `lookup_orchestration_role` resolves by canonical name + cwd — so nothing behavioural depends on it and this stays advisory. Raising it yourself rather than letting it ride was the right call.

**The #555 deferral is accepted.** Your four-point argument holds; a snapshot isn't an authority and a form PR is the wrong place to settle daemon-side arbitration. Please don't reopen that here.

---

## One required change: the suggestion leaks onto panes that aren't orchestrations

`suggest_name_if_orchestration_selected` (`src/ui.rs:1441`) only ever *writes* a name. Nothing restores the previous value when the selection leaves the orchestration, so the suggested name persists onto whatever you land on next.

This isn't an edge case. The cycler order is `No mode → modes… → orchestrations… → schedule → [issues] → [dispatcher]`, so **every arrow-key route to the trailing built-in options passes through the orchestration chip** and picks the suggestion up on the way.

Reproduced end to end against this PR's `target/release` binary, in a sandbox folder `proj` with one orchestration and one plain mode:

| Step | Selected chip | Name field |
|---|---|---|
| form opens | `[No mode]` | `proj` |
| Right ×2 | `[Orch: tdd-cycle]` | `proj-orchestrator-1` — intended |
| Left ×1 | `[chat]` | `proj-orchestrator-1` — wrong |
| Right ×2 | `[schedule]` | `proj-orchestrator-1` — wrong |
| Right ×1 | `[dispatcher]` | `proj-orchestrator-1` — wrong |

Submitting a plain `cat` pane from `[No mode]` produced a pane labelled `proj-orchestrator-1` in both the border and the session card (`1 No agent · proj-orchestrat…`). On `main` it would have been `proj`. `build_new_pane_request` passes `name: form.name.clone()` straight through for the dispatcher and issue-dispatch options too, so those cards inherit it as well.

It also quietly defeats the feature: that plain pane carries the label but isn't a `TabMembership::Orchestration`, so it doesn't reserve the name. I confirmed the next form still suggests `proj-orchestrator-1`, giving two panes with identical labels — the exact thing this PR is for.

`identity/004` already drives this arrow-away sequence with a *touched* name, so the untouched case is close to its existing scaffolding.

---

## One design change: make the suggestion collision-triggered

This is the half that affects the workflow I actually use, and I'd like it changed.

As built, the suffix is unconditional — selecting an orchestration always yields `<folder>-orchestrator-N`, with N=1 when nothing is live. I verified both halves in a clean sandbox:

| Live orchestrations | Folder | Suggested |
|---|---|---|
| none | `proj` | `proj-orchestrator-1` |
| one (`proj-orchestrator-1`) | `proj` | `proj-orchestrator-2` |

In a one-orchestration-per-worktree model that means every tab I ever open gets `-orchestrator-1` appended, permanently, and the digit never varies because it never needs to — a disambiguator with only one value to disambiguate between. The basename was already doing that job, since the worktree name is what distinguishes the lines of work.

It also costs legibility where legibility is the point: with a short fixture name the dashboard card already truncated to `proj-orchestrat…`, cutting the digit that carries the meaning. Real worktree basenames are much longer than `proj`.

**What I'd like instead:** suggest the bare basename as today, and fall to `<folder>-orchestrator-N` only when that basename is already held by a live orchestration. The first orchestration in a directory keeps its clean name; the suffix appears exactly when someone is doing the thing that needs disambiguating, which is when this feature earns its keep. The refusal behaviour stays as you built it.

Useful side effect: with no suggestion applied on the common path, the leak above loses most of its blast radius too.

I know you called the always-suffix form a settled PRD decision — this is me reopening it deliberately, not overlooking it, because the always-on cost lands on the workflow the project recommends.

---

## Non-blocking follow-ups

1. **The stale snapshot also produces false *refusals*.** The list is captured once at form-open and never refreshed. Close an orchestration while a form is open and its name stays refused for that form's lifetime, `[Submit]` gone, no way forward but Cancel and reopen. This is the mirror of Greptile's P1. #555 mentions "the stale-registry-entry case that would otherwise make a name permanently unclaimable" — worth confirming that scope covers this direction explicitly.
2. **The changelog fragment still overstates the guarantee.** You proposed the advisory-scope clause yourself in your first comment and it didn't land — the wording changed from `typed name` to `resulting title`, but nothing notes the concurrent-form window. Your own reasoning applies: a partial guard described as enforcement invites trust in the case it can't see.
3. **Docs contradict the behaviour.** `docs/orchestration.md:413` says concurrent orchestrations are safe "even when two orchestrations share the same `name`", and the §"Two orchestrations with the same project name conflict" section documents same-name as supported. Both stay true of routing, but the form now refuses what the docs call fine, and nothing documents the suggestion.
4. **The new render branch is half-pinned.** `identity/005` asserts `[Submit]` is absent, but nothing asserts the `NAME_COLLISION_WARNING` copy renders and nothing pins the new precedence over `SAME_CWD_ORCHESTRATION_WARNING`. Flip that `if`/`else if` and every gate stays green. `tests/render_orchestration_guard.rs` already owns this seam family.
5. **`suggest_orchestration_name`'s doc comment is inaccurate.** It says N is "counted globally over ALL live orchestrations, not per-directory", and argues a per-cwd counter "would offer `-orchestrator-1` in two different directories at once". Because the candidate embeds the basename, the counter is effectively per-basename and two directories *do* each get `-orchestrator-1` — I saw a sibling folder suggest `proj-orchestrator-2-orchestrator-1` while `proj-orchestrator-1` was live. Comments carry a lot of weight in this codebase; one arguing the opposite of the code is worth fixing.
6. Minor: `render_new_pane_orchestration_name_collision_to_buffer` is `pub` but called only from `src/ui.rs`'s own test module, while its sibling is driven from `tests/render_orchestration_guard.rs`. Either move the caller there (folds into 4) or drop to `pub(crate)`. And #555 has no labels.

---

## Verification

Ran the full gate suite on the merge result (clean, 0 behind `main`):

`fmt` ✅ · `clippy --workspace --all-targets --features e2e` ✅ · `build --release` ✅ · `test-fast` ✅ 2614/2614 incl. all four new `identity_00[2-5]` · `linkage-check` ✅ · `windows-cross` ✅ · `audit` skipped locally (CI's `security` job green). CI is 10/10 on `2568193`, including `build-macos`, `build-windows` and `security`.

The e2e tier reported five failures — `claude_001`, `dispatch_013`, `lock_012`, `shell_activity_005/006`, all real-Claude tests. **Not yours.** Rerun serially in isolation: 5/5 pass, and much faster than they failed (dispatch_013 141s → 11s, lock_012 170s → 28s). That's the load-contention signature already filed as #351, and none of the five touches the new-pane form. Four real-agent tests skipped on Codex model reachability, also unrelated.

I also drove the feature by hand in an isolated sandbox, and **it does what you say it does**: the suggestion appears, skips to `-2` once the first is live, the refusal renders `! This name is already in use by a live orchestration.`, `[Submit]` disappears from the action row rather than dimming, Enter is refused, and the form stays open with no second tab created.

Two things worth calling out on your side of the review: the `name_touched` fix is the right shape — the `idx != selection_index` guard really would have closed only one of the three paths and broken the re-click-refocuses-Mode behaviour — and finding the empty-name bypass yourself was a good catch. Comparing the raw field against a list that filters empties could never match, and it defeated the feature with one user and no concurrency at all.

---

**So: the leak fix and the collision-triggered suggestion, and I'm happy to merge.** Everything under follow-ups can land here or separately, your call.

vfarcic (approved):
Verified locally against the merge result (head 2568193 on main 4520819). All gates green: fmt, clippy, build, test-fast 2721/2721, linkage-check, windows-cross; cargo audit green in CI. e2e reported 4 failures of 8360 — all real-agent LLM tests in the last four concurrent slots, each passing on an isolated rerun, so none is a defect here. Both Greptile P1s disposed of: chip-reselection clobber fixed in 2568193 with identity/004 pinning it; stale-snapshot race deliberately deferred to #555. Rule 12 needs nothing — against the PR's merge-base it touches only src/ui.rs, tests/CATALOG.md and the changelog fragment; PROTOCOL_VERSION stays 7. Manually validated the whole flow in the real binary. One non-blocking follow-up filed on the stale footer hint.

Issue #547, opened 2026-08-13
An ambiguous (partial) first write retains its payload record until TTL (#424 residual)


vfarcic:
Split out of #424 round 9, where the reviewer found it and classified it a follow-up themselves. It fails **closed**, which is why it is not a blocker: the consequence is a refused delivery that is reported on the card and bounded by `PAYLOAD_RECORD_TTL`, never a doubled submit.

## What

A **first** write that comes back `PayloadDelivery::Ambiguous` — a partial write, where some of the payload reached the PTY and we cannot tell how much — leaves a payload record behind that nothing releases. It sits there until the TTL expires.

Three pieces line up to produce it, and each one is individually correct:

1. `AgentPtyRegistry::write_guarded` records the payload for `Ambiguous` exactly as it does for `Applied` (`src/agent_pty.rs:4409-4423`). That is deliberate and should stay — something of ours *is* in that box, and a partial write is the case where a replacement is both most tempting and most dangerous, so it is precisely the case that must stay guarded.
2. `guarded_submit` flattens `GuardedSend::Ambiguous` into a terminal `GuardedOutcome::Refused("ambiguous partial write")` (`src/spawn.rs:1064-1070`). Also correct: the delivery cannot continue from a state it cannot describe.
3. The first-write match in the scheduled delivery path returns on `GuardedOutcome::Refused(_)` (`src/spawn.rs:883-913`) *before* the confirmation loop — and therefore before any `PayloadRecordRelease` or explicit `note_payload_settled` — exists. The other terminal outcomes reach a release; this one has none to reach.

So the delivery is over, but its unit of guard outlives it.

## Consequence

For up to `PAYLOAD_RECORD_TTL` after an ambiguous first write, a later, unrelated delivery of the **same bytes** into the **same pane** is refused as a repeat once the user has typed since — matching a digest whose owning delivery reached a terminal outcome long ago. The delegate worker-task pointer is deliberately the same one-line path across hand-offs, so same-bytes-later is the ordinary case rather than an exotic one.

The refusal is surfaced (`report_user_input_stop` → `DeliveryNotice` on the card) and self-clears at the TTL. Nothing is silently lost, and nothing unsafe is admitted.

## Why it is narrower than it looks

[...]

Issue #544, opened 2026-08-13
An automatic payload concatenates with an unsent user draft and submits both (#424 residual)


vfarcic:
Split out of #424, where both the reviewer and the auditor ruled it a follow-up rather than a blocker: it is a pre-existing limitation of every automatic payload write, not a regression introduced by #424's retry policy.

## What

A pane's input box can hold an unsent user draft. When the daemon writes an automatic payload into that pane — a delegate route, an orchestration hand-off, a scheduled fire, any deck-initiated send — the bytes are appended to that draft and the submit CR sends **both as one turn**. The user's half-finished thought becomes part of a prompt they did not send.

This is the long-documented limitation on `AgentPtyRegistry::write_to_pane_and_submit` ("subsequent calls will submit `{notice text}\n{user prompt}` together — the bytes accumulate in the agent's stdin line buffer"), seen from the user's side rather than the notice's.

## What #424 did and did not close

#424 closed the two cases where the daemon's own retry machinery caused it:

- a **blind submit-only probe** (attempts 3+) fires into a pane the user has typed in — refused, because a probe's entire effect is a CR (`AgentPtyRegistry::user_typed_since_automatic_write`);
- the **one bounded replacement payload** (attempt 2) repeats bytes we already put in that box after the user typed — refused, because that appends our prompt to their draft and submits the pair (`AgentPtyRegistry::user_typed_since_writing_payload`).

Both refusals are terminal and reported on the pane's card (`DeliveryNotice`).

What remains is any **first write of a new delivery**. #424 deliberately did not refuse those, and the reasoning is in the doc comment on `user_typed_since_writing_payload`: a predicate broad enough to catch them would refuse the initial delivery of every prompt into a pane whose user happened to type first, which re-opens the seed-prompt-never-arrives bug #424 exists to fix.

### This is NOT limited to payloads whose bytes differ

An earlier draft of this issue scoped the residual to a "genuinely new, **different**" payload. Both reviewers corrected that in #424 round 9, and the broader reading is the one that matters:

[...]

Issue #543, opened 2026-08-13
Hook events are unauthenticated: standing is not authentication (#424 F4 residual)


vfarcic:
Split out of #424 (F4), where it is a documented residual rather than a fix.

## What

Hook events are unauthenticated. Anything that can reach the daemon's hook socket can post a `SessionStart` naming any pane and agent id. #424 F4 closed the case where an unmarked producer identifies itself only AFTER our prompt was written and uses that to arm a full replacement payload: arming now requires the pane to have declared, BEFORE the write, that it is a launcher with a real agent coming behind it — a `SessionStart` carrying `WRAPPER_FORK_SESSION_START_ORIGIN` that the readiness gate skipped (`SessionStartWait::launcher_handoff`, `AgentPtyRegistry::note_launcher_handoff`).

**Standing is not authentication.** A producer that can post a raw `SessionStart` can post a `wrapper_fork`-marked one first, and then it has the standing. The gate is a real narrowing, not a boundary.

## What the marker costs a sender

Worth stating precisely, because it is why the residual was accepted rather than closed:

- **It withholds readiness.** The readiness gate SKIPS a `wrapper_fork`-origin start — it is a declaration of what a launcher intends to start, not evidence that anything can report a submitted prompt. So the marked event does not satisfy the wait it would otherwise satisfy.
- **It cannot bind a generation.** `latch_generation` does not bind on it, so the sender does not become the delivery's conversation by sending it.
- **It grants exactly one thing, and that thing is already granted.** The standing is the single launcher→successor handoff `latch_generation` already permits for the generation. It authorizes the successor, not the sender.

So the marked path buys an attacker nothing the unmarked path did not already buy them on a socket they can already write to; and the *accidental* case — a misconfigured or duplicated producer on a pane, which is the realistic failure and the one F4 was reported against — does not emit the marker at all, because nothing in an ordinary agent's hook installation sets it.

## Why it still deserves an issue

Every guard on this path is built out of facts a producer asserts. The daemon owns exactly two clocks that a producer cannot move (`user_input_at`, fed by attach STREAM_IN, and `automatic_write_at`, fed by the guarded writer itself, both used by F1) and everything else is somebody's claim. That is a fine trade for a local socket owned by the user, and a bad one the moment the socket is reachable by anything else.

## Suggested direction

[...]

Issue #542, opened 2026-08-13
launcher_handoff_agents (and the two pane-keyed clocks) grow unbounded for a daemon lifetime


vfarcic:
Split out of #424 (round 5), where it is a documented residual rather than a fix.

## What

`AgentPtyRegistry::launcher_handoff_agents` (`src/agent_pty.rs`) is a `Mutex<HashSet<String>>` that only ever grows. `note_launcher_handoff` inserts an agent id; nothing removes it — not `close_agent`, not agent exit, not pane close. A daemon that runs for weeks accumulates one short string per agent it has ever spawned.

## Why it matters

Not much, individually — that is why it was left as a residual. It is the same shape as `user_input_at` and `automatic_write_at`, which are keyed by pane id and likewise never pruned, and each entry is a UUID-sized string. It is worth an issue because there are now three of them and no policy: the honest reason none of them is a leak in practice is that daemons get restarted, which is a property of how the software is used rather than of the code.

Keying by AGENT id (rather than pane id) is deliberate and must stay — pane ids are reused across spawns and a previous occupant's launcher declaration must not grant standing to the next delivery — so any fix has to prune on agent removal, not on pane close.

## Suggested fix

Drop the entry where the agent is removed from the registry, next to the existing cleanup for that agent. The two pane-keyed maps (`user_input_at`, `automatic_write_at`) want the same treatment, but on a different trigger: they are safe to drop only when the pane genuinely goes away, and dropping `automatic_write_at` too eagerly would *weaken* the #424 F1 guard (a missing entry reads as "nothing of ours is in that box"), so that half needs care and a test.

PR #548, opened 2026-08-13
fix(prompt): confirm spawn-time delivery before treating it as delivered

### LINKED ISSUE #424: Spawn-time role prompt counted as delivered when only written to the PTY: a 500 ms fixed buffer loses it silently on MCP-heavy Claude Code startups
## Symptom

An orchestration start role launched as a Claude Code pane came up with **no role prompt at all**. The pane looked completely healthy — deck showed the role as `Working` — but the agent had never received its seed. It only became apparent when the human asked the agent what its role was and it answered as a generic assistant.

Nothing was logged. No error, no status message, no retry.

This normally works in this setup; it started failing recently.

## Environment

- Spawning daemon: `dot-agent-deck 0.35.8` (`daemon serve`), running since 2026-08-07 14:43
- `PATH` binary / hook binary: Homebrew **0.35.7** — this is the one every entry in `~/.claude/settings.json` invokes as `dot-agent-deck hook`
- Source checkout at `v0.35.7-129-g83ff76e`
- macOS 25.5.0 (arm64)
- Pane: `DOT_AGENT_DECK_PANE_ID=25`, `DOT_AGENT_DECK_AGENT_ID=171`, `DOT_AGENT_DECK_SOCKET=/tmp/dot-agent-deck-502.sock`
- Role command: `command = "devbox run claude-opus-devbox --permission-mode plan"`

## What actually happened

Deck did everything right up to and including writing bytes into the PTY. Verified from the failing session's own process table and file metadata:

| Time (2026-08-08) | Event |
|---|---|
| `16:39:58.806` | deck writes `.dot-agent-deck/orchestrator-context.md` — 12.3 KB, correct, current content |
| `16:39:58.837` | deck writes `session.toml` with this pane's `orchestrator_prompt` |
| `16:39:58` | `devbox run claude-opus-devbox --permission-mode plan` spawns (PID 82897) |
| `16:39:59` | devbox gen-script → launcher wrapper (PIDs 83331, 83335) |
| `16:40:00` | `claude --model opus …` actually execs (PID 83363) |
| then | Claude starts **5 MCP servers** and runs a SessionStart hook emitting **10.9 KB** before its input accepts a submit |

So the seed was prepared correctly — `prepare_orchestrator_prompt` (`src/ui.rs:2473`) succeeded, the context file is on disk, and `session.toml` holds the intended pointer:

```
Read .dot-agent-deck/orchestrator-context.md for your role, available agents, and delegation protocol. Acknowledge your role and wait for instructions.
```

What never happened is *submission*.

### 1. Readiness is a fixed 500 ms wall clock, not a readiness signal

`deliver_orchestrator_prompt` (`src/ui.rs:3224-3348`) gates the write on:

- `agent_ready` (`src/ui.rs:3259-3262`) — a session for the pane whose `agent_type != AgentType::None`, which flips only when Claude's **SessionStart hook** arrives; then
- `should_inject_spawn_time_prompt` (`src/ui.rs:1358-1366`), which is nothing but `now - ready_since >= SPAWN_TIME_READINESS_BUFFER`, where that constant is **500 ms** (`src/ui.rs:1344`).

The hazard this was built to close is documented in your own test:

> Claude Code's `SessionStart` hook fires very early in its boot, before its TUI input enters submit-CR-aware mode; … a CR firing immediately on SessionStart is treated as a newline in the input buffer instead of a submit. The user sees the role prompt land in the orchestrator's input box, followed by a blank line — the prompt was never dispatched.
>
> — `tests/spawn_time_role_prompt_submit_after_session_start.rs:8-17`

500 ms is simply not enough budget on a project with a heavy Claude startup. Here, SessionStart fires and *then* 5 MCP servers start plus a hook emitting 10.9 KB. The CR lands far inside the not-ready window. This is exactly PRD #128's original symptom, reproduced by outgrowing the buffer that was tuned for it.

`prds/done/249-delegate-respawn-readiness.md:99-116` already concedes the general point — "a fixed delay cannot *prove* readiness… explicitly the stopgap" — and `src/state.rs:94-95` names #243 and #234 as the durable replacements.

### 2. "Bytes written" is recorded as "prompt delivered"

`src/ui.rs:3301-3315`:

```rust
match pane.write_and_submit_to_pane_with_identity(...) {
    Ok(SendResult::Applied) | Ok(SendResult::Queued) => {
        *orchestrator_prompt = None;
        role_statuses[start_role_index] = OrchestrationRoleStatus::Working;
        ui.orchestration_prompted.insert(tab_id);
        ui.send_retry_backoff.remove(start_pane_id.as_str());
        ui.prompt_delivery.remove(start_pane_id.as_str());
        ui.orchestration_ready_since.remove(&tab_id);
    }
```

`Applied` means the bytes reached the PTY (`src/pane.rs:457-459`), not that the agent submitted them. On that basis deck drops the prompt, marks the role `Working`, and discards **all** retry state.

This is the part that turns a recoverable timing miss into a permanent silent loss. A `delivery_id` is already minted (`src/ui.rs:1465`, `3153-3172`) but nothing ever reconciles it against a subsequent `UserPromptSubmit` event — so "swallowed by a booting TUI" and "delivered and acted upon" are indistinguishable to deck, and the `send_retry_backoff` machinery never gets a chance to fire.

### 3. Secondary path: the 10 s fallback bypasses the buffer entirely

`src/ui.rs:3263-3273` — if `agent_ready` never arrives within 10 s, `timeout_ready` forces `buffer_elapsed = true`, so the write fires with **zero** readiness buffer. The agents least likely to signal readiness get the most fragile delivery.

Relevant here because `AgentType::from_command` (`src/event.rs:116`) only looks through `env`, `sudo`, and `sh|bash|zsh|dash|ksh|ash -c` (`src/event.rs:176-250`) before an **exact basename match** against `detect_basenames` (`src/agent_registry.rs:180`, `:208`, `:228`). It does **not** look through `devbox run`, so `devbox run claude-opus-devbox …` detects as `None` and readiness rests entirely on the hook reaching the daemon from inside the devbox shell. `src/agent_registry.rs:240-243` already acknowledges this launcher class.

### 4. It is undiagnosable after the fact

`init_logging_from_env` (`src/main.rs:1246`) installs a subscriber only when `DOT_AGENT_DECK_LOG` is set, so by default the entire delivery path is silent. `~/.local/state/dot-agent-deck/daemon.log` contained nothing from the failing day. There is no persistent record that a prompt was written, and — because of §2 — no record that one was *lost*.

## Theories checked and ruled out

- **`--permission-mode plan` in the role command** — the dot-agent-deck repo's own config committed the same flag in `9b91960` (2026-08-07 19:28) and has run orchestrator panes with it since.
- **`prepare_orchestrator_prompt` returning `None`** via its swallowed `.ok()?` on `create_dir_all`/`fs::write` (`src/ui.rs:2475`, `:2478`) — excluded: the context file exists on disk with the correct 12.3 KB of current content and a matching mtime.
- **`AUTOMATIC_PROMPT_DEADLINE` abandon** (`src/ui.rs:1455`, warn at `:3246`) — excluded: that path emits a `tracing::warn!` and a status message, and the role would not have been left in `Working`.
- **Project-side `"UserPromptSubmit": []`** in the repo's `.claude/settings.json` — present since January, long predating the working sessions, and an empty contribution rather than a disable.

## Suggested direction

Not proposing an implementation, just noting where the leverage is:

1. **Confirm before clearing.** Don't treat `Applied`/`Queued` as terminal success. Hold `prompt_delivery`/`send_retry_backoff` until a `UserPromptSubmit` carrying the minted `delivery_id` comes back, and let the existing backoff retry when it doesn't. This alone converts every variant of this failure — including ones no buffer tuning would catch — from silent loss into a retry.
2. **Make the buffer adaptive or at least configurable.** A hardcoded 500 ms is a fixed budget against an unbounded quantity (MCP server count, hook output size, machine load). An env override in the shape of `DOT_AGENT_DECK_SEED_FALLBACK_SECS` would be a cheap stopgap ahead of #243/#234.
3. **Give the 10 s fallback the buffer too** — there's no reason the less-certain path should get less protection than the certain one.
4. **Log delivery unconditionally** at info level: prompt written, delivery id, confirmed/unconfirmed. Right now a lost seed leaves no trace anywhere.

## Related, not duplicates

- #415 — `delegate_prompt_injection` PTY/daemon tests starving each other under parallel runs; test-harness flake, not runtime delivery
- #423 — role remits seeded once and decaying over a session; assumes the seed arrived in the first place
- #243, #234 — the durable readiness-observation replacements this issue's §1 is a symptom of waiting for

vfarcic:
Closes #424

### What it fixes

Spawn-time role/seed prompts were treated as delivered the moment bytes reached the PTY. `Applied`/`Queued` mean only
that the write reached the PTY, not that the agent submitted it — so the deck dropped the prompt, marked the role
`Working`, and discarded all retry state. Against a booting Claude Code pane (SessionStart fires early, then MCP
servers start and a hook emits ~11 KB before the input accepts a submit) the CR was swallowed and the pane came up
healthy and Idle with **no prompt at all**, then was permanently abandoned by the 60 s `AUTOMATIC_PROMPT_DEADLINE`.
Nothing was logged.

Delivery is now **provisional**: the prompt, `prompt_delivery` and `send_retry_backoff` are retained until a matching
`UserPromptSubmit` confirms, the existing bounded backoff retries, and `AUTOMATIC_PROMPT_DEADLINE` still abandons a
permanently unconfirmed prompt. Info-level delivery logging was added (written / delivery id / confirmed / unconfirmed
/ abandoned), noting it materializes a subscriber only when `DOT_AGENT_DECK_LOG` is set.

**This landed in three independent delivery paths, not the two the issue documents** — `deliver_orchestrator_prompt`
and `process_pending_seed_prompts` (`src/ui.rs`), plus `spawn::deliver` (`src/spawn.rs`), which is the path
`dispatch --single` actually takes and the one the field repro exercises.

### CLAUDE.md rule 12 — state this explicitly

**No `PROTOCOL_VERSION` bump and no `changelog.d/424.breaking.md` fragment.** The wire shape did not move: no field
added, removed, renamed or retyped, and no stable field changed meaning. Confirmed **twice on paper** by independent
review, and **empirically** by the rule 12 cross-version manual test — a branch TUI against the real previous-release
`v0.36.0` daemon (checksum-verified asset, protocol 7 on both sides so no bump could mask a semantic break), with
`DOT_AGENT_DECK_LOG` isolated alongside the sockets, `HOME` and state dir. Delegate routing, status hooks, work-done,
seed delivery and the new submit-only probe all worked; no behavioural break behind the stable wire.

### Out of scope, deliberately (the issue's own suggested directions)

- **§1** — replacing the fixed 500 ms `SPAWN_TIME_READINESS_BUFFER` with a real readiness signal. Blocked on #243/#234;
  `prds/done/249-delegate-respawn-readiness.md` already concedes a fixed delay cannot prove readiness.
- **§3** — giving the 10 s fallback the readiness buffer. Defensible but separable; filed as **#529**. Note it
  interacts with this change: that fallback is the population that now gets no retry, so the two are worth deciding
  together.
- **Unifying the three delivery paths** into one implementation. Tempting and probably correct eventually, but it turns
  a targeted fix into a rewrite of a seam already losing prompts. Filed as **#528**.

### Follow-up issues filed

#525 (PTY writes block a Tokio worker under the writer mutex), #526 (confirm by prompt digest rather than a 200-byte
prefix), #527 (`delivery_id` count-bounded but not byte-bounded), #528 (unify the three delivery paths), #529 (10 s
fallback writes with no readiness buffer), #532 (`pane_hook_session` treats a second producer's frames as a generation
change), #541 (`git worktree add` races under concurrent dispatch), #542 (pane-keyed maps grow for a daemon lifetime),
#543 (hook events are unauthenticated — standing is not authentication), #544 (an automatic payload concatenates with
an unsent user draft; broadened — digest equality does not make a later delegation a retry), #547 (an ambiguous partial
first write retains its payload record until TTL).

### Gate results

- `cargo test-fast`: **2564 / 2564**.
- `cargo fmt --check` and `cargo clippy --workspace --all-targets --features e2e -- -D warnings`: clean.
- `cargo xtask linkage-check`: ok (523 catalog ids, 451 annotations, 100 allowlisted, 8 rules).
- **Full `cargo test-e2e` on the shipping commit: 7930 passed, 4 failed — all four real-agent/full-tier flakes, three
  of which passed on isolated rerun.** The fourth, `devin_live_001`, failed in isolation twice with the provider
  returning `Permission denied: We're currently facing high demand for this model. Please try again later.` — an
  external capacity refusal, not a branch regression; it passed on this branch in an earlier tier run. Real-agent tests
  are the flaky-tolerant pre-PR tier per rule 4. State this in the PR body; do not hide it.
- New regression coverage includes `scheduler/dispatch/015`, a **real interactive Claude (Haiku)** test that reproduces
  #424 — measured RED on the pre-fix baseline (FAIL in 152 s, all three panes healthy/Idle with no prompt) and green
  after (~32 s, consistently).


vfarcic:
@prageethw when you get a chance, could you take a look at this one? No rush if you're mid-something.

Context for why it'd be good to land: this bug bit a real dispatch today. A `dot-agent-deck dispatch --single` on v0.36.0 spawned the agent fine (SessionStart in the log) but the `--task` prompt never arrived — the agent sat idle with an untouched worktree, exactly the failure #424 describes. MCP-heavy Claude Code startup, which is the condition in the issue title.

CI has been green across all jobs since 2026-08-13 (build, build-windows, build-macos, security, windows-cross-check), and it's currently `MERGEABLE`. Only thing outstanding is the required approval.

prageethw (commented):
## APPROVE WITH COMMENTS — two independent passes at `14000d1c`

Reviewed twice on purpose: one pass on correctness and the rule 12 contract, one on trust/forgery/exhaustion/disclosure, in separate worktrees without sight of each other. They agree on the verdict and found **disjoint** findings, which is the main argument for having done it that way. Nothing blocking. Two items I'd want an explicit answer on before merge; the rest is take-it-or-leave-it.

Both passes were read-only — **no tests were run and no gates executed**, so every conclusion below is from reading code. CI is green on all 10 checks.

One thing worth saying up front: the commit log shows seven prior review/audit rounds already folded in, and it shows. A large share of what an audit would normally raise here is already raised, answered and *documented in-tree* — including the places where an earlier, stronger claim was withdrawn and replaced with a narrower true one. That is unusual and it shaped what a second look could usefully add.

---

## The two worth answering

### 1. Rule 12's mandated test covers one direction; this PR adds something flowing the other way

**The "no `PROTOCOL_VERSION` bump" conclusion is correct, and I checked it rather than taking it.** The `+22/−1` in `daemon_protocol.rs` is three hunks: the in-process `cancel_prompt_confirmation` on `StopAgent`, `expected_session_id` into `delivery_fingerprint`, and `note_user_input` moved before `drop(w)`. The field that *looks* new isn't — `expected_session_id` already exists on `WriteAndSubmitExtras` on `main` (`daemon_protocol.rs:481`, `#[serde(default)]`, PRD #20 R20-003) — and the fingerprint is a daemon-internal ledger key that never crosses the wire. Both passes independently landed on `PROTOCOL_VERSION` staying at 7, and the audit pass went looking specifically for a contradiction and found none. I also checked the one way widening the fingerprint could have bitten — a `delivery_id` reused with a changed session id returning `Conflict` instead of `Replay` — and it's unreachable from both the new TUI (which mints a fresh per-attempt wire id at `ui.rs:3824`) and an old one.

What's untested is the reverse direction. Rule 12 prescribes **new TUI → old daemon**, and there the change is clean: the TUI-side machine needs only `UserPromptSubmit` carrying `user_prompt`, which v0.36.0 already forwards. But **old TUI → new daemon** now carries something new: `install_delivery_notice_sink` (`daemon.rs:1105`) has the daemon synthesize an `EventType::Error` onto a pane's card — a combination no older daemon emitted.

The PR itself is the reason to ask. `apply_daemon_report_event` (`state.rs:3001`) exists *solely* to snapshot and restore `pane_hook_session` / `pane_generation_closures` around applying that event, its doc saying the plain path "rolled `pane_hook_session` BACKWARD". **An old TUI has no `apply_daemon_report_event`** and routes everything through the unguarded `apply_event`.

Two honest qualifications, because they cut against the finding: the specific harm named there ("abandons an in-flight delivery") *cannot* occur on an old TUI, which has no provisional deliveries at all; and under this repo's stated definition nothing here qualifies as a semantic break, since `DELIVERY_NOTICE_METADATA_KEY` is a new key in the free-form `metadata` map that unknown consumers ignore by contract.

**Suggested resolution:** run the rule 12 manual test in reverse — v0.36.0 TUI against the branch daemon, driving a `dispatch --single` prompt through to abandonment — and record the result. If it's clean I'd drop this entirely. **No `.breaking.md` is required on the evidence we have.**

### 2. Capability is read from the agent *type*, not the producer — and `wrap` breaks those apart

`prompt_delivery.rs:308-313`, `wrap.rs:298`/`:304`/`:679-710`, `spawn.rs:931-934`, `ui.rs:3733-3741`

The module's own doctrine says *"capability is a property of the PRODUCER"*, and for Pi the carve-out implements exactly that. But Codex is the only shipped agent launched through `dot-agent-deck wrap` (pinned by `wrap_launch_command_leaves_non_wrapper_agents_bare`), and `wrap`'s emitter declares `agent_type: Codex` while hardcoding `user_prompt: None` on **every** event (`wrap.rs:298`/`:304`), emitting only `SessionStart`/`Thinking`/`Idle`/`Error`. Structurally that is the Pi shape wearing a Codex label.

Codex's real `user_prompt` comes from its native hooks, which `wrap` installs at spawn — but that trust step is explicitly best-effort, and its own warning names the consequence: `"codex: could not record scoped hook trust ({e}); deck events degrade to stdout classification"` (`wrap.rs:702-705`). In that degraded state both delivery paths answer `Reports`, because both read `agent_reports_submitted_prompt(AgentType::Codex) == true`. `evidence_channel_is_unidentified` doesn't rescue it — wrap events carry `agent_id` — and `is_daemon_synthetic` doesn't exclude them.

The consequence is the one this PR is at pains to avoid elsewhere: retries arm against a channel that can never confirm. Attempt 2 rewrites the full payload; attempts 3+ probe submit, submitting attempt 2's payload. So a prompt that *was* delivered can be **submitted twice** and then reported as *"it may never have arrived"* — measured against the PR's own standard that *"a dispatch prompt performs deployments, deletions and issue operations: running one a third and fourth time is strictly worse than a visibly idle pane"*.

**No test reaches it.** The new e2e file drives Claude only — `REAL_AGENT_COMMAND` is Haiku Claude, every synthetic hook posts `--agent claude-code`, and it's gated on `check_claude_available()`. Pi has a unit test; there is no Codex/`wrap` confirmation test in the diff.

**Marked non-blocking for one reason only: we ran nothing, so we could not confirm the precondition is reachable in practice.** If it isn't, this collapses to a doc fix — the doc on `agent_reports_submitted_prompt` asserts *"Codex's wrapper synthesizes the same shape"*, which is true of Codex's native hooks and false of `wrap`'s classifier. The proper fix, if it is reachable, is to have `wrap` declare **provenance** on its classifier events the way `emit_fork_session_start` already does for `SESSION_START_ORIGIN_METADATA_KEY`, and read capability from that.

---

## Smaller findings

**A fourth spawn-time prompt writer.** `arm_seed_fallback` (`agent_pty.rs:121-152`, PRD #201) is untouched. Confirmation is legitimately out of scope there — Pi hardcodes `user_prompt: None` at `main.rs:967`, so `agent_reports_submitted_prompt(Pi) == false` is right. The **identity guard** isn't: it sleeps 15s and then calls the unguarded `write_to_pane_and_submit`, which is the thing `spawn.rs:878-882` argues against for its own first write. It's also unreachable by the new `cancel_prompt_confirmation` / `cancel_all_prompt_confirmations` hooks. Minor knock-on: the changelog says the fix lands "across all three places", which a reader would naturally take to include this one.

**A stale doc paragraph, in the worst place for one.** `confirm_prompt_delivery`'s doc (`spawn.rs:1167-1183`) says a late-identifying producer "arms on that signal". The code (`spawn.rs:1306`) arms only when `accepts_late_producer`, a pre-write launcher-handoff declaration. The F4 comment 40 lines above is correct and contradicts it — and the stale paragraph is the one describing the launcher population the whole issue is about.

**Hook binary as a third version axis.** An old hook against a new daemon never emits the new `SESSION_START_ORIGIN_METADATA_KEY`, so the launcher class silently degrades to pre-PR behaviour. Graceful, but an operator who upgrades the deck without re-seeding hooks gets a partial fix and the changelog doesn't say so.

**The 200-byte truncation collision** (`prompt_delivery.rs:176-195`) — for any prompt over `USER_PROMPT_MAX_LEN`, any submission whose first 200 bytes match confirms the delivery. Already stated precisely in-tree, already tracked as #526, and the reason given for not fixing it here (refusing truncated reports reinstates #424 for every long prompt) is the right one. Noted only for completeness; agreed.

**Cosmetic.** `spawn::deliver` has no `CannotReport`/`Unknown` split, so a Pi pane holds a watch slot for the full 60s (no wrong write). `daemon.rs:1800` is a fifth unguarded automatic writer (delegate result back to the caller pane) — clearly out of scope, maybe a follow-up issue. `PaneInputState`'s `automatic`/`input` maps clear only on pane re-claim (`agent_pty.rs:3830`), never on agent exit, so they grow monotonically — small, and consistent with the accepted policy for the sibling `user_input_at` map. And `prompt/pane-input/029` exists in neither code nor catalog (023–028, 030–032) — a numbering gap, not a dangling reference; linkage-check won't fail.

---

## What came back clean — including some things worth calling out as *well* done

**Prompt content never reaches a log.** All eight `log_prompt_*` fns take path/pane/delivery-id/attempt/reason only. `PromptDelivery` derives `Clone` but **not `Debug`**, so the usual `{:?}` leak is structurally unavailable rather than merely avoided. New `agent_pty` debug lines log `payload_len`, not payload. And `DeliveryNotice.detail` is typed **`&'static str`**, which makes "no interpolation" a type-system guarantee — prompt, repo and role text cannot reach the user-visible notice. The only prompt-bytes surface is the pre-existing opt-in `pane_write=trace` target, present verbatim on `main`.

**Resource exhaustion is bounded on every axis, twice** (key space *and* the 60s deadline). `prompt_delivery`/`send_retry_backoff` keyed by pane and removed on every terminal branch; `CONFIRMATION_TASKS` capped at 256, single-flight per pane, cancelled on pane close (`daemon_protocol.rs:1442`) and daemon shutdown; `payloads` capped at 8/pane with a 60s TTL; backoff bounded at ≤9 attempts, unit-tested. Two paths can't retry one delivery — `spawn_confirmation_task` aborts the previous handle. Retry idempotence is handled properly: payload writes capped at 2, later attempts submit-only probes. Cap exhaustion surfaces on the card instead of a silent log.

**A stray keystroke cannot confirm.** The match requires pane + agent id + generation + watermark + text. Cross-pane confirmation is closed: scoped to the delivery's pane plus identity, text and watermark.

**No agent type waits on a confirmation that cannot arrive.** `agent_reports_submitted_prompt` matches `AgentType` exhaustively with no wildcard, and each answer was verified at its producer — Claude/Devin/Codex via `build_event_typed`, OpenCode via `build_opencode_event`, Pi `false` because `main.rs:967` hardcodes `user_prompt: None`. (Finding 2 above is precisely the seam where the *type* answer and the *producer* diverge.)

**The two `src/ui.rs` paths are not independent implementations** — they share one `prompt_delivery` map, one `capture_prompt_delivery` and one tick loop. TUI vs `spawn::deliver` agree on deadline, backoff, payload-vs-probe, matcher and generation latch, each via a shared call rather than a copy. That was the divergence risk three paths would normally carry, and it isn't there.

**Rule 7 verified mechanically: 74/74** `#[spec]` tests in changed files carry a `/// Scenario:` comment, and all 19 new spec ids are catalogued.

**Rule 4 satisfied on both halves.** `scheduler/dispatch/014` is PTY-attached L2 driving the real binary with an agent that deliberately **swallows the first submission** — the failure mode itself, not a happy path — and it runs in CI. `scheduler/dispatch/015` drives three real interactive Haiku agents behind a launcher with per-pane sentinels, confirming through Claude's native hook, correctly `skip_unless!`-gated out of CI. Abandonment is unit-tested under `tokio::time` control (`dispatch/016`–`021`), which is the right tier for a 60s deadline. Small note: neither is marked ` [reel]`, so the fix ships with no clip although `dispatch/015` looks eligible.

**On the hook socket**, for the record: it's owner-only (`daemon.rs:522-524`, `daemon_protocol.rs:745-747`), so cross-user forgery is blocked and same-uid forgery is a pre-existing non-boundary this PR neither adds an entry point to nor relaxes. It raises the *consequence* of that non-boundary rather than the exposure, a forged confirmation needs pane id + agent id + prompt text, and it buys a silent regression (retries stop) rather than an escalation. You state this in-tree at `ui.rs:3490-3493` and mitigate it correctly by requiring a non-optional `agent_id`. Same for `DELIVERY_NOTICE_METADATA_KEY`: forged events carrying it are excluded from the evidence channel, and it's correctly documented as *not* an authentication marker.

---

## Coverage, honestly

**Read closely:** all 908 lines of `prompt_delivery.rs`; full diffs of `daemon_protocol.rs`, `event.rs`, `hook.rs`; `spawn.rs`'s delivery/confirmation core (900-960, 1167-1330, 1595-1735); `daemon.rs`'s notice sink; `state.rs::apply_daemon_report_event`; `agent_pty.rs` 1929-2275 and the relevant guards; `wrap.rs` 270-320, 679-710; `main.rs`'s Pi event; and the full test-surface cross-check.

**Sampled:** `src/ui.rs` (+3520) — architecture verified and the `PromptDelivery` field docs read in full, but the tick loop (`ui.rs:3085-3407`) and the epoch/`wire_issued` interaction were **not** traced statement by statement; that's where I'd look first for anything missed. Likewise ~2100 lines of `spawn.rs`, `daemon.rs` (+273) by targeted read, and **`src/state.rs` (+1122) — its `latch_generation` / launcher-handoff logic is load-bearing for this design and was not audited end to end.** The ~5500 lines of added unit tests were checked for existence, annotation, catalogue entry and scenario text; their assertions were not verified.

**Not answerable from a read:** real-world reachability of finding 2's precondition, and whether 60s is the right deadline under real MCP-startup latency.

prageethw (approved):
Approving on the strength of the two-pass review posted above — nothing blocking, and the rule 12 conclusion holds independently of the author's own analysis: `expected_session_id` already exists on `main`, the fingerprint never crosses the wire, and `PROTOCOL_VERSION` correctly stays at 7.

The two items I'd still like an answer on are recorded there and neither needs to block a merge:

1. **The reverse cross-version direction** (old TUI → new daemon, where the daemon now synthesizes an `EventType::Error` onto a pane's card and an old TUI has no `apply_daemon_report_event` to guard it). If that manual test comes back clean I'd drop the finding entirely.
2. **`wrap`-launched Codex**, where capability is read from the agent *type* while `wrap`'s classifier hardcodes `user_prompt: None` — so a degraded-trust Codex pane answers `Reports` and can have a delivered prompt submitted twice. Non-blocking only because nothing was run to confirm the precondition is reachable; if it isn't, it collapses to a doc fix on `agent_reports_submitted_prompt`.

Worth saying explicitly since it's unusual: several things here are built better than they had to be. `PromptDelivery` deriving `Clone` but not `Debug` makes the `{:?}` prompt leak structurally unavailable rather than merely avoided, and `DeliveryNotice.detail` being `&'static str` turns "no interpolation" into a type-system guarantee. The seven prior review rounds show in the code.

Scope, stated plainly: both passes were read-only and **no tests or gates were run**. `src/ui.rs`'s tick loop and `src/state.rs`'s `latch_generation` logic were sampled rather than traced end to end — the latter is load-bearing for this design, so that's where I'd look first for anything missed.

vfarcic:
@prageethw — following up on the two items you wanted answered. This merged before both were closed out, so recording the outcomes here.

**Finding 2 (`wrap`-launched Codex) was real.** You marked it non-blocking for one reason — *"we ran nothing, so we could not confirm the precondition is reachable in practice"* — so I went and checked. It is reachable, and not marginally: `list_hooks_in` starts with `Command::new("codex").arg("app-server")` (`src/codex_hooks_manage.rs:522`), which resolves against the **wrapper's own `$PATH`** and returns `NotFound` when `codex` lives only inside the launcher. That is this repo's own configuration — `command = "devbox run codex-big"` (`.dot-agent-deck.toml:182`) — and `installs_hooks` gates on `pane_id.is_some()` (`wrap.rs:686`), which a deck-spawned pane always satisfies, so the trust step runs and shells bare `codex` no matter what the role's launcher is. Every spawn on such a host lands in the degraded state, with a `warn` nobody reads.

So it does **not** collapse to the doc fix. Filed as **#559** with the full chain, the caveats on what I did and didn't verify, and your suggested direction (read capability from provenance the way `emit_fork_session_start` already stamps `SESSION_START_ORIGIN_METADATA_KEY`, rather than from the agent type). The doc correction you identified on `agent_reports_submitted_prompt` is in there too, since it's needed either way.

One honesty note carried into the issue: I did **not** reproduce the double-submit end to end, only the precondition. I also probed `codex app-server` against the installed `codex-cli 0.147.0` and got no answer to `hooks/list` — but that probe proves nothing, because it closed stdin, which is exactly what `list_hooks_in` deliberately avoids for the reason its own comment gives. Only the `$PATH` variant is established.

**Finding 1 (reverse cross-version) is still open.** Old TUI → new daemon, driving a `dispatch --single` prompt through to abandonment, is not yet run. It no longer gates anything, but it's the direction rule 12 actually prescribes and your point stands that this PR added something flowing that way. I'll run it and post the result; if it's clean, that closes the finding as you offered.

Also worth flagging for your other PRs: `main` moved with four merges just now (#467, #466, #548, #553), and **#520 has gone `DIRTY`** as a result.

vfarcic:
@prageethw — finding 1 (reverse cross-version) is **clean**. Per your offer, consider it dropped.

## What was run

Old **v0.36.0 TUI** → new **`main` daemon** (`406df6b`, i.e. this PR merged), driving a spawn-time prompt all the way to abandonment while the old TUI stayed attached.

- The TUI is the **checksum-verified release asset** (`dot-agent-deck-linux-amd64`, `sha256sum -c` OK), not a local rebuild.
- Getting the pair to coexist needs one trick worth recording: PRD #103/#161's handshake SIGTERMs a build-mismatched daemon and respawns from the TUI's own binary when no agents are live, which silently destroys the pairing. The release binary **cannot** be pinned — `src/build_id.rs` gates `DOT_AGENT_DECK_BUILD_ID_OVERRIDE` behind `#[cfg(any(test, debug_assertions))]`, so release builds compile it out. So the *daemon* was pinned instead, to the release's own baked id (`0.36.0-g6a98629`), and the old TUI attached without restarting anything. Only the handshake value is affected; both sides run their real code.
- Rule 12 isolation throughout: `HOME`, `DOT_AGENT_DECK_SOCKET`, `DOT_AGENT_DECK_ATTACH_SOCKET`, `DOT_AGENT_DECK_STATE_DIR` **and `DOT_AGENT_DECK_LOG`** in a scratch dir. The real daemon was confirmed untouched.
- Trigger: `schedule run-now` into a stand-in that posts the same synthetic `SessionStart --agent claude-code` your e2e suite uses (`tests/e2e_prompt_delivery_confirmation.rs:205`), then `exec cat`s and never emits `UserPromptSubmit`.

## Result

The delivery ran the full documented lifecycle under the new daemon:

```
14:24:12  prompt written to pane; provisional until UserPromptSubmit confirms it   attempt=1
14:24:12  ... re-submitting                                                        attempt=1
14:24:13  prompt written to pane; provisional ...                                  attempt=2
14:25:13  ... probing submit without rewriting the payload                         attempt=3
   ...    (attempts 4-8, widening backoff, probe-only)
14:26:12  WARN prompt delivery unconfirmed at the deadline; abandoning   attempts=8
```

Payload writes capped at 2, everything after probe-only, abandoned at the 60 s deadline — exactly as specified.

And the old TUI's screen at the end (vt100 replay of the PTY capture, 200x50):

```
 dot-agent-deck — 1 session(s)
┌ 1 ClaudeCode · xvab3 ────────────────────────────────── ● Error ┐
│Dir:  —                                                          │
└───────────────────────────────────────── Last: 1m 39s  Tools: 0 ┘
 1 active  │  1 error  │  0 tools
```

The daemon-synthesized `EventType::Error` **reached the old TUI, was attributed to the correct pane, and rendered correctly** — `● Error` on the card, rolled up as `1 error` in the status bar — on a client that has no `apply_daemon_report_event` to guard it. No panic, no `ERROR` in the log, clean SIGTERM shutdown. The pair stayed up for the whole ~2.5 minute window.

## Scope, stated plainly

- I verified the event is **applied and rendered**; I could not observe the old TUI's `pane_hook_session` directly, since that's internal state with no external surface. What backs the "no harm" conclusion beyond the clean render is your own qualification: an old TUI has no provisional deliveries, so the harm the guard exists to prevent has nothing to act on there.
- Only the `$PATH` half of finding 2 was reproduced (see #559); the double-submit itself was not driven end to end.
- Four earlier attempts failed for reasons that were mine, not the code's — `dispatch` from a bare shell is rejected (`WARN dispatch from unknown pane`), and a silent stand-in resolves to `CannotReport` no matter what the binary is named, because `spawn.rs:1003` reads `observed_producer` rather than the command. Worth knowing if anyone reproduces this: **the producer has to actually emit an identifying event**, or you never reach the abandonment branch at all.

`python314Packages.pyte` was added to `devbox.json` for the vt100 replay — happy to drop it if you'd rather not carry it.

Issue #84, opened 2026-05-09
PRD: Rendering layer rework — contract-driven render instead of symptom patches


vfarcic:
## PRD: Rendering layer rework

**Problem**: Recurring visual issues — scrambled text in panes (especially near the bottom), empty space on the right after resize, and similar artefacts under layout changes — keep coming back after multiple targeted fixes. Each fix layers in defensive clamps and heuristics, but the underlying class of bug is structural: PTY size, layout rects, and what the terminal widget draws are derived independently and resynced reactively through scattered code paths. There is no single rendering contract.

**Solution**: Define a rendering contract first — single source of truth for layout, PTY sizes derived from layout rects, a render-time widget that draws 1:1 against the area it is given without heuristic clipping or windowing. Catalog every known visual failure mode, validate the new contract against the catalog, then reimplement against it.

**Detailed PRD**: See [prds/84-rendering-layer-rework.md](https://github.com/vfarcic/dot-agent-deck/blob/main/prds/84-rendering-layer-rework.md)

**Priority**: Medium

Issue #54, opened 2026-04-09
Always create dashboard card for every pane, even without an agent


vfarcic:
## PRD: Always Create Dashboard Card for Every Pane

**Problem**: When a pane is created but no agent is launched, no dashboard card appears. Users cannot switch back to or close orphaned panes.

**Solution**: Always create a session/card at pane creation time with a "No agent" state and distinct border color, ensuring all panes are visible and manageable on the dashboard.

**Detailed PRD**: Will be added after PRD file creation

**Priority**: High

PR #556, opened 2026-08-14
chore(prds): de-duplicate two PRD files and repoint their references

prageethw:
## Summary

`prds/54-always-create-pane-cards.md` and `prds/84-rendering-layer-rework.md` were byte-identical to their `prds/done/` twins. This removes the `prds/` copies, keeps the `prds/done/` copies (the archival convention documented in `.claude/skills/dot-ai-prd-close/SKILL.md`), and repoints the three references to `prds/84-rendering-layer-rework.md` at the surviving `prds/done/` copy so no link is left dangling.

`diff prds/54-always-create-pane-cards.md prds/done/54-always-create-pane-cards.md` → no output (identical).
`diff prds/84-rendering-layer-rework.md prds/done/84-rendering-layer-rework.md` → no output (identical).

Repointed references:
- `docs/develop/rendering-contract.md:108`
- `prds/104-snapshot-replay-preserves-pty-dims.md:140`
- `tests/CATALOG.md:1866`

`prds/54-always-create-pane-cards.md` had no references outside `prds/done/`, so it needed no repointing.

`git grep -n "prds/84-rendering-layer-rework\|prds/54-always-create-pane-cards" -- . ':!prds/done'` now returns nothing.

## Test plan

- [x] `cargo fmt --check`
- [x] `cargo clippy --workspace --all-targets --features e2e -- -D warnings`
- [x] `cargo xtask linkage-check` — ok (504 catalog ids, 432 annotations, 100 allowlisted, 8 rules)
- [x] `git grep` confirms no dangling references remain


vfarcic:
Thanks for this — I verified the de-duplication claim independently rather than from the diff, and it holds up cleanly: at the merge base each pair was literally the same git blob (`dfb1008` for #54, `a3e6610` for #84), and both `prds/done/` copies are bit-identical between base and head, so nothing was lost. `prds/done/` is also confirmed as the right survivor per `dot-ai-prd-close/SKILL.md:94-102` (`git mv prds/[number]-[name].md prds/done/`), and both issues are closed.

Heads up on one thing: rather than send you round again for two one-liners, I pushed a small follow-up commit to your branch directly (f8e6ea3). It repoints the two remaining references to the deleted `prds/84-rendering-layer-rework.md`:

- `tests/render_terminal_widget.rs:3` — the path is hyphen-wrapped across two `//!` lines (`prds/84-rendering-` / `layer-rework.md`), so any search for the intact basename slides straight past it. Very easy to miss — Greptile missed it in both review rounds too. This was the one that actually mattered: it resolved before the deletion, and it's the module doc of the very test file whose `tests/CATALOG.md` entry you'd already repointed, so the catalog and the file it describes had drifted apart.
- `prds/done/84-rendering-layer-rework.md:75` — the M2 self-reference Greptile flagged as P2.

The command that catches the wrapped form is `git grep -nE 'prds/(54|84)-'` — worth using instead of a basename search next time something like this comes up. It returns nothing now.

`cargo fmt --check` and `cargo clippy --workspace --all-targets --features e2e -- -D warnings` both pass on the follow-up commit, all 11 checks are green, and Greptile went 4/5 → 5/5 with no inline comments. Approving now — go ahead and merge when you're ready.

vfarcic (approved):
Verified locally at f8e6ea3. Byte-identity of both deleted PRDs confirmed at the git-object level (identical blob SHAs at the merge base, not a textual diff), prds/done/ survivors bit-identical between base and head, and no reference anywhere in the tree still points at either deleted path. prds/done/ confirmed as the archival convention per dot-ai-prd-close/SKILL.md. Local gates green: fmt, clippy (--workspace --all-targets --features e2e), build --release, test-fast, linkage-check, windows-cross. All 11 CI checks green; Greptile 5/5 with no inline findings. Note: the e2e tier was deliberately not run, as the diff is Markdown plus one doc comment.

PR #565, opened 2026-08-14
chore(devbox): add pyte so a PTY capture can be replayed into a screen

vfarcic:
Adds `python314Packages.pyte@0.8.2` to `devbox.json`. Developer tooling only — nothing links it into the shipped binary, and no automated tier changes.

## Why

Rule 12's cross-version manual test asks what an **older TUI** does with a newer daemon's output. That question is only answerable by reading the screen, and a raw `script` capture of a ratatui app is a diff-based stream of cursor moves — stripping escape sequences from it recovers nothing, and tailing it recovers the last repainted *region* rather than the frame. I hit exactly that verifying #548's reverse direction: a 258 KB capture that yielded no readable text at all, which is indistinguishable from "the TUI rendered nothing" unless you can replay it.

`pyte` is a pure-Python VT100 emulator, so the replay is ~10 lines and needs no terminal:

```python
screen = pyte.Screen(200, 50)
pyte.ByteStream(screen).feed(open(path, "rb").read())
print("\n".join(screen.display))
```

That produced the evidence that closed prageethw's finding 1 on #548 — the daemon-synthesized `EventType::Error` landing on the correct pane's card on a v0.36.0 client with no `apply_daemon_report_event` guard:

```
 dot-agent-deck — 1 session(s)
┌ 1 ClaudeCode · xvab3 ────────────────────────────────── ● Error ┐
└───────────────────────────────────────── Last: 1m 39s  Tools: 0 ┘
 1 active  │  1 error  │  0 tools
```

## Notes for review

- **Pinned to `0.8.2`**, matching the convention every other entry follows; `devbox add` initially wrote `@latest` and the lock key now agrees with the pin.
- **The `codex-big` reflow is `devbox add`'s own formatting**, kept rather than reverted so future `devbox add` runs produce no incidental diff. Verified a no-op: `jq -S` output is identical apart from `packages[]`, and the command string hashes the same both ways (`f79b709b…`).
- Existing `changelog.d/` precedent for a slug-named, dev-facing fragment: `dispatched-orchestration-delegate.bugfix.md`, `461.misc.md`, `552.misc.md`.

## Gates

| gate | result |
|---|---|
| `cargo test-fast` | **2600 run, 2600 passed, 0 skipped** |

e2e tier not run — this change touches no Rust, so the pre-PR tier has nothing to exercise that the fast tier and clippy don't already cover.

Closes no issue; there isn't one.


prageethw (approved):
Small, self-contained devbox tooling addition: pins `python314Packages.pyte@0.8.2`, updates the lockfile, and adds a changelog fragment. Nothing links it into the shipped binary and no automated tier is affected — verified in the diff (only `devbox.json`/`devbox.lock`/changelog.d touched). The pin, the kept `devbox add` formatting on `codex-big`, and the `0.8.2` version match the repo's existing conventions. Gates reported green in the description; no e2e needed since no Rust changed.

Issue #530, opened 2026-08-12
dispatch --single exits 0 before its prompt is known to be delivered


vfarcic:
Noticed while fixing issue #424; deliberately not changed there because freeing the caller the instant the bytes are written is the documented, intentional shape of `spawn(detach_delivery)`.

`dispatch --single` returns success as soon as `spawn::deliver` has written the prompt into the PTY. Confirmation — the part that decides whether the agent actually SUBMITTED it — runs in a detached task that outlives the CLI process by up to `AUTOMATIC_PROMPT_DEADLINE` (60s). So the exit code answers "did a pane get created and bytes get written", never "did the dispatch land", which is exactly the distinction #424 exists to make.

Concretely, all of these exit 0 today:

- the prompt was written and then abandoned unconfirmed at the deadline;
- the pane's producer cannot report a submitted prompt, so the write is final and unverifiable;
- the event stream lagged or closed and the delivery stopped `Indeterminate`.

#424 made every one of those *visible* — an `info`/`warn` delivery log line and, for abandonment, a durable unsubmitted notice written into the pane itself — but none of them reaches the caller's exit status, which is what a script or a scheduler branches on.

Options to weigh:
- an opt-in `--await-confirmation[=<secs>]` that keeps the CLI attached to the confirmation outcome and maps abandoned/indeterminate to a nonzero exit;
- leave the exit code alone and document precisely what it does and does not assert.

Whichever way it goes, the current behaviour should be documented rather than implied.

Issue #395, opened 2026-08-07
e2e: find_in_grid never retries, so find_in_grid(..).expect(..) races the render — scheduler form_002 now fails 3/3, and ~51 call sites share the footgun


prageethw:
`e2e_scheduler_manager::form_002_add_spawns_authoring_agent_in_picked_dir` has gone from intermittently flaky to **failing all three retries** on the fork's `e2e` CI job. The immediate cause is a render race in the test, but the same footgun is present at ~51 other call sites, so this is worth fixing as a class rather than one test.

Found while working PRD #393 (fork PR #51). **Unrelated to that branch** — it changes nothing in `tests/e2e_scheduler_manager.rs`, and nothing in the scheduler, the manager, or the form.

## Symptom

```
TRY 1 FAIL [77.338s]  TRY 2 FAIL [91.997s]  TRY 3 FAIL [71.389s]
thread 'form_002_add_spawns_authoring_agent_in_picked_dir'
  panicked at tests/e2e_scheduler_manager.rs:931:10:
the mode-locked schedule form must render a [Submit] button
```

Every try also trips nextest's `SLOW [> 60.000s]` marker before failing.

Trend across three consecutive runs on the same branch:

| run | outcome |
|---|---|
| `31130708274` | `FLAKY` — TRY 1 FAIL at 88.6s, passed on retry |
| `31132847941` | **FAIL 3/3** — 77.3s / 92.0s / 71.4s |

## Diagnosis

`tests/e2e_scheduler_manager.rs:925-931`:

```rust
deck.send_keys(b" ");                  // Space → confirm dir → locked schedule form
deck.wait_for_string("New Schedule");  // waits for the form HEADER
let (scol, srow) = deck
    .find_in_grid("[Submit]")          // single instantaneous scan, no retry
    .expect("the mode-locked schedule form must render a [Submit] button");
```

`find_in_grid` (`tests/common/mod.rs:1459-1468`) takes **one** `snapshot_grid()` and scans it. There is no polling, no timeout, no retry — it returns `Option` immediately:

```rust
pub fn find_in_grid(&self, needle: &str) -> Option<(u16, u16)> {
    let grid = self.snapshot_grid();
    for (row, line) in grid.lines().enumerate() { ... }
    None
}
```

[...]

PR #573, opened 2026-08-14
fix(spawn): arm the retry when the deck itself spawned the pane's agent (#570)

### LINKED ISSUE #570: A producer that claims reporting capability just after the prompt write disarms the retry, so a daemon-spawned dispatch prompt is written but never submitted
A `dispatch --single` spawned its agent and wrote the prompt into the pane, but the prompt was **never submitted**. The agent sat at an empty turn until the deadline. `dispatch` had already exited 0 (#530), so nothing surfaced the loss — it was noticed only because the agent visibly never started work.

The cause is a ~500 ms race between the producer's capability claim and the prompt write. Losing it disables the retry for the whole delivery, and the retry turns out to be **load-bearing on the normal path**, not a safety net.

## Measured, with a natural control

Both deliveries are in one `~/.local/state/dot-agent-deck/deck.log`, 14 seconds apart, same `dispatch --single` shape, same agent type, same daemon.

**The failure** — pane `sched-dispatch-fix-scan-forgery-4`:

```
20:50:55.298  Received event  event_type=SessionStart  agent_type=ClaudeCode
20:50:55.414  prompt written to pane; provisional until UserPromptSubmit confirms it   attempt=1
20:50:55.918  prompt written to pane; delivery cannot be confirmed by this agent, not retrying
              reason="a producer claimed a reporting agent only after the prompt was written
              and this pane declared no launcher handoff before it; holding the write instead
              of retyping into a target that may never report"
20:51:25.261  prompt written to pane; delivery cannot be confirmed by this agent, not retrying
              reason="this agent cannot report a submitted prompt, so nothing could confirm delivery"
```

There is no `attempt=2`. The write was held, watched to the 30 s deadline, and abandoned.

**The control** — pane `sched-dispatch-remove-idle-art-5`, 14 s later:

```
20:51:09.905  prompt written to pane; provisional until UserPromptSubmit confirms it   attempt=1
20:51:10.407  prompt delivery unconfirmed; re-submitting                               attempt=1
20:51:10.561  prompt written to pane; provisional until UserPromptSubmit confirms it   attempt=2
20:51:10.619  Received event  event_type=Thinking
20:51:10.619  prompt delivery confirmed by the agent's submitted prompt                attempt=2
```

**Attempt 1 did not submit in either case.** The delivery that succeeded succeeded *because it retried*. So suppressing the retry is not a conservative fallback here — on this path it is the difference between a working dispatch and a dead one.

## Mechanism

In `src/spawn.rs`:

- `:1003` — `can_report_prompts` is computed **before the write**, from the producer already running plus capability drained from frames queued before the write. Deliberate, per #424 F4: a producer identifying itself afterwards has asserted what it *claims to be*, not that our bytes reached it.
- `:1304` — `let mut armed = can_report_prompts;`
- `:1313` — `let accepts_late_producer = registry.agent_declared_launcher_handoff(&agent_id);` — a pre-write launcher declaration is the *only* thing that lets a late-identifying producer arm the delivery.
- `:1378` — `armed |= can_report_prompts && accepts_late_producer;` — here the producer **did** claim reporting capability, but no launcher handoff had been declared before the write, so this stayed `false`.
- `:1428` — `if !armed { continue; }` — the loop keeps watching to the deadline but never retypes and never submits.

So the bytes sit in the input box, unsubmitted, with the one mechanism that would have submitted them disarmed.

## Why this looks like a gap rather than the intended trade-off

The information the gate needed **was already available before the write**. `SessionStart` reported `agent_type=ClaudeCode` at `20:50:55.298` — 116 ms *before* the write at `.414`, and 504 ms before the producer's own claim at `.918`.

This pane was not an arbitrary process the deck happened to find. **The deck spawned it**, as a `dispatch --single`, having chosen the agent type itself. The gate asks "did this pane declare a launcher handoff?" when for a daemon-spawned dispatch pane it could ask "did *we* spawn a known agent type here?" — a question with a pre-write answer that does not depend on winning a race with the producer's own announcement.

## Suggested direction (not prescriptive)

Treat the deck's own spawn record for a pane it created with a known agent type as satisfying `accepts_late_producer` — a pre-write declaration of what is there, at least as trustworthy as a launcher handoff.

This deliberately does **not** relax #424 F4's actual guarantee. The thing #424 F4 refuses is an *unauthenticated post-write claim* from a producer on a pane whose contents the deck cannot vouch for — the case where the full replacement payload and its submit CRs might be typed into a shell. A pane the deck spawned itself, with an agent type it selected, is not that case.

Whatever shape the fix takes, per `.claude/skills/reproduce-first` it needs a **failing test first** — one that reproduces a producer claiming reporting capability shortly *after* the write on a daemon-spawned dispatch pane with no launcher handoff, and asserts the prompt is ultimately submitted. Note the control above shows a correct test cannot assert on attempt 1 alone.

## Scope notes

- **Not #530.** #530 covers why this was silent — `dispatch --single` exits 0 before delivery is known, and it explicitly lists *"the pane's producer cannot report a submitted prompt, so the write is final and unverifiable"* among the cases that exit 0 today. That is this case, described from the exit-code side. #530 does not address the race that put the delivery into that state, and fixing #530 alone would turn this into a visible failure rather than a working dispatch.
- Adjacent but distinct: #529 (the 10 s readiness fallback skipping the buffer), #243 (retiring the load-bearing `SESSION_START_WAIT_TIMEOUT`), #559 (wrap-launched Codex reading confirmation capability from the agent type), #528 (the three delivery implementations).
- Frequency is unmeasured. Two dispatches, one lost, in a single session — but that is one observation of each outcome, not a rate. Four earlier `dispatch --single` calls in the same session delivered fine.

## Reproduction status

Not deterministically reproduced. The evidence above is one real occurrence plus a natural control from the same log, and the code path was read afterwards to explain it. Anyone picking this up should reproduce it first rather than trusting the reading.

vfarcic:
Fixes #570.

You dispatch a unit, its tab appears with a live agent in it, and the agent sits there having been asked nothing. The task text is in its input box; it was never submitted.

## Reproduced first, then fixed

Issue #570 records that the bug was **not** deterministically reproduced — the diagnosis came from reading the code after a single field occurrence. Per `.claude/skills/reproduce-first` the first deliverable here was a failing test, and the reading turned out to be right.

**`scheduler/dispatch/014` (L2, extended).** A fourth concurrent `dispatch --single` pane, `seed-late-claim`, withholds its `SessionStart` until after the readiness gate has expired and the prompt is already in the pane. Its own delivery log is a line-for-line match of the field report:

```
21:51:21.371  prompt written to pane; provisional until UserPromptSubmit confirms it   attempt=1
21:51:24.224  Received event  event_type=SessionStart  agent_type=ClaudeCode
21:51:24.377  prompt written to pane; delivery cannot be confirmed by this agent, not retrying
              reason="a producer claimed a reporting agent only after the prompt was written
              and this pane declared no launcher handoff before it; …"
```

…while the three panes beside it, identical in every input except *when* their producer identifies itself, do this:

```
21:51:18.371  written attempt=1  →  .873 unconfirmed  →  19.025 written attempt=2  →  19.227 confirmed
```

That is the issue's own control, reproduced: **attempt 1 submits in neither population.** The stand-in swallows the first line by construction, exactly as the booting Claude TUI did in the field, so a confirmation here can only come from a retry — and the assertion is at the user's altitude, on the agent's own `UserPromptSubmit` durably held as `last_user_prompt`, not on bytes having been written.

**`scheduler/dispatch/016` (L1, extended).** The same mechanism, deterministic and in the fast tier, sitting directly beside the `#424 F4` case it must not weaken. Two `/bin/cat` panes, same `can_report_prompts: false`, same unmarked post-write `SessionStart`, differing only in the deck's own pre-write spawn record: the one the deck spawned with no known type must still send zero retry bytes, the one it spawned as ClaudeCode must retype the payload. Before the fix the second failed with `output=""` — nothing ever typed, no retry, nothing for the agent to submit.

Both were red before the fix and green after, and reverting just the one-line gate change turns them red again (verified).

## The fix

`src/spawn.rs` — `accepts_late_producer` now has a second standing:

```rust
let accepts_late_producer = registry.agent_declared_launcher_handoff(&agent_id)
    || registry.agent_spawned_as_reporting_agent(&agent_id);
```

The gate's question is "did anything say, *before we wrote*, what is in this pane?" — because that is what makes a producer appearing afterwards this delivery's expected occupant rather than a stranger's claim. It was asking only about a launcher handoff, when for a daemon-spawned dispatch it had a better answer available: **the deck exec'd that command itself.** In the field occurrence `SessionStart` even reported `agent_type=ClaudeCode` 116 ms *before* the write; the needed fact was already there and simply was not an input to the gate.

The new accessor reads `RunningAgent::spawn_agent_type`, and that choice is the whole security argument: it is the frozen launch-shape identity the spawn site supplied (`SpawnOptions::agent_type`, computed from the command by `AgentType::from_command`), and `set_agent_type` — the learn-from-hook-event upgrade — never writes it. On this path the value comes from `spawn_one`, which derives it from the command in the daemon's own schedule/`default_command` config. No producer, honest or forged, can manufacture the standing for itself.

### Why this preserves #424 F4

F4 refuses an **unauthenticated post-write claim on a pane the deck cannot vouch for** — the case where a full replacement payload and its submit CRs might be typed into a shell. That refusal is untouched: a pane spawned as a bare shell, `cat`, a recorder stand-in, or any command `from_command` could not resolve stays `None` and keeps refusing. `scheduler/dispatch/016`'s forged case is unchanged and green, and now has the positive case next to it so the boundary is pinned from both sides rather than only from the refusing side.

Two further F4 properties are deliberately kept:

- **This is standing, not capability.** `can_report_prompts` is untouched, so the delivery is still not armed at write time. What the record licenses is accepting the producer *when it announces itself* — so the one replacement payload goes in when there is an agent there to receive it, rather than on the retry clock 500 ms after the write. Same reasoning the launcher handoff already uses, and the reason `scheduler/dispatch/015`'s bootstrap launcher is unaffected (its script name resolves to no agent type, so nothing about it changes).
- **Everything downstream of arming is unchanged**: one bounded replacement payload then submit-only probes, the user-draft guards, the identity-guarded write, the pinned generation, the 60 s deadline.

The refusal log line now names both missing facts instead of only the launcher handoff, so a delivery that still holds stays diagnosable.

## CLAUDE.md rule 12 — did the TUI↔daemon contract change?

**No, and no semantic break behind a stable wire.** No message, field or handler changed: `daemon_protocol.rs` is untouched, no `Serialize`/`Deserialize` type moved, and `PROTOCOL_VERSION` is unchanged. The change is confined to the daemon's own detached confirmation task plus one registry accessor over an in-memory field; the TUI never observes `accepts_late_producer`. No `.breaking.md` — bugfix, so **patch** under the 0.x policy.

Cross-version manual test run anyway, in both directions, with `DOT_AGENT_DECK_LOG` isolated alongside `HOME`, both sockets, the lock dir and the state dir:

| | delegate routes | work-done reaches the orchestrator | hook status/type renders |
|---|---|---|---|
| v0.36.1 daemon + branch TUI | ✅ | ✅ | ✅ |
| branch daemon + v0.36.1 TUI | ✅ | ✅ | ✅ |

In both runs the orchestrator pane showed its own shell-out and then the notification typed back into it —

```
bash-5.3$ dot-agent-deck delegate --to worker --task "XVERTASK cross-version probe"
bash-5.3$ Worker worker has completed their task. Read .dot-agent-deck/work-done-worker.md for their full report.
```

— the worker received the delegated task text through its task file, and its card rendered as `ClaudeCode · worker … Idle` from the hook it posted.

## Gates

- `cargo fmt --check` clean; `cargo clippy --workspace --all-targets --features e2e -- -D warnings` clean.
- `cargo test-fast`: 2610 passed.
- `cargo xtask linkage-check`: ok.
- `cargo test-e2e`: 7981 run, 7980 passed, **1 failed — unrelated, and pre-existing.** `scheduler/form/002` (`form_002_add_spawns_authoring_agent_in_picked_dir`) failed on `the mode-locked schedule form must render a [Submit] button` and passes on its own re-run. It is a latent race in the test rather than in the product: `tests/e2e_scheduler_manager.rs:928-931` waits for the string `New Schedule` and then calls `find_in_grid("[Submit]")`, which takes ONE `snapshot_grid()` with no polling, so under full-tier parallel load the frame carrying the title can arrive before the frame carrying the button. It also fails *before* any agent is spawned, which puts it structurally out of reach of a change confined to the post-write confirmation loop. Already tracked as **#395**, which found the same thing by reading the harness rather than from a failure, and scoped it properly: `find_in_grid` has **54 call sites, 51 of which immediately `.expect()`/`.unwrap()` the result**, so this is a latent race at every one of them (the mouse tests are the worst exposure — they feed the coordinates straight into `click()`, where a stale hit can pass for the wrong reason instead of panicking). Not fixed here (unrelated scope). This run is a data point for that issue rather than a new symptom: it matches the corrected model in #395's follow-up comment — contention widens the gap between two paints, so the same test failed once under full-tier parallel load and passed in 1.7s on its own.

## Scope

Only the arming gate. #530 (why the loss was silent — `dispatch --single` exits 0 before delivery is known) is untouched and still worth doing: it would have turned this into a visible failure, where this change makes the dispatch work.


prageethw (approved):
Solid, reproduce-first fix. `scheduler/dispatch/014` (L2) reproduces the exact field-report shape (late `SessionStart` past the readiness gate) with three early-announcing panes as a live control, and `scheduler/dispatch/016` (L1) pins the new standing beside the #424 F4 refusal case it must not weaken — both reported reverted individually to confirm load-bearing.

The security argument for the new `agent_spawned_as_reporting_agent` standing holds up: it reads `RunningAgent::spawn_agent_type`, the frozen launch-shape identity set once at spawn from `SpawnOptions::agent_type`, and confirmed `set_agent_type` (the hook-learned value) never writes that field — so a forged post-write claim still can't manufacture this standing, only a pane the deck itself exec'd as a resolved agent type can. That's the right boundary and keeps F4's refusal intact for bare shells / `cat` / unresolved commands.

Rule 12 addressed correctly (no wire/handler change, patch-level bugfix, cross-version test run in both directions). Gates green apart from one flake (`scheduler/form/002`) that's pre-existing, already tracked as #395, and structurally unrelated (fails before any agent spawns, outside this change's confirmation-loop scope) — reasoning checks out.

PR #586, opened 2026-08-17
fix(dispatch): serialize and retry `git worktree add` under concurrent dispatch (#541)

### LINKED ISSUE #541: git worktree add races under concurrent dispatch (fatal: failed to read commondir: Success)
Split out of #424 (round 5), where it was hit while gating `scheduler/dispatch/015` and is not caused by that PR's changes.

## What

`git worktree add` fails intermittently under concurrent dispatch with:

```
fatal: failed to read '.git/worktrees/<name>/commondir': Success
```

The `Success` in a `fatal:` line is the tell: git is reporting `strerror(errno)` where `errno` was never set, i.e. the read returned short/empty rather than failing. The file is written by the `git worktree add` that creates the entry and read by the next one that scans the worktree list, so a concurrent dispatch (three panes at once in `/015`) can observe a half-created administrative directory.

## Why it matters

It fails a dispatch that has nothing wrong with it, and it fails it in the *setup* step, before any agent runs — so the visible symptom is an e2e test flake with no relation to the code under test. `scheduler/dispatch/015` spawns three real-Claude worktrees concurrently and is the place it shows up.

## Aggravating factor observed

A busy `/var/tmp` made it markedly more likely. When first seen the harness's temp root held **112 stale per-process roots totalling ~8.5 GB** left behind by interrupted e2e runs; `cargo xtask clean-e2e-tmp --apply` cleared them and the race stopped reproducing in that session. That is a correlation, not a mechanism — a slower filesystem widens the window between creating the worktree entry and completing it — but it makes "reap the temp roots first" a reasonable mitigation to try before debugging a `/015` failure.

## Suggested direction

- Serialize worktree creation per repository (the dispatch path already has a per-pane mutex; this needs a per-repo one), or
- retry `git worktree add` once on this specific error after a short backoff, or
- give each concurrent dispatch its own clone rather than sharing one `.git`.

Whichever is chosen, it should be verified by running `/015` repeatedly on a deliberately full `/var/tmp`, since that is the condition under which it reproduces at all.

vfarcic:
Fixes #541.

## The mechanism, pinned

`git worktree add` scans the repository's worktree list **before** creating its own entry, and reads every existing entry's `commondir`. `strace` of git 2.55.0 shows the writer's order inside `$GIT_COMMON_DIR/worktrees/<name>`:

```
mkdir(".git/worktrees/w3")                                  <- entry becomes visible to scanners
openat(".git/worktrees/w3/locked",    O_CREAT|O_TRUNC)
openat(".git/worktrees/w3/gitdir",    O_CREAT|O_TRUNC)
openat(".git/worktrees/w3/commondir", O_CREAT|O_TRUNC)      <- window opens
     write(... "../..")                                      <- window closes, one syscall later
```

A concurrent add that reads `commondir` between those last two calls gets an empty file. Git's reader does `if (strbuf_read_file(...) <= 0) die_errno("failed to read '%s'")`, and `die_errno` appends `strerror(errno)` for an errno **nothing ever set** — which is why the reported line ends in `: Success`. That word is the whole diagnosis: the read did not fail, it came back short.

Confirmed directly by staging an empty `commondir` in one entry and running the neighbours:

```
$ git worktree add ../w2 -b b2
Preparing worktree (new branch 'b2')
fatal: failed to read .git/worktrees/w1/commondir: Success     # exit 128
```

The loser dies in the dispatch's **setup** step, before any agent runs — so the visible symptom is a dispatch failing for reasons unrelated to what it was asked to do.

## Reproduced first, before any fix

**Genuinely, with real concurrent processes.** N concurrent real `git worktree add`s against one repo, on the disk-backed `/var/tmp`:

| concurrency | adds | failures |
|---|---|---|
| 3 (dispatch/015's shape) | 1,320 | 0 |
| 16 | 1,520 | 7 |
| 64 | 5,760 | 14 |
| 128 | 1,024 | 7 |
| **total unserialized** | **9,684** | **28 (0.29%)** |

Every failure carried the reported message verbatim. Failures cluster — five scanners dying on one victim in a single round — which is the signature of a writer descheduled between those two syscalls, and it is why the report's "a busy `/var/tmp` made it markedly more likely" is the right observation: anything that widens the gap widens the race. At N=3 on an idle box it does not reproduce at all, matching the report that this is rare in the field.

**Then deterministically, for the test suite.** ~1,000 real worktree checkouts per observed failure is neither affordable in the fast tier nor a reliable gate, so the tests stage git's *own* intermediate bytes (the `mkdir` + `locked` + `gitdir` + empty `commondir` above — verified to produce the identical `fatal:`) and close the window on a timer. The timing is not a race in the flaky sense: without a retry there is no second attempt for the timer to rescue, so a slow machine cannot turn it green by accident.

## The fix

Two defences in `create_worktree`, each covering what the other cannot:

1. **Serialize per repository, across processes.** An `flock(2)` (named mutex on Windows) via the existing `platform::lock`, on a lock file in `git rev-parse --git-common-dir`. Keyed on the common dir because that is the contended resource, and because `clone_dir` may itself be a linked worktree — asking git means the main checkout, any of its worktrees, and a relative path all land on **one** lock. This binds separate deck **processes**, not merely tasks in one; an in-process mutex would silently not cover concurrent decks.
2. **Retry the one transient signature, bounded.** 4 retries, 100/200/400/800 ms. Serialization cannot help against an add the deck did not start (the user's own, another tool's), because those take no lock — only the retry survives that.

Nothing is swallowed. A `commondir` that stays unreadable — what a crashed add leaves behind — exhausts the attempts and surfaces as `Err`, and every other `git worktree add` failure stays a hard failure with git's own message. The predicate matches on the *file name* rather than git's sentence, because `die_errno`'s format string and `strerror` are both localized while a path component is not. If the lock cannot be taken at all, creation proceeds unserialized rather than wedging the dispatch.

**Second-order defect, found while fixing the first.** An add that dies on the scan has **already created its `-b` branch** (verified: the branch survives the exit-128). A naive retry therefore failed with `a branch named … already exists` — and since a dispatch refuses to reuse an existing branch (it may hold committed work), the dispatch name stayed **poisoned** until the user ran `git branch -D` by hand. The branch is now re-probed on every attempt and attached rather than re-created. Only attempt 1 can report `BranchExists`: reaching attempt 2 means the branch was *proven* absent moments earlier, so anything there now was created by our own failed attempt or by a dispatch racing us — never the "may hold earlier committed work" case that guard exists for.

## Every part proven load-bearing

Reverting one change at a time turns a test red again (skill step 7):

| reverted | result |
|---|---|
| `WORKTREE_ADD_ATTEMPTS = 1` | `Err(… commondir: Success)` — the reported failure |
| branch probe hoisted out of the loop | `fatal: a branch named 'agent/dispatch-alpha' already exists` |
| the lock removed | serialization test fails |

And the serialization was validated at genuine concurrency too — the same N=64 shape, with each add taking an `flock` on the production lock path: **0 failures in 4,800 adds** against 0.29% unserialized.

**One pre-existing misclassification, made reachable by design rather than by luck, so fixed here too** (second commit). `create_worktree` returned `BranchExists` whenever the branch existed and reuse was refused — *including when the worktree directory was still right there*. `dispatch.rs` renders that as "its worktree is already gone" and points the user at `git branch -D`, which for a live claim is false and is destructive advice about a tree another dispatch is working in. Serialization makes the loser of a same-name race always probe the branch **after** the winner created it, where before it might have probed first and been classified correctly. A present worktree dir now reports `AlreadyClaimed`; once the worktree is genuinely gone, a surviving branch still reports `BranchExists`, unchanged and still covered by `second_dispatch_of_a_name_reports_branch_exists_after_cleanup`.

## Tests

Five fast-tier tests in `src/issue_dispatch_run.rs`, extending the existing `create_worktree` unit tests rather than adding a near-duplicate elsewhere:

- `create_worktree_survives_a_concurrent_adds_half_created_entry` — **the reproduction.** Three concurrent dispatches (`scheduler/dispatch/015`'s shape) against a repo holding a half-created entry closed by a *fourth party that holds no deck lock*, so it exercises the case serialization alone cannot fix. Asserts each dispatch's worktree AND its checked-out branch.
- `create_worktree_surfaces_a_half_created_entry_that_never_completes` — **the control.** Same staged state, never repaired: still `Err` naming the file. Bounds the retry (unbounded → this hangs rather than passes) and proves the failure in the test above is attributable to the *transience*, not to the staged state being generally unusable.
- `create_worktree_serializes_per_repository_across_processes` — a held lock makes creation block; releasing it lets it complete. Takes the lock through the same derivation production uses.
- `create_worktree_reports_a_live_claim_not_a_leftover_branch` — the misclassification above; red without the fix (`Ok(BranchExists)` vs `Ok(AlreadyClaimed)`).
- `worktree_scan_short_read_matches_only_the_commondir_signature` — pins the reported message and five genuine failures that must stay hard.

Stable across 15 consecutive runs (~1.6s for the whole group).

## Gates

- `cargo test-fast` — **2615/2615**.
- `cargo test-e2e` — **7985/7985, 0 failures** (full tier, 231s).
- `scheduler/dispatch/014` (deterministic 3-way concurrent dispatch) and `scheduler/dispatch/015` (three real interactive Claude dispatches — where #541 was originally hit) both pass, as do all 87 in `e2e_issue_dispatch`, which cover the real worktree create / close / refire-after-close paths this change sits directly under.

**No TUI↔daemon contract change**, so no `PROTOCOL_VERSION` bump and no `.breaking.md`. The change is confined to `create_worktree`, a daemon-internal function: no wire message, field, or role-map value moved, and no field changed meaning behind a stable wire. An older TUI against a newer daemon observes identical behaviour, so I judged the cross-version sandbox run not applicable here — flagging it explicitly rather than silently, in case you read it differently.

The one new on-disk artifact is `<git-common-dir>/dot-agent-deck-worktree.lock` (empty, never in `git status`). It is not a compatibility contract: an old deck simply does not take the lock while a new one does, which degrades to the retry rather than breaking anything.

## Scope notes

- **`src/spawn.rs` was not touched.** The issue points at it, but its only mention of `git worktree add` is a comment — `create_worktree` in `src/issue_dispatch_run.rs` is the sole production call site (callers: `dispatch.rs:293`, `issue_dispatch_run.rs:400`). This also keeps the diff clear of #573, which is in `spawn.rs`'s prompt-confirmation loop; the two PRs share no file.
- **`git worktree list` has the same reader-side hazard** — `src/worktree_reclaim.rs:274` dies identically on a half-created entry (verified). Left alone as out of scope for this issue; worth its own issue if you want it covered.
- **`git worktree remove` is not serialized.** A concurrent remove is a reader too, so the retry covers the add against it, but add-vs-remove is not mutually excluded. Deliberate scope call, easy follow-up.

## One thing worth knowing

The full `cargo test-e2e` tier initially came back with **59 failures** spread across unrelated areas — mouse, render contract, session restore, scheduler manager — with **every real-Claude test among them**. That was the machine, not the change: it had accumulated 277 stale harness temp roots totalling **6 GB** plus leftover agent processes, the exact aggravating condition #541 itself calls out. `cargo xtask clean-e2e-tmp --apply` reclaimed it, after which the same tier ran **7985/7985 green**, and `dispatch_015` passes with this change in 32.5s against 33.4s on a stashed clean tree.

So the report's "reap the temp roots first" advice generalises well beyond `/015`: on a polluted box the e2e tier fails in ways that look like anything except a full disk. Worth reaching for `clean-e2e-tmp` before debugging any broad e2e redness.


prageethw (approved):
Excellent reproduce-first work. The root-cause diagnosis (git's own `die_errno` printing `strerror` for an errno nothing set — the giveaway `: Success`) is pinned with a real `strace`, then confirmed by staging an empty `commondir` directly, then reproduced with genuine concurrent `git worktree add`s at multiple concurrency levels with measured failure rates. The fix is the right shape — per-repo cross-process `flock` for the deck's own races, bounded retry keyed on the file-name signature (not the localized message) for races the deck didn't start — and every piece is proven load-bearing by individual revert.

The second-order branch-poisoning fix and the `BranchExists`-vs-`AlreadyClaimed` misclassification fix are both real, in-scope improvements found honestly while fixing the primary issue, not scope creep — both come with their own tests and reasoning for why serialization makes the misclassification newly reachable.

Rule 12 handled correctly, including explicitly flagging the judgment call to skip the cross-version sandbox run rather than silently omitting it. Full e2e tier green (7985/7985) after correctly diagnosing the initial 59-failure run as stale harness temp roots rather than the change itself — good instinct to reach for `clean-e2e-tmp` before concluding the change was broken. Scope notes (spawn.rs untouched, `worktree list`'s matching hazard left out of scope, add-vs-remove not mutually excluded) are exactly the kind of honest boundary-drawing that makes a diff easy to trust. All CI checks green.

PR #591, opened 2026-08-18
fix(release): stop assemble-changelog.sh deleting fragments it never validated (#582)

### LINKED ISSUE #582: `assemble-changelog.sh` deletes fragments in `changelog.d/` subdirectories without suffix-checking them
## The inconsistency

Two `find` invocations over `changelog.d/` in `scripts/assemble-changelog.sh` disagree on depth:

- **`:41`** — the unknown-suffix validation loop uses `find "$CHANGELOG_DIR" -maxdepth 1 -name '*.md'`.
- **`:61`** — the collection/processing loop uses `find "$CHANGELOG_DIR" -name "*.$type.md"` **without** `-maxdepth 1`.

## The consequence

A fragment in a subdirectory (`changelog.d/foo/12.bugfix.md`) is **never suffix-checked** by the validation loop, because that loop cannot see it — but it *is* reached by the processing loop, which ends in `rm -f "$f"` (`:114`).

So such a file is deleted at release time having passed through no gate at all. If its suffix were unrecognised, the validation loop that exists specifically to catch that never gets the chance — which is the same failure shape that guard was written to prevent, reachable through a path the guard does not cover.

## Why it has not bitten

Nothing puts fragments in subdirectories today, and `changelog.d/` is flat by convention. So this is latent rather than active — low priority rather than merely low severity.

## Expected

The two loops should agree on depth. Either add `-maxdepth 1` at `:61` so the processing loop matches the validation loop and a subdirectory fragment is ignored rather than silently consumed, or drop it from `:41` so nested fragments are validated too. The first is the smaller change and matches the flat-directory convention; the second is more permissive and would need the rest of the script checked for path assumptions.

Whichever is chosen, a comment saying the two `find` calls are deliberately in lockstep would stop them drifting apart again.

## Why this is filed here

Verified present on `main` at `e6e6dfea`: both `find` invocations are quoted above from upstream `scripts/assemble-changelog.sh`, at the line numbers given, and the `rm -f` at `:114` is upstream too. It surfaced in a fork (`prageethw/dot-agent-deck`, issue #363) during an audit that read the script whole rather than only a diff, and was deliberately left out of that PR's scope. The script is shared and unmodified in the relevant part. Reporting it here rather than only fixing it downstream, so the fix lands where every user of this project benefits.

vfarcic:
Closes #582.

## The defect

`scripts/assemble-changelog.sh` scanned `changelog.d/` twice with two `find` invocations that disagreed on depth:

- the guard that rejects an unrecognized type suffix used `find "$CHANGELOG_DIR" -maxdepth 1 -name '*.md'`;
- the loop that renders fragments into `CHANGELOG.md` and then `rm -f`s them used `find "$CHANGELOG_DIR" -name "*.$type.md"`, unbounded.

So a fragment below a subdirectory was invisible to the guard and visible to the deletion. It could be consumed at release time having passed through no gate at all, and an unrecognized one — `changelog.d/nested/702.fix.md`, where `fix` is not a recognized type and `bugfix`/`fixed` are — was silently dropped while its siblings were rendered and deleted around it. That is precisely the v0.24.3 failure the guard was added to prevent (fragments ignored, empty release body, nobody told), reached by the one path the guard did not cover.

## Reproduced first

`tests/assemble_changelog.rs` was written before the script was touched, and confirmed red for the reported reason: against the unfixed script the run exited **0**, rendered and deleted `700.feature.md` and `nested/701.bugfix.md`, wrote `CHANGELOG.md`, and left `nested/702.fix.md` sitting there unreported.

Two controls ran green both before and after, which is what pins the failure to *depth* rather than to a generally broken guard: the same unrecognized fragment at the **top level** is rejected (it always was), and the **flat happy path** assembles, deletes, and keeps `.gitkeep`.

After the fix, each half was reverted individually and confirmed to turn the reproduction red again — restoring `-maxdepth 1`, and restoring basename-only reporting.

## Which way the two loops converge, and why

The issue offered both directions. **They converge on recursive-with-validation**, not on `-maxdepth 1`.

The invariant that actually matters is *nothing is deleted without having passed the guard* — i.e. validation's coverage must be a superset of collection's. Both directions satisfy that. What separates them is what happens to the nested fragment afterwards:

- **Bounding both to depth 1** stops the deletion, but the nested fragment is then neither validated, nor rendered, nor removed. It is silently ignored — which is the same silent-drop shape the guard exists to prevent, just with the file left on disk instead of deleted. It fixes the deletion and keeps the silence.
- **Recursing in both** leaves no silent path at all: every `*.md` anywhere under `changelog.d/` is either recognized (rendered, then deleted, having passed the gate) or loudly rejected with a non-zero exit. Validation's `*.md` is a strict superset of collection's `*.$type.md` at every depth, so the invariant holds by construction rather than by the two globs happening to line up.

It is also the smaller behavioural claim than it looks: the guard already covered *more* than the deletion at its own depth, and this keeps that relationship while extending it downward. The rest of the script was checked for path assumptions — the processing loop reads `< "$f"`, the deletion uses `"$f"`, and `sort -z` orders full paths, all of which are depth-agnostic. The one real assumption was in the **error message**, which reported `basename "$f"` and would have named a nested offender `702.fix.md` with no indication of which directory to look in; it now reports the path relative to `changelog.d/`, which for the flat layout the repo actually uses is still just the filename.

Deliberately **not** included: a `-type f` predicate on either `find`. It would exclude a symlinked fragment from both loops, which reintroduces a silent-ignore path — the thing this PR is removing. A directory named `foo.bugfix.md` remains a loud failure (the redirect fails under `set -e`), which is acceptable.

Each `find` now carries a comment recording that the two are deliberately in lockstep, so the next edit to one has to consider the other.

## Verification

- `cargo fmt --check` — clean
- `cargo test-fast` — 2727 passed, 0 skipped (4 new)
- Smoke-tested against the repo's **real** `changelog.d/` in a scratch copy: same headings (`Added`/`Fixed`/`Miscellaneous`), every fragment consumed, `.gitkeep` kept, exit 0 — before and after the new `582.misc.md`.

## A pre-existing bash-4 requirement that this PR surfaced but does not fix

The first push turned **`build-macos`** red, with all four new tests failing identically at `assemble-changelog.sh: line 12: added: unbound variable`. Line 12 is the `declare -A TYPE_HEADERS=(` block — it predates this branch, and the #582 diff starts at line 28. macOS ships bash 3.2, which has no associative arrays, so `[added]="Added"` is parsed as an *arithmetic* subscript and `added` is an unset variable under the script's `set -u`. Reproduced on bash 5 by dropping the `declare -A`: identical message, identical variable — the mechanism, not just a correlation with the platform.

So the assembler has **always** required bash 4+; `release.yml` runs it on `ubuntu-latest` and nothing else invokes it, so the limit has never mattered. These tests were simply the first thing ever to execute the script on a macOS runner.

The tests are therefore gated `#![cfg(target_os = "linux")]` rather than `unix` — a test that runs where its subject does not is a test whose red says nothing about the subject's correctness. Making the script bash-3.2-clean is a genuine improvement, but it is a change to the highest-blast-radius script in the repo and does not belong inside a #582 bugfix, so it is filed as **#593** with the proposed `case`-function fix and a request that whoever takes it re-widen this file back to `unix` in the same change — that file is the ready-made verification.

## Review

Greptile's check-run completed with **no inline findings** (confidence 5/5); the inline-comment API was queried directly rather than reading the green check as the review, per CLAUDE.md rule 8.

The four new tests are fast-tier shell-level tests (one `bash` invocation against a scratch directory each, no network, no sleeps, ~0.16s total). They are deliberately **not** `#[spec]` catalog tests: the subject is release tooling with no TUI surface, so CLAUDE.md rules 4 and 7 do not bind. Scratch directories come from `test_temp::tempdir()` per linkage-check check 8.

Per CLAUDE.md rule 12: this touches no daemon, protocol, orchestration, or hook code, so there is no contract change, no `PROTOCOL_VERSION` bump, and no cross-version run. Release tooling only — the shipped binary is unchanged.


prageethw (approved):
Traced the full script: the two find invocations now agree on recursion depth, and the deletion loop runs strictly after validation completes with no path that deletes before validating. Four new regression tests cover the reported nested-unknown case, a depth control, the flat happy path, and nested-but-recognized fragments, with a documented RED/GREEN methodology. No prior review activity to reconcile, CI green.

PR #592, opened 2026-08-18
fix(ui): cut a producer-controlled session_id on a char boundary (#574)

### LINKED ISSUE #574: Byte-index slice on a producer-controlled `session_id` panics the whole TUI
## Summary

`render_session_card` byte-slices a producer-controlled `session_id`, so a session id whose 11th **byte** is not a UTF-8 char boundary panics the render loop — killing the entire TUI, not just that card.

```rust
// src/ui.rs:17634
&session.session_id[..11]
```

`str` indexing is by byte and panics on a non-boundary index. `session_id` reaches this line with no validation: `apply_event` keys the session map on the event's `session_id` verbatim, and there is no length, charset, or boundary check anywhere between the hook socket and the render.

## Impact

A session id of nine `α` characters (2 bytes each, so byte 11 falls mid-character) panics `render_session_card`. Because this is inside the render loop, **every frame panics and the whole deck dies**, taking every other agent's card with it.

The hook socket is owner-only (0600), so the threat set is same-uid processes — which on this product means every agent running on the deck. One agent can terminate the operator's entire TUI with a single hook post. It is equally reachable by accident: any agent that generates a non-ASCII session id trips it without malice.

## Second site, same shape

`src/ui.rs:12742` — `format!("{}...", &url[..57])` on a URL taken from the embedded pane's OSC-8 hyperlink map, i.e. text an agent wrote into its own PTY. Reached when the operator clicks the link.

(The adjacent `open::that(&url)` on that same agent-controlled URL is a separate question and is not part of this report.)

## Expected

Truncation on a char boundary, never a panic. `crate::prompt_delivery::truncate_on_char_boundary` already exists in this codebase and is already used for exactly this purpose in `src/hook.rs`. Both sites should use it.

## Suggested tests

- A `SessionState` whose `session_id` is 9 multi-byte characters renders without panicking.
- The same for the `src/ui.rs:12742` URL path.
- Ideally a property/fuzz-style case over arbitrary Unicode session ids through the card render.

## Why this is filed here

The defect is in shared code that this repository owns, verified present on `main` at `e6e6dfea` at the line cited above — it is not a downstream modification. It was found in a fork (`prageethw/dot-agent-deck`, issue #389) during a security audit of an unrelated change, but nothing about it is fork-specific: the slice, the missing validation, and the render-loop blast radius are all upstream code. Reporting it here rather than only fixing it downstream, so the fix lands where every user of this project benefits.

Worth recording how close it stayed to invisible in the fork: an e2e test there explicitly works *around* the same 11-char limit, with a comment about **truncation** and no awareness of the panic.

vfarcic:
Closes #574.

## The defect

`render_session_card` shortened a long `session_id` with `&session.session_id[..11]` — a **byte** index. `str` indexing panics when the index is not a UTF-8 char boundary, so a session id of nine `α` characters (2 bytes each; byte 11 lands mid-character) panicked the render. Because the slice runs **inside the render loop**, every frame panicked the same way and the whole deck died, taking every other agent's card with it.

Nothing validates the id on the way in: `AppState::apply_event` keys the session map on whatever the hook socket sent, with no length, charset or boundary check between the socket and the render. The socket is owner-only (0600), so the threat set is same-uid processes — which on this product is every agent running on the deck. It is equally reachable by accident from any agent that mints a non-ASCII id.

Second site, same shape: `format!("{}...", &url[..57])` on the Ctrl+click `Opened: …` status line, where the URL comes from the embedded pane's OSC-8 hyperlink map — text an agent wrote into its own PTY.

## The fix

Both sites now go through `prompt_delivery::truncate_on_char_boundary`, which `src/hook.rs` already uses on prompts for exactly this reason.

The set of strings that get shortened **at all** is unchanged — the `> 11` and `> 60` gates are preserved, so short ids and URLs are still reported verbatim. Only the cut itself moves to a character boundary, and it is now marked with `…` (replacing the URL path's literal `...`).

The URL arm is extracted into `opened_link_status` so the test drives the same function the live click path calls rather than an agreeing copy — the pattern `scroll_focused_agent_pane` already uses for the mouse arms.

**Swept the rest of `src/`**: no other byte-index slice or `String::truncate` is boundary-unsafe. `orchestrator_ext.rs:87` and `login_shell.rs:176` are `find`-derived; `state.rs:618` is ASCII by construction; `daemon_client.rs:225` and `ui.rs:5456` already snap back to a boundary.

## Reproduced first

Per `/reproduce-first`, the failing tests came before the fix. Both failed with the reported panic at the reported site:

```
pane_011      -> panicked at src/ui.rs:17791: end byte index 11 is not a char boundary; it is inside 'α'
hyperlink_001 -> panicked at src/ui.rs:5409:  end byte index 57 is not a char boundary; it is inside 'α'
```

Each fix was then reverted **individually** and the corresponding test went red again while the other stayed green, so neither change is decorative.

## Tests (rule 4 — L1)

- **`dashboard/pane/011`** — replays a hook event carrying a non-ASCII `session_id` through the real `AppState::apply_event`, then renders the resulting card **between two healthy ones**. The blast radius is the point, so the assertion is that all three cards survive with their titles, not just that the poisoned card draws.
- **`mouse/hyperlink/001`** — `opened_link_status` on the URLs an agent can emit.

Both sweep 2-, 3- and 4-byte characters (`α`, `あ`, `𝄞`, `😀`) at every ASCII offset that moves the cut inside a character — the property/fuzz-style coverage the issue asked for. Expected values are computed by walking `chars()`, a deliberately different algorithm from the production truncator, so the test states what a boundary cut *is* rather than restating the code under test.

## Rule 12

No TUI↔daemon protocol or handler-contract change — this is a render site and a status string. No `PROTOCOL_VERSION` bump and no `.breaking.md`; `changelog.d/574.bugfix.md` added.

## Gates

- `cargo fmt --check` ✅
- `cargo xtask linkage-check` ✅ (535 catalog ids, 463 annotations)
- `cargo test-fast` ✅ **2723 passed, 0 skipped**

### DIFF
--- a/changelog.d/574.bugfix.md
+++ b/changelog.d/574.bugfix.md
@@ -1,0 +1,7 @@
+## One agent's session id can no longer take the whole deck down
+
+A session id containing non-ASCII characters killed the entire TUI. The dashboard shortens a long id to fit the card title, and it did so by cutting at the 11th *byte* — which in Rust panics outright when that byte lands in the middle of a character. Nine Greek letters were enough. Because the cut happened inside the render loop, every subsequent frame panicked the same way, so the deck died taking every other agent's card with it, not just the one card that could not be drawn.
+
+Nothing validated the id on the way in: the deck stores whatever a hook posts to its socket verbatim. That socket is owner-only, so this was never a route in from outside the machine — but on this product every agent running on the deck is a same-uid process, which made it one agent's ability to terminate the operator's whole session with a single hook post. It was equally reachable by accident, from any agent that happened to mint a non-ASCII id.
+
+The same shape sat on the "Opened: …" status line that a Ctrl+click on an agent-emitted OSC-8 hyperlink leaves behind, where the URL — text the agent wrote into its own PTY — was cut at byte 57. Both now cut on a character boundary and mark the cut with an `…`, using the truncation the hook binary has always used on prompts. Ids and URLs short enough to fit are untouched.
--- a/src/ui.rs
+++ b/src/ui.rs
@@ -5398,6 +5398,26 @@ fn screen_row_offset(screen: &vt100::Screen, pane_rect: Rect) -> u16 {
     effective_rows.saturating_sub(inner_h) as u16
 }
 
+/// The status line a Ctrl+click on an OSC-8 hyperlink leaves behind, with the
+/// URL shortened to fit the one-row status area.
+///
+/// Extracted out of the mouse arm so the L1 seam (`mouse/hyperlink/001`)
+/// exercises the same code the live click path runs rather than an agreeing
+/// copy of it.
+pub fn opened_link_status(url: &str) -> String {
+    // Issue #574: the URL is whatever the agent wrote into its own PTY as an
+    // OSC-8 hyperlink, so byte 57 is as arbitrary as the string — `&url[..57]`
+    // panicked the event loop the moment it landed inside a character. The
+    // `> 60` gate is kept so the set of URLs that get shortened at all is
+    // unchanged; only the cut itself moves to a character boundary.
+    let display = if url.len() > 60 {
+        crate::prompt_delivery::truncate_on_char_boundary(url, 57)
+    } else {
+        url.to_string()
+    };
+    format!("Opened: {display}")
+}
+
 /// Extract text from a vt100 screen for the given selection region.
 /// Selection coordinates are widget-relative; `row_offset` maps them to screen rows.
 fn extract_selection_text(screen: &vt100::Screen, sel: &TextSelection, row_offset: u16) -> String {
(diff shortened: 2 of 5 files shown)


prageethw (approved):
Both byte-slice sites now route through the existing, correct truncate_on_char_boundary helper — verified the boundary-scan logic handles the zero-boundary edge case without panicking. Two regression tests (dashboard/pane/011, mouse/hyperlink/001) sweep 2/3/4-byte chars at every offset that lands mid-character, with independently-computed expected values and confirmed red-before/green-after. One non-blocking note for the author: the PR body claims a sweep found no other unsafe byte-slice in src/, citing five specific file:line locations (orchestrator_ext.rs:87, login_shell.rs:176, state.rs:618, daemon_client.rs:225, ui.rs:5456) — worth double-checking those five directly before relying on the sweep being exhaustive, though it doesn't block this fix.

PR #594, opened 2026-08-18
fix(config): find the feature-flag config by walking up to the project root

### LINKED ISSUE #577: `features_config_path()` resolves against the process cwd, so a deck launched outside its project dir reads feature flags from the wrong file
## Summary

`features_config_path()` resolves the feature-flag config against the **process's own working directory**, while every other config read in the deck resolves against an **explicit project directory**. A deck launched from anywhere other than its project directory therefore reads its `[features]` table from the wrong file — usually a file that does not exist — and every experimental surface silently resolves off, no matter what the project's `.dot-agent-deck.toml` says.

This is not theoretical. Downstream it was found because it disabled a feature on a running deck, and the symptom was indistinguishable from the feature having been removed.

## The asymmetry

```rust
// src/config.rs:1179
pub fn features_config_path() -> PathBuf {
    if let Ok(p) = std::env::var("DOT_AGENT_DECK_FEATURES_CONFIG") {
        return PathBuf::from(p);
    }
    std::env::current_dir()
        .unwrap_or_else(|_| PathBuf::from("."))
        .join(crate::project_config::CONFIG_FILE_NAME)
}
```

Every other consumer goes through `load_project_config(dir)` with a directory it was handed — a session cwd, a surface cwd, or a canonicalised project path. The feature flag is the only config read keyed to the process's own cwd.

The doc comment above the function states the assumption that breaks: *"the file in the current working directory, so the TUI and daemon (launched in the same dir) read the same source of truth."* That holds for TUI-vs-daemon agreement and fails the moment the deck is launched from anywhere but the project dir. Since the deck is explicitly a multi-project tool — per-pane cwd, per-session project config — a single process-global cwd-derived flag path is structurally the wrong shape for it.

## Reproduction

1. Put `[features]\nexperimental = true` in a project's `.dot-agent-deck.toml`.
2. Launch the deck from a directory that is not that project (e.g. `$HOME`), pointing it at the project.
3. Every experimental surface resolves off. No warning is printed; the config file that *was* read does not exist.

## Expected

The feature-flag table should resolve from the same explicit project directory every other config read uses, so a deck's features do not depend on which directory the operator happened to launch it from.

## Why this is filed here

Verified present on `main` at `e6e6dfea` — the function is quoted above verbatim from upstream. It was found in a fork (`prageethw/dot-agent-deck`, issue #303) on a live deck, but the defective resolution is entirely upstream code. Note the fork has since added an ancestor-walk variant with its own trust checks; that hardening is downstream-only and is **not** what is being reported here. What is reported is the upstream form above, where the cwd join is unconditional. Reporting it here rather than only fixing it downstream, so the fix lands where every user of this project benefits.

vfarcic:
Fixes the directory-dependence half of #577. **Deliberately does not close it** — see [Residual](#residual-577-stays-open).

> **Rev 2**, after @prageethw's review. He was right: rev 1 made the resolution explicit but left `launch_project_dir()` as bare `current_dir()`, so the reported symptom reproduced unchanged and only the silence was cured, while the title and `Closes #577` claimed more than that. This revision makes the behaviour match the claim, and the claim smaller than #577.

## What changed

`resolve_project_dir(start)` walks up from the launch directory to the nearest ancestor holding a trusted `.dot-agent-deck.toml`, and that directory is the project. Launching from `repo/src`, `repo/crates/app`, or anywhere else inside the project now finds `repo`'s flags — the flag depends on which *project* you are in rather than on which of its directories you were standing in. Nearest config wins, so a nested crate or worktree overrides the one above it.

With nothing at or above the launch directory the result is that directory unchanged — the pre-#577 path byte for byte, so a deck launched outside any project reads exactly what it read before.

## The trust check

Walking upward reads directories the operator never named and may not own. `/tmp/project` sits under a world-writable `/tmp` where any local user can drop a `/tmp/.dot-agent-deck.toml`, so a candidate is accepted only as a **regular file owned by the current uid**; anything else is skipped and the walk continues past it. An attacker cannot create a file owned by us, which is why ownership is the check that matters.

The decision is split into the pure `config_owner_is_trusted(file_uid, our_uid)`, mirroring `fsperm::endpoint_owner_is_trusted`, so the deletion-unsafe direction is unit-testable without root. Windows has no uid and the `fsperm` counterpart takes a kernel `HANDLE` rather than a path; the walk there climbs an ACL-protected user profile before reaching an admin-writable `C:\`, so it accepts any regular file and the divergence is recorded at the check rather than left to be inferred from the `cfg`.

## Reproduced first — at the altitude you asked for

Your third point was that rev 1's test only exercised a `project_dir != cwd` combination that occurs at no real call site. **`features/gating/004`** is an L2 PTY test driving the *real* binary from `nested/deep/` inside a fixture whose project-root config carries `[features] experimental = true`, with **no env var**, waiting for the `experimental: on` footer on the rendered grid.

Authored RED:

```
did not see "experimental: on" within 10s.
```

with a **control** launch at the project root that was **green before the fix** — so the failure is attributable to where the deck was started, not to the fixture or the footer. Both halves then proved load-bearing: reverting only the `launch_project_dir` wiring turns it red again, reverting only the resolver turns the unit tests red.

Plus five fast-tier units: nearest ancestor found; nearest wins over a higher one; fallback to the start directory; a *directory* named `.dot-agent-deck.toml` is stepped over; and the trust rule.

`with_launch_subdir` is a new opt-in on the L2 harness, defaulting to `None`, so every existing test launches at the fixture root exactly as before.

## Verified against the real binary

Isolated sandbox (own `HOME`, `STATE_DIR`, both sockets, `DOT_AGENT_DECK_LOG` — rule 12), real `daemon serve`:

| Launched in | Result |
|---|---|
| project root | `ON (from <sb>/proj/.dot-agent-deck.toml)` |
| **3 levels below the root** — the #577 case | `ON (from <sb>/proj/.dot-agent-deck.toml)` |
| outside any project | `OFF (from <sb>/nowhere/deep/.dot-agent-deck.toml)` — the preserved fallback |
| below a **root-owned** planted config in world-writable `/var/tmp` | `OFF` — skipped, not adopted |

The last row used a genuinely `uid 0`-owned file, not a simulation.

## Residual — #577 stays open

A deck launched where **no project sits above it** — `$HOME`, say — with its panes pointed into a project elsewhere still does not adopt that project's flags. The walk cannot help there, and the honest reason is that the `[features]` table is one process-global toggle (CLAUDE.md rule 9 / PRD #139: "exactly one experimental toggle") while some gated surfaces — the dashboard footer among them — belong to no project at all, so there is no per-pane answer to give. That is a design question, not a bug fix, so this PR carries no `Closes` keyword and #577 stays open for it. `DOT_AGENT_DECK_FEATURES_CONFIG` names one file outright and is the escape hatch meanwhile.

Stated in the same terms in the changelog fragment, in `docs/develop/experimental-flag.md`, and in the catalog entry's **Does not assert**.

## Also in rev 1 (unchanged)

`features_config_path` takes an explicit `project_dir` instead of reaching for the process cwd, so the resolution is the caller's decision; the startup line names the file the value came from — `experimental flag: ON (from …)` — because #577's real sting was that nothing distinguished "read the wrong file" from "set to false". Both overrides untouched.

No protocol change, no bump. `PROTOCOL_VERSION`, `daemon_protocol.rs`, the handlers, hooks and orchestration are untouched; the daemon-side edit is which directory a *process-local* config read starts from. Filed as `changelog.d/577.bugfix.md`.

## Gates

- `cargo fmt --check` — clean
- `cargo test-fast` — **2727 passed, 0 failed**
- `cargo xtask linkage-check` — ok (534 catalog ids)
- `cargo test-e2e --test e2e_experimental_flag` — 83 passed (both `gating/003` and the new `gating/004`)

The full e2e tier was not run — the task was scoped to the fast tier and that suite bills real LLM APIs. This change spawns no agents; the L2 tests that cover its exact surface are green.


prageethw (changes requested):
The refactor itself is clean and well-tested at the unit level (features_config_path now takes an explicit project_dir, init_and_watch threads it through, and the new log line names the source file). Confirmed there's no set_current_dir anywhere in main.rs and no other callers of these functions, so the change is behavior-preserving at both real call sites -- launch_project_dir() still just resolves to current_dir().

That's the actual problem: this PR's own manual-verification table shows the reported symptom from #577 -- "a deck launched outside its project dir reads feature flags from the wrong file... every experimental surface silently resolves off" -- still reproduces identically after this fix. The "elsewhere/ (no config)" row still resolves OFF; only the silence is cured (there's now a log line naming which file was read). Issue #577's own "Expected" section asks that "a deck's features do not depend on which directory the operator happened to launch it from" -- that expectation isn't met, since launch_project_dir() is still bare current_dir() with no other resolution strategy.

The title ("resolve... against an explicit project dir") and the `Closes #577.` keyword both claim full resolution, but per the PR's own scope note this is honestly an observability improvement, not a directory-independence fix. Requesting changes on the framing, not the code: either (a) retitle/rescope as an observability fix, drop the `Closes #577.` keyword, and leave #577 open since the underlying symptom is unchanged, or (b) if directory-independence is actually intended, resolve project_dir some way other than bare current_dir() at the entry points (e.g. walking up for a .dot-agent-deck.toml, or documenting the launch-directory contract explicitly as the fix). The new unit test also only exercises a project_dir != cwd combination that never occurs at either real call site, so it doesn't pin an end-to-end regression against the reported bug.

vfarcic:
@prageethw — you were right on all three, and I've pushed `c85296e` rather than argue any of them. Taking your option (b), and part of (a) as well.

**1. "the reported symptom still reproduces; only the silence is cured."** Correct, and my own verification table was the evidence. `launch_project_dir()` now walks up from the launch directory to the nearest ancestor holding a trusted `.dot-agent-deck.toml`, so a deck started at `repo/src` or `repo/crates/app` resolves `repo`'s flags. Nearest wins, so a nested crate or worktree overrides the one above it. With nothing at or above the launch directory the result is that directory unchanged — the pre-#577 path byte for byte.

Since walking up reads directories the operator never named and may not own, a candidate is accepted only as a regular file owned by the current uid; `/tmp/project` under a world-writable `/tmp` is the case that motivates it. The rule is split into a pure `config_owner_is_trusted`, the same shape as `fsperm::endpoint_owner_is_trusted`, so the unsafe direction is testable without root. I verified it end to end with an actual `uid 0`-owned config planted in a world-writable ancestor: skipped, not adopted.

**2. "the title and `Closes #577` claim full resolution."** Retitled, and the `Closes` keyword is gone — **#577 stays open**. The walk does not fix your literal repro: launched from `$HOME` with panes pointed into a project elsewhere, there is no config above the launch directory to find. I don't think that one is reachable as a bug fix, and I'd rather say so than quietly widen the walk until it looks fixed. The `[features]` table is one process-global toggle (rule 9 / PRD #139), and the dashboard footer belongs to no project at all — so a per-pane answer doesn't exist for every gated surface. That's a design question and it deserves the issue staying open. It's written down as the residual in the PR body, the changelog fragment, the docs, and the catalog entry's **Does not assert**.

**3. "the unit test exercises a combination that never occurs at either real call site."** The fair hit, and the one I should have caught. There's now `features/gating/004`: an L2 PTY test driving the real binary from `nested/deep/` inside a fixture whose project-root config sets `[features] experimental = true`, with no env var, asserting the `experimental: on` footer on the rendered grid. Authored RED (`did not see "experimental: on" within 10s`), with a control launch at the project root that was **green before the fix**, so the failure is attributable to where the deck was started rather than to the fixture or the footer. Reverting only the `launch_project_dir` wiring afterwards turns it red again; reverting only the resolver turns the units red. Five fast-tier units cover the walk's own edges (nearest-wins, fallback, a *directory* named `.dot-agent-deck.toml`, the trust rule).

`with_launch_subdir` is a new opt-in on the L2 harness, defaulting to `None`, so every existing test launches at the fixture root exactly as before.

Gates: `fmt` clean, `clippy --workspace --all-targets --features e2e -D warnings` clean, `test-fast` 2727 passed, `linkage-check` ok, `test-e2e --test e2e_experimental_flag` 83 passed. Full e2e tier not run — scoped to the fast tier and it bills real LLM APIs; this change spawns no agents.

Re-requesting review. Thanks for pushing back on the framing — the first version would have closed your issue without fixing it.

prageethw (approved):
Rev 2 addresses my prior request in full: the resolver now genuinely walks up to the nearest ancestor with a trusted `.dot-agent-deck.toml` instead of only adding a log line, with a proper uid-ownership trust check (unit-tested), a real L2 PTY reproduction (authored RED, verified GREEN, both halves load-bearing on revert), and honest scope framing — no `Closes` keyword since the whole-home-directory case is correctly left open as a design question. Checked `src/config.rs`/`src/main.rs` directly; the walk is lexical (no symlink re-resolution surprise) and the fallback to the pre-#577 path when nothing is found is byte-identical. CI green.

PR #596, opened 2026-08-18
fix(ui): escape control characters in the exit session-warning flush

### LINKED ISSUE #576: The `session_warnings` exit flush writes daemon-supplied text raw to the terminal
## The surface

`ui.session_warnings` is flushed by an `eprintln!` loop that runs **after `ratatui::restore()`** (`src/ui.rs:13067`), so it writes straight to a real terminal with no widget layer in between.

The in-session TUI sink is affirmatively safe: `ratatui-core` filters `!symbol.contains(char::is_control)` in both `Buffer::set_stringn` and `Span::styled_graphemes`, so control characters in a rendered string cannot reach the terminal that way. **The exit flush has no such protection.**

Several push sites interpolate values the deck does not control — daemon-supplied `display_name` / `agent_id` reaching the loop via a saved pane's name, plus the saved pane's directory (e.g. `src/ui.rs:11538`, `:13042`, `:13046`). A value carrying ANSI escapes, `\r` or `\n` can corrupt the post-restore output or forge additional lines.

## Expected

Control characters in a warning should be escaped or stripped before they reach the terminal, exactly as they are on the in-session path.

## Why fix it at the flush loop rather than per site

There are ~13 push sites and they only grow — the most recent one in the fork was added by following the established idiom exactly, which is the correct thing for a contributor to do. Sanitising at each producer means every future site must remember; sanitising **once at the flush loop** fixes all of them by construction and cannot be forgotten by a later addition.

Worth deciding explicitly whether to strip or escape: escaping preserves evidence that something odd was in the name, stripping produces cleaner output. Escaping seems better for a diagnostic surface. This codebase already reasons about this class elsewhere — see the `\uXXXX` escaping in `tests/common/mod.rs` (`toml_escape`).

## Severity

Low. It requires an attacker-influenced orchestration or pane name, and the payload only lands when the user quits the deck. Filed so it is fixed once rather than argued at each new push site.

## Why this is filed here

Verified present on `main` at `e6e6dfea`: the push sites and the post-`restore()` flush loop are all upstream code, unmodified downstream. It was raised by an auditor on a fork PR (`prageethw/dot-agent-deck`, issue #319) as pre-existing scope that PR should not have to carry. Nothing about it is fork-specific. Reporting it here rather than only fixing it downstream, so the fix lands where every user of this project benefits.

vfarcic:
Closes #576.

## The defect

`ui.session_warnings` is flushed by a loop that runs **after `ratatui::restore()`**, so it writes straight to a real terminal with no widget layer in between. The in-session sink is affirmatively safe — `ratatui-core` filters `!symbol.contains(char::is_control)` in both `Buffer::set_stringn` and `Span::styled_graphemes` — but the exit flush printed each warning verbatim.

Several push sites interpolate values the deck does not control: a daemon-supplied `display_name` / `agent_id` arriving via a saved pane's name, and the saved pane's `dir`. A value carrying ANSI escapes could repaint the shell the user is dropped back into, a `\r` could overwrite the line just printed, and a `\n` could forge an additional line that reads as the deck's own output.

## The fix — one sanitisation point, at the flush loop

Fixed **at the flush loop, not per push site**, which is what the issue argues for and the reason is load-bearing: there are ~13 push sites and they only grow — the most recent was added by following the established idiom exactly, which is the right thing for a contributor to do. Per-producer escaping means every future site must remember; one point at the single consumer fixes them all by construction and cannot be forgotten by a later addition. Push sites keep pushing plain, readable strings.

- `flush_session_warnings(&[String], &mut impl Write)` writes one line per warning through `escape_control_chars`.
- The predicate is deliberately the **same `char::is_control`** ratatui-core filters on, so the exit path is safe on exactly the terms the in-session path already is (Unicode `Cc`: C0, DEL, C1).
- **Escaped, not stripped** — the issue asks the question explicitly. This is a diagnostic surface, and a name that silently loses characters reads as a *different* name; `\r` in the output preserves the evidence. Spelling is Rust's own: `\n`/`\r`/`\t`, `\u{XX}` for the rest.
- A literal backslash is deliberately **not** escaped: nothing here needs an injective encoding (a human reads this, nothing parses it back), and doubling would cost real legibility on the push sites that interpolate a saved pane's `dir` — every Windows path would come back as `C:\\Users\\…`.
- Clean warnings — the overwhelming majority — borrow rather than allocate.
- Write errors are dropped instead of panicking the way `eprintln!` does; this is the last statement of a clean shutdown.

## Reproduced first

`session/restore/015` (L2, real binary on a PTY) stages a `session.toml` whose single saved pane points at a directory that does not exist — the "skipping pane … directory … not found" branch, which interpolates the saved pane NAME — with an ESC, a CR and an LF in that name, each bracketed by a unique sentinel. It detach-quits, drains the process to exit, and asserts on the final byte stream.

On the old code it reported **all three** control characters reaching the terminal:

```
  ESC (staged after "DAD576-ESC>") reached the terminal as '\u{1b}' …
  CR  (staged after "DAD576-CR>")  reached the terminal as '\r' …
  LF  (staged after "DAD576-LF>")  reached the terminal as '\r' …
```

It passes on the new code. Two details worth reviewing:

- The checks are "the character right after each sentinel is not a control", **not** `contains("MARK\n")` — the PTY's ONLCR rewrites `\n` to `\r\n` on the wire, so a literal-LF needle can never match and that assertion would have been silently unfalsifiable.
- A `stream.contains(ESC_MARK)` guard first proves the warning was flushed at all, so the negative assertions cannot pass vacuously.

Five unit tests in `src/ui.rs` pin the escape spelling, the C1/DEL coverage, the Windows-path behaviour and the one-line-per-warning framing in the fast tier.

## Gates

- `cargo fmt --check` — clean
- `cargo test-fast` — 2726 passed
- `cargo xtask linkage-check` — ok (534 catalog ids, 8 rules)
- `cargo nextest run --workspace --features e2e -E 'test(/^restore_0/)'` — 12/12, i.e. the new test plus every neighbouring session-restore test (including the two that already assert on flushed warnings). The full e2e tier was **not** run — it spawns real agents against LLM APIs and is the pre-release gate.

No daemon / protocol / orchestration / hook surface is touched, so no `PROTOCOL_VERSION` bump and no cross-version run (CLAUDE.md rule 12).


vfarcic:
_Automated deep verification via `/verify-pr` (local worktree at the `origin/main` merge, full gate suite incl. the e2e tier). **Advisory only — this is not an approval**, and per CLAUDE.md rule 8 the PR still needs a second maintainer._

---

# PR #596 — fix(ui): escape control characters in the exit session-warning flush

**Verdict: MERGE**

The fix sits exactly where issue #576 asked for it — at the single post-`ratatui::restore()` consumer, not at the producers — and it works: I reverted it locally and the PR's own e2e test went red, naming all three control characters, then went green again when the fix was restored. Every gate that could run locally passed; the only two e2e failures are real-Claude prompt-adherence flakes in areas this PR does not touch, and both pass on isolation rerun (rule 6). Greptile left no inline findings, so nothing is outstanding there.

- Author: vfarcic (OWNER) · Head: `eb6b33e` · Merge result: **clean** (merged `origin/main` @ `2ad70ca` into the review worktree; 10 commits had landed on `main` since the branch point, no conflicts)
- Report of a local run at `6427077` (merge commit) on Linux/x86_64, rustc 1.97.1, nextest 0.9.143
- PR-only diff vs `origin/main`: 4 files, +352/−7

## Checks

| Step | Result | Time | Note |
|---|---|---|---|
| fmt | **PASS** | 3s | rule 2 |
| clippy | **PASS** | 39s | rule 2, `--workspace --all-targets --features e2e -D warnings` |
| build --release | **PASS** | 67s | |
| test-fast | **PASS** | 88s | rule 5 fast tier |
| linkage-check | **PASS** | 7s | rule 7 — `/// Scenario:` present on the new `#[spec]` |
| windows-cross | **PASS** | 10s | type-check only; CI owns real Windows |
| audit | **SKIPPED** | — | `cargo-audit` not installed locally. **CI's `security` job passed**, so this gap is CI-covered |
| e2e | **FAIL → attributed** | 338s | 8558/8560 passed. 2 failures, both flakes — see below |
| e2e-real-coverage | **ATTENTION** | — | 4 real-agent tests printed `SKIP` and still counted as passed. All four are Codex-auth skips on this host (`gpt-5.1-codex-mini` unreachable). **None covers this PR's surface** — the surface here is covered by `restore_015`, which genuinely ran |

### The two e2e failures are not this PR's

| Test | In-suite | Isolation rerun (rule 6) |
|---|---|---|
| `dispatch_015_three_real_claude_seeds_are_genuinely_confirmed` | FAIL 155s | **PASS 41.0s** |
| `shell_activity_006_real_claude_bash_call_crossing_the_cap_keeps_the_badge_working` | FAIL 133s | **PASS 128.5s** |

Both are real-Claude tests whose failure text is self-diagnosing — `shell_activity_006` printed *"PRECONDITION NOT MET: no ToolStart/Bash event observed … the model never issued the Bash tool call at all. This is a harness flake (prompt adherence), NOT a badge regression: rerun rather than reading this as BADGE WRONG."* `dispatch_015` confirmed 2 of 3 seeds and timed out on the third. Both live in prompt-delivery / shell-activity-badge code, which this PR does not touch (its only production change is 12 executable lines in `run_tui`'s teardown). Attributed as flakes, consistent with rule 5's flaky-tolerant tier and the resource contention of an 8560-test run.

## CI

All 10 check-runs **green on the head SHA `eb6b33e`**: `build`, `build-macos`, `build-windows`, `security`, `nix`, `devbox`, `windows-cross-check`, `changes`, `label`, `Greptile Review`. The three the local run cannot replace — **`build-macos`**, **`build-windows`** (real build + clippy + `cargo nextest run` on the real OS) and **`security`** (`cargo audit`) — all passed, so those are verified by CI rather than unverified.

`reviewDecision: REVIEW_REQUIRED`, 0 reviews. Per CLAUDE.md rule 8 this still needs @prageethw's approval; this report is advice, not an approval.

## Blocking findings

**None.**

## Verification of the four points the fix had to get right

**1. Sanitisation is at the single flush point, and push sites cannot bypass it. ✅**

`src/ui.rs:13213` is now one call: `flush_session_warnings(&ui.session_warnings, &mut std::io::stderr().lock())`. Grepped the whole tree: `session_warnings` has **15** push sites (`src/ui.rs:10123, 11203, 11349, 11358, 11377, 11383, 11389, 11446, 11459, 11558, 11618, 11628, 11684, 13188, 13192`) and **exactly one** consumer. The field is module-private (`src/ui.rs:1980`, no `pub`), it is never read by any render path, and after this change the flush is the only producer-fed writer to the post-restore terminal in `ui.rs`. A new push site is sanitised by construction because push sites do not print. The PR also corrected its own commit message's site count from "thirteen" to "fifteen" (`eb6b33e`) — the count in the doc comment is accurate.

*Residual (non-blocking, and the only architectural one):* the guarantee is "one consumer", not type-enforced. Nothing stops a future contributor adding a **second** consumer that prints raw — the invariant is held by a doc comment at the seam, not by the type system. A `SessionWarnings` newtype owning `push` + `flush` would make it unforgeable. Not worth blocking on: the issue asked for the flush-loop fix, this is it, and the doc comment sits exactly where someone adding a second sink would read it.

**2. Full control-character set, not just ANSI introducers. ✅**

The predicate is `char::is_control` — Unicode category `Cc`: C0 `U+0000..=U+001F` (ESC, CR, LF, BS, BEL, NUL …), DEL `U+007F`, and C1 `U+0080..=U+009F` (including 8-bit CSI `U+009B`). Pinned by `escape_control_chars_covers_del_and_c1_but_not_printable_unicode`, which asserts DEL, `U+009B`, NUL and TAB. Not covered: `Cf` bidi overrides (`U+202E`) and homoglyphs — correctly so, since neither is a terminal *control* and the in-session ratatui sink does not filter them either. The CATALOG entry declares that exclusion explicitly rather than leaving it implied.

**3. Consistency with the in-session path. ✅ (same predicate, deliberately different action — and correctly so)**

`ratatui-core` **drops** control characters (`!symbol.contains(char::is_control)`); this **escapes** them. That is the same predicate with a different action, and it is the right call for two reasons. First, issue #576 posed strip-vs-escape explicitly and reasoned that escaping is better on a diagnostic surface — a name that silently loses characters reads as a *different* name. Second, the "same warning rendered two ways" risk does not arise: `session_warnings` is written by the exit flush and nothing else (no render path reads the field), so no warning string ever reaches both sinks. The shared property that matters — *no control character reaches the tty* — is now identical on both paths, and the doc comment names ratatui's predicate as the reason it chose that one.

**4. A genuine regression test that fails on the old code. ✅ — falsified directly**

I reverted `flush_session_warnings`'s body to the pre-fix `writeln!(out, "{warning}")` in the review worktree and ran `restore_015`:

```
FAIL [2.258s] restore_015_flushed_warning_escapes_control_characters_in_pane_name
  issue #576: control characters from the saved pane name reached the real terminal …
    ESC (staged after "DAD576-ESC>") reached the terminal as '\u{1b}'
    CR  (staged after "DAD576-CR>")  reached the terminal as '\r'
    LF  (staged after "DAD576-LF>")  reached the terminal as '\r'
```

and the captured raw PTY stream shows the exploit landing on a real terminal, forged line and all:

```
…\u{1b}[?1049lWarning: skipping pane 'DAD576-ESC>\u{1b}[31mDAD576-CR>\rSPOOFEDDAD576-LF>\r\n
Warning: everything is fine<DAD576-END' — directory …/vanished-project not found\r\n
```

Restoring the fix → `PASS [0.874s]`. So it is a real red-to-green regression test through the real binary on a real PTY, not an assertion about a helper in isolation.

One detail worth calling out as *good* work: `first_control_after()` (`tests/e2e_session_restore.rs:913`) exists because a literal `contains("MARK\n")` needle would be **unfalsifiable** under the PTY's ONLCR. The falsification run proves that concern was real — the log shows the staged LF arriving as `'\r'`. A naively-written test here would have passed against the live bug.

The `+352` also holds up on inspection: ~12 lines of executable production code, ~60 lines of production doc comment, ~95 lines of `src/ui.rs` unit tests, ~176 lines of e2e test plus the TOML-escape widening the new test forced, 7 CATALOG, 7 changelog. No sprawl.

**5. Rule 4 TUI coverage. ✅**

L2 PTY test `session/restore/015` drives the real binary through a PTY against a hand-staged snapshot, plus five L1 unit tests in `src/ui.rs`. Rule 7's `/// Scenario:` comment is present and actually describes the test (linkage-check green). Decision 17 naming (`restore_015_…`) means rule 6's single-test filter picks exactly one — I relied on that throughout. Correctly **not** marked `[reel]`, since no real agent spawns. Rule 4's real-agent bar targets major user-facing *features*; this is a bugfix on an exit path with no agent in the loop, so a real-agent test would prove nothing extra here.

### Phase 4 — I also saw it as a user does

Ran the release binary in an isolated sandbox (own `HOME`, sockets, lock dir, state dir, and `DOT_AGENT_DECK_LOG` per rule 12's isolation lesson) with a staged `session.toml` whose pane name carried ESC/CR/LF, then Ctrl+C → Enter (Detach). Hexdump of the real stderr:

```
5c 75 7b 31 62 7d  ->  \u{1b}     (escaped, not raw 1b)
5c 72              ->  \r         (escaped)
5c 6e              ->  \n         (escaped)
e2 80 94           ->  em-dash    (printable Unicode preserved)
```

One warning, one line, exactly one framing `0a`. Legible, and the payload text is visibly inert.

## Other rules checked

- **Rule 12** — not triggered and correctly so: no daemon, protocol, orchestration or hook code touched; `run_tui` still returns `std::io::Result<()>`; no `PROTOCOL_VERSION` or `.breaking.md` needed. `bugfix` fragment → patch bump, which matches the policy.
- **Rule 3** — no milestone/PRD prefixes in filenames (test added to the existing `tests/e2e_session_restore.rs`).
- **Rule 9** — no new user-visible surface; experimental-flag question does not apply.
- **Rule 10** — the changelog fragment's prose paragraphs are single lines (82/538/469/441 chars). Not hard-wrapped.
- **Nothing silently removed** — the 7 deleted lines are fully accounted for (the 3-line `eprintln!` loop, 3 doc-comment lines, 1 body line). `#[spec]` annotations: +1/−0. CATALOG entries: +1/−0. Perfectly paired, which is the exact shape that has broken `main` here before.
- **Deps** — untouched; no pin bumps, so the `crossterm` / `portable-pty` pin-comment obligations do not apply.
- **Security** — no `.claude/**`, `.github/**`, `build.rs`, `.cargo/**`, `xtask/**`, `scripts/**` or `devbox.json` changes; nothing logged, no new network. Phase 0 flagged nothing (`READ_DIFF_BEFORE_RUNNING=none`), and no CI runs were held for approval.

## Follow-up

- **`toml_basic_escape` is now a byte-for-byte duplicate.** `tests/e2e_session_restore.rs:47-70` and the harness's `tests/common/mod.rs:3598-3616` `toml_escape` have identical bodies; the new copy's own doc says *"Mirrors the harness's private `toml_escape`."* Widening the test-local copy was unavoidable (the staged name has to survive TOML round-tripping, or the test stages nothing and passes vacuously), but two copies can now drift independently. Proposed issue: *"Make `common::toml_escape` public and drop the `e2e_session_restore` copy."* Non-blocking.

## Nits

- `stage_session_snapshot`'s doc comment (`tests/e2e_session_restore.rs:28-33`) still says `dir` "must already exist on disk so the restore path's dir-exists check does not skip them", which `restore_015` deliberately inverts to reach the skip branch. The call site comments the deviation clearly, so this is only the helper's own contract text lagging its callers.

## Already covered by Greptile

**Nothing to carry over.** Per rule 8 I read the inline comments — the surface where P1/P2 findings actually live — via `gh api repos/vfarcic/dot-agent-deck/pulls/596/comments`, and it returned `[]`. There are also **0 PR review objects**. Greptile posted only its summary issue comment (confidence **5/5**, *"The PR appears safe to merge. No blocking failure remains."*), listing two review passes with the last reviewed commit being the current head `eb6b33e`. So the green `Greptile Review` check is, in this case, backed by an empty findings list rather than masking one — but that was checked, not assumed.

## NOT verified

- **The rule 12 cross-version manual test** — not run. Not required: the diff touches no daemon, protocol, orchestration or hook code.
- **4 skipped real-agent e2e tests** (Codex auth on this host — `gpt-5.1-codex-mini` unreachable with the local credential). They count as passed and did not really run. None of them covers this PR's surface, which is covered by `restore_015` — a test that genuinely ran, and that I independently proved red-on-old-code.
- **macOS / Windows / `cargo audit`** are **not** in this list: `build-macos`, `build-windows` and `security` all went green on head `eb6b33e`, so they are verified by CI.

prageethw (approved):
Correct, well-tested terminal-injection fix. Sanitizing at the single flush-loop consumer rather than at ~15 push sites is the right call, the escape predicate deliberately matches ratatui-core's own `char::is_control` filter so the exit path is safe on the same terms as the in-session one, and the reproduce-first L2 PTY test (`session/restore/015`) correctly avoids the ONLCR \n->\r\n rewrite trap by asserting on the character after each sentinel rather than a literal-LF needle. Unit tests cover borrow-vs-allocate, DEL/C1 range, and the deliberate non-escaping of literal backslashes for Windows paths. CI green.

Issue #12, opened 2026-03-29
Dashboard UI: Dir truncation and plain digit keybindings


vfarcic:
## PRD: Dashboard UI Improvements

**Problem**: Two usability issues in the dashboard: (1) Long `Dir` values overlap with `Last` and `Tools` fields on the same row, and (2) deck selection requires Alt+N even when the dashboard pane is focused and plain number keys are unused.

**Solution**: Truncate `Dir` with ellipsis when it would overlap, and allow plain digit keys (1-9) to select decks in Normal mode.

**Detailed PRD**: See [prds/12-dashboard-ui-improvements.md](https://github.com/vfarcic/dot-agent-deck/blob/main/prds/12-dashboard-ui-improvements.md)

**Priority**: Medium

PR #597, opened 2026-08-18
fix(daemon): one wedged agent no longer starves its siblings of their SIGKILL

### LINKED ISSUE #581: `shutdown_all_graceful` phase 3: one wedged agent's serialized unbounded `wait()` starves every later agent of its SIGKILL
## What

`AgentPtyRegistry::shutdown_all_graceful` (`src/agent_pty.rs:5430`) runs phase 3 **serially** over the agent vector:

```rust
for mut agent in agents {
    crate::platform::proc::force_kill_child_and_wait(...)
}
```

and `force_kill_child_and_wait` ends in an unbounded `let _ = child.wait();` (`src/platform/proc/unix.rs:136`, `:171` — the comment at `:90` already concedes the wait is unbounded).

A child wedged in uninterruptible kernel I/O does not die on SIGKILL until that I/O completes — the stuck-NFS case. One such agent blocks the loop indefinitely, and **every agent positioned after it in the vector never receives its phase-3 SIGKILL at all.**

## Why it matters

It is silent: the starved agents produce no log line, and the daemon exits immediately afterwards (all production callers of `shutdown_all_graceful` are terminal), so nothing runs later to notice or clean up. The result is leaked agent processes after what looks like a clean shutdown.

## Expected

Phase 3's *signal* delivery should be independent of any one agent's *reap* latency, so a single wedged agent cannot starve its siblings.

## Suggested direction

Two shapes, both already precedented in this codebase:

1. Signal every survivor's process group first, in one pass, and only then reap in a second pass.
2. Reap through a bounded, guaranteed-reaping helper rather than a bare `child.wait()`.

## Why this is filed here

Verified present on `main` at `e6e6dfea`: the serial phase-3 loop at `src/agent_pty.rs:5430` and the unbounded `child.wait()` at `src/platform/proc/unix.rs:136`/`:171` are upstream code, unmodified downstream. It was raised as a LOW finding by a security audit on a fork PR (`prageethw/dot-agent-deck`, issue #211) whose subject was *which agents phase 3 actually reaches*; the audit recommended a follow-up rather than widening that PR. Nothing about it is fork-specific. Reporting it here rather than only fixing it downstream, so the fix lands where every user of this project benefits.

vfarcic:
Closes #581.

## The defect

`AgentPtyRegistry::shutdown_all_graceful` phase 3 ran serially over the agent vector, and `force_kill_child_and_wait` ends in an unbounded `child.wait()`. A child wedged in uninterruptible kernel I/O does not die on SIGKILL until that I/O completes (the stuck-NFS case), so the loop parks in that agent's `wait()` and **every agent behind it in the vector is never signalled at all**. It fails silently: the starved agents log nothing and every production caller is terminal, so the daemon exits and nothing runs later to notice the still-running agent processes.

`shutdown_all` — the SIGKILL-without-grace path reached from `Drop` and idle shutdown — had the identical loop, so it is fixed in the same change rather than left as the next issue.

## Which shape, and why

The issue offered two. **This takes both**, because they fix different halves of the same starvation, and taking only one leaves the other half in place:

1. **Signal every survivor first, reap second.** This is what the issue's *Expected* section actually asks for — phase 3's signal delivery becomes independent of any one agent's reap latency. It is precedented directly above it: phases 1 and 2 are already a signal-everyone pass followed by a poll-everyone pass.
2. **Reap through a non-blocking poll rather than a bare `child.wait()`.** Shape 1 alone still leaves the *reap* serialized behind the wedged agent, so every sibling that did receive its SIGKILL stays an unreaped zombie for as long as the wedge lasts. The reap is now a shared `try_wait` poll — the shape `crate::wrap`'s reap loop already uses (its "finding #12" comment: *"reap through a NON-blocking loop (never a bare blocking `child.wait`)"*).

**The reap is not dropped.** An agent leaves the vector only once its `try_wait` has reported an exit status, or reported an error (meaning there is no status left to collect) — so this does not trade the leaked-process bug for a leaked-zombie one. A genuinely wedged child still holds the function until the kernel lets its `wait` complete, exactly as before; what changed is that it no longer takes its siblings with it. Bounding the daemon's own exit on a D-state child was deliberately left out of scope: the only way to stop waiting is to stop reaping, which is the trade the issue warns against.

Cost: the reap poll runs at the same 50 ms cadence as the grace poll above it, and costs at most one tick — a shutdown whose agents exited during the grace window clears the whole vector on the first `try_wait` pass and never sleeps.

## Shape of the change

- `src/platform/proc/{unix,windows}.rs` grow `force_kill_child_group` — the signal half (`killpg(SIGKILL)`) / terminate half (`TerminateJobObject`) of `force_kill_child_and_wait`, which is now defined in terms of it. Every existing caller of `force_kill_child_and_wait` is untouched.
- `AgentPtyRegistry::force_kill_and_reap_all` is the new two-pass helper, shared by `shutdown_all` and phase 3 of `shutdown_all_graceful`.

## Reproduce-first

Written before the fix, and each is verified **red on the pre-fix code**, green after. The wedge itself has to be modelled — a real child always returns from `wait` once SIGKILL lands, so the stuck-NFS shape is not reachable with a real process — but everything the teardown path *does* stays production code.

| Test | What it pins |
|---|---|
| `..._kills_every_real_agent_process_even_when_a_reap_wedges` (unix) | The operator-visible symptom. Two **real** `sh` processes, each in its own process group and made SIGTERM-proof with `trap "" TERM; exec sleep`, so only phase 3's real `killpg(SIGKILL)` can end them; death is observed by EOF on each one's inherited stdout pipe. Pre-fix: `died? = [true, false]`. |
| `..._force_kills_every_agent_even_when_a_reap_wedges` | The same property portably, via the pid-unavailable `Child::kill` fallback both backends document — this is the half that also covers the Windows job-object path. |
| `..._reaps_without_parking_in_an_unbounded_wait` | Shape 2: children that answer `try_wait` but whose blocking `wait` never returns must still all be reaped and the shutdown must finish. |
| `..._force_kills_and_reaps_every_agent_when_nothing_wedges` | **Control** — passes on the pre-fix code, so "both agents were killed" cannot be read as "this whole path is broken". |

Both tests use two wedged agents rather than one wedged plus one live, so the assertion does not depend on `HashMap::drain` order: exactly one kill escapes the pre-fix loop whichever agent it happens to touch first, and requiring both is deterministically red.

Two traps worth recording, since both made an earlier draft pass vacuously:

- The first draft's "was this child killed?" observable was a boolean — but **phase 1's SIGTERM ask reaches the same fallback**, so it was already true for every agent before phase 3 ran. It is now a count, asserted strictly above a documented per-platform phase-1 baseline.
- The stand-in shells raced their own startup against phase 1's SIGTERM and sometimes died *before* installing their trap (measured: both gone ~1 ms after spawn, no phase 3 involved, test green). There is now a readiness handshake — the shell writes one byte after `trap`, and the test blocks on it before starting the shutdown.

Each half of the fix was then reverted independently to confirm it is load-bearing: reverting the split pass turns the two starvation tests red; reverting the non-blocking reap turns the unbounded-`wait` test red.

## CLAUDE.md rule 12 — cross-version contract

**The TUI↔daemon contract did not change.** No frame, field, handler, or role-map value is touched: the change is entirely inside the daemon's own process-teardown ordering, below any wire surface, and `src/daemon_protocol.rs` is not in the diff. So `PROTOCOL_VERSION` is unchanged, no `.breaking.md` fragment applies, and per the bump policy this is a **patch**. There is likewise no semantic (same-wire, different-meaning) break to classify — an old TUI and a new daemon exchange byte-identical frames, and the changed code runs only after the last client is gone.

## Gates

- `cargo fmt --check` ✅
- `cargo test-fast` ✅ — 2725 passed
- `cargo nextest run --features e2e -E 'binary(e2e_daemon_orphan_exit)'` ✅ — the daemon's real SIGTERM graceful-shutdown path, which is the caller of the changed code
- The full `cargo test-e2e` tier was **not** run (it spawns real agents against LLM APIs); CI plus the targeted e2e binary above cover the changed path.
- The Windows target could not be type-checked locally — `aws-lc-sys` needs an MSVC cross-toolchain this box does not have — so the `build-windows` CI job is the check for it. The Windows-side change is two lines calling an existing helper, and the new test double's `#[cfg(windows)] as_raw_handle` mirrors the existing `test_child::StdChild` impl in `src/platform/proc/mod.rs` exactly. macOS shares the Unix path verbatim.

A user-facing `changelog.d/581.bugfix.md` fragment is included.


prageethw (approved):
Correct fix for the shutdown starvation. The two-pass split (signal all, then reap via non-blocking `try_wait` poll rather than per-agent blocking `wait()`) is the right shape and matches the precedent in `crate::wrap`'s reap loop. Verified the reap-drop condition is sound (`retain_mut` keeps only `Ok(None)`, so both a reported exit and a reported error clear the agent — no leaked-zombie trade). Real dual-process tests correctly model the stuck-NFS wedge (SIGTERM-proof children, group-killed via real `killpg`) with a control test that passes pre-fix to prove the assertion isn't vacuous. CI green, both Unix and Windows backends symmetrically split.

PR #598, opened 2026-08-18
fix(hook): retry transient socket read errors instead of reporting no daemon

### LINKED ISSUE #564: socket_004 is timing-flaky on the macOS fast tier, and a job-level re-run erases the evidence
`hook::tests::socket_004_slow_but_replying_daemon_still_returns_the_reply` (`src/hook.rs:1093`) failed once on the macOS fast tier and passed on a job-level re-run of the **same commit**.

## Evidence

PR #558, head `3e179daf`:

| | |
|---|---|
| Run | [`31807225052`](https://github.com/vfarcic/dot-agent-deck/actions/runs/31807225052) — `run_attempt=2`, conclusion `success` |
| **Attempt 1** | **`failure`** — job [`build-macos`](https://github.com/vfarcic/dot-agent-deck/actions/runs/31807225052/job/94788966425), step `Run cargo nextest run --workspace` |
| Attempt 2 | `success`, same commit, no code change |

**Not attributable to that branch.** `3e179daf`'s diff touches only `xtask/linkage-check/src/repo_state.rs` and `xtask/linkage-check/src/main.rs`. `src/hook.rs` is untouched, and `build-macos` passed on both earlier heads of the same branch (`c0c52bfe`, `4c9177b3`, each `run_attempt=1`) with that code identical.

## Why it is worth an issue rather than a re-run

**A job-level re-run overwrites the conclusion**, so `gh run list` and the PR's check board both show `build-macos success` and the failure leaves no visible trace. It is recoverable only via `actions/runs/<id>/attempts/1`. That is the same masking #402 recorded for `delegate_011` — *"a job-level re-run of the above"* — and it means the observed rate of these is systematically under-counted.

**`retries = 0`** in `.config/nextest.toml`'s default profile, and `build-macos` is a **blocking** job. So a single timing blip hard-reds a whole PR, and the only remedy on offer is a re-run that erases the evidence.

**This one is a plain unit test, which makes it a different case from #402.** `delegate_011` is an integration-shaped test in `tests/`. `socket_004` lives in `src/hook.rs`'s own `mod tests` — it spawns a daemon thread, talks to it over a socket, and asserts a reply arrives inside `GET_SEED_REQUEST_TIMEOUT`. A wall-clock bound between two threads on a shared CI runner is load-sensitive by construction, so the fast tier's fastest and most trusted layer contains a test that can lose a race to the scheduler.

## What I am not claiming

I have one occurrence. I am not asserting a rate, and I have not reproduced it locally — a load-dependent timing test is exactly the kind that will not reproduce on demand. Filing it because the evidence is about to become invisible: attempt-1 logs age out, and once they do there is nothing left to point at.

## Possible directions

- Make the assertion wait on the daemon thread's own readiness signal rather than a wall-clock bound, so it cannot lose to the scheduler.
- Or scope a `slow-timeout`/retry override to this test by name in `.config/nextest.toml`, the way `route_001` and `new-pane/016` already are — bounded, and it keeps the strict default for everything else.
- Either way, worth checking whether other `src/**` unit tests bound a cross-thread wait on wall-clock time; if `socket_004` is not alone, the class matters more than the instance.

Found while working #557/#558; unrelated to that change.

vfarcic:
Fixes #564.

## Where the defect is: the production timeout logic, not the test's timing assumption

The issue's leading hypothesis was that `socket_004` is a wall-clock race between two threads that a loaded macOS runner can lose. The durations refute that. Both occurrences failed **fast**:

| Occurrence | Head | nextest line | Budget |
|---|---|---|---|
| PR #558 | `3e179daf` | `FAIL [ 0.435s] ( 417/2628)` | `GET_SEED_REQUEST_TIMEOUT` = 5s |
| PR #464 | docs-only (`prds/*.md`) | `FAIL [ 0.401s] ( 405/2826)` | `GET_SEED_REQUEST_TIMEOUT` = 5s |

**This PR's own CI run then closed the argument.** The new artifact (below) records what a *healthy* macOS run of this test costs: `socket_004` **passed on `build-macos` in 0.420s** — sitting between the two failures at 0.401s and 0.435s. So the failing attempts were not slow at all; they took the same wall clock as a passing one and simply returned the wrong answer at the moment the reply was due. Whatever ended them was not the clock. (The honest cost is the test's own 300ms stub-daemon sleep plus platform overhead: 0.307s on Linux, ~0.42s on macOS.) Neither attempt spent a tenth of the budget. Starving this test into a false negative would take **5+ seconds** of scheduler denial — what actually happened returned in 0.4, about 100ms *after* the reply was due. The deadline was never in play, so the "wait on a readiness signal instead of a wall clock" direction would have papered over a real bug.

An abort there is reachable exactly one way. Every other terminal path in `read_reply_line` either needs the full ~5s (deadline expiry) or returns the reply — both EOF branches do, since a partial line is returned as `Some(partial)` and a completed one breaks out. So a `read` returned `Err`, and the loop did:

```rust
let n = stream.read(&mut buf).ok()?;
```

`.ok()?` makes **every** `io::Error` terminal. Two classes carry no information about whether the exchange is over:

- **`Interrupted` (`EINTR`)** — a signal landed on the waiting thread. It says nothing about the peer and nothing about the clock, and `std` does not retry it here the way `Write::write_all` does on the send side.
- **`WouldBlock`/`TimedOut`** — the socket's own per-read `SO_RCVTIMEO`, which is a *different clock* from the total-operation deadline this function advertises.

`read_reply_line` computed `deadline` only to re-arm the next read; it never consulted it when deciding a read failure was final. So the operation ended on the socket's word rather than the deadline's, inverting the contract its own doc comment states ("tracks wall-clock elapsed time against `deadline` … so the operation as a whole … cannot run past `deadline`"). The deadline bounded the operation from above but was not what *ended* it.

The user-visible cost is silent: `get-seed` treats "no reply" and "no daemon" alike and falls back to PTY injection, so a reply that was written and dropped is indistinguishable from an older daemon that never knew the verb. `delegate` shares the transport via `send_and_await_reply`.

## (a) The fix

A read failure is now reconciled with the deadline before it ends anything (`is_transient_read_error`): `Interrupted` always retries; `WouldBlock`/`TimedOut` retry while the operation's own budget has time left; everything else stays terminal. The loop head still returns once the budget is gone, so the operation stays bounded and **the 5s bound is unchanged** — `socket_003` and `socket_005` still return in 5.01s.

`read_reply_line` now returns a typed `ReplyReadError` (`DeadlineExpired` / `ClosedWithoutReply` / `Io` / `InvalidUtf8`) instead of a bare `Option`. **Every variant still folds into `SocketReply::NoReply` at the boundary**, so no caller sees a new outcome; the reason is logged and made available to the tests, because the ambiguity between those four is exactly what made this take two occurrences and a log excavation to place.

## Reproduced — the class, not the macOS failure

I have **no macOS box and did not reproduce the CI failure**, as the issue itself anticipated. What I did reproduce locally is the defect class, deterministically: **`error/socket/007`** drives the same slow stub daemon while `pthread_kill`ing the *client* thread with `SIGUSR1` 25 times at 10ms intervals, under a handler installed deliberately **without `SA_RESTART`** so the interrupted read really surfaces as `EINTR` rather than being restarted by the kernel.

Against the pre-fix logic it fails — and the signature matches CI:

```
FAIL [ 0.310s] socket_007_signal_interrupted_read_still_returns_the_reply
  got NoReply (read error: Some(Io(Os { code: 4, kind: Interrupted, message: "Interrupted system call" })));
  the stub daemon reports: writing the reply failed: Broken pipe (os error 32)
```

0.310s of a 5s budget, next to CI's 0.401s and 0.435s. It passes after the fix, and is stable over repeated runs including pinned to a single core (`taskset -c 0`, 4x, 0.312–0.314s).

To be explicit about what this does **not** establish: it does not prove which errno macOS returned. That is why the fix covers the whole transient class rather than one errno, and why the assertions now print the cause.

## Test hardening

`socket_004` no longer collapses the result into `Option<String>` before asserting — a bare `left: None` is a value that `NoReply` and `Unreachable` produce alike. It now reports the full `SocketReply`, the `ReplyReadError`, **elapsed vs. budget**, and the stub daemon's own outcome (did its write actually succeed?). Had that message existed, the diagnosis above would have come out of the failure instead of out of attempt-1 log archaeology.

The 300ms sleep is kept: the evidence says it was never the problem, and "slow daemon" is the scenario under test.

## (b) A job-level re-run no longer destroys the failing attempt

Two changes:

1. **All three nextest jobs upload their JUnit XML**, named `nextest-<os>-attempt-${{ github.run_attempt }}`. The attempt in the *name* is load-bearing twice over: artifacts are scoped to the run rather than the attempt, and upload-artifact v4+ refuses to overwrite an existing name, so a fixed name would make the re-run's upload fail outright. With it, attempt 1's results sit alongside attempt 2's and stay readable after the conclusion has been overwritten in place.

   The payload is the only structured record of a run — per-test wall clock plus the full failure message. Verified locally that a failing run's XML carries `time="0.325"` and the complete panic text. **The duration is not incidental; it is what diagnosed this issue**, and it currently appears nowhere but a nextest log line that ages out.

   Applied to `build`, `build-windows` and `build-macos`: the masking is a property of re-running any job, not of macOS.

   **Verified live on this PR's first run**, rather than assumed: it produced `nextest-linux-attempt-1`, `nextest-windows-attempt-1` and `nextest-macos-attempt-1` (37-55KB each, 90-day retention), and the macOS one is where the 0.420s figure above came from — a number that would previously have existed only in a log line. That is the mechanism doing exactly its job on its first outing.

2. **Fixed the JUnit path.** nextest resolves `[profile.default.junit] path` relative to the profile's store directory, so `"target/nextest/default/junit.xml"` was actually writing `target/nextest/default/target/nextest/default/junit.xml`. Nothing read the file, so the doubling went unnoticed — which is the point.

## Deliberately not done

- **No `slow-timeout`/retry override for `socket_004`** (the issue's second suggested direction). The evidence says the test is right and the code was wrong; a retry would have hidden a genuine production defect, and `retries = 0` stays the strict default.
- **`UnixStream::connect` still does not retry `EINTR`.** `connect(2)` on an already-bound local AF_UNIX socket completes without blocking, so this is theoretical; `socket_007`'s 50ms lead-in keeps signals out of that window on purpose, and the assertion message names `Unreachable` explicitly so that case would not be mysterious.
- **I did not sweep other `src/**` unit tests for wall-clock cross-thread waits** (the issue's third direction). Worth its own pass; the class here turned out to be a production bug rather than a test-timing one, so the sweep would be looking for something different than originally framed.

**Did this change the TUI↔daemon contract? No.** No wire shape moved, no field changed meaning, no role-map value type shifted; `SocketReply`'s variants and their semantics are identical and the daemon side is untouched. The change is confined to the client's read loop deciding when to stop reading. So no `PROTOCOL_VERSION` bump and no `.breaking.md`.

Because nothing on the wire moved, I did **not** run the two-daemon cross-version sandbox — flagging that explicitly rather than silently, so it can be required if you read the boundary differently.

## Gates

- `cargo fmt --check` clean
- `cargo clippy --workspace --all-targets --features e2e -- -D warnings` clean
- `cargo test-fast` — 2722 passed, 0 failed
- `cargo xtask linkage-check` — ok (534 catalog ids, 462 annotations)
- `cargo test-e2e` **not** run (rule 5 bounds it to pre-release; this change is fast-tier only)

Changelog fragment: `changelog.d/564.bugfix.md`.

### DIFF
--- a/.config/nextest.toml
+++ b/.config/nextest.toml
@@ -3,8 +3,17 @@ retries = 0
 slow-timeout = { period = "60s", terminate-after = 3 }
 fail-fast = false
 
+# Issue #564: `path` is resolved RELATIVE TO the profile's store directory
+# (`target/nextest/<profile>/`), so the previous value of
+# "target/nextest/default/junit.xml" actually wrote
+# `target/nextest/default/target/nextest/default/junit.xml`. Nothing read the
+# file, so the doubling went unnoticed — and that is the point: this is the one
+# structured, per-test, machine-readable record of a run (per-test durations and
+# the full failure message), and CI now uploads it per run_attempt so that a
+# job-level re-run cannot erase a failing attempt's evidence the way it erased
+# the first `socket_004` occurrence.
 [profile.default.junit]
-path = "target/nextest/default/junit.xml"
+path = "junit.xml"
 
 # PRD #140 M5.1: `orchestration/route/001` boots SIX interactive Haiku panes
 # across two orchestration tabs and drives two concurrent
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -199,6 +199,17 @@ jobs:
       # inert and the gap survived #436. Costs nothing measurable: all 48 are
       # under 40ms and the tier's wall clock is set by its slowest test.
       - run: cargo nextest run --workspace
+      # Issue #564: keep this attempt's results after a job-level re-run
+      # overwrites the conclusion. Full rationale on the identical step in
+      # `build-macos` below — the masking is a property of re-running any job,
+      # not of macOS, so all three test jobs carry it.
+      - name: Preserve this attempt's test results
+        if: always()
+        uses: actions/upload-artifact@v7
+        with:
+          name: nextest-linux-attempt-${{ github.run_attempt }}
+          path: target/nextest/default/junit.xml
+          if-no-files-found: warn
       # CLAUDE.md rule 7 has always SAID this runs in CI ("CI's linkage-check
       # rule 7 fails the build if the Scenario comment is missing or the
       # generator fails"), and the build-macos/build-windows comment says these
@@ -349,6 +360,17 @@ jobs:
       # dependency (`symlinks_named_like_owned_dirs_are_skipped`) is already
       # `#[cfg(unix)]`; the rest are in-memory or tempdir-based.
       - run: cargo nextest run --workspace
+      # Issue #564: keep this attempt's results after a job-level re-run
+      # overwrites the conclusion. Full rationale on the identical step in
+      # `build-macos` below — the masking is a property of re-running any job,
+      # not of macOS, so all three test jobs carry it.
+      - name: Preserve this attempt's test results
+        if: always()
+        uses: actions/upload-artifact@v7
+        with:
+          name: nextest-windows-attempt-${{ github.run_attempt }}
+          path: target/nextest/default/junit.xml
+          if-no-files-found: warn
 
   # macOS (host = aarch64-apple-darwin) compile + clippy + unit-test gate,
   # mirroring build-windows. macOS is otherwise built ONLY in the release
@@ -388,6 +410,45 @@ jobs:
       # there is nothing to match-and-narrow here — the clippy divergence is
       # about `--all-targets --features e2e`, not about package selection.
       - run: cargo nextest run --workspace
+      # Issue #564: preserve THIS attempt's test results, so a job-level re-run
+      # cannot destroy the evidence from a failing one.
+      #
+      # A job-level re-run overwrites the job's conclusion in place. After
+      # re-running this job on the same commit, `gh run list` and the PR's
+      # check board both read `success`, and the failure that prompted the
+      # re-run survives only under `actions/runs/<id>/attempts/<n>`, whose logs
+      # age out. So the standard remedy for a one-off timing blip — and with
+      # `retries = 0` in .config/nextest.toml and every build job blocking, a
+      # re-run is the ONLY remedy on offer — is also the thing that erases the
+      # record of it, which makes the observed rate of such failures
+      # systematically under-counted. #564 nearly lost its first occurrence
+      # that way, and #402 recorded the same masking for `delegate_011`.
+      #
+      # `${{ github.run_attempt }}` in the artifact NAME is the load-bearing
+      # part, twice over: artifacts are scoped to the run rather than the
+      # attempt, and upload-artifact v4+ refuses to overwrite an existing name,
+      # so a fixed name would make the re-run's upload fail outright. With the
+      # attempt in the name, attempt 1's results sit alongside attempt 2's and
+      # the failing attempt stays readable after its conclusion is gone.
+      #
+      # The payload is nextest's JUnit XML — the only structured record of a
+      # run: per-test WALL CLOCK plus the full failure message. The duration is
+      # not incidental, it is what diagnosed #564. Both macOS failures of
+      # `socket_004` took ~0.4s against that test's 5s budget, which is what
+      # showed the deadline was never in play and a read error had ended the
+      # exchange instead. That number appears nowhere but this artifact and a
+      # nextest log line.
+      #
+      # `always()` so it runs on the failing path, which is the only path that
+      # matters here; `if-no-files-found: warn` because a job that dies before
+      # the test step has no results and must not fail on that.
+      - name: Preserve this attempt's test results
+        if: always()
+        uses: actions/upload-artifact@v7
+        with:
+          name: nextest-macos-attempt-${{ github.run_attempt }}
+          path: target/nextest/default/junit.xml
+          if-no-files-found: warn
 
   # `cargo audit` reads Cargo.lock, which a devbox-only PR cannot touch.
   security:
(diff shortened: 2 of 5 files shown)


vfarcic:
**Note for review: Greptile did not actually inspect this PR.**

Its review object is present but is a credit notice, not a review:

> Greptile has paused reviews on this repository — it used its 300 free open-source review credits for this billing period. Reviews resume automatically on August 25.

So there are no inline comments to respond to (`gh api repos/vfarcic/dot-agent-deck/pulls/598/comments` returns 0), and none will arrive on a re-push before Aug 25 unless an org admin enables billing. Per CLAUDE.md rule 8 the findings normally live in the inline comments, so flagging explicitly that **the automated review layer is absent here** rather than clean — this PR has had no bot review at all, and human review is the only review it will get.

CI is green on `4f97690` across all nine checks, including `build-macos` twice.

prageethw (approved):
Excellent root-cause work — the duration evidence (0.401s/0.435s failures vs 0.420s healthy run, all well under the 5s budget) correctly rules out the wall-clock race the issue suspected and locates the real bug in `.ok()?` treating every read error as terminal, including EINTR and the per-read SO_RCVTIMEO firing early relative to the operation deadline. The fix reconciles read failures against the deadline before ending the exchange, and `is_transient_read_error` is correctly bounded (WouldBlock/TimedOut only retry while deadline has remaining budget, so no infinite-spin risk). Deterministic local repro via SIGUSR1 without SA_RESTART is a good substitute for an unavailable macOS box. The JUnit-artifact and double-path fixes are a solid diagnostic-infrastructure bonus. CI green.

Issue #103, opened 2026-05-22
PRD: Local daemon build-version handshake + daemon stop CLI


vfarcic:
## PRD: Local daemon build-version handshake + `daemon stop` CLI

**Problem**: After v0.25.0 (PRD #93, always-external daemon), a user who upgraded the binary while a daemon from a previous build was still running had their local TUI silently attach to the stale daemon. The local attach path performs no version handshake (it only checks the socket's mode/uid), and `PROTOCOL_VERSION=1` did not change even though rounds 9-12 mutated handler-side schemas (e.g. `pane_orchestration_map` value type changed from `String` to `(String, String)`; `TabMembership::Orchestration` gained `orchestration_cwd`). Delegate signals reached the stale daemon but silently no-op'd in its role-map lookup. Remote (`dot-agent-deck connect`) was unaffected because each ssh-connect spawns a fresh daemon at the remote's installed binary version.

**Solution**: Close the local upgrade-path gap on two axes: (1) extend the existing `AttachRequest::Hello` / `AttachResponse` handshake to carry a `build_version: String` alongside `server_version: u32`, and add a TUI-side check after `ensure_external_daemon_or_die` that bails with a clear error when the daemon's build differs from `env!("DAD_VERSION")`; (2) ship `dot-agent-deck daemon stop` (and `daemon restart`) so users have a documented, non-`kill -9` way to recycle the local daemon — refusing without `--force` if managed agents are still alive. This delivers the "equivalent local command" promised by PRD #93 line 39.

**Detailed PRD**: See [prds/103-local-daemon-build-version-handshake.md](https://github.com/vfarcic/dot-agent-deck/blob/main/prds/103-local-daemon-build-version-handshake.md)

**Priority**: High

PR #599, opened 2026-08-18
fix(ipc): bound the sync clients' connect step by the caller's deadline (#435)

### LINKED ISSUE #435: IpcClient::connect is uninterruptible on Unix, so both sync IPC paths can hang past their advertised deadline

vfarcic:
Fixes #435.

## The defect

Both synchronous IPC paths opened the connection with a blocking, uninterruptible call and applied their deadline only afterwards:

- `src/hook.rs` — `request_from_socket_at` (the `get-seed` and `delegate` verbs), whose doc comment claims to bound the **whole exchange**;
- `src/ui.rs` — `send_daemon_request_blocking_with_timeout` (the manager dialog's run-now and the 250 ms `Ctrl+n` hint), on the synchronous TUI key path.

On Unix `IpcClient::connect` was a straight lift of `std::os::unix::net::UnixStream::connect`, which honours no timeout — so the connect step sat outside every budget those callers advertise.

## Reproduced first

`connect(2)` to a Unix socket whose listener has a full accept queue and never calls `accept()` **parks indefinitely**. It does not report `ECONNREFUSED`, and it does not report `EAGAIN` either — that is what a *non-blocking* socket gets, which is precisely why only the blocking path was exposed.

The new `platform::ipc::unix::tests::connect_against_a_saturated_listener_returns_within_the_deadline` builds exactly that shape (a raw `listen(2)` with backlog 1, filled until the queue stops taking connections, never accepted). Against the code as it stood it **sat parked past a 5 s watchdog**; with the fix it returns `TimedOut` in ~250 ms. Reverting `connect_timeout` to delegate to `connect` turns it red again, so the fix is load-bearing rather than incidental.

Its control — the same client, the same code path, a listener with a free backlog slot — connects in 6 ms both before and after, so the red above is attributable to saturation and not to "the deck can no longer talk to an un-accepted socket".

## The fix

`IpcClient::connect_timeout(endpoint, timeout)` on both backends; both sync callers now use it.

**Unix** creates the socket non-blocking and retries `connect(2)` against the budget with a 1 ms → 20 ms backoff. The obvious shape — connect once, `poll(2)` for writability, read `SO_ERROR` — is the `TcpStream` recipe and is silently wrong here, which I found by measurement rather than by reasoning: `connect(2)` documents `EINPROGRESS` as the pollable "started, not finished" state and states that "UNIX domain sockets failed with `EAGAIN` instead", and an *unconnected* Unix socket already polls writable, so that version returned `Ok` — with a client connected to nothing — against a saturated listener in single-digit milliseconds. Retrying is also the right behaviour for the condition: a full accept queue is transient, and the Windows backend already retries its exact analogue (`ERROR_PIPE_BUSY`) against a budget.

Error kinds are preserved: everything except "try again" comes straight from `connect(2)`, so a missing socket file is still `NotFound` and a stale inode still `ConnectionRefused`. Blowing the budget is `TimedOut`, which no caller had a way to see before. The socket is created `SOCK_CLOEXEC` (with an `fcntl` fallback off Linux) to match what the stdlib's connect does — the deck `exec`s an agent per pane, so a leaked daemon-socket fd would be a real handle escape.

**Windows** could never hang here, but it could not honour a budget *shorter* than its own 2 s retry ceiling either; `connect_timeout` takes the smaller of the two, so the 250 ms hint path no longer spends two seconds on a busy pipe. This only tightens the path.

**`hook::send_to_socket` deliberately keeps the unbounded `connect`.** It advertises no deadline, and dropping an agent event on a momentarily-full queue is a worse trade than waiting for a slot. That is now stated on `connect`'s own doc comment, so the choice between the two entry points is documented at the seam rather than left to be rediscovered.

**Did this change the TUI↔daemon contract? No.** No request or response variant, field, or framing moved, and no field changed meaning — this is how a connection is *opened*, not what is said over it. `PROTOCOL_VERSION` stays at **7** (both `v0.36.2` and this branch report `server_version: 7` from `daemon hello`), and no `changelog.d/*.breaking.md` fragment applies. Bump class: **bugfix → patch**.

Verified rather than asserted. A `v0.36.2` daemon was started in a fully isolated sandbox (sockets, lock dir, state dir, config, `HOME`, **and `DOT_AGENT_DECK_LOG`**), and this branch's binary was driven against it:

| Path | Result |
|---|---|
| Attach socket — TUI attach + `daemon status` | both panes listed |
| `Ctrl+n` form (`send_daemon_request_blocking_with_timeout`, 250 ms) | form opened, orchestration offered and created |
| `delegate` (`send_and_await_reply` → `connect_timeout`) | routed; task file written with the exact task text; worker pane received it |
| `agent-event --type running` | arrived; the worker card flipped to Thinking |
| `work-done --done` | arrived; the daemon retired the delegation and cancelled its idle watch |
| `get-seed` (`request_from_socket` → `connect_timeout`) | answered (`has_seed=false`) |

One wrinkle worth recording, because it makes the naive form of this test vacuous: the **first** attempt did not test cross-version at all. The branch build's `DAD_BUILD_ID` differs from the release's, so the PRD #103 handshake did what it is supposed to do — the branch TUI `SIGTERM`'d the older daemon on attach and lazy-spawned its own, and everything after that was same-version. The run above therefore builds the branch with `DAD_VERSION=0.36.2 DAD_BUILD_ID=0.36.2-g4bb47a5` so the handshake sees a current daemon and attaches to it; the `v0.36.2` daemon PID was confirmed still serving after every step. Anyone repeating this check on a daemon-adjacent PR should mask the build id the same way, or they will silently test a new binary against itself.

## Tests

- `platform::ipc::unix::tests::connect_against_a_saturated_listener_returns_within_the_deadline` — the reproduction. Asserts against what the platform actually did, because Linux and the BSDs genuinely differ here: Linux parks a blocking connect (`TimedOut` after the budget), while `sonewconn` on the BSDs refuses an over-limit connection outright (`ConnectionRefused`, fast). Either way the client must come back inside its budget, which is the property under test.
- `platform::ipc::unix::tests::connect_against_a_listener_with_room_still_succeeds` — the control.
- `platform::ipc::unix::tests::connect_retries_until_the_listener_drains_its_queue` — the failure direction alone is not enough: a `connect_timeout` that could never connect *at all* would satisfy the reproduction above, and returning `TimedOut` on the first `EAGAIN` does exactly that while leaving it green. This test saturates the same listener, starts a connect with a 5 s budget, then drains the queue until the client gets in — and was confirmed to fail under that stubbed-out implementation while the reproduction stayed green.
- `platform::ipc::tests::connect_timeout_is_a_drop_in_for_the_unbounded_connect` — un-gated, so both backends are asserted on all three CI jobs: a live endpoint still round-trips, and an absent one still fails with `NotFound` immediately rather than burning the budget.

No TUI test is added under rule 4: nothing user-visible changes except in the pathological wedged-daemon case, which has no rendered surface.

## Gates

`cargo fmt --check`, `cargo clippy --workspace --all-targets --features e2e -- -D warnings`, `cargo test-fast` (2724/2724), and `cargo xtask linkage-check` all clean.

### DIFF
--- a/changelog.d/435.bugfix.md
+++ b/changelog.d/435.bugfix.md
@@ -1,0 +1,9 @@
+## The deadlines on the deck's synchronous daemon requests now cover connecting, not just talking
+
+Two paths in the deck ask the daemon a question and wait for the answer on the spot: the TUI's own synchronous requests (the manager dialog's run-now, and the hint the `Ctrl+n` form asks for as it opens), and an agent's `delegate` / `get-seed` over the hook socket. Each states a deadline — five seconds, or a quarter of a second for the hint — and each applied it only *after* the connection was open. Connecting itself was unbounded.
+
+That is not a theoretical gap on Unix. A blocking `connect(2)` to a socket whose listener has a full accept queue does not fail: it parks, indefinitely, until something accepts. It does not report "connection refused", and it does not report "try again" either — that is what a non-blocking socket gets, which is exactly why only the blocking path was exposed. So a daemon whose runtime was starved long enough for connections to queue behind it could hang a keystroke in the TUI, or an agent's `delegate`, with no ceiling at all, while the code and its comments said the whole exchange was bounded.
+
+The connect step now honours the caller's budget on both platforms. On Unix the socket is created non-blocking and the connect retried until it succeeds, fails for a real reason, or the budget runs out — a full accept queue is a transient condition worth waiting out, just not forever. Failure kinds are unchanged, so a missing socket file and a stale one still read as "no daemon" exactly as before; blowing the budget is a timeout, which no caller could see previously. Windows could never hang here, but it can now honour a budget *shorter* than its own two-second retry ceiling, so the 250 ms hint path no longer spends two seconds on a busy pipe.
+
+Nothing on the wire moved: this is how the deck opens a connection, not what it says once it is open. A build carrying this fix and one without it interoperate exactly as before.
--- a/src/hook.rs
+++ b/src/hook.rs
@@ -505,8 +505,8 @@ pub const DELEGATE_REPLY_TIMEOUT: std::time::Duration = std::time::Duration::fro
 /// connect/write/read. An earlier draft of this function was exactly that, and
 /// it carried no deadline; sharing the one implementation gives `delegate` the
 /// total-operation bound, the half-close, the read-exactly-one-line behaviour
-/// PRD #163 M4 needed for Windows, and any future fix to the connect phase
-/// (issue #435) for free.
+/// PRD #163 M4 needed for Windows, and — since issue #435 — a connect step that
+/// is inside the deadline rather than beside it, for free.
 pub fn send_and_await_reply(json: &str) -> SocketReply {
     request_from_socket_inner(json, Some(DELEGATE_REPLY_TIMEOUT))
 }
@@ -567,7 +567,10 @@ const GET_SEED_REQUEST_TIMEOUT: std::time::Duration = std::time::Duration::from_
 /// `Some(String::new())`.
 ///
 /// [`GET_SEED_REQUEST_TIMEOUT`] bounds the **whole exchange**, not just a
-/// single idle read: [`request_from_socket_inner`]'s read loop measures
+/// single idle read — and since issue #435 that is literally rather than
+/// approximately true, because the connect step goes through
+/// [`crate::platform::ipc::IpcClient::connect_timeout`] instead of a blocking
+/// `connect(2)` that no deadline reached. The read loop measures
 /// elapsed wall-clock time against it and shrinks each individual blocking
 /// read's own timeout to whatever is left, so a peer that keeps dribbling a
 /// byte at a time without ever completing the reply line can no longer keep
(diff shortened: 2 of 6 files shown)


prageethw (approved):
Careful, correct low-level fix with real measurement backing the design choice — good catch that the poll+SO_ERROR TcpStream recipe is silently wrong for AF_UNIX (an unconnected socket already polls writable, so it would return Ok against nothing). The retry-based connect_timeout correctly distinguishes EISCONN (success), EINTR (always retry), and EAGAIN/EINPROGRESS/EALREADY (retry only while budget remains, else TimedOut), with SOCK_CLOEXEC handled per-platform. The reproduction (saturated accept queue via raw listen/backlog) plus the queue-drains-until-connects test (which would catch a trivial "always return TimedOut" stub) are exactly the right pair. Genuine cross-version verification against a real v0.36.2 daemon, including catching the DAD_BUILD_ID same-version trap that would have made the first attempt vacuous. CI green.

Issue #524, opened 2026-08-12
A worker that exits without StopAgent leaves its routing mappings behind, so a delegate to it reports success


vfarcic:
Split out of the [PR #466 review](https://github.com/vfarcic/dot-agent-deck/pull/466#discussion_r3742633846) (greptile P1, adjudicated in-thread).

## The mechanism

`AppState::unregister_pane` — which clears `pane_role_map`, `pane_cwd_map`, `orchestrator_pane_ids` and `pane_orchestration_map` — has exactly one daemon-side caller: the `AttachRequest::StopAgent` handler (`src/daemon_protocol.rs`). The `src/ui.rs` call sites operate on the TUI's own `AppState`, not the daemon's.

So a worker that exits **without** going through the explicit pane-close path — it crashed, or the agent quit on its own — leaves its role mappings in place. `delegate_targets` still matches that role, the pane lands in `DelegateResponse::delivered`, and `dot-agent-deck delegate` exits 0. The actual PTY write happens later inside `dispatch_one_owned`, in a detached `tokio::spawn`, so its failure cannot reach the response.

The staleness itself predates PR #466 — `delegate_targets` has always been able to resolve a dead pane. What that PR changed is that it now has a *consequence*: before, `delegate` exited 0 unconditionally, so there was no success claim to be wrong.

## Scope of the false success — narrower than it first looks

`write_to_pane_internal` filters on `!exited` (round-11 reviewer #A), so a write to a dead worker really does return `AgentPtyError::NotFound`. But that is only reached for a `clear = false` role. For `clear = true` — the default — `dispatch_one_owned` calls `respawn_agent_for_pane` first, and that function *deliberately* omits the `exited` filter:

> The `exited` filter is deliberately omitted: a dead-but-not-yet-reaped agent should also be replaced — its registry entry is the place the new agent's identity (display_name, tab_membership, etc.) lives, and `clear = true` on a crashed agent should still produce a fresh worker.

So for the common config the delegate genuinely lands: the crashed worker is replaced and the fresh one receives the task. The false success is real for `clear = false` roles, and for a pane whose registry entry is gone entirely.

[...]

Issue #205, opened 2026-07-10
Reminding Agents to deliver results as queued messages


ss89:
I am just trying out dot-agent-deck and sometimes my reviewer/auditor agent does not report back to the orchestrator.

Is there some setting that needs to be set to remind it or to let the orchestrator ask the remaining agent to deliver results upon "task done"?

Would be neat to otherwise have this implemented, e.g. by a hook that forwards the last agent message when it stops working if the dot-agent-deck work-done command was not called automatically by the model.

vfarcic:
Apologies for the long silence — notifications about new issues weren't reaching me, so this never surfaced until I audited the backlog today. Sorted now.

Good news: this is implemented, and it shipped in **0.35.0** (today). You asked for it on 2026-07-10, so you were 18 days ahead of it — and the design you sketched is close to what landed.

## What it does

The daemon tracks every outstanding delegation — role, worker pane, which orchestrator delegated it, and when. If no `work-done` arrives within the timeout, it injects **one** self-describing prompt into the orchestrator's session telling it that a delegated worker has gone silent, so the orchestrator can chase it. Delivery happens at a turn boundary, never mid-reasoning, and each silent worker gets its own report rather than one aggregated message.

It's one report per delegation, not a repeating nag, and the watch is cancelled the moment `work-done` arrives (or the delegation is superseded, or the pane closes) — so a worker that reports normally costs nothing.

## The setting you asked about

`worker_response_timeout_minutes`, in your project's `.dot-agent-deck.toml`:

```toml
worker_response_timeout_minutes = 20
```

- Default **120** minutes, so it is already on — no setting is *required*
- Range `1`–`10080` (one minute to seven days); an out-of-range value falls back to the default rather than being clamped
- **`0` disables the detector entirely** — it does not mean "report immediately"
- There's also `DOT_AGENT_DECK_WORKER_RESPONSE_TIMEOUT_MS` if you want to override it per-run

**One trap worth knowing about**, because it's the most likely reason this appears not to work for you: `worker_response_timeout_minutes` is a **top-level scalar**, so in TOML it must appear *above* the first table header — above your first `[[modes]]` or `[[orchestrations]]`. Appended to the end of a config it silently becomes a key of whatever table came last, where it means nothing. The config still parses, `dot-agent-deck validate` still reports `Config is valid.`, and you quietly keep the 120-minute default with nothing telling you otherwise.

Full details are in `docs/idle-workers-and-notifications.md`.

[...]

PR #602, opened 2026-08-19
fix: signal when a worker exits without calling work-done

prageethw:
## Summary

A delegated worker whose process ends on its own — no crash, no explicit stop, no work-done — leaves both the idle-worker and silent-worker timeout watches armed for their full window (up to two hours) with nothing telling the orchestrator the worker was already gone. `pump_reader` already observes PTY EOF the moment the process dies; this signal now retires any armed `OutstandingDelegation`/`SilenceWatchRecord` for the pane immediately and reports a new fixed-text "worker exited without work-done" notice to the orchestrator, naming only the daemon pane id — no worker-authored content is interpolated.

Retirement is identity-gated on both sides (worker and orchestrator), closing a narrow pane-id-reuse window where a dead predecessor's sweep could otherwise drain a live successor's records.

Fixing this also surfaced a related identity-gate gap in the delegate write path: when a worker's identity couldn't be resolved (a plausible pane-id-reuse race), the code fell through to `write_and_submit_guarded_detailed` with a bare `None` expected identity, which that primitive treats as "no check" rather than "refuse". The call site now refuses the write itself instead.

## Relation to existing issues

- Extends #205 — addresses the *detection* half (a worker that stops reporting is now signaled instead of silently timing out). The *forwarding* half (relaying the worker's last message) is not implemented here, so #205 is only partially addressed and stays open.
- Related to #524 — that issue asks for `unregister_pane` to be called when the daemon observes a pane's process die, not only on an explicit stop, and specifically names the `clear = true` respawn race this PR's `is_agent_still_registered` gate answers. This PR builds that observation point and solves that race, deliberately scoped to retiring the two timeout watches; `unregister_pane` cleanup and `delegate_targets` staleness are left to #524.

## Notes for review

- `AgentPtyRegistry::spawn_agent` and `respawn_agent_for_pane` move from `&self` to `self: &Arc<Self>` (needed so the PTY reader thread can hold a `Weak` back-reference to the registry without creating a reference cycle against the registry's own `Drop`-triggered shutdown). This is source-breaking for any external consumer holding a bare `&AgentPtyRegistry`; all in-repo call sites are updated.
- Adds `scheduler/idle-worker/016` (real-PTY integration coverage of the natural-exit path) plus unit coverage for both sides of the identity gate, the delegate-path refusal, and the new notice composer.

Co-authored with Claude Sonnet 5.


vfarcic:
Independent verification by @vfarcic — no blocking findings in this PR.

**The code is sound; the hold-up is ordering, not this change.**

1. `main` is currently broken (`E0063: missing field 'pane_handed_over'`, a #597/#607 semantic merge conflict — see #609). This PR only inherits it through the merge.

2. ⚠️ **This PR's green checks are stale.** They ran at 03:11Z, before #607 introduced the field at 11:33Z. With `strict_required_status_checks_policy: false` nothing forces the PR up to date, so it would merge on that stale green onto a red `main`.

Verified locally against `origin/main` with #609's fix applied (everything else this PR as-is): fmt, clippy (`--workspace --all-targets --features e2e -D warnings`), build, test-fast, linkage-check, windows-cross all pass. The 4 red real-agent e2e tests all rerun green in isolation — contention under the full 8,624-test run, plus one instance of Devin API variance; `devin_live_001` was 3/3 green on this branch and 3/3 on a clean baseline.

Rule 12: no TUI↔daemon contract change — `daemon_protocol.rs` untouched, PROTOCOL_VERSION stays 7, no `.breaking.md` needed.

Follow-up (non-blocking): `scheduler/idle-worker/016` uses a shell-stub worker and a `cat` orchestrator. A PTY-attached real-agent variant would meet rule 4's "as a user actually sees it" bar and make the feature reel-eligible. Happy for that to be a separate issue.

Suggested order: merge #609 → re-run CI here → approve → merge.

vfarcic (approved):
Approving after a full `/verify-pr` run at `440d93e` (tree `a9b338d`).

**Gates (local, macOS arm64, merged tree):** fmt, clippy (`--workspace --all-targets --features e2e -D warnings`), `build --release`, linkage-check and `cargo audit` all clean; `test-fast` **2942/2942**. All 15 e2e failures were attributed away from this PR by explicit reruns — 2 (`sigterm_001/002`) reproduce identically at `origin/main` in the main checkout, 8 are this host's inability to authenticate a real agent (no `~/.claude/.credentials.json`; `delegate_014` says so itself), and 5 pass when rerun in isolation.

**CI:** the original green was stale (2026-08-19, pre-#609/#625). Branch refreshed; all four required checks — `build`, `build-macos`, `build-windows`, `security` — are green on the current head. `devbox` failed inside the third-party install action before any project code ran; it is transient and not required.

**Rule 12 — no contract change.** `daemon_protocol.rs` untouched, `PROTOCOL_VERSION` stays 7, and neither `OutstandingDelegation` nor `SilenceWatchRecord` is serialized or referenced there, so the new `worker_agent_id` never reaches the wire. `bugfix` categorization is right; no `.breaking.md` owed.

The cross-version manual test needed care: the PRD #103 build-id handshake makes an attaching TUI restart the daemon to match its own build, so a naive run silently collapses to a same-build pair (2x `Daemon listening` + a `SIGTERM` in the log). With the pairing genuinely forced via a debug build and `DOT_AGENT_DECK_BUILD_ID_OVERRIDE`, both directions pass: delegate routes and `work-done` arrives, and a v0.36.2 TUI renders the new notice correctly against the branch daemon. Killing a worker's process externally put the notice in the orchestrator pane ~5s later instead of after the timeout window.

**Design:** the `is_agent_still_registered` gate holds across all four kill paths — `close_agent`, `respawn_agent_for_pane`, `shutdown_all` and `shutdown_all_graceful` all remove the entry before killing, so no spurious notice on quit or on a `clear = true` respawn. Identity gates are pinned in both directions including the stale-`seq` bind. No silent deletions; `#[spec]` 474 -> 475. The CATALOG entry correctly carries no `[reel]` marker.

Follow-ups to file separately, none blocking: (1) `sigterm_001/002` are red on `main` itself and invisible to CI since e2e never runs there; (2) the orchestrator side of the silence-watch EOF drain is ungated on `orchestrator_agent_id`, accepted as a lost safety net rather than a misdelivery; (3) `idle-worker/016`'s fixed `sleep 2.0` race margin; (4) no real-agent test for this surface — the closest coverage is a shell stub with a `cat` orchestrator.

Issue #606, opened 2026-08-19
Delegating to a role just after closing its pane permanently breaks that role (respawn NotFound)


vfarcic:
Found while running the rule-12 cross-version test for #454, and **verified pre-existing** — it reproduces identically on released **v0.36.2**, so it is not caused by that branch. Filed so a later reader does not mis-attribute it to the ownership rewrite.

## Symptom

Closing a worker pane and delegating to that role shortly afterwards (~0.2s) permanently breaks delegation to that role for the rest of the session.

Observed on both daemons, identically:

```
WARN … delegate: respawn for clear=true failed; surfacing high-level notice in orchestrator pane
     and skipping the subsequent prompt write pane_id=sched-orch-2-r2 role=stub
     error=Agent sched-orch-2-r2 not found
```

The orchestrator pane receives `⚠ respawn failed for role 'stub' on pane sched-orch-2-r2 (see daemon log for details)`, the pane row disappears from `dot-agent-deck daemon status`, and every later delegate to that role is rejected with `delegate from pane … reached no worker for role(s): stub …`.

## Cause

`close_agent` removes the registry entry **before** spending its termination grace (`src/agent_pty.rs:4863-4870`). So a concurrent `respawn_agent_for_pane` — the `clear=true` delegate path — looks the pane up during the grace window, finds nothing, and fails with `NotFound` rather than waiting for the close to finish or re-creating the pane.

## Reproducer

With an orchestration whose worker role is a stand-in (`cat` or a sleeper), close the worker pane and issue `delegate --to <role>` about 200ms later. The role is then unreachable for the remainder of the session.

## Why it matters

The failure is silent from the user's point of view apart from one notice line, the role stays dead, and the only recovery is recreating the orchestration. The delegate path already has a "surface a notice and skip the prompt write" branch for this, so the failure is handled but not repaired.

## Suggested direction

Either have `close_agent` retain the registry entry until termination completes (so a concurrent respawn waits rather than 404s), or have `respawn_agent_for_pane` treat `NotFound` during an in-flight close as "spawn a fresh pane for this role" instead of a hard failure. Whichever way, the role should not end up permanently unreachable.

Issue #605, opened 2026-08-19
RUST_LOG=dot_agent_deck=debug is silently discarded by an appended default directive


vfarcic:
Found while running the rule-12 cross-version test for #454. Pre-existing; unrelated to that branch.

## Symptom

`RUST_LOG=dot_agent_deck=debug` has **no effect** — the process still logs at `info`, silently. There is no warning and no error, so it reads as "debug logging produced nothing interesting" rather than "your filter was discarded".

## Cause

`init_logging_from_env` (`src/main.rs:1394`) builds the filter as:

```rust
EnvFilter::from_default_env().add_directive("dot_agent_deck=info")
```

`add_directive` **overrides** an identical-target directive coming from `RUST_LOG`, so a user-supplied `dot_agent_deck=debug` is replaced by the appended `dot_agent_deck=info`. Only *more specific* targets survive — which is why the repo's own tooling uses forms like `dot_agent_deck::daemon=debug`.

## Why it matters

`dot_agent_deck=debug` is the form a newcomer will reach for first, and it is the form most projects support. Losing it costs real debugging time: during the #454 cross-version run it produced an apparently-quiet daemon until the filter was switched to

```
dot_agent_deck::agent_pty=debug,dot_agent_deck::daemon=debug,dot_agent_deck::state=debug,dot_agent_deck::spawn=debug,dot_agent_deck::prompt_delivery=debug
```

which immediately surfaced the lines being looked for.

## Suggested direction

Only apply the default `dot_agent_deck=info` directive when `RUST_LOG` does not already constrain that target — or build the default filter first and let `RUST_LOG` layer on top of it, which is the usual precedence. Failing that, emit a warning when a user-supplied directive is discarded, so the override is at least visible.

Issue #601, opened 2026-08-19
SessionStart pre-authorizes any pane id, so managed_pane_ids is not an ownership proof


vfarcic:
Found independently by a reviewer and an auditor while working #454. **Pre-existing and unchanged by that PR**, which is why the hardening is filed here rather than folded in — but #454's PR corrects source comments and a changelog fragment that had overclaimed the guarantee, so the prose and the code now disagree deliberately and this issue is the record of that gap.

## The gap

`AppState::apply_event` admits a non-`SessionStart` event only for a pane in `managed_pane_ids` — that is the admission control which exists so an arbitrary process cannot drive daemon session state for a pane the daemon does not own.

But the `SessionStart` arm (`src/state.rs:3905-3922`, insert at `3919-3921`) inserts **any** pane id that does not begin with the reserved `__dead-slot__-` prefix into `managed_pane_ids`, with **no check against `AgentPtyRegistry`**. So a same-UID process can pre-authorize an arbitrary pane id by sending one forged `SessionStart`, after which every subsequent non-`SessionStart` event for that id is admitted as owned.

`run_hook_loop` elsewhere already logs exactly this case, describing a foreign agent that "will register a card with no local pane" — so the behaviour is known and partly intentional (it is what lets an external/foreign agent surface a card at all).

Consequence: `managed_pane_ids` does not mean, and has never meant, "exactly the panes this daemon spawned". Any reasoning that treats it as proof of ownership is unsound.

## What an attacker achieves

Constrained to the **same OS user**. For an existing pane, agent and pane ids are enumerable through the owner-only `ListAgents` socket and `daemon status`, so status, active-tool, prompt-history and generation state can be spoofed on a card. The forged event does **not** confer an orchestration role — `pane_role_map`, `pane_orchestration_map` and `orchestrator_pane_ids` are written only by `register_orchestration_role`, so `handle_delegate` cannot be induced to resolve the pane.

## Why this is not being treated as an urgent vulnerability

Both sockets are owner-only; `StartAgent` already treats the same OS user as having arbitrary execution capability; an attach-socket peer can already attach to and write to agents directly; and hook broadcasts reached attached TUIs before the daemon-side admission decision even prior to #454. So this is a hardening / design-clarity issue, not a cross-user privilege escalation.

## Suggested direction

Decide what `managed_pane_ids` is actually meant to assert, then make the code and the prose agree:

[...]

Issue #600, opened 2026-08-19
Partial orchestration spawn leaves orphaned children with no handle to close them


vfarcic:
Found by a reviewer while working #454. **Pre-existing — not introduced by that PR**, and deliberately left out of it because it is a question about `spawn`'s error semantics rather than about pane admission.

## Symptom

`crate::spawn`'s orchestration loop calls `spawn_one` with `?` for each role (`src/spawn.rs:539-571`, `src/spawn.rs:586-598`). If role N fails after earlier roles have already spawned successfully, `spawn` returns immediately and:

- the already-launched agents remain alive in `AgentPtyRegistry`;
- no `SpawnHandle` is returned, so the caller has nothing to close them with;
- the scheduler / issue-dispatch caller reports the spawn as failed.

The user is left with orphaned child processes and no surfaced tab or handle through which to close them. A later reconnect can still encounter the registry records.

## Reproducer

Configure a two-role orchestration with role 1 as `/bin/cat` and role 2 as a nonexistent bare executable. Role 1 stays alive in the registry after role 2 returns `SpawnError`.

## Why it is filed separately

#454's PR makes state registration happen per-successful-spawn, so daemon `AppState` and the registry stay consistent on a partial failure. That closes the *admission* half. It deliberately does **not** roll back already-spawned agents, because deciding whether a partial orchestration spawn should be all-or-nothing — versus surfacing what did start so the user can act on it — is a design question with its own tradeoffs, and predates that PR.

## Suggested direction

Either roll back already-spawned agents on a later role failure, or return a partial `SpawnHandle` so the caller can close or surface them. Whichever way, the choice deserves a test covering the partial-failure path.

PR #607, opened 2026-08-19
fix(daemon): restore live status for ordinary panes in daemon status and on reconnect

### LINKED ISSUE #454: daemon status and PRD #162 reconnect show no live status: ListAgents live-join yields None for TUI-spawned panes
Found while verifying #420 (`dot-agent-deck daemon status`). **Pre-existing — not introduced by that PR**, but #420 is the first user-facing surface that exposes it as an empty column.

## Symptom

Two surfaces disagree about the same agent:

- The TUI card renders `● Thinking`.
- `dot-agent-deck daemon status` prints `STATUS=-  TOOL=-`, and `--json` omits the keys entirely.
- After a TUI detach/reattach, the card restores as `● Idle`, not `Thinking` — so PRD #162's own reconnect restore is affected too, not just the new CLI.

## Reproducer

Against a fresh daemon in an isolated sandbox, with one TUI-spawned pane:

```bash
DOT_AGENT_DECK_PANE_ID=1 DOT_AGENT_DECK_AGENT_ID=1 dot-agent-deck agent-event --type running
dot-agent-deck daemon status
```

Both env vars are exactly what the daemon injects into a spawned pane (confirmed via `/proc/<pid>/environ`). The wire payload was sniffed off the hook socket and does carry the ids:

```json
{"session_id":"1-session","agent_type":"pi","event_type":"thinking",...,"pane_id":"1","agent_id":"1"}
```

## Where it goes wrong

`AgentRecord.live` is populated by the join in `src/daemon_protocol.rs:1183-1196`, which requires BOTH `s.agent_id == Some(record.id)` AND `s.pane_id == record.pane_id_env`. In the repro, `daemon status` reports `AGENT=1` and `PANE=1` (i.e. `record.id == "1"`, `record.pane_id_env == Some("1")`) and the event carried the same pair — yet `live` comes back `None`.

**Root cause not established.** I stopped at localization once it was clear the behaviour predates #420: the identical flow against an untouched **v0.35.9** daemon behaves the same way. So it is either the join itself or how `SessionState.agent_id` is keyed/updated on ingest (`src/state.rs` around the session-upsert path, ~3236-3240 / ~3634).

## Why the tests miss it

`tests/daemon_status.rs` (and the PRD #162 coverage) construct a **raw** `AgentEvent` with `agent_id` set to the registry's spawn id, so the join is guaranteed to succeed in-test. No test drives the real `agent-event` CLI path against a TUI-spawned pane.

## Why it matters

`changelog.d/daemon-status-command.feature.md` advertises "live status, and active tool" — both render `-` in an ordinary single-pane session.

### LINKED ISSUE #455: daemon status --json emits active_tool.detail, which the human table deliberately withholds
Found while verifying #420.

`src/daemon_status.rs` serializes `Option<ActiveTool>` whole into the `--json` document, so `active_tool.detail` rides along. The human table deliberately prints only the tool **name** (`src/daemon_status.rs:158-160`), and `daemon/status/004` explicitly scopes the no-prompt-text requirement to the human view.

Per `src/hook.rs:183-228`, `detail` is:

- `Bash` / `shell` / `exec` → the **first line of the command being run** (truncated to 120 chars)
- `Read` / `Edit` / `Write` → the **file path**
- `Grep` / `Glob` → the **search pattern**
- `Agent` → the **subagent description**
- generic fallback → the **first string-valued key** of arbitrary tool input

This is not a privilege escalation: the attach socket is per-user and the TUI already renders `detail`. But #420's stated contract is a privacy one ("It never prints prompt text"), and the two output modes currently disagree about what a status query may reveal.

**Decide deliberately:** either drop `detail` from the JSON, or keep it and document that `--json` is strictly more revealing than the table. Whichever way, extend `daemon/status/004` to cover `--json` so the choice is pinned by a test.

### LINKED ISSUE #459: daemon_status comments cite a gitignored design doc that is not in the repo
Found while verifying #420.

Three source comments cite the feature's design rationale as `.dot-agent-deck/47-status-query-design.md` "in the root checkout":

- `src/daemon_status.rs:6`
- `src/main.rs:1593`
- `tests/daemon_status.rs:3`

`.dot-agent-deck/` is gitignored (`.gitignore:6`), so that file is not in the repository and no reader of the merged source can follow the reference. Several comments lean on it as the justification for non-obvious choices (why no `PROTOCOL_VERSION` bump, why prompt text is excluded, why the 3s deadline never retries).

Either inline the load-bearing rationale into the comments, or drop the path so it does not read as a followable citation.

vfarcic:
Fixes #454. Also fixes #455 and #456, and closes #459.

## The bug

`dot-agent-deck daemon status` printed `STATUS=-  TOOL=-` for an agent whose dashboard card was visibly `● Thinking`, and `--json` omitted the `status` and `active_tool` keys entirely. The same root cause broke PRD #162's reconnect restore: detaching and reattaching rebuilt a busy agent's card as `● Idle` rather than the status the daemon already knew. Pre-existing — it reproduces against v0.35.9 — and merely exposed by the `daemon status` CLI.

## Root cause

The issue localized this to the `ListAgents` live-join. The join was not the problem; the failure is earlier.

The daemon keeps **two** views of the panes it owns: `AgentPtyRegistry`, which spawns and tracks them, and its own `AppState`, which ingests their lifecycle reports. Only `AppState::register_orchestration_role` ever announced a pane to the second one, so an ordinary dashboard pane existed in the registry and nowhere else. `AppState::apply_event` admits a non-`SessionStart` event only for a pane in `managed_pane_ids` — and the real `agent-event` CLI emits `running`/`waiting`/`finished`, never `SessionStart`. So every report an ordinary pane made was dropped, and `ListAgents` had nothing to join.

That also explains why orchestration panes never had this bug, and why the existing tests missed it: they construct a raw `AgentEvent` with `agent_id` hand-set to the registry's spawn id, so the join could not fail in-test. No test drove the real CLI path against a TUI-spawned pane. One now does.

A second seam had the identical gap and was not in the issue: `crate::spawn`'s **single-agent** branch, serving scheduler and issue-dispatch cards. Left alone it would have shipped the same bug for every scheduled card.

## How it is fixed

Three iterations, because the first two were structurally unable to hold:

1. **Register at the spawn seams.** Rejected in review: registration is *late* (the child can report before its spawner returns) and *permanently stale* (a naturally-exited child notifies nothing, and `agent_records()` filters exited entries, so even a later `StopAgent` could not clean up).
2. **Resolve ownership live from the registry via an oracle.** Fixed staleness, but the oracle was **pane-keyed** — a pane is a slot, not an identity, and the registry deliberately lets a new agent take over a dead one's pane. No pane-keyed answer can say which occupant an event belongs to.
3. **Key ownership on the generation** — the shipped design.

A **generation** is one spawn: a registry id, from the instant its spawn is *reserved* (before the child is forked) until its record is *reaped*. Registry ids only increment and are never reused. An event is admitted when the registry answers `Owned` for the generation it names.

- The answer is **tri-state** (`Owned` / `Unclaimed` / `Unknown`). A poisoned lock or dropped `Weak` yields `Unknown`, and every caller that reads the answer as a grant denies on it; the one caller that reads *absence of a claim* as its own grant requires a positive `Unclaimed`. A process that installed **no** oracle (the TUI, bare `AppState`) is a distinct fourth case and keeps the historical rule unchanged.
- Ownership survives the child's death until the pane changes hands or the record is reaped, so a generation's own final `SessionEnd` — written in the same breath as its exit — still lands. Clock-free: no timer, no grace window, no reaper.
- Disownership is **monotone**. `pane_handed_over` is recorded on the disowned record inside the same lock hold that publishes a successor and is never cleared, so a successor exiting in turn cannot hand the pane back.
- `managed_pane_ids` is **demoted**: a registration confers nothing while the registry claims the pane, however the pane got into the set.
- A pane whose cleanup is in flight is **held** (`PaneCleanupHold`), excluding it from all spawns, so the authorization a pane-scoped teardown was granted cannot go stale under the teardown acting on it.

## Scope: why #455, #456 and #459 are here and #457, #458 are not

The issue asked for this decision to be explicit.

**Included, because this fix changes their facts:**

- **#455** (`--json` leaks `active_tool.detail`) — before this fix `live` was always `None` for ordinary panes, so `active_tool` was never emitted and the field never leaked. **This fix activates the disclosure.** Shipping #454 alone would knowingly land it. `detail` is now projected away for both output modes; `SCHEMA_VERSION` goes 1 → 2 for the field removal, keeping `active_tool` an object with a `name` key so a v1 consumer reading `active_tool.name` still works.
- **#456** (pin the `--json` field names) — the `status` and `active_tool` keys only materialise *because* of this fix, so this is the moment to pin them, and the schema test also guards against regressing back to omitted keys. `daemon/status/002` now pins the full public key set and every supported status string, and asserts the exact `schema_version`.
- **#459** (comments cite a gitignored design doc) — a three-line cleanup in files this PR already edits; a separate PR would re-touch the same lines.

**Deferred to their own PRs:**

- **#457** (document `daemon status` in the published docs) — docs-only, no dependency on this fix, and deserves the validated-examples docs flow rather than riding a bugfix.
- **#458** (the 3s timeout branch has no test) — an entirely disjoint code path this change does not touch.

**The TUI↔daemon contract did not change. `PROTOCOL_VERSION` stays 7; no `.breaking.md`.**

`AttachRequest`, `AttachResponse` and `AgentRecord` are byte-identical to v0.36.2; `src/event.rs` is untouched; no new attach request. `AgentOwnership` is a process-internal trait object and is never serialized.

Verified dynamically, not only statically, in an isolated sandbox (sockets, `HOME`, state dir **and** `DOT_AGENT_DECK_LOG` all redirected, isolation confirmed by diffing the real log afterwards). Both directions — branch TUI against a v0.36.2 daemon and v0.36.2 TUI against the branch daemon — completed delegate routing, work-done and status hooks end to end, verified **in the target pane's PTY**, not merely in a daemon log line.

One same-wire behaviour change is observable and is deliberate: while a `StopAgent` tears a pane down, a concurrent `StartAgent` on that pane is refused with `DuplicatePaneId` in a window where it previously succeeded. Measured: zero for an ordinary child, ≤3s for one ignoring `SIGTERM`, self-clearing ~0.15s after the teardown reply. A real v0.36.2 TUI driven into it shows one recoverable command-bar error and stays fully responsive, with the pane id immediately reusable. The change moves toward safety — on the old daemon that spawn *succeeds* and leaves a live generation whose pane state the concurrent stop then deletes.

## Tests

New `#[spec]` coverage: **`daemon/status/005`** (the real `agent-event` CLI path with only the env pair the daemon injects — the issue's reproducer), **`session/live/011`** (reconnect restores live status), **`session/live/012`** (PTY-attached L2: real `StartAgent`, real CLI event, real TUI, detach, fresh TUI, card still `Thinking`), **`scheduler/spawn/007`** (the `RunNow → spawn_or_reuse → spawn` seam). Modified: `daemon/status/002` and `daemon/status/004`.

`session/live/006` could not have caught this — it seeds daemon state with a synthetic `SessionStart`, which has always auto-registered an unknown pane, so it passed straight through the bug.

Every fix in the final commit carries a mutation-verified regression test. The most telling one: under the reverted `StopAgent` fix, the new test fails **and** the pre-existing `stopping_a_retired_agent_leaves_its_panes_new_occupant_alone` passes — direct evidence the old test structurally could not catch that ordering.

## Gates

- `cargo fmt --check` — clean
- `cargo test-fast` — **2752 / 2752**
- `cargo test-e2e` (recorded) — **8,393 run, 8,392 passed**; the single failure is a known real-Claude flake that passes its isolated rerun

## Follow-ups filed, not silently dropped

Found during review/audit/cross-version work, all pre-existing and deliberately out of scope: **#600** (partial orchestration spawn leaves orphaned children), **#601** (`SessionStart` pre-authorizes any unclaimed pane id, so `managed_pane_ids` is not an ownership proof — this PR narrows the source comments and changelog that overclaimed it, and neither fixes nor widens the behaviour), **#605** (`RUST_LOG=dot_agent_deck=debug` silently discarded), **#606** (delegating to a role just after closing its pane permanently breaks that role).

One known residual, ruled non-blocking in review: a paneless agent whose `HistoryOnly` declaration is lost still falls back to `agent_writable`'s `Live` default. Pre-existing, present on `main` and unchanged here.


prageethw (approved):
This is a thorough, carefully-reasoned fix and it shows the scars of real internal review (round-2/round-3 audit findings are visibly folded in, not just claimed). Read the core of it directly rather than trusting the writeup: the generation-keyed ownership model in `AgentPtyRegistry::generation_ownership` correctly returns the tri-state `Ownership` (Owned/Unclaimed/Unknown) rather than a bool specifically so a poisoned-lock/dropped-Weak failure can't be conflated with genuine non-ownership at any call site that treats absence as a grant — that distinction is applied consistently everywhere I checked (`owns_pane_event`, `admits_paneless_event`, the SessionStart auto-registration match arm, `generation_disowned`). The `pane_handed_over` monotone flag correctly fixes the resurrection bug the round-2 live-lookup version had (A exits, B takes the pane, B exits -> A's generation must not silently become owner again). `SpawnReservation`/`PaneCleanupHold` are sound RAII guards with correct poisoned-lock handling in Drop. The pre-fork pane-exclusivity check (reservation + cleanup_holds + live agents) closes the concurrent-StartAgent race the post-fork check alone couldn't. Scope decisions (#455/#456/#459 included because this fix activates/creates the facts they depend on; #457/#458 correctly deferred as unrelated) are well-argued, and follow-up issues were filed for everything found-but-out-of-scope (#600/#601/#605/#606) rather than silently dropped. Cross-version contract verified dynamically in both directions against a real v0.36.2 daemon, including flagging the one deliberate same-wire behavior change (DuplicatePaneId during a StopAgent race window) with measured bounds. CI green (8392/8393 e2e, the one failure a known flake with a passing isolated rerun).

Issue #513, opened 2026-08-11
A PR can merge on CI that predates the gate which would have caught it — and rapid merges cancel main's verification


vfarcic:
Redirected from a narrower follow-up I first proposed while verifying #464. Investigating #511 showed the Windows test bug was a *symptom*; the mechanism that let it reach `main` is general and will recur on any platform.

## What happened

#511's broken test sat in #472 and merged green. Verified timeline:

| When (UTC) | What |
|---|---|
| 2026-08-10 17:27 | #472's CI passes on head `3d1b6f4`. `build-windows` ran `cargo nextest run` **without `--workspace`**, so `xtask/*` tests never executed |
| 2026-08-11 00:57 | `fbfb12f` — "ci: select the whole workspace in the test gates (#489, #470)" — adds `--workspace` |
| 2026-08-11 16:09 | **#472 merges on that ~23h-old run**, which predates the new gate |
| 2026-08-11 16:21 | `main` at `3cf4b68` executes the `xtask` tests on Windows for the first time → **red** |

`build-windows` was fully capable of catching this. It never got the chance, because nothing required #472's CI to be re-run after the gate that selects those tests had changed.

### 1. Stale PR CI is merge-eligible

`main` has **no required status checks** (see CLAUDE.md rule 8 for why they were removed — CI's direct pushes to `main` hit `GH006`), so nothing compares a PR's CI age against the base. A day-old green run merges exactly like a fresh one.

The targeted setting is GitHub's *require branches to be up to date before merging*, which in a ruleset is `strict_required_status_checks_policy` — a **parameter of the `required_status_checks` rule**, so it cannot be adopted standalone. Two things to weigh, both already noted in CLAUDE.md rule 8:

- Required checks are now *less* dangerous than when they were removed, since `RELEASE_TOKEN` gives CI an admin identity the ruleset lets bypass — **but that combination has never been tried**, so it is unverified rather than safe.
- It creates a real asymmetry: @prageethw has **write**, not admin, so required checks would genuinely gate his merges while remaining bypassable for the owner. That is a governance decision, not a config tweak.

### 2. Rapid merges cancel `main`'s own verification

`ci.yml:21-23`:

```yaml
concurrency:
  group: ci-${{ github.ref }}
  cancel-in-progress: true
```

[...]

PR #609, opened 2026-08-19
fix(agent-pty): set pane_handed_over on the synthetic test agent — unbreak main

vfarcic:
## `main` is RED — this unblocks a pending release. Please prioritise.

`main` at 776fab8 does not compile. Four of eight jobs in run [32249065340](https://github.com/vfarcic/dot-agent-deck/actions/runs/32249065340) fail — `build`, `build-macos`, `build-windows`, `windows-cross-check` — all on one hard compile error, not a flake:

```
error[E0063]: missing field `pane_handed_over` in initializer of `agent_pty::RunningAgent`
    --> src/agent_pty.rs:6011:13
```

It surfaces in the **lib-test** target, which is why it takes out all four: `cargo clippy --workspace --all-targets --features e2e` on Linux, `cargo nextest run --workspace` on macOS/Windows, and the Windows cross type-check.

## Root cause: a semantic merge conflict between two independently-green PRs

Neither PR touched a line the other touched, so neither CI run could have seen the break:

- **#607** (`715bc95`, issue #454) added `pub pane_handed_over: bool` to `RunningAgent` and updated **every construction site that existed at that time**.
- **#597** (`5c9d08b`, issue #581) added a **new** construction site — the `insert_test_agent` seam, the synthetic wedged-child test agent — which of course does not set a field that did not exist on its branch.

Both merged green; the combination is red. This is exactly the failure mode tracked in **#513** ("A PR can merge on CI that predates the gate which would have caught it"). Not fixing #513 here — this is the unbreak only.

## The fix

One field, at the one site that lacks it.

`false` is the *correct* value, not just the one that compiles. `pane_handed_over` is a **monotone latch**: it is set at PUBLISH (src/agent_pty.rs:4323-4330), only when a *successor* takes over a pane this record still holds, and it is never cleared — that permanence is the whole point, so a retired generation cannot get its pane back once both records die (read site: src/agent_pty.rs:5594-5605). The synthetic agent carries `pane_id_env: None`, so it holds no pane at all and nothing can ever hand one over. This matches the production site's own birth value and its comment at src/agent_pty.rs:4374-4378: *"a fresh generation has not been handed over yet."*

## Other sites, and the mirror-image hazard — both checked

`git grep -n "RunningAgent {"` gives five hits, and all five are accounted for:

| Site | Kind | Status |
|---|---|---|
| src/agent_pty.rs:1569 | struct declaration | n/a |
| src/agent_pty.rs:1730 | `impl` block | n/a |
| src/agent_pty.rs:4353 | production construction | already correct (`pane_handed_over: false`) |
| src/agent_pty.rs:5201 | **destructuring** | already correct — see below |
| src/agent_pty.rs:6011 | test construction | **fixed here** |

The mirror-image hazard is already covered: the respawn destructuring at :5201 binds `pane_handed_over: _` **explicitly** rather than eliding with `..`, so it is exhaustive and a future field addition still breaks loudly there rather than silently dropping a field. `RunningAgent` is constructed nowhere outside `src/agent_pty.rs` — every other `RunningAgent*` reference in the tree is `RunningAgentsSummary`, an unrelated type.

## Verification

Reproduced first on unpatched `main` (single E0063, as quoted above), then all gates run locally on the fix:

| Gate | Result |
|---|---|
| `cargo clippy --workspace --all-targets --features e2e -- -D warnings` | clean (exit 0) |
| `cargo test-fast` | **2887 passed, 0 failed, 0 skipped** |
| `scripts/windows-cross-check.sh` | clean (exit 0) |

The last one was run locally too, since `windows-cross-check` is one of the four red jobs. The four tests that actually drive `insert_test_agent` (src/agent_pty.rs:10472, :10528, :10611, :10663 — #581's wedged-reap shutdown coverage) are in that passing run.

## Changelog fragment: deliberately none

Both colliding commits post-date `v0.36.2`, and each already carries its own fragment (`changelog.d/454.bugfix.md`, `changelog.d/581.bugfix.md`). So the two features that collided will be described in the next release notes, while the collision itself never reached a released build — no user could have encountered it. A third fragment would describe a defect that only ever existed between two unreleased commits on `main`.

## Scope

Deliberately minimal and surgical — an urgent unbreak, not a refactor. No restructuring of the helper, no renames, no unrelated lint fixes. The diff is five added lines: one field and its rationale comment.

### DIFF
--- a/src/agent_pty.rs
+++ b/src/agent_pty.rs
@@ -6032,6 +6032,11 @@ impl AgentPtyRegistry {
                 pty_rows: 24,
                 pty_cols: 80,
                 exited: Arc::new(AtomicBool::new(false)),
+                // Issue #454: `false` is the birth value — the flag latches to
+                // `true` only when a *successor* takes this record's pane, and
+                // this synthetic agent holds no pane at all (`pane_id_env:
+                // None`), so nothing can ever hand one over.
+                pane_handed_over: false,
                 pending_seed: None,
                 seed_delivered_native: false,
             },


vfarcic:
Flagging urgency: this unbreaks `main`, and only @prageethw can approve it.

`main`'s tip (776fab8) currently fails CI — `build`, `build-macos`, `build-windows` and `windows-cross-check` are all red with the same `error[E0063]: missing field 'pane_handed_over' in initializer of 'RunningAgent'` (run 32249065340). Reproduced locally with `cargo clippy --workspace --all-targets --features e2e`.

Cause is a semantic merge conflict between two PRs that were each green alone: #597 added the `insert_test_agent` test seam that constructs `RunningAgent`, and #607 added the `pane_handed_over` field but branched before #597 landed, so it never saw that initializer. Textually clean, semantically broken — a burst of merges between 11:33–11:44Z cancelled each other's `main` runs, so it only surfaced on the last one.

Consequences while this sits open:
- the pending release cannot be cut from `main`
- #602 is blocked behind it

I authored this, so I can't approve it. @prageethw — one-line fix, all 9 checks green.

prageethw (approved):
Verified: the single added field (`pane_handed_over: false`) matches the exact `E0063` compile error breaking `main` right now, the rationale for `false` as the correct (not just compiling) value is sound given this synthetic agent holds no pane, and all 5 `RunningAgent` construction/destructuring sites in the file were checked. Minimal, correctly scoped, CI green. Approving — this unblocks CI on several other open PRs.

prageethw:
Heads up — this is currently blocking ~9 other open PRs from going green (#464, #615, #616, #618, #619, #620, #623, #624, #625 all fail the same 4 required checks, confirmed as this exact `pane_handed_over` compile error inherited from `main`). This PR is approved and green, so whenever it's convenient to merge it'll unblock the rest once each is rebased.

PR #613, opened 2026-08-19
feat(ui): pane scrolling is per-agent, and the deck now says so (#611)

### LINKED ISSUE #611: PRD: Pane scrolling is per-agent, and the deck should be honest about it (codex panes cannot scroll)
## Reported symptom

Scrolling a pane with a two-finger trackpad swipe works on a `claude` pane and does nothing at all on a `codex` pane — "as if it's locked". Reported on macOS/Ghostty; the mechanism is agent-side and not OS-specific. `fn`+Up/Down (PageUp/PageDown) behaves the same way: works on claude, nothing on codex.

## Diagnosis

**No line of deck code is misbehaving.** Both branches of `scroll_focused_agent_pane` (`src/ui.rs:17719`) do exactly what they were designed to do. The two agents use *opposite* scrolling models, and the deck supports one of them well and the other not at all.

| | claude | codex |
|---|---|---|
| Mouse tracking enabled | all four: `1000h`, `1002h`, `1003h`, `1006h` | **none** — only `1004h` (focus reporting, not mouse) |
| Alternate screen (`1049h`) | — | **no** |
| Who owns the transcript | **the app** | **the terminal** |
| Responds to PageUp / `ESC[5~` | yes | **no** |

- **claude is app-managed.** It requests mouse tracking precisely so it can scroll its own transcript. The deck forwards the wheel while in `PaneInput`, claude scrolls itself. Works.
- **codex is terminal-managed.** It requests no mouse events, so no terminal ever sends it any. It expects the *terminal* to hold scrollback — while contributing none.

### Codex contributes zero scrollback, measured

A real 139 KB interactive codex session (v0.147.0) was captured with `script -qfc codex` and replayed through the repo's own `vt100` 0.16 at several parser heights, asking for 3 lines the way `FOCUSED_PANE_SCROLL_LINES` does (`vt100` clamps `set_scrollback` to the real buffer, so the readback is authoritative):

| parser height | codex buffer depth |
|---|---|
| **51 rows (what codex rendered for)** | **0 lines** |
| 44 rows | 182 lines |
| 24 rows | 187 lines |
| 12 rows | 225 lines |
| *control: 2000 plain streamed lines @ 51 rows* | *1950 lines* |

Codex sizes its scroll regions to the exact terminal height (164 DECSTBM calls: `ESC[1;50r`, `ESC[40;51r`, …) and repaints its whole transcript in place — 4971 cursor-position sequences against 95 newlines. **It never lets a line scroll off the top**, so there is nothing for any terminal to retain. The 182–225 lines at smaller heights are an artifact of the parser being shorter than codex believed the terminal to be; they do not occur in reality. The control proves the parser and measurement are sound.

### Confirmed: codex's transcript is unscrollable in Ghostty too

Scrolling up during a long codex session in plain Ghostty reaches the content that was on screen *before* codex started — never an earlier part of the conversation. Codex does not use the alternate screen, so the pre-codex screen stays in the host terminal's scrollback above its UI. That is what appears to be "scrolling working outside the deck". A deck pane spawned for the agent has no pre-existing content, so nothing moves at all.

**The deck therefore cannot provide scrollback for such an agent.** Two routes were considered and rejected: retaining overwritten frames (codex repaints the whole transcript region, so frames are near-duplicates — paging redundant screenshots, not scrollback), and handing codex a taller virtual terminal than the pane (codex lays out against the height it is told, so its composer would render off-screen). A real fix can only happen upstream in codex — filed separately.

### Hypotheses falsified along the way

Recorded so nobody re-treads them:

1. *Codex enables mouse tracking and ignores the forwarded wheel* — no, it enables none, so the forwarding branch is never taken.
2. *Codex's pane is scrollback-starved at startup while claude's is not* — the startup measurement showed the opposite (codex 9 lines, claude 0); both startup captures were unrepresentative.
3. *Mouse-mode is seeded per agent type* — no; only test seams take a bool, the real spawn path always seeds `false`.
4. *`ESC[5~` reaches codex and scrolls it* — no, codex ignores PageUp entirely.

Note that `~/.local/state/dot-agent-deck/deck.log` has **no scroll or mouse instrumentation at all**, which is why none of this was diagnosable from artifacts.

### M1 — Docs tell the truth about per-agent scrolling
`docs/keyboard-shortcuts.md` currently reads as though pane scrolling works uniformly (see the "Scrolling back through a pane" section). State the per-agent reality: app-managed agents (mouse tracking; the agent scrolls its own transcript) versus terminal-managed agents that contribute no scrollback, and that for the latter the deck has nothing to scroll — by wheel or by key, in any mode.

### M2 — The deck says so on screen instead of no-opping silently
Today the wheel and PageUp both do nothing, with no feedback, on a pane whose buffer is empty. Working-as-designed still reads as broken. Add a brief dismissable on-screen notice, modelled on the existing `COMMAND MODE · Ctrl+D to type` banner.

Design constraints, decided deliberately:

- **Reactive, not proactive.** Fire when a scroll is *attempted* and cannot land — not on pane focus or spawn. Proactive nags every codex pane at startup including the many times nobody meant to scroll. This matches the existing banner philosophy already documented for command mode ("a key that isn't bound to anything keeps it up, because that is the moment you most likely thought you were talking to the agent").
- **Must not swallow the keystroke.** Visual only, cleared by a timer or by the next key *without consuming it* — otherwise this fix introduces a dropped-character bug in `PaneInput`.
- **Detect the condition, not the agent name.** Trigger on "this pane's scrollback is empty AND the agent has produced substantial output", which distinguishes "nothing yet" (a claude pane one second after spawn) from "this agent never produces any". Do not special-case Codex in a registry entry that will go stale — codex is not special, its *rendering model* is, and a future repainting agent should be covered automatically.
- **Message says what is true and does not mislead.** Do not suggest PageUp as an alternative; it fails for the same reason.
- **Scope.** Dashboard panes and the orchestrator pane. Worker panes in an orchestration are mostly autonomous; with the reactive design this largely takes care of itself, since a pane nobody scrolls never shows it.
- **Not behind the `experimental` flag** (CLAUDE.md rule 9 — decided): it is a small explanatory message that makes existing behaviour legible, not a new mode. Shipping it flagged-off would hide the explanation from exactly the confused user it exists for.

### M3 — Derive mouse state correctly (`scan_mouse_mode`)
`scan_mouse_mode` (`src/embedded_pane.rs:2722`) has **zero tests** and three defects, each verified directly:

1. **Combined SET form undetected.** Apps commonly emit `ESC[?1000;1002;1006h`; none of the four single-mode patterns is a substring of it, so mouse mode goes entirely undetected.
2. **Enable beats disable within a chunk**, regardless of byte order — all four enable patterns are checked and `return`ed on before any disable is considered, so a child turning reporting *off* in the same read stays flagged on.
3. **Split sequences invisible.** `contains_bytes` scans a single PTY read with no carry-over, so `ESC[?100` + `2h` across a chunk boundary is missed. PTY read boundaries are arbitrary.

This is not cosmetic and it is coupled to M2: `scan_mouse_mode` derives `mouse_mode_enabled`, the exact input to the branch M2 reacts to. An agent that enabled tracking via the combined form would be mis-derived, the deck would wrongly take the scrollback branch, and it would produce **this same "locked" symptom for an agent that would otherwise have worked**. claude emits all four modes, so it is the agent most exposed.

`mode/scroll/001` pins the routing matrix but feeds it a *synthetic* boolean — its catalog entry explicitly does not assert "real terminal mouse-report decoding" — so nothing anywhere covers this function.

## Testing

Per CLAUDE.md rule 4, M2 is user-visible TUI behaviour and needs harness coverage; extend the existing `mode/scroll/00X` family rather than adding a near-duplicate. M3 is pure unit tests. A regression fixture should be **synthetic** — a byte stream reproducing the DECSTBM-plus-repaint pattern — rather than the real captured session, which contains actual conversation content.

## Not in scope

Making codex's transcript scrollable. That requires codex to use the alternate screen, enable mouse tracking, or bind a scroll key, and belongs upstream.

vfarcic:
## What it does

Three milestones:
- **M1** — docs stop claiming pane scrolling works uniformly across agents.
- **M2** — adds a reactive on-screen notice when a scroll cannot land.
- **M3** — rewrites `scan_mouse_mode` from substring matching to a real escape-sequence scan, and gives it its first unit tests.

Closes #611.

## Stacked on an unmerged PR

The base commit `55bf6d3` is **PR #609's one-line fix, not ours**. `origin/main`'s tip does not compile its lib-test target — `E0063: missing field pane_handed_over`, from #597 and #607 landing in an order where each was green on its own base. Without it, `cargo test-fast` and `cargo clippy --workspace --all-targets --features e2e` fail for reasons unrelated to this diff, and CI's `build` job would be red on arrival.

#609 is green on all checks and `MERGEABLE`, waiting only on approval. **When #609 merges, rebasing this branch drops that commit automatically** — identical patch. Please don't review `55bf6d3` as part of this PR — it belongs to #609.

## Open question for @prageethw

`prds/611-pane-scrolling-is-per-agent.md:105` records that **vt100 and xterm disagree about one DECRST shape**, and the PRD deliberately files it as an open question with a recorded default rather than a settled decision:

- vt100 0.16.2's `clear_mouse_mode` clears reporting **only when the DECRST names the protocol currently in force**. xterm clears unconditionally on a DECRST of any of 9/1000/1002/1003.
- They disagree on exactly one sequence: `ESC[?1000h` `ESC[?1002h` `ESC[?1000l` — vt100 leaves reporting **on**, xterm turns it **off**.
- vt100's model shipped, because it is the emulator this deck's own rendering derives from and it is mechanically checkable (the differential test covers that shape as one of fifteen streams).
- The honest counter-argument: the child forms its expectations from the **real** terminal it believes it is talking to, not from the deck's rendering parser — so xterm is arguably the better model for a *forwarding* decision, and it errs conservative, which is the direction this PRD's own safety asymmetry favours.
- It is a **one-line change** (`} else if state.protocol == protocol {` → `} else {`), though the differential test would then need an xterm oracle instead of the vt100 one.

Would appreciate your read on which model to ship with.

## A deliberate scope expansion, declared

#611 named three verified `scan_mouse_mode` defects. This PR fixes a fourth found in review: `1006` is an SGR **encoding** selector, not an event-enabling mode, so `ESC[?1006h` alone must not mark a pane mouse-driven. The published flag now requires an active protocol **and** SGR encoding, because the deck only emits SGR reports and forwarding one to a child that selected another encoding injects bytes it will mis-parse. It was in scope to fix rather than defer because this PR *rewrites that function* and had pinned the wrong behaviour in a new test.

## Rule 12 (cross-version contract)

The **TUI↔daemon contract is unchanged**. `mouse_mode_enabled` is derived client-side from `KIND_STREAM_OUT` bytes the daemon already forwards; `src/daemon_protocol.rs` is untouched; `PROTOCOL_VERSION` stays at **7**; the fragment is `.feature.md`, not `.breaking.md`. Verified independently by both the reviewer and the auditor.

## Rule 9 (experimental flag)

Ships **visible by default**, not behind `experimental`. Decided by the user before work started: it is a small explanatory message making existing behaviour legible, not a new mode, and shipping it flagged-off would hide the explanation from exactly the confused user it exists for. No `features::show_*` wrapper, no `graduate-*` follow-up.

## e2e results (honest)

Rule 5's pre-PR tier was run with recording on: **8,636 tests, 8,635 passed**, 164 casts recorded. Two caveats, both investigated and neither ours:

- `dispatch_015_three_real_claude_seeds_are_genuinely_confirmed` failed in the full run, then **passed 3/3 on re-run**. Nondeterministic real-agent flake — the failing pane moved between runs, `dispatch_014` exercises the same concurrent-seed retry machinery and passed, and the failed cards showed `● Error` with `Tools: 0` (agent process death, not undelivered seeds). The recorded pane body carries no diagnostic text, so the underlying cause is unidentified.
- **Four real-Codex tests silently skipped and counted as passes** in the first run, because this host's Codex auth cannot reach the default `gpt-5.1-codex-mini`. Re-run with `DOT_AGENT_DECK_CODEX_TEST_MODEL=gpt-5-nano` and `DOT_AGENT_DECK_REQUIRE_REAL_E2E=1` (which makes a skip panic): `delegate_009` passed; `codex_worker_001`, `codex_hooks_001` and `codex_live_001` failed or timed out — **and fail identically at baseline `55bf6d3`**, so they are the substitute model's behaviour, not this branch. Note honestly that real-codex e2e therefore did **not** validly exercise this change on this host, which is worth knowing given the PRD is about codex.


prageethw (approved):
Thorough work — L1 specs with proper scenario comments, 22 unit tests for the mouse-mode scanner rewrite including a differential test against the vt100 parser it shares a stream with, docs updated to stop overclaiming uniform scrollability. No protocol/wire changes. On the open question you flagged (vt100 vs xterm DECRST interpretation): I don't have a strong domain opinion either way and the documented alternative is a one-line change if it turns out wrong in practice, so I'm comfortable shipping with vt100's model as written — happy to revisit if it causes real friction. Approving.

Issue #421, opened 2026-08-08
Automatic issue labelling: priority, size, and a machine-written in-progress claim that records its claimant


prageethw:
**Detailed PRD**: See [prds/421-issue-triage-labels-and-dispatch-claims.md](https://github.com/vfarcic/dot-agent-deck/blob/main/prds/421-issue-triage-labels-and-dispatch-claims.md)

---

## Problem

Issues carry no priority, no size, and no machine-written status. Triage is entirely manual, and the deck itself contributes nothing even though it is the thing doing the work.

**The product gap:** `issue_dispatch` (PRD #120) **reads** labels — it builds `gh issue list --label <x>` in `issue_list_argv` (`src/issue_dispatch.rs:118`) — but **never writes one**. When the deck dispatches an issue to a worker, it leaves no mark on the issue at all. Nothing on GitHub shows that an agent is working on it.

That matters more than status display, because of how dispatch decides what to pick up. `dispatch_decision` (`src/issue_dispatch.rs:277`) uses two idempotency signals:

1. the per-issue worktree already exists on disk, or
2. an open PR's head branch is `agent/issue-<n>`.

**Both are inferred side-effects.** Neither is visible to a human scanning the issue list, and neither is checkable by a *second* deck instance before it begins work. There is no explicit claim anywhere.

## Proposal

**1. Write `in-progress` at dispatch, and read it as a third idempotency signal.**

Write it *after* the worktree and spawn succeed, so a failed dispatch never leaves a false claim — the per-issue error boundary already in the fire flow is where this is won or lost. Then have `dispatch_decision` read it. This turns the claim from an inferred side-effect into an explicit, visible one.

**2. Record which orchestrator claimed it.**

A bare `in-progress` says *someone* claimed the issue, not *who* — so a second instance cannot tell "another orchestration owns this, stay out" from "this is my own earlier claim, resume it". The identity to record already exists: `mint_orchestration_id()` (`src/agent_pty.rs:383`) mints a per-tab token and `OrchestrationIdentity::Instance { id, name }` carries it with the config name. Its own doc states the property this needs — *"two tabs of the SAME orchestration in the SAME directory are two distinct routing groups"* — so the **name alone is not sufficient**, since it is ambiguous in exactly the same-name-same-directory case.

[...]

Issue #589, opened 2026-08-17
New-pane form footer promises Enter: submit in the name-collision state, where Enter is silently refused


vfarcic:
## The problem

When the new-pane form is in the name-collision state introduced by #539, the action row correctly drops `[Submit]` and the `! This name is already in use by a live orchestration.` warning renders — but the modal footer still reads:

```
  Tab: switch  ◀▶: mode  Enter: submit  Esc: cancel
```

Pressing Enter there hits the collision guard at `src/ui.rs:7851` and returns `Action::Continue` with no feedback at all. The form advertises a key that does nothing.

Observed directly in the real binary while verifying #539 (isolated sandbox, orchestration `tdd-cycle` in `/tmp/.../myproj`, with `myproj-orchestrator-1` already live):

```
│  Name:    myproj-orchestrator-1                              │
│                                                              │
│  ! This name is already in use by a live orchestration.      │
│                                                              │
│ [Cancel]                                                     │
│  Tab: switch  ◀▶: mode  Enter: submit  Esc: cancel           │
```

`[Submit]` is gone from the button row while the footer beneath it still promises `Enter: submit`.

## Cause

`new_pane_form_hint()` (`src/ui.rs:16805`) chooses between `Enter: submit` and `Enter: next` from `name_submits` alone:

```rust
if has_mode_field {
    if name_submits {
        "  Tab: switch  ◀▶: mode  Enter: submit  Esc: cancel"
    } else {
        "  Tab: switch  ◀▶: mode  Enter: next  Esc: cancel"
    }
}
```

It has no knowledge of `name_collision`. The render path already computes that value one function away (`src/ui.rs:17265`) in order to drop the button, so the information is available at the call site.

## Suggested fix

Thread the collision state into `new_pane_form_hint()` and render wording that does not promise a refused action — dropping the `Enter:` clause entirely in that state would match how the button row handles it (removed, not dimmed).

## Why this is a consistency issue, not a matter of taste

The repo already holds itself to this standard in two places:

- `src/ui.rs:17400` carries a comment about the generic `Enter: next` wording being *"misleading in that state"* — the same class of defect, already recognised and fixed once.
- `src/ui.rs:28036-28049` has tests pinning which hint wording appears in which state (`Enter: submit`, `Enter: next`, `Enter: next/confirm`).

The collision state is simply a new state that this existing, already-enforced standard was never extended to.

## Severity

[...]

Issue #222, opened 2026-07-25
Scheduled issue-dispatch parity with the interactive orchestration-open path (role names, unattended startup, prompt order) + repo default + issue sort


vfarcic:
## PRD: Scheduled issue-dispatch parity with the interactive orchestration-open path

**Problem**: Scheduled issue-dispatch reimplements orchestration spawning in a separate daemon-side primitive (`src/spawn.rs`) instead of reusing the new-dialog path (`TabManager::open_orchestration_tab`). That divergence causes real defects: dispatched orchestration panes lose their role names (shown as `sched-<task>` instead of `orchestrator`/`coder`/…), non-auto-approving agents (claude/opencode/codex) block on first-run trust/permission/onboarding prompts and never come up unattended, and it is unclear the orchestrator's role `prompt_template` (generic instructions) is applied before the schedule prompt. Separately, the authoring flow hardcodes an example repo instead of inferring the current one and offers no way to choose issue ordering.

**Solution**: Converge scheduled issue-dispatch onto the same canonical orchestration-open routine the interactive new-dialog uses — so role naming, role-prompt ordering, and startup handling are shared, not duplicated — and make dispatched agents come up unattended. Add additive authoring UX: a smart default repo (inferred from the current repo) and a first-class issue `sort` field (current order as default). Ship behind the existing `experimental` flag; re-test after release; graduate via #193.

**Detailed PRD**: See [prds/222-issue-dispatch-orchestration-parity.md](https://github.com/vfarcic/dot-agent-deck/blob/main/prds/222-issue-dispatch-orchestration-parity.md)

**Relationships**: Fixes the dispatch/orchestration behavior behind PRD #120's `issue_dispatch`; blocks graduation #193 (remove the `experimental` flag) on a post-release re-test; the ceiling-vs-top-up selection-depth limitation is named and deferred.

**Priority**: High

vfarcic:
## The "prompt order" parity item, measured — and it is bigger than issue-dispatch

Found while using PRD #220's `dispatch` by hand (PR #232). The same defect hits **both** producers, because they share `src/spawn.rs`. Recording the mechanism here so whoever picks this up does not have to re-derive it.

### What is wrong

A daemon-spawned orchestration starts with **no delegation protocol**, so only its orchestrator ever acts:

- `spawn`'s orchestration branch delivers `req.prompt` to exactly one pane — `orchestrator_role_index(&roles)` (`src/spawn.rs:318`, `:378`).
[...]

PR #614, opened 2026-08-20
feat(triage): add adaptive issue labeling

### LINKED ISSUE #603: introduce prioritisation and size tags

vfarcic:
## Summary

- add a guarded live GitHub Agentic Workflow that classifies new issues and same-repository pull requests, then learns from maintainer label corrections
- add a staged single-item preview worker and a manual batch dispatcher that previews by default and serializes memory-writing apply runs
- enforce the repository taxonomy through safe-output allowlists, strict threat detection, serialized memory access, and semantic validation against the triggering event and structured classifier output
- document activation, security, costs, maintenance, and Apache-2.0 attribution for the adapted Dosu workflow

## Safety and rollout

- automatic runs remain disabled unless `ISSUE_LABELER_ENABLED` is exactly `true`
- fork pull requests are rejected before credentialed execution
- the agent receives read-only GitHub tools; separate generated jobs own label and memory writes
- authority and lifecycle labels remain blocked, generated diagnostic issues are disabled, and bot-authored label events are skipped before AI execution
- classifier and threat detector are each capped at 10 AI credits; automatic runs have a 100-credit rolling guardrail, while the documented manual batch ceiling is 400 credits
- the read-only trial classified a synthetic crash as `bug` and left the issue unchanged

## Taxonomy ownership vs. PRD #421

PRD #421 proposes a second mechanism for the same `priority:*` / `size:*` / `needs-triage` vocabulary — a scheduled `triage-issues` skill running on a maintainer's deck. `docs/develop/issue-labeling.md` now records the precedence: **this workflow owns classification labeling for the repository**, and PRD #421 Phase 1 must not be enabled alongside it. PRD #421's Phases 2–3 (the `agent-deck/claimed` dispatch claim) are a different job and cannot collide — the safe-output `allowed` list is a closed allowlist that excludes that label, and the workflow has no label-removal capability at all. The activation procedure now carries the interlock as an explicit precondition. See the review thread for the reasoning and for the part that is a maintainer call.

## Verification

- `gh aw compile issue-labeler issue-labeler-preview --validate --strict --actionlint --approve`
- `gh aw validate --strict`
- actionlint and yamllint on `.github/workflows/issue-labeler-batch.yml`
- the post-agent memory validator is covered by `xtask/linkage-check/src/issue_labeler_memory.rs` (10 tests, in `cargo test-fast`): the script is extracted from the Markdown source and run under `node` against fixture memory trees — two accepts (a labeling-mode prediction; a six-label correction, where a prediction caps at five) and six rejects (a row smuggled in for an untouched issue, a row for another repository, a non-taxonomy label, an extra file in the memory clone, feedback about an item never classified, and a feedback-mode run that applies labels). A companion test asserts the extracted script is embedded verbatim in the generated `.lock.yml`, so an edit landing without a recompile fails the suite instead of leaving it testing a script no runner executes.
- `cargo fmt --check`
- `cargo clippy --workspace --all-targets --features e2e -- -D warnings`
- `cargo test-fast` (2,953 passed)
- `cargo xtask linkage-check`

`origin/main` is merged in (`10f9c71`). It had accumulated two conflicts since 2026-08-20, and had already fixed both independently, so this branch's versions are dropped rather than kept: `.env.vals.yaml` (main added `OPENAI_API_KEY` above `OPENROUTER_API_KEY`; this branch had added it below) and the one-line `src/agent_pty.rs` fixture initialization this PR previously carried, which main now supplies with its issue #454 comment. Both files are byte-identical to main again. Until that merge the PR was `mergeable: CONFLICTING`, so GitHub was running neither the `pull_request`-triggered CI nor Docs on it.

Closes #603
Refs #421


vfarcic:
Pre-merge execution note: GitHub does not allow a newly introduced `workflow_dispatch` workflow to be dispatched from this repository until the workflow path exists on the default branch. `--ref` can select a branch revision only after GitHub has registered that workflow from `main`.

The branch implementation was nevertheless exercised with `gh aw trial` against real issue #589 through a temporary private host repository. The staged preview completed successfully, proposed `bug`, `source`, `priority:low`, and `size:low` with high confidence, passed strict threat detection, and used 1.11479 AI credits. Because this was a staged trial, #589 was intentionally not modified; the temporary host was deleted afterward.

The production-repository smoke test must therefore happen after merge while automatic activation remains disabled:

1. Dispatch `issue-labeler-batch.yml` with `item_type=issue`, `numbers=589`, and `apply=false`; confirm the same staged labels and no issue mutation.
2. If the preview is correct, dispatch it with `apply=true`; confirm the labels are applied and the memory update succeeds.
3. Remove the test labels from #589 if they should not remain, then enable `ISSUE_LABELER_ENABLED` only when automatic classification is ready to roll out.

prageethw (commented):
Two non-blocking items worth resolving before merge rather than after:

1. **Product overlap with #421**: this repo already has `priority:*`/`size:*`/`needs-triage` labels, and #421's PRD (open PR #464) proposes an *in-app* scheduled skill to apply the same taxonomy. This PR implements the same labeling independently via an external GitHub Actions workflow driven by Codex. Both ship without conflict today because #421 Phase 1 hasn't landed and this PR's automatic runs are gated behind `ISSUE_LABELER_ENABLED` (currently unset), but if both go live they could double-label or conflict on the same fresh issue. Worth deciding now which mechanism should own this, rather than discovering the collision later.
2. The PR description cites verification via "semantic memory validation fixtures for valid predictions, six-label feedback, and tampering rejection" — I couldn't find any fixture files matching that description anywhere in the diff. Could you confirm whether those fixtures exist elsewhere (and should be included here) or the description should be corrected?

Everything else looks solid: the feature ships fully disabled by default, the fork-safety argument (`pull_request` not `pull_request_target`, so a fork PR can't reach the credentialed classifier) holds structurally, and the post-agent JS validator recomputing the expected memory diff is a real enforcement boundary rather than relying on model restraint. CI is green.

vfarcic:
@prageethw thanks — both were worth catching. Fixes pushed in cbf24aa.

---

## 1. Overlap with #421

**Scope first, because it is narrower than it looks.** The only surface where the two mechanisms meet is #421's **Phase 1** — the `triage-issues` skill + cron that applies `priority:*` / `size:*` / `needs-triage`. Phases 2–3 (`agent-deck/claimed`, the claimant record, claim-aware `dispatch_decision`) cannot collide with this workflow even in principle: the safe-output `allowed:` list is a **closed allowlist** that does not contain `agent-deck/claimed`, and the workflow has no label-*removal* capability at all. So whatever is decided about triage, #421's actual product value — the dispatch claim — is unaffected. I've written that down rather than left it inferable.

**The collision is worse than double-labeling**, which is the part I think justifies deciding now rather than later:

1. **Divergent vocabulary, not duplicate labels.** The repo's live labels are colon-form (`priority:high`, `size:high`) — that is what this workflow's allowlist enforces and what the label descriptions were written for. PRD #421 records the **hyphenated** form (`priority-high`) as *settled*, in its header and again at M1.1. If Phase 1 ships as written it creates a second, parallel taxonomy; neither mechanism can read the other's, and no `label:` query matches both. This one is already live-vs-document, not hypothetical.
2. **The correction memory would be poisoned.** This workflow treats every non-Bot `labeled`/`unlabeled` event as maintainer ground truth. #421's triage agent applies labels through `gh` under a maintainer's own credential, so GitHub reports `sender.type == "User"` and the bot skip in the `if:` does not fire. The second mechanism's output would be ingested and *trained on* as if a human had corrected the classifier — a silent quality regression, not a visible conflict.
3. Each of those events is also a metered feedback run against the same rolling credit guardrail.

**Where I land: this workflow should own classification labeling; #421 should keep the dispatch claim and drop Phase 1.** The reasoning, including the parts that cut the other way:

- **#421's stated rationale for classifying locally is now stale.** "Why the triage classifier runs locally" rests on *"There is no LLM credential in CI and no intent to add one."* `OPENAI_API_KEY` was added to this repo's Actions secrets on 2026-08-20. That premise was correct when written; it is not correct now, and it is load-bearing for the whole phase.
- **The second half of that rationale — "a CI job can never pause to ask a human" — is explicitly not load-bearing in #421's own design.** Its uncertainty rule is a label, not a question: *"correctness never depends on someone watching."* That is exactly `needs-triage`, which this workflow implements.
- **#421's own durability note points the same way.** Its three artifacts live in three places and only one is versioned; the schedule sits in `~/.config/dot-agent-deck/schedules.toml` on whichever machine runs the daemon and *"vanishes with the machine."* This workflow's trigger is in the repo.
- **Enforcement.** The safe-output allowlist rejects out-of-taxonomy labels at infrastructure level whatever the model says. #421 puts *"Any restriction on what the triage agent may run"* out of scope — the skill scopes behaviour by prompt only.
- **Coverage.** This also labels PRs and applies type/area labels, which #421 never touches, and it learns from corrections, which #421 does not.

**The one axis where #421 Phase 1 is genuinely better, and I don't want to paper over it:** it sweeps the *existing* backlog on a cron. This workflow is event-driven and will never reach an issue opened before it was enabled, so Success Criterion 1 ("every open issue … carries either a `priority-*` + `size-*` pair or `needs-triage`") is not something it satisfies by itself. The mitigation is the batch workflow — manual, 20 items per run, so roughly six runs for the current backlog — and then event-driven coverage from there. That is a real downgrade from "maintained without anyone remembering to do it" to "one deliberate backfill", and it is the honest cost of the recommendation.

**This last part is a product call I don't think I should make unilaterally** — M2.0 was settled jointly by you and @vfarcic, #464 is merged, and retiring a phase of it is your and Viktor's decision, not a side effect of this PR. So, stated as a single yes/no:

> **Retire PRD #421 Phase 1 (M1.1–M1.4) and rescope #421 to the dispatch claim, letting this workflow own the classification taxonomy?**

If **yes**: #421 drops M1.1–M1.4, Success Criterion 1 is restated against this workflow (or accepts the manual backfill), the hyphenated-separator decision in the header is retired in favour of the colon form the repo already uses, and Open Question 3's `sort` interaction with #222 resolves against the colon form — which does need quoting (`label:"priority:high"`), the one concrete cost of that separator, and worth recording there.
If **no**: this workflow stays off, its `ISSUE_LABELER_ENABLED` variable is never set, and #614 lands as documentation plus a preview/batch tool only — in which case say so and I'll add that to the docs instead.

**What I changed on #614 so this can't happen by accident either way** (`docs/develop/issue-labeling.md`):

- a new **"Who owns the triage taxonomy (PRD #421)"** section stating the precedence, the three collision mechanisms above, and why Phases 2–3 are structurally excluded;
- the activation step now carries the interlock as a precondition — set `ISSUE_LABELER_ENABLED` *only after* confirming no `triage-issues` schedule points at this repo;
- a matching line in `CONTRIBUTING.md`'s developer-docs index.

---

## 2. The fixtures

**You were right — they did not exist.** Nothing matching that description was in the diff, elsewhere in the repo, or on any other branch. The description overclaimed, and the post-agent validator — the boundary you correctly identified as the one that doesn't rely on model restraint — shipped with no test of any kind.

Correcting the description alone felt like the wrong fix for exactly the reason you were probing, so the coverage now exists: **`xtask/linkage-check/src/issue_labeler_memory.rs`, 10 tests, in `cargo test-fast`** (~0.12s; `SKIP:`s where `node` is absent, following the `verify_pr_stream.rs` precedent).

It extracts the `Validate feedback memory` script from the Markdown source and runs the **real** script under `node` against fixture memory trees, rather than testing a transcription:

- **Accepts** — a labeling-mode prediction matching the classifier output; and a **six-label correction**. That second one is the case worth having: a prediction caps at 5 labels while a correction snapshots whatever the item now carries and may exceed it, so a suite that only ever passed ≤5 would not notice the two caps being collapsed into one. Its fixture also carries a blocked `PRD` label in the event payload, which must be filtered back out rather than stored.
- **Rejects**, each asserting the *specific* error so a fixture that fails for the wrong reason doesn't read as a pass — a row smuggled in for an issue the run never touched, a row pointing at another repository, a non-taxonomy label, an extra file in the memory clone, feedback about an item the workflow never classified, and a feedback-mode run that tries to apply labels.

One thing I want to flag explicitly, because it is what makes the rest mean anything: the tests read the **Markdown**, and GitHub runs the **lock**. So `issue_labeler_validator_lock_and_source_agree` asserts the extracted script is embedded verbatim in `issue-labeler.lock.yml`. Without it an edit landing without a `gh aw compile` would leave ten green tests exercising a script no runner ever executes. I don't have the pinned `gh-aw` extension installed, so **no workflow file was touched and no lock was regenerated** — the two absolute runner paths the script hard-codes are redirected into a `TempDir` at test time, and `redirect()` asserts each literal occurs exactly once so a rename fails loudly instead of silently testing an unsandboxed path.

Sanity-checked by mutation rather than by assertion count: adding `"PRD"` to the validator's `allowedLabels` in the source turns three tests red (the drift guard, the non-taxonomy reject, and the six-label accept). Reverted.

Gates: `cargo fmt --check`, `cargo clippy --workspace --all-targets --features e2e -- -D warnings`, `cargo test-fast` (2,897 passed), `cargo xtask linkage-check` all green locally. PR description corrected — the fixtures bullet now describes what actually exists.

vfarcic:
One thing found while pushing the above, which matters more than either review item: **this PR was not running CI at all.**

`main` moved after the last green run on 2026-08-20 — #464 among others — and the branch went `mergeable: CONFLICTING`. GitHub does not evaluate `pull_request`-triggered workflows on a PR whose merge commit cannot be computed, so neither `CI` nor `Docs` had run since; only `PR Labeler` had, because it is `pull_request_target` and therefore unaffected. The green checks on the PR page were from 2026-08-20, against a `main` that no longer exists. Worth knowing as a general trap: a stale-green board and a not-running board look identical from the PR page.

`main` is now merged in (10f9c71). Both conflicts were cases where `main` had already made the same fix independently, so this branch's versions are dropped rather than kept, and both files are byte-identical to `main` again:

- **`.env.vals.yaml`** — `main` added `OPENAI_API_KEY` alongside `ANTHROPIC_API_KEY` above `OPENROUTER_API_KEY`; this branch had added the same key below it. Git's auto-merge took both sides and produced a **duplicate YAML key**, which is why this one needed resolving by hand rather than by taking a side — it was not reported as a conflict.
- **`src/agent_pty.rs`** — the one-line `pane_handed_over: false` fixture initialization this PR carried is now supplied by `main` with its issue #454 comment. The PR description's note about that line repairing a `main` compilation failure was stale and has been updated.

**On the way to green, `build-macos` failed twice** on `hook::tests::socket_007_signal_interrupted_read_still_returns_the_reply` — the macOS load-flake already tracked as #642, which `main`'s own 23:02 run hit independently. `src/hook.rs` is untouched by this branch, and `build` and `build-windows` passed on every attempt. Attempt 3 passed with no code change, which is the fail→fail→pass signature #642 documents; I've added both attempt URLs there so the re-run does not erase them. **All ten checks are green now.**

Finally, a correction to my own comment above: I cited the hyphenated-separator decision as living in PRD #421's header "and again at M1.1". The accurate citation is the **header** (`**Label separator**: **hyphenated** … **settled** with M2.0`), **In Scope** (`priority-high|medium|low`, `size-high|medium|low`), and the **Open Question 4** comparison table; M1.1 itself does not restate it. The substance is unchanged — the document settles on hyphens, the repository's live labels are colons.

prageethw (approved):
Both points from my prior review are now addressed: the PR body's new "Taxonomy ownership vs. PRD #421" section documents the precedence, and the fixture-based verification I couldn't find is now in `xtask/linkage-check/src/issue_labeler_memory.rs` (10 tests, confirmed present in the diff) with the two-accepts/six-rejects coverage described. CI is green. LGTM.

PR #615, opened 2026-08-20
fix(logging): let RUST_LOG override the crate's default info directive

### LINKED ISSUE #605: RUST_LOG=dot_agent_deck=debug is silently discarded by an appended default directive

vfarcic:
Closes #605.

## The defect

`RUST_LOG=dot_agent_deck=debug DOT_AGENT_DECK_LOG=1 dot-agent-deck` produced a log file byte-identical to one written with no `RUST_LOG` at all — no warning, no error, nothing to tell "your filter was discarded" from "debug logging had nothing to say".

`init_logging_from_env` built the filter as

```rust
EnvFilter::from_default_env().add_directive("dot_agent_deck=info".parse().unwrap())
```

`add_directive` **replaces** an equally specific directive (`Directive`'s `Ord` compares specificity, not level, so `DirectiveSet::add`'s `binary_search` hits `Ok` and overwrites). The default therefore won over the user's directive every time, and only a *more specific* target survived — which is why the workaround in the repo's own tooling is to enumerate `dot_agent_deck::daemon=debug`, `dot_agent_deck::spawn=debug`, … one subsystem at a time. That is the form nobody reaches for first, and it cost a real debugging session on #454.

## The fix

Same parser, same replacement rule, opposite order — the deck's defaults go in first and `RUST_LOG` is layered on top:

```rust
EnvFilter::new(format!("error,dot_agent_deck=info,{}", rust_log.unwrap_or_default()))
```

- `RUST_LOG=dot_agent_deck=debug` now raises the crate to debug, and `=warn` quiets it — precedence runs both ways, and neither worked before.
- A directive that does **not** name the `dot_agent_deck` target still leaves the default alone, so a bare `RUST_LOG=debug` moves only the dependency floor. That is the issue's stated constraint and it is pinned by a test.
- With no `RUST_LOG` the filter is exactly what it was: the deck at `info`, dependencies at `error`.

The construction moved out of `main.rs` into `logging::env_filter` for one reason: `main.rs` is the binary and integration tests link the library only, so the precedence rule was unreachable from the fast tier. `main.rs` now calls `logging::env_filter_from_env()`.

### One deliberate behavior change beyond the report

The dependency `error` floor is now named explicitly instead of inherited from `EnvFilter::from_default_env`'s default directive, which only fires when the parsed input is empty — and the assembled string never is. That makes the floor **unconditional**. Previously it silently vanished the moment `RUST_LOG` was set at all, so asking for deck debug logs also switched dependency errors *off*. Keeping the old asymmetry would have meant reproducing the builder's empty-input rule by hand; making it unconditional is both smaller and the direction nobody would argue for. It has its own test (`logging_filter_user_directive_keeps_the_dependency_error_floor`).

## Reproduced first

Per the repo's `reproduce-first` skill, the tests came before the fix. Against the **pre-fix** builder, extracted verbatim:

- `logging_filter_explicit_crate_directive_enables_debug` — **red**, for the reporter's reason (`RUST_LOG=dot_agent_deck=debug did not enable crate debug logging`; only INFO/WARN in the capture).
- `logging_filter_explicit_crate_directive_can_quiet_the_crate` — **red**, same defect, other direction.
- The four controls — no `RUST_LOG`, empty `RUST_LOG`, a more specific target, a bare level — **green on both sides**, which is what attributes the failure to the exact-target override rather than to a broken harness.

Both halves of the fix were then revert-checked one at a time: dropping the composition order turns the two precedence tests red, and dropping the `error` floor turns the two floor tests red.

The tests assert at the reporter's altitude — whether the line lands in the log — not which directives the filter holds. They are pure lib units, not `#[spec]` catalog tests, following the `tests/features.rs` precedent (no scenario comments, no catalog entries); `cargo xtask linkage-check` is clean.

## Verified against the real binary

Not only the unit tests. A sandboxed `daemon serve` (isolated `HOME`, sockets, state dir and `DOT_AGENT_DECK_LOG`, per rule 12's isolation note):

| `RUST_LOG` | DEBUG lines | INFO lines |
|---|---|---|
| *(unset)* | 0 | 4 |
| `dot_agent_deck=debug` | 1 | 4 |

## Scope checks

- **Rule 4 (TUI tests):** no user-visible TUI behavior changes — this only affects the opt-in `DOT_AGENT_DECK_LOG` file. Verified against the final diff: nothing under `src/ui.rs`, `src/tab*.rs`, `src/pane*.rs` or the render path is touched.
- **Rule 12 (cross-version contract):** no wire change. This is how a process filters its own log file, not anything the TUI and daemon say to each other — no field, no shape, no meaning moved. `PROTOCOL_VERSION` untouched, no `.breaking.md`; `605.bugfix.md` bumps patch.

## Gates

- `cargo fmt --check` ✅
- `cargo test-fast` ✅ 2894/2894
- `cargo test-e2e` ✅ 8619/8619 (7 slow, 0 skipped)

**Note on how those were run:** `origin/main` is currently red from an unrelated missing `RunningAgent::pane_handed_over` initializer, which breaks the `lib test` target and therefore every tier. PR #609 is the reviewed one-line unbreak and is still open. To gate this branch honestly I applied #609's hunk **locally and uncommitted**, ran all of the above, then reverted it — it is deliberately **not** in this branch and this PR does not duplicate or bundle it. Once #609 lands, this branch wants an update from `origin/main`; until then CI here will fail on #609's error, not on anything in this diff.

## Docs

`RUST_LOG` was documented nowhere, so a "Turning the verbosity up" subsection was added to Troubleshooting › Enabling Debug Logs, plus a row in Configuration › Environment Variables. Both note that `RUST_LOG` selects *what* is logged while `DOT_AGENT_DECK_LOG` decides *whether* there is a log file, since neither is useful alone.


vfarcic:
## State of this PR — please read before reviewing

**The four red jobs are not this diff.** `build`, `build-macos`, `build-windows` and `windows-cross-check` all fail on the identical, pre-existing error from `main`:

```
error: could not compile `dot-agent-deck` (lib test) due to 1 previous error
```

That is #609's one-line fix, still open and unapproved. It breaks the `lib test` target, so it takes down every tier on every platform for every branch cut from `main`. Nothing else appears in any of the four logs. The jobs that do not build the test target — `security`, `nix`, `devbox`, `docs-build`, `changes`, `label` — are green.

Rather than bundle #609 here, I applied its hunk **locally and uncommitted**, ran the full gate set, and reverted it. It is not in this branch:

| gate | result |
|---|---|
| `cargo test-fast` | 2894/2894 |
| `cargo test-e2e` | 8619/8619 (7 slow, 0 skipped) |
| `cargo xtask linkage-check` | ok — 547 catalog ids, 475 annotations, 8 rules |

**Sequencing this needs:** once #609 lands, this branch wants an update from `origin/main` to turn CI green. That update is a push, and `dismiss_stale_reviews_on_push` is `true` — so an approval given now will be dismissed and need re-granting. Reviewing the substance now is still useful; just expect the second click.

## Greptile has not run, and it is not this PR

No `Greptile Review` check-run and no summary comment appeared here after ~25 minutes. That is repo-wide, not specific to this branch — Greptile last published on **2026-08-18T13:51Z** (#597). Every PR opened since has none of its four surfaces:

| PR | opened | Greptile check | Greptile comment |
|---|---|---|---|
| #594 | 08-18 01:35 | yes | yes |
| #596 | 08-18 13:40 | yes | yes |
| #597 | 08-18 13:51 | yes | yes |
| #598 | 08-18 22:43 | — | — |
| #599 | 08-18 22:53 | — | — |
| #607 | 08-19 06:10 | — | — |
| #609 | 08-19 12:10 | — | — |
| #615 | 08-20 04:34 | — | — |

`greptile.json` is unchanged (`triggerOnUpdates: true`), so this looks like an outage or a lapsed installation rather than a config change — worth checking the app's status on the repo. Per CLAUDE.md rule 8 I did not block on a reviewer producing neither a comment nor a pending check. There were consequently no inline findings to address or threads to resolve; if Greptile comes back and posts on a later push, its findings still need reading before merge.

prageethw (approved):
Root cause and fix are correct — tracing-subscriber's add_directive replaces on equal specificity, so appending the crate's own default directive after RUST_LOG silently discarded a same-specificity user override; reordering fixes it. Good test coverage at the right altitude (asserting actual emitted log lines, not internal filter state) covering both directions of precedence plus the one deliberate side effect (the dependency error floor becoming unconditional), which is called out and justified rather than smuggled in. The four red build checks are unrelated pre-existing main breakage (verified: identical E0063 on pane_handed_over, fixed by #609, now approved) — will need a rebase to go green, not a code change here. Approving on content.

Issue #561, opened 2026-08-14
The absolute-path fallback is POSIX-quoted, so cmd.exe and PowerShell cannot run it


vfarcic:
Split out of the verification of #520 ([report](https://github.com/vfarcic/dot-agent-deck/pull/520#issuecomment-5294199953)). Raised in the earlier round-4 review as a follow-up, acknowledged by the author in the changelog fragment, but never filed.

## Problem

`platform::paths::shell_quote_if_needed` omits `\` from its safe byte set:

```rust
b.is_ascii_alphanumeric()
    || matches!(b, b'/' | b'.' | b'_' | b'-' | b'+' | b'=' | b':' | b'@' | b'%' | b',')
```

Every Windows absolute path contains backslashes, so *every* Windows path takes the single-quoting branch. When `binary_name()`'s fallback fires on Windows, the generated command example becomes:

```
'C:\Users\foo\dot-agent-deck.exe' work-done --task-file '.dot-agent-deck/...'
```

Single quotes are a POSIX-shell construct. This is correct in git-bash and WSL, and **inert in `cmd.exe` and PowerShell** — PowerShell treats a single-quoted string as a value, not a command, and needs the call operator (`&`) to execute it.

## Scope, and why this is not urgent

- Only the **fallback** branch is affected. A normal on-`PATH` Windows install takes the bare-name branch, which is unquoted and runs fine everywhere.
- The generated blocks are ` ```bash `-fenced and the worker footer explicitly says to run the command via Bash, so an agent following the instructions as written lands in a POSIX shell.

So this is a real correctness gap in a degraded-mode branch, not a broken shipping path.

## Possible directions

1. Quote per-platform: use the POSIX form only when the generated text targets a POSIX shell, and emit a `cmd`/PowerShell-appropriate form (or bare, since Windows paths rarely need quoting unless they contain spaces) otherwise.
2. Leave the quoting alone and make the surrounding generated prose state the shell assumption explicitly, so a Windows agent knows to invoke through bash.

Option 1 is more correct; option 2 is nearly free. Worth deciding deliberately rather than by default.

Issue #560, opened 2026-08-14
binary_name()'s fallback can emit a relative path on macOS, defeating its own cwd-independence


vfarcic:
Split out of the verification of #520 ([report](https://github.com/vfarcic/dot-agent-deck/pull/520#issuecomment-5294199953)). Not a regression introduced there — a gap in the fallback that PR adds.

## The claim the code makes

`platform::paths::binary_name()` emits the running binary's bare file name only when it is shell-safe *and* `$PATH`-identity-matches. Otherwise it falls back to `current_exe()`'s path, and the justification — stated in the doc comment, the changelog fragment, and the reply on Greptile's stale-path thread — is that **an absolute path is independent of whatever `$PATH` or cwd the consuming agent's shell ends up with**.

That justification holds only if the path is actually absolute.

## Why it may not be

`resolve_binary_name` (`src/platform/paths.rs:403`) passes `current_exe()`'s result straight to `shell_quote_if_needed` at `src/platform/paths.rs:417`, with no absolutising step:

```rust
match path.to_str() {
    Some(path_str) => shell_quote_if_needed(path_str),
    None => DEFAULT_BINARY_NAME.to_string(),
}
```

On Linux `current_exe()` reads `/proc/self/exe`, which the kernel resolves to an absolute path, so this is always fine. On **macOS** it is backed by `_NSGetExecutablePath` — and `binary_name()`'s own doc comment says that call "reports the path the process was INVOKED as". If that is accurate, then invoking the deck through a relative path yields a relative `current_exe()`, and the fallback emits a relative command word.

## Concrete failure

- **Platform:** macOS.
- **State:** deck launched as `./target/release/dot-agent-deck`; the bare name `dot-agent-deck` fails the identity gate (not on `$PATH`, or a different build shadows it — the exact case the fallback exists to serve).
- **Result:** the generated worker task file's footer reads `./target/release/dot-agent-deck work-done --task-file …`. The worker's shell resolves that against the *worker's* cwd — an orchestration directory or git worktree, not the deck's launch dir — so it fails.
- **Symptom:** none. `work-done` writes to the unversioned hook socket, the call site is fire-and-forget, and the daemon drops unparseable frames without logging. The signal silently never arrives — precisely the failure mode #520 exists to eliminate.

## Status: unconfirmed

[...]

PR #612, opened 2026-08-19
docs(prd): pilot Dosu for repository knowledge and documentation drift

vfarcic:
## Summary

- add the documentation-first PRD for a bounded 30-day Dosu pilot
- evaluate repository Q&A, issue discovery, and documentation-drift detection
- define permission, privacy, cost, safety, success, rollback, and go/no-go gates
- keep automated issue mutation and repository writeback disabled during the pilot

## Stacked dependency

This draft is temporarily stacked on #609 because current `main` does not pass the repository's mandatory feature-enabled Clippy gate. The branch includes #609's unchanged commit `55bf6d36` followed by this PR's PRD-only commit `fbf95b8`.

Do not merge this PR before #609. After #609 merges, merge the updated `main` into this branch, rerun the gates below, push, review the resulting diff, and then mark the PR ready.

## Validation

- `cargo fmt --check`
- `cargo clippy --workspace --all-targets --features e2e -- -D warnings`
- `env -u DOT_AGENT_DECK_EXPERIMENTAL devbox run -- cargo test-fast` — 2,886 passed

The experimental environment variable was removed for the fast-test run because the affected feature-resolution test explicitly exercises its documented no-environment-variable precondition; the ambient shell had it enabled.

Documents #610.

PR #616, opened 2026-08-20
test(paths): pin shell_quote_if_needed's contract with direct unit tests (#563)

### LINKED ISSUE #563: shell_quote_if_needed is shared by three call sites but has no direct unit test
Split out of the verification of #520 ([report](https://github.com/vfarcic/dot-agent-deck/pull/520#issuecomment-5294199953)). Test-coverage gap, no user-visible defect.

## Gap

#520 promoted `shell_quote_if_needed` into `src/platform/paths.rs` as the canonical shared copy, replacing duplicates that had been sitting in `codex_hooks_manage.rs` and `devin_hooks_manage.rs`. It now has three call sites:

- the Codex hook installer,
- the Devin hook installer,
- `resolve_binary_name`'s absolute-path fallback (the new one, which is what justified deduplicating rather than triplicating).

Nothing tests it directly. It is exercised only incidentally — `orchestration/delegate/019` asserts a fallback result that happens to be quoted, which pins the composed behaviour but not the helper's own contract.

For a `pub(crate)` helper whose output is interpolated into shell command text that agents execute verbatim, and which three separate subsystems now depend on, that is thin.

## Worth pinning

- A path needing no quoting is returned **unchanged** (no gratuitous quotes).
- A path containing a space is single-quoted.
- An embedded single quote is escaped as `'\''` and the result still parses as one shell word.
- The empty string is quoted rather than returned bare — `!path.is_empty()` in the safe-path guard means it already is, but nothing asserts it, and an empty command word silently expanding to nothing is exactly the class of failure #520 exists to prevent.
- Each byte in the safe set (`/ . _ - + = : @ % ,`) stays unquoted, so a future narrowing of that set fails a test instead of silently adding quotes to every generated path.
- Backslash is **not** in the safe set, so Windows paths always quote — the property #561 is about. Worth pinning as *observed* behaviour so that whichever way #561 resolves, the change is deliberate and visible in a diff.

These are pure-data assertions: no filesystem, no environment, no daemon. They belong beside the existing `platform::paths` unit tests and cost nothing in the fast tier.

vfarcic:
Closes #563.

## What

`platform::paths::shell_quote_if_needed` is a `pub(crate)` helper whose output is interpolated into command text that agents execute verbatim. #520 promoted it here as the canonical shared copy, and it now has three call sites — the Codex hook installer, the Devin hook installer, and `binary_name`'s absolute-path fallback. Nothing tested it directly: its only coverage was incidental, via `orchestration/delegate/019`, which asserts a *composed* fallback result that happens to be quoted and so pins the composition rather than the helper's own contract.

This adds eight fast-tier unit tests beside it. **Test-only — no production code changed.** The diff is one purely additive hunk inside `mod tests`.

## What is pinned

| Property | Test |
| --- | --- |
| A path of only safe bytes comes back unchanged, not gratuitously quoted | `..._returns_a_safe_path_unchanged` |
| A space forces the single-quoted form | `..._single_quotes_a_path_containing_a_space` |
| An embedded `'` is spliced as `'\''` and the result is still ONE shell word | `..._escapes_an_embedded_single_quote_into_one_word` |
| The empty string is quoted (`''`), not returned bare | `..._quotes_the_empty_string_rather_than_returning_it_bare` |
| Every byte of the safe set (`/ . _ - + = : @ % ,` + alphanumerics) stays unquoted | `..._leaves_every_byte_of_the_safe_set_unquoted` |
| Every *other* ASCII byte is quoted — the split pinned in both directions | `..._quotes_every_ascii_byte_outside_the_safe_set` |
| Backslash is not in the safe set, so Windows and UNC paths always quote | `..._currently_quotes_backslashes_observed_behavior_for_issue_561` |
| A non-ASCII path quotes, with no multi-byte character cut in half | `..._quotes_a_non_ascii_path_without_mangling_it` |

The empty-string case is smaller than it looks: a bare empty string interpolated into command text expands to *nothing*, so the next word silently becomes the command. `!path.is_empty()` already guards it; nothing asserted it.

The non-ASCII case is separate because the predicate reads `path.bytes()` while a path is UTF-8: every byte of a multi-byte character is `>= 0x80`, so the ASCII sweep cannot reach any of them, and bytes-vs-chars is the obvious question to ask of a helper that mixes the two.

The ASCII sweep is what covers the more dangerous direction. A per-byte allowlist test alone catches a *narrowed* safe set (silently adding quotes everywhere) but not a *widened* one — letting a `$`, a backtick or a space through unquoted turns an interpolated path into shell syntax.

## Two deliberate choices

**The test module re-declares the safe set instead of reusing the production `is_safe` predicate.** Reusing it would make the test agree with whatever the helper currently does rather than pin what it is supposed to do, so neither narrowing nor widening the set would fail anything.

**The "still one shell word" claims are checked by a test-local POSIX word reader, not by spawning a shell.** It understands exactly the two constructs the helper can emit — bare safe bytes, and single-quoted runs spliced with `'\''` — and reports anything else as unsafe rather than quietly accepting it. That keeps the whole addition pure data: no filesystem, no environment, no daemon, no sleeps. All eight tests together run in ~0.04s.

## Scope: #560 and #561 are deliberately untouched

The backslash test records **observed** behavior, not endorsed behavior. Whether POSIX single-quoting is the right dialect for a command line `cmd.exe` or PowerShell will run is #561, and resolving it is explicitly out of scope here. Pinning the current behavior *is* the contribution: whichever way #561 lands, the change then has to appear as a deliberate edit to a named test instead of passing silently. #560 (the macOS relative-path fallback) is not touched at all.

## Verification

Each assertion was confirmed to bite, by mutating the production helper and checking the intended test turned red:

| Mutation | Tests that failed |
| --- | --- |
| Narrow the safe set (drop `%`) | safe-set, ASCII-sweep, unchanged-path |
| Widen the safe set (add `$` and `\`) | ASCII-sweep, backslash |
| Drop the `!path.is_empty()` guard | empty-string |
| Replace the `'\''` splice with a naive `format!("'{path}'")` | single-quote-escape, ASCII-sweep |

Gates, all run with PR #609's one-line fix applied to the working tree (see below) since `main` does not currently compile without it:

- `cargo test-fast <the new tests>` — 8 passed
- `cargo test-fast` — 2895 passed, 0 failed
- `cargo fmt --check` — clean
- `cargo test-e2e` — see below

### The e2e tier is flaking on its real-agent tests, independently of this branch

Four full `cargo test-e2e` runs, all on this branch:

| Run | Result | Failures |
| --- | --- | --- |
| 1 (7-test tree) | 3 failed | `codex_wrap_001`, `delegate_015`, `dispatch_015` |
| 2 (7-test tree) | **8619 passed, 0 failed** | — |
| 3 (8-test tree) | 2 failed | `dispatch_013`, `dispatch_015` |
| 4 (8-test tree) | 4 failed, 1 timed out | `codex_hooks_003`, `pi_live_002`, `dispatch_015`, `shell_activity_006`, `chain_smoke_pi_002` (timeout) |

**Every one of those nine failures passes when rerun alone** (CLAUDE.md rule 6), and `dispatch_015` was additionally rerun on a baseline tree with this branch's work stashed — where it also passed — and then again with the work restored, where it passed too.

The cause is environmental, not a regression, on five independent grounds: run 2 was green end to end on the same production code; the failing set is different every run; each failure passes in isolation; run 4's timeout shows `Error: Request timed out.` in the pane, i.e. the LLM API itself; and `src/platform/paths.rs`'s additions are `#[cfg(test)]` **lib** code, which is not linked into any of the integration-test binaries those tests live in. The degradation tracks the number of consecutive real-agent runs — four full tiers inside ~40 minutes — which is exactly the load rule 5 bounds the tier to avoid. `codex_hooks_003` is the only non-real-agent name in the list and it passes alone in 0.26s, consistent with temp-dir contention under that load rather than with anything in this diff.

Reported rather than papered over: the tier is not reproducibly green on this machine right now, and the honest claim is "green once, and every individual failure green in isolation" rather than "green".

Per CLAUDE.md rule 12, this touches neither the daemon, the TUI↔daemon protocol, orchestration, nor hooks — it adds test code only — so there is no contract question and no `PROTOCOL_VERSION` bump. No user-visible TUI behavior changes.

> [!NOTE]
> **`main` is red independently of this PR, and this branch does not carry the fix.**
> `src/agent_pty.rs:6011` is missing the `pane_handed_over` field in a synthetic test agent's initializer, so `cargo clippy --all-targets` and every test tier fail to compile on `main`. #609 is the reviewed one-line fix and is green and mergeable but not yet merged. It was applied to the working tree *temporarily* to run the gates above and reverted before committing, so it is deliberately absent from this diff rather than duplicated into it. This branch will need an update from `origin/main` once #609 lands; CI here will stay red until then, for that reason alone.


vfarcic:
Status notes for whoever picks this up.

**The four red required checks are not this PR.** All four fail with the identical `error[E0063]: missing field 'pane_handed_over' in initializer of 'RunningAgent'` at `src/agent_pty.rs:6011` — `main`'s own compile error, which #609 fixes in one line. That fix is deliberately absent from this diff rather than duplicated into it; it was applied to the working tree temporarily to run the local gates (all clean: `fmt --check`, `clippy --workspace --all-targets --features e2e -D warnings`, `test-fast` 2895 passed, `xtask linkage-check` ok) and reverted before committing. CI here stays red until #609 lands.

**Suggested order, because of `dismiss_stale_reviews_on_push`.** #609 first, then this branch gets updated from `origin/main` — and that push voids any approval already sitting here. So reviewing this before the refresh buys a guaranteed second round trip. Happy either way; just flagging it so the choice is deliberate rather than accidental.

**Greptile did not run.** No check-run (not even pending), no summary comment, no review object, no inline comments. That is repo-wide rather than specific to this PR: its last review anywhere was #598 at 2026-08-18T22:43Z, and #602, #609, #612, #613 and #614 have all had nothing since. `greptile.json` is unchanged. Per CLAUDE.md rule 8's closing note I stopped waiting rather than hanging to a timeout on a signal that never arrives — but it does mean this PR carries no automated review, so the human read is doing all the work here.

The e2e-tier flakiness is written up in the description; short version is that every failure passes when rerun alone, the failing set differs every run, one full tier was green end to end, and the additions here are `#[cfg(test)]` lib code that is not linked into any of the integration-test binaries involved.

prageethw (approved):
Solid test-only PR — 8 well-reasoned unit tests for shell_quote_if_needed covering the safe-set boundary in both directions (the full ASCII sweep is the one that would catch a dangerous widening, not just a narrowing), the byte-vs-char distinction for non-ASCII paths, and explicitly scoping the Windows/backslash dialect ambiguity to issue #561 rather than silently endorsing current behavior. No production code touched. The four red build checks are unrelated pre-existing main breakage, fixed by #609 (now approved) — will need a rebase. Approving on content.

Issue #475, opened 2026-08-09
Contain the remaining ~82 bare tempfile constructors in src/ unit tests (or close as wontfix)


vfarcic:
## What

After #322 (PR #472), **everything under `tests/` allocates its temp directories under the disk-backed private parent** `/var/tmp/dad-e2e-<uid>`, enforced mechanically by `linkage-check` rule 8, which scopes by directory.

One gap remains: roughly **82 bare `tempfile::tempdir()` calls across 22 files in `src/`** — everything except `src/dispatch.rs` (converted) and `src/test_temp.rs` (the helper itself). These are `#[cfg(test)]` unit tests in the lib target, and they still allocate in the OS temp dir, which on this dev box is a 14 GB RAM-backed tmpfs.

Filing so the documented gap has an issue number rather than only a paragraph in `docs/develop/e2e-temp-dirs.md`.

## Why it was left

Deliberate, on measured grounds:

- **No measured leak behind it.** Across three full recorded `cargo test-e2e` runs with continuous `/tmp` sampling (50 ms primary plus a 20 ms supplemental pass), every stray attributed to a `tests/` file. The `src/` unit tests never showed up as a meaningful allocation — the ones that did (`src/dispatch.rs`, which clones repositories) were converted.
- **Large mechanical diff** — 82 sites, 22 files — for no demonstrated benefit.
- It would **move fast-tier churn onto `/var/tmp`**, where these small `TempDir`s currently drop cleanly on the normal path anyway.

The conversion itself is trivial per site: `crate::test_temp::tempdir()` is already available in the lib target.

## Why it might still be worth doing

- **Uniformity.** "Everything under `tests/` is contained, `src/` is not" is a line someone has to remember. Rule 8 could cover `src/` too, and then there is no line.
- **On SIGKILL these survive as untagged `.tmp*`**, which the default reaper will not remove — `--include-untagged` is opt-in for good reason. A pre-run snapshot during #322 work held **80** such directories, so they do accumulate, small as each one is.

## Suggested disposition

Reasonable to close as wontfix if nobody measures a leak behind it. If it is done, the natural shape is: convert the sites, then extend rule 8's scope from `tests/` to include `src/` so the guarantee is uniform and a new file inherits it.

Discovered while implementing #322 (PR #472). See the `## What is outside the root` section of `docs/develop/e2e-temp-dirs.md`, which records the current boundary and the measurements behind it.

PR #618, opened 2026-08-20
test(linkage-check): guard src/test_temp.rs against `crate::` paths (#474)

### LINKED ISSUE #474: Guard src/test_temp.rs's self-containment — seven crates depend on it having no crate:: references
## What

`src/test_temp.rs` is a ~40-line `#[cfg(test)]` helper that resolves the disk-backed `/var/tmp/dad-e2e-<uid>` parent. It is included by **seven** crates — the lib target plus six integration-test crates — via

```rust
#[path = "../src/test_temp.rs"]
mod test_temp;
```

That `#[path]` trick is only possible because the file is **self-contained**: it contains no `crate::` references. If it ever gains one, every one of those six includes stops compiling.

## Why it matters

The self-containment is load-bearing economics, not a style preference.

Containing those six files was originally rejected on measured grounds: doing it with `mod common;` would have pulled the PTY harness into six more binaries and duplicated roughly **530** fast-tier executions. The `#[path]` route cost **+12** (measured: `cargo nextest list` went 2,315 → 2,327, two per crate).

So the difference between a 12-execution solution and a 530-execution one rests entirely on one file staying free of `crate::` — and that property is currently enforced by **a comment**.

Someone adding an innocuous `crate::something` will get six compile errors with no explanation of why the file is unusual, and the obvious "fix" is to unpick the arrangement.

## Proposed fix

A linkage-check rule: **no `crate::` in `src/test_temp.rs`**. Roughly five lines next to the existing rules, in the same style as rule 8. It turns an invisible constraint into a mechanical one with a message that explains itself.

Worth pairing with a sentence in the file's own header pointing at the rule, so the constraint is discoverable from either direction.

## Related

A second, smaller inconsistency in the same file, worth deciding on at the same time: **`test_temp` deliberately ignores `DAD_E2E_TMPDIR`.** It resolves only `/var/tmp/dad-e2e-<uid>`.

That was a deliberate call — honouring a security-sensitive variable with a second, weaker validation path (no `openat`/`O_NOFOLLOW` walk, judgement by name) was judged worse than not honouring it at all. The consequence is real though: someone who moves the harness base with `DAD_E2E_TMPDIR` will still find these unit-test directories under `/var/tmp`. It is documented in `docs/develop/e2e-temp-dirs.md`, but it is a genuine surprise.

Closing it properly means sharing the validated resolver between `tests/common/` and `src/`, which is a real refactor and would itself threaten the self-containment above. Filing the two together because any fix to the second has to respect the first.

Discovered while implementing #322 (PR #472).

vfarcic:
Closes #474.

## What

`src/test_temp.rs` resolves the disk-backed `/var/tmp/dad-e2e-<uid>` scratch parent for tests that do not link the PTY harness. It is compiled into the lib target as an ordinary module and, via `#[path = "../src/test_temp.rs"]`, into **ten** integration-test crates as well. That only works because the file is self-contained — in those crates `crate::` is the *test* binary's own root, where nothing the library defines is in scope.

The property is load-bearing economics, not style. Containing those fast-tier crates with `mod common;` was priced at pulling the PTY harness into six more binaries and duplicating ~530 fast-tier executions; the `#[path]` route cost **12** (measured, `cargo nextest list` 2,315 → 2,327). The whole difference between those numbers rested on one file staying free of `crate::`, and it was enforced by a comment.

This adds **linkage-check rule 9**: no `crate::` path in `src/test_temp.rs`, with a diagnostic that explains itself, plus a header note in the file pointing at the rule so the constraint is discoverable from either direction.

## Why a check rather than the compiler

Two reasons, the second measured on this branch:

1. It otherwise fails as N identical `E0433`s that explain nothing about why the file is unusual, so the obvious "fix" is to unpick the arrangement rather than to drop the reference.
2. **Whether it fails at all depends on the consumer's own root names.** Adding a `crate::features::experimental_enabled()` probe to the module and running `cargo check --tests --keep-going`, nine of the ten including crates failed with `cannot find features in crate` — and `tests/features.rs` compiled **clean**, because its own `use dot_agent_deck::features::{self, Features};` puts a `features` at that test crate's root. So a `crate::` path can land green under whichever filtered `cargo test-fast` the author ran and break the next consumer to include the file.

## Scope

Deliberately narrow, per the issue: the guard only. No refactor of `DAD_E2E_TMPDIR` handling, and none of #475's broad `tempfile` conversion. Production runtime behaviour is unchanged — the only `src/` edit is comments.

## Implementation notes

- The scan runs over the **comment-stripped** source (`strip_rust_comments`), so the file's own header note and this rule's message can name `crate::` freely. A `crate::` inside a *string literal* is not exempt; the rustdoc says so.
- `\bcrate\s*::` — the `\b` keeps `some_crate::x` out (`_` is a word character); `$crate::` from a `macro_rules!` body **is** matched, deliberately, since it expands to the defining crate's root and is non-portable for the same reason; the spaced `crate ::` the compiler accepts is matched too.
- The guarded file is read directly rather than folded into the existing `tests/`+`src/` scan, so **an unreadable file is itself a failure**. A rename that left `SELF_CONTAINED_PATH` behind would otherwise turn the rule into a no-op that still prints `ok` — the same shape of silence the rule exists to end.
- The trailing consumer list in `src/test_temp.rs` said seven crates; it is ten now, so it is refreshed in the same commit.

## Tests

Five in `xtask/linkage-check`, all against **synthetic sources** rather than the live checkout — whose copy of the file is clean by construction and would make them pass for a reason unrelated to the rule working:

- a rejected `use crate::…` and a rejected `crate::…` call, each reported at its own raw line number;
- the `$crate::` and spaced `crate ::` forms;
- an accepted self-contained module — `std`, `libc`, `tempfile`, `super::`, a crate whose name merely *ends* in `crate`, and comments that name the forbidden path;
- `check_self_contained` reading a clean and then a violating file under a synthetic root;
- the guarded file going missing.

Also verified end to end against the real checkout: `cargo xtask linkage-check` prints `ok (… 9 rules)`, and injecting `use crate::config::Config;` into `src/test_temp.rs` makes it exit 1 with `[9] …/src/test_temp.rs:48: …`.

## Gates

- `cargo nextest run -p xtask-linkage-check` — 147 passed.
- `cargo test-fast` — 2,892 passed.
- `cargo test-e2e` — see the comment below.

## ⚠️ CI will be red until #609 lands

`origin/main` is currently red from an unrelated pre-existing break: `RunningAgent::pane_handed_over` was added without an initializer in `insert_test_agent`, so `cargo clippy --all-targets` and every test target fail with `E0063` on `src/agent_pty.rs:6011`. #609 is the fix and this branch deliberately does **not** duplicate or bundle it.

The clippy / `test-fast` / `test-e2e` results above were produced with #609's one-field fix applied to the working tree only — never committed here. This branch will be updated from `origin/main` once #609 merges, and review will be requested after CI is green.


vfarcic:
### `cargo test-e2e`

**8,617 tests, 8,613 passed, 4 failed — all four pass on an isolated rerun** (AGENTS.md rule 6):

| test | outcome alone |
|---|---|
| `e2e_codex_wrapper codex_wrap_001_synthetic_jsonl_reaches_dashboard` | pass (16.9s) |
| `e2e_scheduler_manager manager_010_blank_default_command_falls_back_to_claude` | pass (3.7s) |
| `e2e_shell_activity_real_agent shell_activity_006_…badge_working` | pass (140.3s) |
| `e2e_prompt_delivery_confirmation dispatch_015_three_real_claude_seeds_are_genuinely_confirmed` | failed twice, passed on the third (42.3s) |

The last two are the real-agent, flaky-tolerant shape rule 5 keeps out of CI, and both said so in their own diagnostics — `shell_activity_006` printed `PRECONDITION NOT MET: … the model never issued the Bash tool call at all. This is a harness flake (prompt adherence), NOT a badge regression: rerun`, and `dispatch_015` failed with two of its three real Claude panes confirming the seed and `real-seed-beta` not submitting within the 150s window.

None of them can be attributable to this branch, and that is checkable rather than assertable: `git diff origin/main HEAD -- src/ tests/ Cargo.toml Cargo.lock build.rs` yields **zero non-comment lines**. The only executable change in the PR is inside `xtask/linkage-check`, which no test target links.

### The four red CI jobs are `main`'s pre-existing break, not this branch

`build`, `build-macos`, `build-windows` and `windows-cross-check` all fail on the same line, verified in each job's log:

```
```

That is `src/agent_pty.rs`'s `insert_test_agent`, and #609 is the fix. This branch deliberately does not duplicate or bundle it. Local `clippy --workspace --all-targets --features e2e`, `test-fast` (2,892 passed) and the `test-e2e` run above were all produced with #609's one-field fix applied to the **working tree only** — it is not in any commit here.

I will update this branch from `origin/main` once #609 lands, and request review after CI is green.

prageethw (approved):
Good addition — converts a comment-only invariant (no crate:: in test_temp.rs, since it's #[path]-included into 10 different test crate roots where crate:: resolves differently) into a mechanical linkage-check rule. The empirical demonstration that a probe broke 9/10 consumers while one happened to compile clean is a genuinely useful catch that justifies the check. Comment-stripped scanning, word-boundary-correct regex, and failing loudly on a missing/renamed guarded file are all handled correctly. The four red build checks are unrelated pre-existing main breakage, fixed by #609 (now approved) — will need a rebase. Approving on content.

Issue #502, opened 2026-08-11
e2e reliability: the tier runs in no workflow, and the same flake classes keep recurring — decide a permanent strategy


vfarcic:
## Why this issue exists

Individual e2e flakes and red-on-main tests each get their own issue, and there are now enough of them that the *pattern* deserves one. This is a question-issue: decide a permanent strategy, then let the individual issues be closed by it rather than one at a time.

Surfaced while verifying #500 (a full `cargo test-e2e` run: 3738 passed, 6 failed — three of which pass on rerun, three of which are #460 and are red on `main` too).

## The structural cause: the tier has no automated signal at all

**No workflow runs the e2e tier.** Verified on `746403b`: the only `schedule:` trigger in `.github/workflows/` is `stale.yml`, and the only mentions of `--features e2e` in `ci.yml` are the *clippy* gate plus a comment that says as much — "which is not in CI (rule 5)". Per CLAUDE.md rule 5 that is deliberate: the tier spawns binaries and hits LLM APIs, so it is bounded to the pre-PR step.

The consequence is not deliberate, though. Nothing anywhere observes this tier, so:

- **`main` can be red for weeks and nobody knows.** #460 (three tests), #488 (`devin_live_001`), #418 (`sigterm_001/002`, macOS) all accumulated this way. #460 has now survived 12 commits since it was filed.
- **There is no flake-rate data**, so "is this test flaky or broken?" is re-litigated by hand every time, per test, per contributor. That question is answerable in one query if anyone were recording it.
- **The cost lands on whoever opens the next PR**, who must decide whether 6 failures are theirs. On #500 that meant checking out `origin/main` and rerunning to prove they were pre-existing — the right thing to do, and pure overhead that every PR author repeats.

Rule 5's reasoning (don't run this per-task, don't run it on every PR) does not imply *never observe it*. A scheduled run is a different question from a per-PR gate.

## The recurring flake classes

The open issues sort into four kinds, and only the first two are ordinary engineering:

**A — wall-clock deadlines and render races.** #364 (`delegate_011` wall-clock-sensitive), #247 (`spawn_005`, 5s deadline), #413 (card snapshots read the clock at render; the margin bounds the flake rather than removing it), #395 (`find_in_grid` never retries — ~51 call sites share the footgun), #244 (sentinel reads wait for existence, then read content).

[...]

PR #619, opened 2026-08-20
fix(config): delete the nextest e2e profile that nothing ever selected (#494)

### LINKED ISSUE #494: nextest: `[profile.e2e]` is never selected — remove it or start using it
## Problem

`.config/nextest.toml` defines a `[profile.e2e]` with its own `slow-timeout` and a list of `[[profile.e2e.overrides]]`, but **nothing ever selects that profile**, so none of it takes effect.

Evidence:

- `.cargo/config.toml:3` — `test-e2e = "nextest run --features e2e"`. No `-P` / `--profile`, so it runs under the **default** profile.
- No `-P e2e` / `--profile e2e` appears anywhere in `.github/`.

The `--features e2e` flag selects the cargo *feature*; it has nothing to do with the nextest *profile* of the same name. That collision of names is most of why this is easy to misread.

## Why it matters

It is misleading rather than harmful. A contributor adding a long real-agent test reasonably adds a carve-out under `[[profile.e2e.overrides]]`, believes the test is protected from the `3 x 60s = 180s` default kill, and it is not — the test silently dies mid-wait with a bare `TIMEOUT` and no diagnostics.

This is not hypothetical. It came up directly in #390: 3 of the 5 pre-existing `[[profile.default.overrides]]` entries (`route_001`, `new_pane_016`, `orchestration_dispatch_002`) have no `e2e` counterpart, while 2 (`devin_live_001`, `version_001`) do — an inconsistency that only makes sense once you know the `e2e` profile is never consulted. #390 added entries to both lists for symmetry, and the author explicitly flagged the profile as possibly dead config and asked for a ruling.

## Options

1. **Delete `[profile.e2e]` and its overrides**, and move anything unique in it into `[profile.default]`. Simplest, and matches how the suite is actually run.
2. **Start using it** — change the `test-e2e` alias to `nextest run --features e2e -P e2e`. Note this changes real behaviour: `[profile.e2e]` currently sets `slow-timeout = { period = "120s", terminate-after = 2 }` (a 240s kill) versus the default's 180s, and its override list is not a superset of the default's, so some tests would move to different windows.

Whichever way it goes, a short comment at the top of the profile saying who selects it would keep this from being re-derived.

## Context

Surfaced while verifying #390 (PRD #386).

vfarcic:
Closes #494.

`.config/nextest.toml` defined a `[profile.e2e]` with its own `120s x 2` kill window and four `[[profile.e2e.overrides]]`, and **nothing ever selected it**. `cargo test-e2e` is `nextest run --workspace --features e2e`: `--features e2e` names the *cargo feature*, and there is no `-P` / `--profile` — not in either alias, not in any `cargo nextest run` in `.github/workflows/`.

## Proof of the current selection behaviour

Measured rather than reasoned, via nextest's store directory (`target/nextest/<profile>/`):

nextest names the profile in its own banner, and a `--features e2e` run says:

```
Nextest run ID 3c7faf96-… with nextest profile: default
```

Corroborated by the store directory (`target/nextest/<profile>/`):

| command | store dir created |
|---|---|
| `cargo nextest run --workspace --features e2e` | `target/nextest/`**`default`**`/junit.xml` |
| `cargo nextest run --workspace --features e2e -P e2e` | `target/nextest/`**`e2e`**`/` |

A `--features e2e` run lands in `default` and picks up `[profile.default.junit]`. It has never touched the `e2e` profile.

## Override inventory

| test | `[profile.default]` | `[profile.e2e]` |
|---|---|---|
| `route_001` | 60s x 8 = 480s | — |
| `new_pane_016` | 60s x 5 = 300s | — |
| `orchestration_dispatch_002` | 60s x 7 = 420s | — |
| `version_001` | 120s x 10 | 120s x 10 (identical) |
| `devin_live_001` | 60s x 6 | 60s x 6 (identical) |
| `shell_activity_006` | 60s x 8 | 60s x 8 (identical) |
| `shell_activity_005` | 60s x 6 | 60s x 6 (identical) |

Every `e2e` override was a **value-identical duplicate** of a `default` entry, so nothing unique needed preserving. The only non-duplicate line in the profile was its `120s x 2` base window — which, since nothing selected the profile, has never applied to any test.

## Why delete rather than wire up

Wiring it up (`-P e2e`) was the option that would have **changed real behaviour**: the three carve-outs with no `e2e` counterpart (`route_001`, `new_pane_016`, `orchestration_dispatch_002`) would have dropped from 480s / 300s / 420s to that profile's 240s. Deleting is the no-op on the tier that actually runs.

## What changed

- `.config/nextest.toml` — `[profile.e2e]` and its four overrides deleted. **`[profile.default]` is byte-untouched**: the diff contains zero `-` lines from it.
- A short comment at each place the name collision reads wrong: the top of `.config/nextest.toml`, and the `test-e2e` alias in `.cargo/config.toml`.
- `changelog.d/494.misc.md` (`misc` per `pyproject.toml`: internal tooling; users don't directly see it).

Not a `breaking` fragment and no `PROTOCOL_VERSION` bump — CLAUDE.md rule 12's axis is the TUI↔daemon contract, and this touches no source, no wire shape and no semantics.

## Verification

- `-P e2e` now fails fast: `error: profile 'e2e' not found (known profiles: default, default-miri)`. Nothing in the repo referenced it, so nothing else breaks.
- All **seven** surviving `[[profile.default.overrides]]` filters still resolve to **exactly one** test each (`cargo nextest list --features e2e -E 'test(<name>)'` per filter).
- `cargo fmt --check` clean; `cargo clippy --workspace --all-targets --features e2e -- -D warnings` clean.
- `cargo test-fast`: 2887/2887 passed.
- `cargo test-e2e`: **8610/8612 passed**. The two failures are the known real-agent flake cluster (#351 / #502), deliberately left out of scope:

| test | failed at | its window, before **and** after this PR | rerun alone (rule 6) |
|---|---|---|---|
| `dispatch_015` | 164.7s | 180s — no override under either profile | **PASS** in 37.3s |
| `shell_activity_006` | 136.7s | 480s — `[[profile.default.overrides]]`, unchanged | **PASS** in 132.6s (flaky: 2 of 3 solo runs green) |

  Neither was killed by a timeout; both failed their own assertions well inside their windows. `shell_activity_006`'s diagnostic declares itself: *"PRECONDITION NOT MET: no ToolStart/Bash event observed … This is a harness flake (prompt adherence), NOT a badge regression: rerun rather than reading this as BADGE WRONG."* That the diagnostic printed at all is the widened window doing precisely the job it was added for.

  This PR cannot influence either result: the diff touches **zero** `.rs` files, the profile in force is `default` before and after, and `[profile.default]` is byte-identical — so every test's resolved timeout is unchanged.

## Note on CI

This branch is cut from `main` at `776fab8`, which is currently **red** — `error[E0063]: missing field 'pane_handed_over' in initializer of 'RunningAgent'` at `src/agent_pty.rs:6011`, unrelated to this issue and fixed by #609. The Rust jobs here will fail on that until #609 lands; I deliberately did **not** bundle its fix. Every gate above was run with #609's one-line patch applied to the working tree only, never committed. Once #609 merges I will update this branch from `origin/main` and re-verify.


vfarcic:
CI attribution: the four red jobs are `main`, not this PR.

All four fail on the **same** error, confirmed from each job log:

```
  --> src/agent_pty.rs:6011:13
```

| job | failing step |
|---|---|
| `build` | `cargo clippy --workspace --all-targets --features e2e -- -D warnings` |
| `build-macos` | `cargo nextest run --workspace` |
| `build-windows` | `cargo nextest run --workspace` |
| `windows-cross-check` | Type-check the workspace for Windows from Linux |

That break is on `main` at `776fab8`, which this branch is cut from, and #609 is the one-line fix. `changes`, `devbox`, `label`, `nix` and `security` all pass here.

This PR **cannot** cause it: the diff touches zero `.rs` files — only `.config/nextest.toml`, `.cargo/config.toml` and a changelog fragment. Locally, `cargo clippy --workspace --all-targets --features e2e -- -D warnings` on this branch reports exactly that one error and nothing else, and is clean the moment #609's patch is applied.

I deliberately did not bundle #609. Every gate reported in the description was run with its one-line patch applied to the working tree only, never committed. Once #609 merges I will update this branch from `origin/main` and re-verify before this is ready to approve.

prageethw (approved):
Verified independently rather than taking the removal claim at face value: no workflow or script anywhere passes -P/--profile, every CI nextest invocation writes to target/nextest/default/junit.xml, and all four deleted profile.e2e.overrides entries were value-identical duplicates of profile.default.overrides. Dead config, correctly removed. One tiny non-blocking note: prds/done/77-tui-testing-harness.md still mentions profile.e2e as historical record — fine to leave as-is. The four red build checks are unrelated pre-existing main breakage, fixed by #609 (now approved) — will need a rebase. Approving on content.

Issue #621, opened 2026-08-20
A stale refusal names no generation, so a prompt is abandoned when the event stream drops a SessionStart


vfarcic:
A `stale` refusal from the atomic write-and-submit guard does not tell the caller **which** conversation to rebind to, and the TUI's event subscriber does not replay what it missed across a reconnect. When those two meet, an automatic prompt is abandoned after 60 seconds rather than delivered.

Found while reviewing the #608 branch. The #608 fix is correct and this is not a regression it introduced — but #608 widens the set of requests answered `stale`, so it makes this path easier to reach, and the branch documents the limitation in code rather than leaving it implied.

## The mechanism

Each premise below was verified against the tree, not inferred:

- **`Stale` carries no generation.** `SendResult::Stale` is a bare variant; the response has no current-generation field. A refused caller's only route to the generation is its own event stream.
- **The subscriber does not resynchronize.** `spawn_event_subscriber` (`src/main.rs:1627-1680`) breaks its inner loop on `Err` / `Ok(None)`, sleeps, and calls `subscribe_events()` again. There is no replay, catch-up, or resync anywhere in that path — a `SessionStart` that arrives while disconnected is simply never applied to the client `AppState`.
- **So the client's view can stay `None` forever.** With `pane_hook_session_id` never populated for that pane, every retry goes out unnamed, and every unnamed retry against a pane that *has* a generation is refused `stale` by design.
- **The failure is bounded and logged, not silent.** `AUTOMATIC_PROMPT_DEADLINE` is 60s (`src/prompt_delivery.rs:62`), checked at `src/ui.rs:3292`, and expiry calls `log_prompt_abandoned`. Nothing is mis-delivered — but the prompt is not delivered either.

The ordinary race is fine and should stay that way: a session landing after the caller's snapshot costs exactly one safe `stale`. That refusal writes nothing and does not bump `attempts`, so `bind_delivery_generation`'s `attempts == 0` precondition still holds and it binds on the next render pass. Only the dropped-event case fails to recover.

## Remedies

**Return the daemon's current generation on `stale`.** A response field rather than a new state machine: it closes the hole for this path without touching the subscriber, and lets `bind_generation_before_retry` bind from the refusal itself instead of waiting on a snapshot that may never arrive. Additive to the wire, so it is a `PROTOCOL_VERSION` question — an older client ignores the field and behaves exactly as today.

[...]

Issue #617, opened 2026-08-20
Pane-keyed writes are still permissive on absent identity at three call sites outside the #608 handler


vfarcic:
Issue #608 closed the permissive-on-`None` identity gate in `compute_write_and_submit_outcome`'s **paned** branch. A security audit of that fix (PR for #608) found the same "permissive on absent identity" shape at **other** pane-keyed write sites, plus two residual holes in the #608 fix itself that were deliberately left open because closing them needs new daemon state.

Issue #608 anticipated this family and deferred it, and the #608 PR keeps that scope: fixing these means changing a shared signature and touching several call sites, which is the broad delivery refactor that issue deliberately avoided.

All findings below were produced by an audit of commit `e57c48b` and verified against the tree; the sections marked *verified* were checked by reading the code paths, not inferred.

---

## 1. HIGH — `write_guarded` still authorizes a paned send with no expected agent

**Files:** `src/agent_pty.rs:4767-4789`, `src/state.rs:3273-3275`, `src/state.rs:3321-3325`, `src/state.rs:3444-3447`

`compute_write_and_submit_outcome` now refuses an absent `expected_agent_id`, but the lower-level `AgentPtyRegistry::write_guarded` still resolves a paned target first and runs its identity comparison only inside `if let Some(expected) = expected_agent_id`. Callers that bypass the handler therefore keep the old permissive behavior.

`dispatch_one_owned` is a production caller that passes `expected_worker_agent_id.as_deref()`. That option is populated from `pane_current_agent_id`, which can legitimately return `None`, and the function's own comment at `src/state.rs:3444-3447` assumes the guarded call will then return `NoLiveTarget`. It does so only while the pane remains empty. If a process claims the reusable pane between the lookup and `write_guarded`'s target resolution, the call resolves that newcomer, skips the absent identity check, and can return `Applied`. The orchestration revalidation does not close the hole when either orchestration identity is unknown — `orchestration_still_matches` deliberately returns `true` on absence.

**Scenario.** A `clear = false` delegate resolves no live worker and records `expected_worker_agent_id = None`. Before its detached dispatch task reaches the guarded send, a new agent starts on the same pane slot. The task pointer is written and submitted into that new agent even though no identity ever authorized it, and a tool-capable agent may act on a delegation intended for the missing predecessor.

[...]

PR #625, opened 2026-08-20
fix(daemon): fail closed on an absent identity in the paned guarded send (#608)

### LINKED ISSUE #608: compute_write_and_submit_outcome's paned branch is permissive-on-None for both identity axes
## Repro

`compute_write_and_submit_outcome`'s **paned** branch in `src/daemon_protocol.rs` (currently around the `async fn compute_write_and_submit_outcome` definition) is permissive-on-`None` for both identity axes when writing to a pane:

1. **Agent id**: `extras.expected_agent_id.as_deref()` is forwarded straight into `registry.write_and_submit_guarded(...)` with no gate — if it's `None`, the write primitive still proceeds keyed only by `pane_id`, so it can write to whatever currently occupies the pane rather than refusing.
2. **Session id**: the re-validation closure only checks `expected_session` when it is `Some`:
   ```rust
   if let Some(expected) = expected_session.as_deref() {
       match guard.pane_hook_session_id(&pane_for_check) {
           Some(current) if current != expected => return false,
           Some(_) => {}
           None if has_live_attach => return false,
           None => {}
       }
   }
   ```
   When `expected_session` is `None`, the whole block is skipped and the write is allowed with no session check at all — even on an attached, live-interactive pane, where the comment directly above this block already claims a `None` current-session "is refused too... never a silent accept." The code doesn't implement that for the case where the *caller* supplied no `expected_session` at all.

## Expected behavior

Compare this to the **paneless** branch a few lines earlier in the same function: it fails **closed** on an absent agent id — the `Writable` resolution for a paneless target already maps `None` → `Writable::None` before the write is ever attempted. The paned branch has no equivalent upstream gate forcing `Some`, so a caller that (accidentally or otherwise) supplies no identity for a paned write still gets a delivery.

## Why this matters

A same-uid actor is already inside this product's trust boundary, so the realistic risk here is the *accidental* path — a caller (legacy client, or any code path that doesn't populate the guard fields) landing a write in a pane it didn't mean to target — rather than deliberate steering. But "write lands in the wrong pane with no identity check" is exactly the failure class this delivery-guard mechanism (PRD #20's guarded-send machinery) exists to prevent for every OTHER case.

## Evidence this is undiverged, not fork-specific

Filed from a fork (`prageethw/dot-agent-deck`) that found this while fixing a related instance (a different call site) of the same bug shape. Confirmed via `git grep -nF` on the exact defective literals that both the paned-branch code and the "cross-version fail-safe" comment describing the intentional identity-less degrade path are byte-identical between this fork's `HEAD` and `upstream/main` at the time of filing — this is not fork-diverged code.

## Related

- A related instance in the same function's *paneless* branch is already handled correctly (fails closed) and needs no change.
- A sibling family of the same "permissive on absent/stale identity" shape exists at other pane-keyed write call sites (`deliver_on_idle` in `spawn.rs`, the seed-fallback path in `agent_pty.rs`, a work-done feedback write, and a dispatch-result delivery) — being filed as a separate, related upstream issue from the same fork investigation.

vfarcic:
Closes #608.

## ⚠️ CI will show `build` (and likely `build-macos` / `build-windows`) as RED — this is not caused by this branch

`origin/main` currently does not compile its test targets: `AgentPtyRegistry::insert_test_agent` (a `#[cfg(test)]` fn in `src/agent_pty.rs`) omits the `pane_handed_over` field. The fix is #609 — reviewed, green, `MERGEABLE`, but still **open** awaiting `@prageethw`. This PR does not duplicate or bundle that fix; please do not review/comment on the compile error here. Once #609 merges, this branch will be updated and re-pushed.

## Summary

The paned branch of `compute_write_and_submit_outcome` was permissive-on-`None` for both identity axes, so an identity-less paned write could be delivered to whatever agent currently occupied the pane.

- **Axis 1** (`expected_agent_id`): now maps an absent expected id to `Writable::None` → `NoLiveTarget` before any write is attempted, instead of allowing the write through.
- **Axis 2** (attach/session identity): now refuses (`Stale`) when the pane has a current hook session that the caller did not name.
- `(None, None)` — no expected id and no current session — still delivers, **deliberately**: this is the path for agents that never emit a session generation at all, and refusing it would break them outright.

## Rule 12 — cross-version / protocol verdict

- Wire shape is unchanged: `PROTOCOL_VERSION` stays **7**.
- This is a **semantic break behind a stable wire** — the meaning of an identity-less paned write changed, not its shape. Documented in `changelog.d/608.breaking.md`.
- Per the 0.x bump policy, a semantic break drives a **minor** version bump.
- Cross-version tested:
  - v0.36.2 and v0.36.1 — unaffected.
  - **v0.36.0 is the single release the break reaches** — it predates #424, so it names the agent but never binds the conversation.
  - v0.35.10 and earlier are already refused at the handshake, so they're unaffected by this change.

## Gate results

- `cargo test-fast`: **2887 tests run, 2887 passed, 0 skipped**
- `cargo fmt --check`: clean
- `cargo clippy --workspace --all-targets --features e2e -- -D warnings`: clean
- `cargo test-e2e` (pre-PR L2 tier): **8612 tests run, 8610 passed, 2 failed, 0 skipped** — both failures reproduce green in isolation and are unrelated to this branch:
  - `chain_smoke_pi_001_orchestrator_delegates_to_real_worker` — died at 3.2s on `PermissionDenied` copying the host's `~/.claude/.credentials.json` for the *claude worker*, before any model call, on a line this branch does not touch. Isolated rerun: 1 passed.
  - `dispatch_015_three_real_claude_seeds_are_genuinely_confirmed` — one of three concurrent real-Claude panes came up `● Error` with `Tools: 0` and never got a turn; the mechanism under test reported `all_first_attempts_swallowed=true`. Isolated rerun: 1 passed.

## Side-work riding along in this PR

Four of the eight commits are deliberate side-work the user asked to include here, not part of the #608 fix itself:

- `487cf8a` — three legacy identity-less e2e clients converted to guarded clients.
- `e90434e` — dev-env config: route the pi reviewer to OpenAI, give vals a durable credential.
- `b80aa1c` — dev-env config: add `ANTHROPIC_API_KEY` for the real-pi e2e tier.
- `5d0cf61` — move the pi e2e tier to Anthropic Haiku **temporarily**, while the GPT accounts are dry. This is provider-agnostic; reverting means changing back `PI_MODEL`, the `--provider` flag, and the credential/key name — three things, no more.

## Follow-ups filed from this work

- #617 — the sibling family of permissive-on-absent-identity write sites (five sites), deliberately left out of scope for this PR.
- #621 — a `Stale` refusal names no generation, so a prompt can be abandoned when the event stream drops a `SessionStart`.


prageethw (approved):
This closes a real permissive-on-None gap in compute_write_and_submit_outcome on two identity axes (missing expected_agent_id falling through to a pane-id-only match; an attached pane with a live session but no session named by the caller skipping the session check entirely), and does it well: the new tests hit the fail-closed paths through a real socket against a real in-process daemon and assert no bytes leaked, not just the return code, and several production e2e tests were migrated to the new identity-bearing call so a regression in real callers would be caught. The residual gaps you documented (the just-cleared-session window, has_live_attach's staleness direction) are called out candidly rather than glossed over, and the new unnamed-session refusal is deliberately written not to depend on the stale-in-the-permissive-direction value. Correctly classified as a same-wire semantic break per your own versioning discipline (breaking.md fragment, no PROTOCOL_VERSION bump, wire shape unchanged) with a clear who's-affected section. The four red build checks are unrelated pre-existing main breakage, fixed by #609 (now approved) — will need a rebase. Approving on content.

PR #639, opened 2026-08-21
fix(ui): restore the basename when cycling away from an orchestration (#638)

### LINKED ISSUE #638: New-pane form: the `-orchestrator-N` suggestion survives cycling away from the orchestration, so plain panes, mode panes and scheduled tasks get orchestrator names
Reported on **v0.36.2**, the release that shipped #538/#539.

## Symptom

Open `Ctrl+n`, pick a directory, and the Name field pre-fills with the directory basename (`dot-agent-deck`) as it always has. Now cycle the Mode field. The moment the selection lands on an orchestration the Name becomes `dot-agent-deck-orchestrator-1` — correct, that is #538. But cycling **away** from the orchestration leaves that name in the field. Select "No mode" and submit, and you get a plain pane named `dot-agent-deck-orchestrator-1`.

The reporter hit this while creating a pane with **no mode at all**, which is what made the suggestion look like it had replaced the ordinary basename pre-fill outright.

## Why it is easy to hit rather than a corner case

The cycler orders options `No mode → …modes… → …orchestrations… → schedule → schedule: issues → dispatcher` (`mode_option_name`). The orchestrations sit **between** "No mode" and the built-in `schedule`/`dispatcher` options, so reaching those from the left necessarily passes over an orchestration. In this repo (one mode, one orchestration) every one of these ends up named `dot-agent-deck-orchestrator-1`:

- a plain pane (`No mode`)
- a `dev`-mode pane
- a `schedule` authoring card
- a `dispatcher` card

## Cause

`NewPaneFormState::suggest_name_if_orchestration_selected` (`src/ui.rs`) is a one-way overwrite — it applies the suggestion on an orchestration and returns early on everything else, so nothing ever restores the pre-fill:

```rust
fn suggest_name_if_orchestration_selected(&mut self) {
    if self.name_touched {
        return;
    }
    if self.selected_orchestration().is_some() {
        self.name = self.suggest_orchestration_name();
    }
    // ← no else: the generated name outlives the selection that generated it
}
```

`name_touched` is only set by the two text-edit key arms, so the stale value is *not* protected as a human edit — it is a generated default leaking across selections.

`orchestration/identity/004` covers arrowing away and back without clobbering a **typed** name, so the leave path was exercised — but only ever asserted against a name the user had typed, never against the generated default. That is the gap.

## Fix

Make the re-suggest two-way: remember the basename the form opened with (`name_prefill`) and restore it whenever the selection is not an orchestration. The `name_touched` guard is unchanged, so it still never overwrites a typed name.

## Scope

Presentation only — the new-pane form's Name field. No daemon, protocol, hook or orchestration-runtime change, so no `PROTOCOL_VERSION` bump and no `.breaking.md` (CLAUDE.md rule 12).

## Not in scope

Whether the suffix should appear on the *first* orchestration at all, and whether `N` should count per-directory instead of globally, were raised alongside this and **deliberately left as-is** — that is what #538 specified.

vfarcic:
Fixes #638.

## What was wrong

The new-pane form's Name field pre-fills with the picked directory's basename, and #538/#539 made selecting an orchestration replace that with the suggested `<folder>-orchestrator-N`. That replacement was **one-way** — `suggest_name_if_orchestration_selected` applied the suggestion on an orchestration and `return`ed early on everything else, so nothing ever restored the pre-fill:

```rust
if self.selected_orchestration().is_some() {
    self.name = self.suggest_orchestration_name();
}
// ← no else: the generated name outlives the selection that generated it
```

So a pane created with **no mode at all** — or with an ordinary workload mode, or as a `schedule` / `dispatcher` card — ended up named `<folder>-orchestrator-1` despite never having been an orchestration.

## Why it was easy to hit

`mode_option_name` orders the cycler `No mode → …modes… → …orchestrations… → schedule → schedule: issues → dispatcher`. The orchestrations sit **between** "No mode" and the built-in options, so reaching any of those from the left necessarily passes over an orchestration — and passing over one was enough. The plainest path to it was opening the form, cycling through to see what is on offer, settling back on "No mode", and submitting.

## The fix

Rename to `resuggest_name_for_selection` and make it two-way: an orchestration suggests `<folder>-orchestrator-N` as before, anything else restores `name_prefill` — the basename the form opened with, captured in the constructor.

The `name_touched` guard is deliberately untouched, so a name the user typed is still never overwritten in either direction. `orchestration/identity/004` continues to pin that and passes unchanged.

## Reproduce-first

Test written and confirmed failing **before** the fix, against `v0.36.2` — the released build this was reported on:

```
assertion `left == right` failed: leaving the orchestration must restore the
basename pre-fill — a plain pane must not be named `-orchestrator-N`
  left: "myproj-orchestrator-1"
 right: "myproj"
```

`orchestration/identity/006` carries a **control** assertion — landing on the orchestration still suggests `-1` — which passes first, so the failure is attributable to the leave path specifically and not to the suggestion feature as a whole. The fix was then confirmed load-bearing by reverting only the `else` branch and watching the same test go red again.

## Gates

- `cargo test-fast` — **2944 passed, 0 failed** (includes `orchestration/identity/002`–`005` unchanged)
- `cargo fmt --check` — clean

## Contract check (CLAUDE.md rule 12)

Presentation only — the new-pane form's Name field. No daemon, TUI↔daemon protocol, orchestration-runtime or hook change, so no `PROTOCOL_VERSION` bump and no `.breaking.md`. Cross-version run not applicable.

## Deliberately not in scope

Whether the suffix should appear on the **first** orchestration at all, and whether `N` should count per-directory rather than globally, came up alongside this and were decided to stay as-is — that is what #538 specified, and the global counter is what stops `-orchestrator-1` being offered in two directories at once.

https://claude.ai/code/session_0181zWG1w5h7SebEmLBuSpEQ

### DIFF
--- a/changelog.d/638.bugfix.md
+++ b/changelog.d/638.bugfix.md
@@ -1,0 +1,7 @@
+## A plain pane is no longer named after an orchestration you only cycled past
+
+The new-pane form's Name field pre-fills with the picked directory's basename, and selecting an orchestration replaces that with the suggested `<folder>-orchestrator-N`. Cycling *away* from the orchestration left the suggestion behind, so a pane created with no mode at all — or with an ordinary workload mode, or as a `schedule` or `dispatcher` card — was named `<folder>-orchestrator-1` despite never being an orchestration.
+
+This was not a corner case, because the Mode cycler orders the project's orchestrations *between* "No mode" and the built-in `schedule` / `schedule: issues` / `dispatcher` options. Reaching any of those from the left means passing over an orchestration, and passing over one was enough: the suggestion applied on the way through and nothing ever put the basename back. The likeliest way to meet it was the plainest one — open the form, cycle through the options to see what is on offer, settle on "No mode", and submit something now claiming to be an orchestrator.
+
+The suggestion is now two-way: landing on an orchestration suggests `<folder>-orchestrator-N` as before, and landing on anything else restores the directory basename the form opened with. A name you typed yourself is still never overwritten in either direction.
--- a/src/ui.rs
+++ b/src/ui.rs
@@ -1074,13 +1074,25 @@ struct NewPaneFormState {
     /// opened — set by the two text-edit key arms in
     /// `handle_new_pane_form_key` (`FormField::Name` only), never by the
     /// basename pre-fill `transition_after_dir_pick` applies. While `false`,
-    /// [`Self::suggest_name_if_orchestration_selected`] is free to overwrite
+    /// [`Self::resuggest_name_for_selection`] is free to overwrite
     /// `name` with the generated suggestion; once `true`, that fn leaves the
     /// field alone — a generated default may replace a generated default,
     /// never a human edit. Guards re-clicking the selected chip, arrowing
     /// between orchestrations, and arrowing off and back onto one, all of
     /// which land on the same call.
     name_touched: bool,
+    /// The Name the form opened with — the directory basename
+    /// `transition_after_dir_pick` pre-fills. Kept so that cycling AWAY from an
+    /// orchestration can put it back: the `-orchestrator-N` suggestion belongs
+    /// to the orchestration selection, and a plain pane, a workload mode or the
+    /// built-in `schedule`/`dispatcher` options must not inherit it. Without
+    /// this, the suggestion was a one-way overwrite — and since the cycler
+    /// orders orchestrations BEFORE `schedule`/`dispatcher`, merely passing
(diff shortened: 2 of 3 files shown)


prageethw (approved):
Clean, well-scoped fix: the one-way suggestion is correctly made two-way via `name_prefill`, the `name_touched` guard is untouched, and the new test (`orchestration/identity/006`) covers both the backward-to-"No mode" and forward-past-the-orchestration paths with a clear control assertion. CI green.

PR #641, opened 2026-08-21
fix(renovate): close the automerge bypass lane on the Rust toolchain pin

### LINKED ISSUE #640: Renovate: the manual-review hold on the Rust toolchain pin has an automerging bypass lane
## Summary

`renovate.json` holds the Rust toolchain pin (and the `cargo-nextest` pin) for manual review on purpose. That hold has a **bypass lane**: the same `toolchain:` string is detected a second time by Renovate's `github-actions` manager, which routes it into the automerging `GitHub Actions` group. The hold has therefore been advisory rather than enforced.

## Evidence

Two open PRs propose the identical 1.97.1 → 1.98.0 bump, from different lanes, with different automerge verdicts:

| PR | Lane | depName | depType | Branch | Label | Automerge |
|---|---|---|---|---|---|---|
| #624 | `custom.regex` (ours) | `rust-lang/rust` | — | `renovate/rust-lang-rust-1.x` | `manual-review` | **Disabled by config** |
| #623 | `github-actions` | `rust` | `uses-with` | `renovate/github-actions` | none | **Enabled** |

Their diffs are byte-identical — the same seven `toolchain: 1.97.1` → `1.98.0` edits across `ci.yml` (5), `release.yml` (1) and `aarch64-crossbuild-check.yml` (1); both branches produce the same blob hashes (`9596228..a8688de` etc.).

#623 did not automerge itself only because CI happened to be red from an unrelated `main` breakage (fixed by #609).

## Root cause

`renovate.json`'s customManager description asserted:

> The github-actions manager cannot see this: it reads the `uses:` ref

That was true when written and is false now. Renovate's `github-actions` manager carries a registry of community actions whose version lives in an **input** rather than the ref — `lib/modules/manager/github-actions/community.ts`. `dtolnay/rust-toolchain` is in it:

```ts
// https://github.com/dtolnay/rust-toolchain
'dtolnay/rust-toolchain': {
  datasource: RustVersionDatasource.id,
  packageName: 'rust',
  withSchema: valSchema('toolchain'),
},
```

Every dep produced that way is tagged `depType: 'uses-with'`. The `GitHub Actions` packageRule matched `matchManagers: ["github-actions"]` with **no depType filter**, so it automerged all six depTypes that manager can emit (`action`, `docker`, `container`, `service`, `github-runner`, `uses-with`) — tool versions read out of action inputs included.

The Dependency Dashboard (#253) shows both lanes side by side: `rust 1.97.1 → 1.98.0` seven times under `github-actions`, and `rust-lang/rust 1.97.1 → 1.98.0` seven times under `regex`.

## Why the hold matters

Per the packageRule's own description: the workflow toolchain pins must stay in lockstep with `devbox.json`'s `rustc`/`cargo`/`clippy`/`rustfmt`, nothing enforces that lockstep automatically, and a Rust release is where new deny-by-default lints arrive. 1.98.0 ships a deny-by-default `invalid_runtime_symbol_definitions` plus warn-by-default `suspicious_runtime_symbol_definitions` and `c_void_returns`.

This is not hypothetical for 1.98.0: **nixpkgs has no Rust 1.98.0 yet** (`devbox search rustc@1.98.0` → `No results found`; latest is 1.97.1). Had #623 automerged, CI would have moved to 1.98.0 while every `devbox shell` stayed on 1.97.1 — breaking the "green locally and green in CI are the same claim" property that `ci.yml`'s own comment says the pinning exists to provide.

Worse, the group is a *group*: with any ordinary action bump pending, a Rust toolchain change would have arrived bundled inside a PR titled "update GitHub Actions" and merged without being read on its own.

## Fix

Narrow the `GitHub Actions` automerge rule to `matchDepTypes: ["action"]` — a whitelist, so anything this manager learns to detect in future is held by default rather than needing a new exclusion — plus an explicit, labelled hold rule for the non-action depTypes.

`taiki-e/install-action` is **not** in the community registry today, so the `cargo-nextest` pin has the same exposure latently rather than actively; the whitelist covers it the day that changes, with nobody having to notice.

vfarcic:
Closes #640.

## The gap

`renovate.json` holds the workflow Rust toolchain pin for a human even on a patch — because it must move in lockstep with `devbox.json`'s `rustc`/`cargo`/`clippy`/`rustfmt` by hand, and because a Rust release is where new deny-by-default lints arrive. **That hold has been advisory rather than enforced.**

The customManager's description asserted:

> The github-actions manager cannot see this: it reads the `uses:` ref

True when written; false now. Renovate's `github-actions` manager carries a registry of community actions whose version lives in an **input** rather than the ref — `lib/modules/manager/github-actions/community.ts`:

```ts
// https://github.com/dtolnay/rust-toolchain
'dtolnay/rust-toolchain': {
  datasource: RustVersionDatasource.id,
  packageName: 'rust',
  withSchema: valSchema('toolchain'),
},
```

It reads the identical `toolchain:` string the custom manager was written for and emits it as depName `rust`, depType `uses-with`. The `GitHub Actions` packageRule matched `matchManagers: ["github-actions"]` with **no depType filter**, so it automerged all six depTypes that manager can emit (`action`, `docker`, `container`, `service`, `github-runner`, `uses-with`) — tool versions read out of action inputs included.

Both lanes were sitting in the open, side by side:

| PR | Lane | depName | depType | Branch | Label | Automerge |
|---|---|---|---|---|---|---|
| #624 | `custom.regex` (ours) | `rust-lang/rust` | — | `renovate/rust-lang-rust-1.x` | `manual-review` | Disabled by config |
| #623 | `github-actions` | `rust` | `uses-with` | `renovate/github-actions` | none | **Enabled** |

Byte-identical diffs — same seven `toolchain: 1.97.1` → `1.98.0` edits, same blob hashes. #623 did not merge itself only because CI was red from an unrelated `main` breakage, since fixed by #609.

## What that would have cost

Not hypothetical for 1.98.0: **nixpkgs has no Rust 1.98.0 yet.** `devbox search rustc@1.98.0` → `No results found`; 1.97.1 is the newest it offers, for all four of `rustc`/`cargo`/`clippy`/`rustfmt`. An automerged #623 would have moved CI to 1.98.0 while every `devbox shell` stayed on 1.97.1 — breaking the property `ci.yml`'s own comment says the pinning exists to provide, that `cargo test-fast` locally and the same command in CI are the same claim.

And because the rule carries a `groupName`, a toolchain change could have arrived bundled inside a PR titled "update GitHub Actions" alongside a handful of `@v7` bumps, and merged without being read on its own.

## The fix

**Whitelist, not blacklist.** The `GitHub Actions` automerge rule is now scoped to `matchDepTypes: ["action"]` — `uses: owner/repo@ref` and nothing else. Excluding the one dep name observed misbehaving would have been correct only until Renovate's registry grew its next entry; a whitelist leaves anything the manager learns to detect tomorrow **outside** the automerging rule with no edit required. Being wrong this way costs a PR someone has to click; being wrong the other way costs an unreviewed toolchain change on `main`.

**The hold is stated, not implied.** A second rule holds the non-action depTypes explicitly and labels them `manual-review`, following the convention the 0.x-minor rule set: a hold that exists only as an absence is invisible on the PR page and impossible to grep for.

**Both stale claims corrected.** The toolchain customManager's description records what actually changed and why it had teeth. The `cargo-nextest` manager makes the same claim, and that one is still true — `taiki-e/install-action` is **not** in the registry today — so it is now written as an expiry date rather than a property, with the depType rules covering it the day that changes.

**The duplicate PR is deliberately not suppressed.** Disabling the `github-actions` lane for `rust` would have removed the noise and the redundancy together, leaving the custom manager's regex as the only detector of a pin whose named failure mode is *silent rot* — a reformatted `toolchain: "1.98.0"` (quoted, which the regex does not match) would take both lanes down at once. Two held PRs per Rust release is the cheaper mistake, and they self-resolve: the lanes edit the same lines, so merging either makes the other conflict and `rebaseWhen: conflicted` autocloses it, exactly as happened to #252.

Net effect on what automerges: every action ref still does. `rust` and the docs site's `node` (`actions/setup-node`'s `node-version:`, the repo's only other `uses-with` dep) move to held-and-labelled. That is the intended reading — both are toolchains.

## The 1.98.0 bump is deferred, on purpose

Only `renovate.json` moves here. The bump cannot be done in lockstep today, because nixpkgs has no 1.98.0 (above), and moving the workflow pins alone is precisely the divergence the hold exists to prevent. #624 stays open on the correct lane, waiting for nixpkgs.

Measured what the bump *would* cost, so whoever lands it is not starting cold — under a rustup-installed 1.98.0, against this branch:

- `cargo +1.98.0 fmt --check` — clean.
- `cargo +1.98.0 clippy --workspace --all-targets --features e2e -- -D warnings` — clean, zero warnings.
- None of 1.98.0's three new lints fire: deny-by-default `invalid_runtime_symbol_definitions`, warn-by-default `suspicious_runtime_symbol_definitions` and `c_void_returns`.

So the bump is a pure version change whenever nixpkgs catches up. No code work is pending.

## PR housekeeping

#623 closed as the redundant lane, with a comment pointing at #624 and at this PR. #624 kept.

## Unrelated finding, not fixed here

The `cargo-nextest` lockstep has **already drifted**: `devbox.json` is at `0.9.143` (automerged via the `Devbox packages` rule, #510) while all three workflow pins are still `0.9.140`. crates.io's `max_stable_version` is `0.9.143`, but the Dependency Dashboard shows the custom regex manager reporting no update for those three deps. Worth its own issue; deliberately left out of a config-fix PR.

## Verification

- `npx renovate-config-validator` — validates as repository config. Confirmed the check is meaningful by injecting a bogus key first, which it rejected.
- `cargo test-fast` — 2943 passed, 0 failed.

Changelog fragment: `changelog.d/640.misc.md`.

https://claude.ai/code/session_01BBQhMDzxzdKR1tV1RSbPev


prageethw (approved):
Solid config fix — the whitelist-over-blacklist reasoning (matchDepTypes: ["action"] rather than excluding just "rust") is the right call given Renovate's community-actions registry only grows, and the deliberate non-suppression of the duplicate PR lane (self-resolving via rebaseWhen: conflicted) is well justified. Verified via renovate-config-validator, CI green.

PR #643, opened 2026-08-22
fix(hook): stop discarding a buffered reply when macOS refuses the timeout re-arm (#642)

### LINKED ISSUE #642: socket_007 is timing-flaky on the macOS fast tier under concurrent CI load
`hook::tests::socket_007_signal_interrupted_read_still_returns_the_reply` (`src/hook.rs:1422`) fails intermittently on the macOS fast tier under concurrent CI load.

**This is one data point for [#502](https://github.com/vfarcic/dot-agent-deck/issues/502)'s Class A** (wall-clock deadlines and render races), not a separate strategy proposal — see *Relationship to existing issues* below. Filed as its own issue only so the instance and its evidence are recorded before the CI logs age out.

## Confirmed flaky: same commit, fail → fail → pass

Run [`32535763053`](https://github.com/vfarcic/dot-agent-deck/actions/runs/32535763053) on `fix/638-…` @ `0ece2b7`, three attempts, **no code change between them**:

| Attempt | `build-macos` | Job |
|---|---|---|
| 1 | **fail** | `96936283678` |
| 2 | **fail** | `96937723336` |
| 3 | **pass** | `96940445038` |

Deterministic breakage cannot produce that sequence.

## Not attributable to any branch

`socket_007` is present in every commit below (verified at `5f78810`); only the *timing* differs.

| Time (08-21) | Branch | `build-macos` |
|---|---|---|
| 21:33 | four `agent/dispatch-issue-*` branches | **pass** (×4) |
| 21:51 | `main` | **pass** |
| 23:02 | `main` @ `8997363` | **fail** |
| 23:08 | `fix/638-…` @ `0ece2b7` (attempt 1) | **fail** |
| 23:21 | `fix/638-…` (attempt 2, same commit) | **fail** |
| 23:25 | `agent/dispatch-renovate-toolchain-lane` @ `5f78810` | **pass** |
| 23:36 | `fix/638-…` (attempt 3, same commit) | **pass** |

Every failure falls inside one 23:02–23:21 window during a burst of concurrent runs, and it passes either side. `0ece2b7` touches only `src/ui.rs`, `tests/CATALOG.md` and `changelog.d/`; `src/hook.rs` is untouched.

## Failure

```
FAIL dot-agent-deck hook::tests::socket_007_signal_interrupted_read_still_returns_the_reply
2943 tests run: 2942 passed, 1 failed
```

```
a read interrupted by a signal must be retried, not mistaken for an absent daemon —
got NoReply (read error: Some(Io(Os { code: 22, kind: InvalidInput,
message: "Invalid argument" }))); the stub daemon reports: wrote and flushed the reply
line. `SocketReply::Unreachable` here would mean a signal landed during connect/write
instead of during the read, which the 50ms lead-in is sized to prevent.
```

The assertion names its own fragility: correctness depends on a **50ms lead-in** landing the signal inside the read rather than during connect/write. On a loaded macOS runner that window is not reliable.

## A possible product bug hiding behind the flake

The errno is **`EINVAL` (22), not `EINTR`**. Worth checking independently of the timing: if the retry path added by #598 does not recognise how macOS surfaces a signal-interrupted read, then the load window is *exposing* a real gap rather than being the whole story. That half would not be fixed by any amount of test-timing work.

## Evidence is being masked, as #564 warned

#564 recorded that a job-level re-run overwrites the conclusion so the failure leaves no trace on the check board. That happened twice here — the run is on `run_attempt=3` and the board now reads green. The failures survive only at `/actions/runs/32535763053/attempts/1` and `/attempts/2`, plus `main`'s at [`32535373581`](https://github.com/vfarcic/dot-agent-deck/actions/runs/32535373581).

## Why it still matters

`retries = 0` in `.config/nextest.toml`'s default profile and `build-macos` is one of the four **required** checks, so one timing blip hard-reds every PR open at that moment — here it red-flagged `main` and an unrelated TUI-only PR simultaneously. The remedy on offer erases the evidence.

## Relationship to existing issues

- **[#502](https://github.com/vfarcic/dot-agent-deck/issues/502)** — the umbrella. Its Class A already lists #364, #247, #413, #395, #244; this is another member. #502 asks to *decide a permanent strategy, then let the individual issues be closed by it rather than one at a time*, so the class-level decision belongs there, not here. Note #247 is also a **fast-tier** wall-clock flake, so Class A already spans both tiers.
- **[#564](https://github.com/vfarcic/dot-agent-deck/issues/564)** (closed) — `socket_004`, the sibling test in the same `mod tests`. Its closing note asked whether `socket_004` was alone; it is not, and both share the cross-thread wall-clock bound.

## Instance-level directions, if #502's strategy does not subsume them

- Wait on the daemon thread's own readiness signal instead of a wall-clock bound, so the test cannot lose to the scheduler.
- Or scope a `slow-timeout`/retry override to this test by name in `.config/nextest.toml`, as `route_001` and `new-pane/016` already are.
- Audit the rest of `src/hook.rs`'s `mod tests` for cross-thread wall-clock bounds — two instances now.

Found while working #638/#639; unrelated to that change.

vfarcic:
Closes #642.

## What actually produces the `EINVAL`

#642 flagged it as an open question — "the errno is **`EINVAL` (22), not `EINTR`** … if the retry path added by #598 does not recognise how macOS surfaces a signal-interrupted read, then the load window is *exposing* a real gap". It is a real gap, and it is not in how the read reports a signal. **It is the per-read timeout re-arm, one line above the read.**

XNU's `sosetoptlock()` refuses **every** `setsockopt` with `EINVAL` once a socket carries both `SS_CANTSENDMORE` and `SS_CANTRCVMORE` — *"the socket has been shutdown, no more sockopt's"*. Both get set in the ordinary course of a `get-seed`:

- `SS_CANTSENDMORE` — set by **us**. `request_from_socket_at_detailed` half-closes its write side (`shutdown_write`) before it reads, so the daemon's line reader sees EOF. That has been there since PRD #163 M4.
- `SS_CANTRCVMORE` — set when the **peer** closes. `soisdisconnected`/`socantrcvmore` sets it immediately, with the peer's bytes still sitting in our receive buffer.

So on macOS, from the moment the daemon closes, every trip round `read_reply_line`'s loop hits a re-arm that fails with `EINVAL`, and #598 propagated that as `ReplyReadError::Io` → `SocketReply::NoReply` — **discarding a reply that was already complete in the kernel buffer.** Linux's `sock_setsockopt` has no shutdown rule, which is why this is macOS-only. It is also why the failure text reads `NoReply (read error: Some(Io(Os { code: 22 … })))` rather than `Unreachable`: the assertion message's guess ("a signal landed during connect/write") was the wrong suspect, and `IpcClient::connect_timeout` has retried `EINTR` since #599 anyway.

**This is not only a test problem.** The daemon's hook loop answers `get-seed`, reads EOF from our half-close on its very next pass, and drops the connection (`src/daemon.rs`, `run_hook_loop`) — so "the peer is already gone by the time the read loop goes round again" is the *ordinary* ending of a get-seed, not a corner case. Any second pass on macOS — an `EINTR`, an early `SO_RCVTIMEO` fire, a reply split across two reads — silently lost the seed and fell back to PTY injection. `socket_007` under CI load was just the first thing loud enough to notice.

## The evidence, before the fix

This PR opens **deliberately red on `build-macos`**, at [`9926a49`](https://github.com/vfarcic/dot-agent-deck/commit/9926a49), which is tests-only. `error/socket/008` reproduces the mechanism deterministically — one thread, no sleeps, no signals: connect, accept, write the reply, half-close our write side, close the peer, then call `read_reply_line`. On macOS the very first re-arm fails and the buffered reply is thrown away; on Linux and Windows it passes either way. That red run is the confirmation #642 asked for and could not get by re-running a race.

The fix lands in the following commit and the same job must go green.

## What changed

**`src/hook.rs` — a failed re-arm is no longer fatal.** It is logged and the read is attempted anyway. Safe rather than merely hopeful, on two counts: a socket in the state that produces this error *cannot block* in `read(2)` (`SS_CANTRCVMORE` means the next read returns buffered bytes or EOF at once); and more generally the re-arm only ever **tightens** a bound that is already in place — the caller arms the socket before it writes — so a failed one leaves the previous, never-larger-than-the-budget timeout standing rather than leaving the read unbounded, while the loop head still returns `DeadlineExpired` the moment the budget is gone. `socket_003`/`socket_005` still return at ~5.0s, unchanged.

**`socket_007` — both wall-clock bets deleted, not widened.** #642 asked for determinism over a bigger window, and there were two bets, not one:

| was | now |
|---|---|
| `SIGNAL_LEAD_IN` = 50ms, hoping connect+write finished | **handshake** — the stub daemon reports having read the request line, which *proves* connect and the write are behind us |
| `REPLY_DELAY` = 300ms, i.e. the daemon answered at the very moment the 25th signal was sent | **handshake** — the reply is withheld until the last signal has been sent, so it cannot race the storm |
| 25 sends, nothing checked | the handler **counts its own deliveries**; the loop signals until it has seen 20, and the count appears in the failure message |

That last row is a strengthening, not a wash: the old test could deliver zero signals into the read and still pass, i.e. degrade silently into a tautology. Now "a signal really landed inside the read" is observed. The only wall clock left is a 2s backstop on the signalling loop, present so a machine that cannot deliver 20 signals fails saying *that* rather than surfacing the client's own `DeadlineExpired`. Runtime went from ~0.31s to ~0.06s.

Nothing was loosened: the assertion is still exactly `SocketReply::Line("{\"seed\":\"abc123\"}")`, no outcome was widened, nothing is `#[ignore]`d, and `retries` stays 0.

## Rule 12 — checked, does not apply

The change is confined to the **client's** read loop in `src/hook.rs`. No message shape, field, or field meaning moves; the daemon is untouched; `PROTOCOL_VERSION` is unchanged and needs no bump, and there is no same-wire/different-meaning break, so no `.breaking.md`. An old TUI against a new daemon and the reverse both see the identical `get-seed` exchange — the new build just stops discarding the reply on macOS. `changelog.d/642.bugfix.md` covers it as a bug fix (patch).

## Relationship to #502

#642 filed itself as one data point for #502's **Class A** (wall-clock deadlines and render races) and asked that the class-level decision be made there. This PR does the Class A mechanical work for this instance — replace wall-clock bounds with happens-before edges — and does **not** pre-empt #502's strategy question. It also splits a genuine product defect out of that class: half of #642 was never a timing problem at all.

## Gates

`cargo fmt --check`, `cargo clippy --workspace --all-targets --features e2e -- -D warnings`, `cargo xtask linkage-check`, `cargo test-fast` and `cargo test-e2e` — results in the comments below.

https://claude.ai/code/session_017ah6h1p1JYQs4FjFFVBRTb


vfarcic:
## The experiment, run on `build-macos`

The two commits were pushed separately on purpose, so the same CI job could be observed on both sides of the fix.

### Before — [`9926a49`](https://github.com/vfarcic/dot-agent-deck/commit/9926a49), tests only ([job `96948476464`](https://github.com/vfarcic/dot-agent-deck/actions/runs/32540148633/job/96948476464))

```
PASS [   0.373s] ( 459/2943) hook::tests::socket_007_signal_interrupted_read_still_returns_the_reply
FAIL [   0.026s] ( 460/2943) hook::tests::socket_008_reply_survives_a_peer_that_closed_before_the_read_re_armed

a reply already buffered by the kernel must survive a per-read timeout that can no longer be
re-armed — got Err(Io(Os { code: 22, kind: InvalidInput, message: "Invalid argument" })).

Summary [ 129.317s] 2943 tests run: 2942 passed, 1 failed
```

**`code: 22` — the identical errno from #642 — reached in 0.026s on one thread with no sleeps, no signals and no second thread.** That is the whole claim demonstrated: the `EINVAL` comes from the timeout re-arm, not from a signal-interrupted read. `socket_007` passed in the same job, which is the control: the flake's *trigger* was the signal, its *cause* was not.

`build` (Linux) and `build-windows` were green on that same commit, so the reproduction is macOS-specific exactly as predicted.

### After — [`c452b10`](https://github.com/vfarcic/dot-agent-deck/commit/c452b10), the fix

`build-macos` green, and re-run four more times on the same commit for independent samples:

| run | result | `socket_007` | `socket_008` |
|---|---|---|---|
| [`96949380236`](https://github.com/vfarcic/dot-agent-deck/actions/runs/32540472063/job/96949380236) | 2943/2943 | pass | pass |
| [`96950131718`](https://github.com/vfarcic/dot-agent-deck/actions/runs/32540472063/job/96950131718) | 2943/2943 | 0.170s | 0.013s |
| [`96950764625`](https://github.com/vfarcic/dot-agent-deck/actions/runs/32540472063/job/96950764625) | 2943/2943 | 0.180s | 0.018s |
| [`96951417961`](https://github.com/vfarcic/dot-agent-deck/actions/runs/32540472063/job/96951417961) | 2943/2943 | 0.212s | 0.013s |
| [`96952138465`](https://github.com/vfarcic/dot-agent-deck/actions/runs/32540472063/job/96952138465) | 2943/2943 | 0.341s | 0.035s |

**What that confidence is and is not.** Five green macOS runs do *not* by themselves retire a flake that passed on plenty of runs before — #642's own table has four green `build-macos` runs on either side of the failing window. What carries the weight here is the red/green pair on `socket_008`: a deterministic failure that is now deterministically fixed, on the exact errno the flake reported. The five greens are corroboration, and they also say the machine was not under the load that produced the original failures (`socket_007` at 0.17–0.34s against the 1.394s of the 2026-08-21 23:55 occurrence), so they are weak evidence about the load regime specifically. The claim that `socket_007` is de-flaked rests on the code — both wall-clock bets are now happens-before edges — not on those five runs.

## Local gates

| gate | result |
|---|---|
| `cargo xtask linkage-check` | pass — 550 catalog ids, 478 annotations, 9 rules |
| `cargo test-fast` | **2944/2944 passed** |
| `cargo test-e2e` | 8669 run, 8663 passed, **6 failed** — see below |

### The six e2e failures are pre-existing, and cannot be this change

On rerun, `mode_live_002_real_haiku_user_journey`, `lock_012_real_agent_gated_by_lock_state` and `idle_worker_008_closed_orchestrator_pane_id_reuse_receives_nothing` all passed. The three that stayed red — `shell_activity_005`, `shell_activity_006`, `card_stats_005` — are the ones #502 names by hand as **Class C** (real-agent/model non-determinism); `shell_activity_006` printed its own `PRECONDITION NOT MET: … the model never issued the Bash tool call at all. This is a harness flake (prompt adherence), NOT a badge regression`, which is the 1-in-3 precondition CLAUDE.md rule 4 and `.config/nextest.toml` both document.

More decisively than any rerun: **the production change is a no-op on Linux by construction.** The only edit is that a failed `set_timeouts` is logged instead of returned, and on Linux `setsockopt(SO_RCVTIMEO)` with a valid non-zero duration on a live socket does not fail — there is no shutdown rule in `sock_setsockopt`. The new branch is unreachable there, so nothing in a Linux e2e run can move because of it. That is also why `socket_008` passes on Linux before and after, and only bites on macOS.

prageethw (approved):
Correct root-cause diagnosis (XNU's sosetoptlock refusing setsockopt once both SS_CANTSENDMORE/SS_CANTRCVMORE are set) and a safe fix: reading anyway on a failed re-arm is sound since the re-arm only ever tightens an already-in-place bound and the loop head still enforces DeadlineExpired. The socket_007 rewrite from sleep-based to handshake-based timing is a real determinism improvement, not just a wider window, and the new socket_008 pins the exact mechanism single-threaded. CI green.

Issue #554, opened 2026-08-14
Orchestration tab silently rebuilds role cards from stale daemon metadata when the local config no longer lists the orchestration


prageethw:
## What happens

Rename an orchestration in `.dot-agent-deck.toml` (or remove it) while one of its tabs is live, then reattach the TUI. The tab is rebuilt from the daemon's recorded role metadata instead of the project config, and nothing on screen says so — the role cards no longer correspond to the file that was just edited, and the only trace is a `tracing::info!` line.

The user-visible symptom is "the roles in my deck don't match the roles in my toml", with no obvious way to connect that back to the rename.

## Where

On `main` at the time of writing (`src/ui.rs`, line numbers approximate):

- `:9660-9668` — the reconnect-hydration path distinguishes "local config absent" (legitimate remote-reconnect) from "local config exists but does not list this orchestration — config drift or stale", and logs both at `info`.
- `:9671` — the drift case then calls `resolve_orch_config_for_hydration` (`:2546`), which synthesises an `OrchestrationConfig` from the bucket's daemon-supplied role slots (`OrchestrationConfig::synthesize_from_bucket_metadata`) and rebuilds the tab from that.
- No `session_warnings` entry is pushed on either branch.

## Why this looks like a bug rather than a deliberate trade-off

The snapshot-restore path in the same file treats the identical drift class as user-visible: `resolve_orchestration_for_restore` returns an `Err` naming the orchestration, and the caller pushes it to `session_warnings` and falls back to plain panes (`:9867` and nearby). Hydration reaches the same conclusion — "the saved tab no longer matches the config" — and says nothing.

The synthesis fallback itself is clearly intentional and correct for its motivating case (PRD #111: a laptop TUI attached to a remote daemon whose `bucket.cwd` doesn't exist locally). The gap is only that the *drift* case — where the config file loaded fine and simply doesn't contain that orchestration — is indistinguishable from it on screen.

## Suggested fix

Push a `session_warnings` entry in the `cfg.is_some() && local_orch_config.is_none()` branch only, naming the orchestration and cwd and saying the cards come from the live daemon's roles rather than the project config. Leave the `cfg.is_none()` branch quiet so remote reconnects aren't spammed.

No change to which config wins — re-slotting live panes against a changed role list is a separate decision, and the synthesised config is the right answer for the remote case.

A pure helper along the lines of

```rust
fn hydration_drift_warning(
    config_present: bool,
    local_found: bool,
    name: &str,
[...]

PR #644, opened 2026-08-22
fix: fit every role card, or say what is hidden (#588)

### LINKED ISSUE #588: Orchestration tab silently renders only some role cards when the dashboard is too short
## Symptom

An orchestration tab renders only a **subset** of its roles when the dashboard is too short, with nothing on screen indicating anything was omitted.

A 7-role orchestration showed 5 cards — `orchestrator, developer, tester, reviewer, releaser` — with `researcher` and `documenter` absent. Config, saved session and daemon were all correct:

- `.dot-agent-deck.toml` lists 7 roles
- `~/.config/dot-agent-deck/session.toml` records the same 7
- `dot-agent-deck daemon status` shows 7 live panes, all in the right cwd

The two missing cards are simply never painted. On screen this is **indistinguishable from a role that failed to start**, which is the real cost of the bug.

## Root cause

The two layout axes are decided independently, so neither can compensate for the other:

```rust
let cols    = grid_columns(dashboard_area.width);           // src/ui.rs:15627 — width ONLY
let density = choose_density(sessions.len(), cols, height); // src/ui.rs:15632 — height ONLY
```

- `grid_columns` (`src/ui.rs:19514`) is a pure width→count step function: `<100 → 1`, `>=100 → 2`, `>=180 → 3`. Height is never an input.
- `choose_density` (`src/ui.rs:142-155`) walks `Spacious(11) → Normal(9) → Compact(6)` and returns the first tier that fits all rows — **but when none fits it returns `Compact` anyway**, and the caller then silently slices to whatever fits (`src/ui.rs:15735-15737`):

```rust
let visible_rows = (available_for_density / card_height).max(1);
let end  = (ui.scroll_offset + visible_rows).min(total_rows);
let rows = &all_rows[ui.scroll_offset..end];
```

### Worked example (the reported case)

7 roles, ~30 usable rows, dashboard width < 100:

| Step | Value |
|---|---|
| `sessions.len()` (tab-scoped, from toml) | 7 ✅ |
| `cols = grid_columns(width)` | 1 |
| `choose_density(7, 1, 30)` | Spacious 77 ✗, Normal 63 ✗, Compact 42 ✗ → returns `Compact` |
| `card_height` | 6 |
| `visible_rows = 30 / 6` | 5 |
| `rows = all_rows[0..5]` | **5 painted, 2 never drawn** |

A **2-column** layout needs only `ceil(7/2) = 4` rows × 6 = **24 rows** and would fit comfortably. `grid_columns` refuses to consider it purely because the window is under 100 columns wide. The information needed to fit everything is available at render time — the algorithm never looks.

## What is NOT the cause

Recording this explicitly, because it has now been re-derived twice by separate investigations:

- **Per-tab scoping is correct.** `src/ui.rs:13176-13213` already filters each orchestration tab to its own `role_pane_ids` and sorts by that tab's config order. `sessions.len()` is correctly 7.
- The 5 rendered cards are exactly **positions 1–5 of the toml**, which is the signature of a viewport cut. A scoping bug would have leaked another tab's role names or broken the ordering.
- It is **not** the hydration-drift path in #554 either — that was the first misdiagnosis. A short window produces an identical symptom.

## Suggested fix

Search `(cols, density)` jointly for the first combination that renders **all** N cards, instead of fixing columns from width and then density from height:

```
for cols in 1..=max_cols_for(width):      # width sets the CEILING, not the value
    for density in [Spacious, Normal, Compact]:
        rows   = N.div_ceil(cols)
        needed = rows * density.card_height()
        if needed <= available_height and cols * MIN_CARD_W <= width:
            return (cols, density)
return densest_possible                    # only here may anything be hidden
```

Two things this needs:

1. **An explicit `MIN_CARD_W`.** `grid_columns`'s 100/180 thresholds encode a minimum card width implicitly; deriving `max_cols_for(width)` from a named constant is the substantive part of the change.
2. **`ui.columns` must stay in sync.** `grid_columns` has a second caller at `src/ui.rs:13271` (`ui.columns = grid_columns(dashboard_width)`) driving left/right card navigation. If the two disagree, selection movement desyncs from what is drawn.

**Open design question:** when both axes can't be satisfied at full richness, prefer *more columns at a richer density*, or *fewer columns with smaller cards*? Suggest preferring more columns first — the prompt/tool lines are the card's actual content, and horizontal space is the cheaper sacrifice.

### Considered and rejected

A 4-row density tier below `Compact` (role + dir + border) also fits 7 cards in 28 rows, but it doesn't address the width/height independence and it touches invariants the source itself flags as overlap-prone:

- the blank separator is gated on `density != Compact` at both `src/ui.rs:19856` and `src/ui.rs:119` — they must move together or reserved height desyncs from rendered content;
- the `is_placeholder` branch (`src/ui.rs:19838`) pushes a line unconditionally, ignoring density, overflowing a 4-row card by one.

## Why scrolling is not an acceptable resolution

The grid does scroll (`ui.scroll_offset`, `src/ui.rs:15720-15733`), so the cards are reachable. That is not sufficient:

- A tab's role set is **small, fixed and fully known from the toml** at render time. It is not an unbounded list — it should always be rendered in full.
- A hidden role reads as a dead worker. That ambiguity is what caused two separate misdiagnoses, one of which reached a hydration-drift theory and a filed issue before being ruled out.

If a genuinely tiny window makes fitting impossible, the overflow must at minimum be **signalled**. The pattern already exists in this file: the scheduled-tasks header appends `(↑12 ↓34)` when rows are hidden (`src/ui.rs:18734-18742`), while the card grid renders its title plain (`src/ui.rs:15748`). That asymmetry is worth closing regardless of the layout fix.

## Workaround

Widen the window past 100 columns — `grid_columns` returns 2 and all 7 fit in 24 rows.

## Environment

- Observed on Homebrew `vfarcic/tap` **0.36.0**
- Line references above are against fork `prageethw/dot-agent-deck` at `1da70756` (post-v0.39.0); the same code paths are present upstream

## Related

- #554 — the hydration-drift defect this symptom was initially and incorrectly attributed to

vfarcic:
## What this fixes

An orchestration tab painted only a subset of its role cards when the dashboard was too short, with nothing on screen saying so. The reported case: seven roles, five cards — `orchestrator, developer, tester, reviewer, releaser` — while `researcher` and `documenter` were never drawn, and the title, the toml and `dot-agent-deck daemon status` all correctly said seven.

The cost is not the missing pixels, it is the ambiguity: **a hidden role is indistinguishable from a role that failed to start**. That is what sent two separate investigations after a hydration defect (#554) that was not there.

## Which behaviour I chose, and why

The issue offered two end states — every card reachable, or an unmistakable omission indicator. **This does both, and the split between them is the design.**

**Fit everything, whenever a layout exists that can.** The two layout axes were decided independently and neither could compensate for the other. `grid_columns` was a pure width→count step function (`<100 → 1`, `>=100 → 2`, `>=180 → 3`) with height never an input; `choose_density` walked Spacious→Normal→Compact and returned `Compact` *even when Compact did not fit either*, after which the caller sliced to whatever rows were left. Seven single-column cards needed 35 rows and had 25. Two columns needed 20 and would have fitted comfortably — the information was available at render time and the algorithm never looked.

`choose_grid_layout` now searches both axes together for the first `(columns, density)` pair that renders **every** card. Scrolling already made the cards technically reachable; that is exactly what the issue argues is insufficient, and it is why the primary fix is fitting rather than signalling.

**Signal only what genuinely cannot fit.** On a window too small at any layout the deck keeps the columns it has always had and the title carries `— 7 session(s)  (↓2)` — the same `(↑a ↓b)` marker the Scheduled Tasks header has always used, now from one shared `scroll_indicator` serving both surfaces. That closes the asymmetry the issue calls out as worth closing regardless.

Three sub-decisions worth reviewing explicitly:

- **`MIN_CARD_W = 40`, measured rather than guessed.** The old 100/180 thresholds encode a minimum card width implicitly *and generously* — 50 columns per card at two columns, 60 at three. Those describe a comfortable card, not the narrowest useful one, and conflating the two is what pinned a sub-100-column deck to one column. I rendered a role card across widths to pick the floor: at 40 you still get ten columns of role name (`ClaudeCode · documente…`), the full-width `Dir:` row and the widest ` Last: 0s  Tools: 14 ` bottom-border rung; by 30 the role name has vanished entirely (`ClaudeCode ·…`), which would defeat the point of painting the card.
- **Widening is spent only on completeness.** The search *starts* at `grid_columns` and only widens from there, so a deck that already fits is untouched — the grid never narrows its cards merely to look busier. This is why all 2950 fast-tier tests and every existing snapshot passed unchanged; the only layouts that move are ones that were truncating.
- **When nothing fits, do not escalate.** Narrowing every card is a real cost paid for completeness. If completeness is out of reach the trade buys nothing, so the deck keeps its familiar layout and the indicator — not a squeezed grid — resolves the ambiguity. Answering the issue's open question the way it suggests, *within* a column count the richest density wins: prompt and tool lines are the card's content, horizontal space is the cheaper sacrifice.

## Keeping render and navigation in agreement

This was the trap, and it is worth stating what was actually wrong. `ui.columns` — which drives left/right card navigation — was computed by a **second** `grid_columns` call in the main loop, against a `dashboard_width` re-derived from a hardcoded `33%`. The orchestration split is 34%, or 25% when narrow, so that number was already not the `dashboard_area.width` the cards were laid out in. Two independent answers to one question, with the navigation half being the one nobody could see was wrong.

Rather than teach the second site to compute the same thing, I removed it: **`ui.columns` now has exactly one writer**, `render_card_grid`, which assigns the column count it just drew with. It runs inside the loop's `terminal.draw`, i.e. before any key is handled, so navigation never reads a value from before the first frame. Structural agreement, not a duplicated calculation.

## Tests

Per the `reproduce-first` skill, the failing test came first and each change was verified load-bearing by reverting it alone.

| id | what it pins | red when |
|---|---|---|
| `dashboard/grid/001` | the report, reproduced exactly on a 90x27 deck, then all seven roles painted in two columns. Plus a 90x60 **control** that already fitted — proving the fixture is not what the assertion measures, and that a fitting deck keeps **one** column | column escalation reverted: five cards, `researcher`/`documenter` absent |
| `dashboard/grid/002` | `UiState::columns` vs the card columns counted in the rendered buffer, across six geometries | nav path left unfixed: `left: 1, right: 2` |
| `dashboard/grid/003` | `(↓5)` / `(↑5)` in the title, the full `7 session(s)` count kept, and no column escalation when nothing fits | indicator reverted: bare `— 7 session(s)` |

Plus unit tests for the `max_columns_for_width(w) >= grid_columns(w)` invariant (swept 0..=400 — if that ever inverted, the fix would have become a regression), the widen-only-to-fit rule, the `MIN_CARD_W` floor, and the shared indicator.

**A new L1 seam was needed.** Every pre-existing dashboard seam renders cards the caller has already laid out — `render_dashboard_cards_to_buffer` stacks them one per row at a density the test names — so the layout decision itself, the thing that was dropping cards, had L1 coverage nowhere. `render_card_grid_to_buffer` drives the real `render_card_grid`. Its probe deliberately carries **one** field: everything else is read out of the drawn cells (cards by their corner glyphs, columns by the widest row, overflow by the title), so no assertion rests on a second copy of the layout arithmetic. `nav_columns` is the exception because it is state, not pixels.

## Rule 12 — checked, does not apply

Presentation only. The diff is `src/ui.rs` render code, tests, one docs page and a changelog fragment; it touches no daemon, protocol, orchestration or hook contract, changes no wire shape and no field's meaning, and `PROTOCOL_VERSION` is unchanged. No `.breaking.md` fragment, and no cross-version manual test is owed.

## Gates

- `cargo fmt --check` — clean
- `cargo test-fast` — **2950 passed, 0 failed**, no snapshot churn
- `cargo xtask linkage-check` — ok (552 catalog ids, 9 rules)
- `cargo test-e2e` — **8674/8675**. The one failure is `shell_activity_006`, whose own assertion says so in-band: *"PRECONDITION NOT MET … the model never issued the Bash tool call at all. This is a harness flake (prompt adherence), NOT a badge regression: rerun rather than reading this as BADGE WRONG."* It passes on rerun. Earlier contended runs produced disjoint sets of real-agent flakes (Pi `Connection error`, prompt-adherence timeouts); every one that touched layout or dispatch — `e2e_card_layout` in full, `orchestration_dispatch_002`, `idle_worker_008` — passes cleanly in isolation.

## Docs

`docs/troubleshooting.md` gains *"The deck is missing cards"*: read the title, `(↓2)` means scrolled off rather than dead, `j`/`k` brings them back — and if there is no marker, the count in the title versus `daemon status` is what tells you the role never started. Given the diagnosis this bug cost twice, the page seemed worth more than the code comment.

Closes #588

https://claude.ai/code/session_011aGfDmT9SMUGJjQVbrNv41


prageethw (approved):
Good fix on both axes: searching columns×density together (rather than deciding them independently) directly addresses the reported card-dropping, and collapsing ui.columns to a single writer in render_card_grid closes a real nav/render desync that was invisible before. The widen-only-when-needed invariant (untouched layout when it already fits) is verified by the max_columns_for_width >= grid_columns sweep. CI green including e2e (8674/8675, the one failure a documented harness flake with rerun confirmation).

Issue #236, opened 2026-07-28
Worktree removal safety + reclamation: never force-remove a dirty tree; reclaim the ones we keep


vfarcic:
## PRD: Worktree removal safety + reclamation

**Problem**: Dispatched worktrees are removed with `git worktree remove --force` on tab close with no dirty/PR/merged check, so uncommitted and untracked work in the tree is destroyed silently; and there is no reclamation path for worktrees that are *not* removed — issue-dispatch relies on the next scheduled fire's idempotency check, while a user-driven dispatch has no next fire, so a leak is permanent.

**Solution**: Fail toward leaking — gate removal on a clean tree (drop `--force` or check `git status --porcelain`) and keep dirty trees with a visible `worktree kept at <path>: uncommitted changes` message — then add the necessary other half: a reclamation path (prune verb / stale-tree sweep / a surface that lists orphaned worktrees) so deliberately-kept trees don't accumulate forever.

**Detailed PRD**: See [prds/236-worktree-removal-safety-reclamation.md](https://github.com/vfarcic/dot-agent-deck/blob/main/prds/236-worktree-removal-safety-reclamation.md)

**Relationships**: Fixes a pre-existing hazard in #120 (issue-dispatch, the only feature shipping this code today); blocks widening it via #220 (dispatcher mode, which makes worktree creation interactive and agent-driven); the same policy will govern #174 (cross-project dispatch).

**Priority**: High

vfarcic:
Audited against `origin/main` while picking a PRD to work on. **Both halves are implemented.** Closing as delivered, with one in-scope item split out as a follow-up rather than silently dropped.

## Half 1 — removal gating: done

`RemovalPolicy` (`src/issue_dispatch_run.rs:95-119`) now makes the choice explicit, and the reasoning this PRD argued is recorded in the code:

> `RemovalPolicy::KeepIfDirty` — PRD #220 dispatch. The name is chosen by an LLM and the tree is a sibling of the user's own checkout, so Ctrl+W reads as "close this view", not "destroy uncommitted work". A leaked worktree costs disk; a force-removed one costs work, and that asymmetry decides it.

[...]

PR #645, opened 2026-08-22
fix: a partial orchestration spawn is all-or-nothing, and the dispatch rollback never deletes a live agent's cwd (#575, #600)

### LINKED ISSUE #575: Spawn-failure rollback force-removes a worktree with live PTY children in a multi-role orchestration
## The hazard

`handle_dispatch`'s spawn-failure rollback runs `git worktree remove --force` (`RemovalPolicy::Force`) on the assumption that no agent has been handed the worktree yet:

```rust
// src/dispatch.rs:397
// `Force` on the rollback path, unlike the tab-close path: we created
// this worktree seconds ago and the agent never started, so there is
// no user work to protect ...
remove_worktree(&paths.worktree_dir, &clone_dir, RemovalPolicy::Force).await;
```

**That assumption does not hold for a multi-role orchestration.** `spawn()`'s orchestration branch spawns roles in a loop (`src/spawn.rs:521`, results zipped per-role at `:579`). If role 2 fails to spawn, roles 0 and 1 are already **live PTY children whose cwd is that worktree**. The rollback then force-removes the directory out from under them, and — unlike the tab-close path — it does not consult `worktree_still_in_use`, so those agents also remain in the registry with a deleted cwd.

## Expected

The rollback should not destroy a worktree that live agents are rooted in. Guarding it with `worktree_still_in_use` — the same check the tab-close path already performs — would leave the tree and report, rather than deleting it.

Worth deciding alongside that: whether the earlier-spawned roles should be closed as part of the failed dispatch, since today they survive with a deleted cwd.

## Reproduction sketch

1. Configure an orchestration with three or more roles.
2. Arrange for a later role's spawn to fail (an unresolvable command is the cheapest way).
3. Observe that earlier roles are live PTY children rooted in the dispatched worktree, and that the rollback force-removes it.

## Relationship to #236

#236 is the tab-close path — *"never force-remove a dirty tree"* — and the comment quoted above explicitly contrasts itself with it. This is the **other** call site, and the hazard is different in kind: not uncommitted content, but live processes whose working directory is deleted underneath them. A fix for that issue does not reach this one.

## Why this is filed here

Verified present on `main` at `e6e6dfea`: the `Force` call is at `src/dispatch.rs:397` and the per-role spawn loop at `src/spawn.rs:521`. Both are upstream code with no downstream modification involved.

It was found in a fork (`prageethw/dot-agent-deck`, issue #332) by an auditor pass during that fork's work on #236. Note the fork corrected the *justifying comment* in its own tree so the code stopped denying the hazard; **the comment quoted above is the current upstream text, which still asserts the assumption that is false.** The behaviour was deliberately left unchanged in both trees, because guarding the rollback is a real behaviour change with its own test and review needs and did not belong folded into a reviewed #236 branch.

Reporting it here rather than only fixing it downstream, so the fix lands where every user of this project benefits.

### LINKED ISSUE #600: Partial orchestration spawn leaves orphaned children with no handle to close them

vfarcic:
Two coupled defects in the same per-role spawn loop, fixed together because fixing either alone leaves the other half incoherent.

A multi-role orchestration spawns its roles one at a time. When a **later** role failed:

- **#600** — the loop `?`d straight out of `spawn_one`, so roles `0..N` stayed live in `AgentPtyRegistry` as PTY children while `spawn` returned `Err`. No `SpawnHandle` came back, so the caller had nothing to close them with, and the `OrchestrationSurface` broadcast that would have given the user a tab sits **past** the early return — so those children were unreachable by every close path that exists.
- **#575** — `handle_dispatch`'s rollback then force-removed the worktree on the stated assumption that *"we created this worktree seconds ago and the agent never started, so there is no user work to protect"*. That is true for a single-agent dispatch and false for an orchestration: the survivors' cwd **is** that tree.

Closes #575. Closes #600.

---

## The design decision

Both issues flag this as a design question and both were deliberately kept out of #454 for that reason. The choice was between rolling the already-spawned agents back (all-or-nothing) and returning a partial `SpawnHandle` so the caller can close or surface what did start.

**Chosen: all-or-nothing.** `Err` from `spawn` now means *nothing is running*.

Three reasons, in order of weight:

1. **A half-spawned orchestration is not a usable product.** It is not a working set minus one member — the orchestrator has already been handed an `orchestrator-context.md` naming co-workers that do not exist, and `handle_delegate` cannot route to a role with no pane. Surfacing it hands the user something that looks live and structurally cannot finish the work. Worse, the failure can be the orchestrator role itself, in which case `agents[orch_idx]` — where the prompt is delivered — was never spawned at all.
2. **No caller can act on a partial handle.** All three (`handle_dispatch`, `issue_dispatch_run`, `spawn_or_reuse`) treat `Err` as "the dispatch failed" and each would have to close the partial set anyway. `handle_dispatch` additionally *needs* the worktree gone or the dispatch name is wedged for every later dispatch of that name — a partial handle defeats the rollback's whole purpose and leaves a half-orchestration in a lingering worktree.
3. **It restores the invariant the rollback always asserted** rather than weakening the rollback to accommodate a state that should not exist. `issue_dispatch_run`'s failure path gets the same fix for free: it drops the worktree registry entry and leaves the directory as an idempotency signal, which was correct only if nothing was running in it.

The cost is honest and worth stating: on a partial failure the user loses the roles that did start and gets an error instead of a tab. That is the right trade when what they would otherwise get cannot do its job.

## What the rollback now does when live agents occupy the tree

`handle_dispatch`'s inline rollback became `rollback_dispatched_worktree`, which asks **`agents_rooted_in_worktree`** — the counting form of `worktree_still_in_use`, the *same* predicate the tab-close path in `daemon_protocol` already asks — before removing anything.

- **Nothing rooted there** → unchanged behaviour: `git worktree remove --force`, `git branch -D`, drop the registry entry. `Force` is still correct; the guard is what makes it correct.
- **Live agents rooted there** → **nothing** is cleaned up. Not the tree, not the branch, and *not the worktree-registry entry* — dropping that entry would be the worse half of the bug it replaces, because the survivors' eventual tab close is what triggers cleanup and it finds the tree by that entry. Retaining the branch follows for free (`git branch -D` refuses a branch checked out in a live worktree). The dispatch message says so, with the count:

  > `dispatch: spawn failed: … (the worktree /…/repo-dispatch-x was LEFT IN PLACE: 2 agent(s) are still running in it, and removing it would delete their working directory underneath them. Close them to release it. Branch agent/dispatch-x is retained too, so this dispatch name stays wedged until then.)`

With #600's teardown in place this is defence in depth rather than the ordinary path — but the invariant is cheap to verify and catastrophic to assume, and a role whose child outlived its teardown must cost a retained directory, not a deleted working directory.

As #575 asks, **the false justifying comment is corrected** rather than left asserting the hazard away. Note this is *not* #236: that is the tab-close path's "never force-remove a dirty tree", and the hazard differs in kind — live processes, not uncommitted content.

The teardown itself uses the same close protocol as the `StopAgent` handler (`begin_pane_close` → `close_agent` on a `spawn_blocking` hop → `unregister_pane` → `finish_pane_close`), not a bare `close_agent`: the roles that started are already in `pane_role_map`, so a delegation could in principle be armed against one. State is unregistered **after** the child is closed, which is why #454's mid-loop registration stays load-bearing — a live child must never be missing its routing entry.

## Rule 12 — the contract answer

**Neither a `PROTOCOL_VERSION` bump nor a `.breaking.md` fragment. This is a bugfix → patch.**

- **Wire shape: unchanged.** `src/daemon_protocol.rs` is untouched by this branch (`git diff main -- src/daemon_protocol.rs` is empty). No request/response type, field or enum moved. Both binaries report `server_version: 7`.
- **Semantics behind a stable wire: unchanged.** No field's meaning shifts. The behaviour change is entirely daemon-*internal* — what the daemon does with its own children when its own spawn fails. The only observable difference to any client is that `ListAgents` no longer returns orphan records after a failed orchestration dispatch, which is strictly less and is the fix. `agents_rooted_in_worktree` is a new in-process helper, not a wire type. The `DispatchResult.message` gains a suffix, but that is human-readable prose the dispatching agent relays, not a contract.

### The cross-version manual test

Run with sockets, `HOME`, state dir **and `DOT_AGENT_DECK_LOG`** all isolated into a per-run sandbox on a disk-backed path (rule 14). Full harness: `xver2.sh` / `probe.sh`, driving the real binaries under `tmux`.

**A finding worth recording first.** "Launch the branch TUI at an older daemon" does *not* by itself produce a mixed pair: `build_version_handshake` silently restarts a build-mismatched daemon **when no agents are running** ("nothing to lose"). My first two runs converged the pair to whichever build the *client* was, which is why an early run showed the new daemon apparently orphaning — it had been replaced by the release binary. Rule 12's *"with an agent under it"* is precisely the condition that makes the pairing reachable: with an agent live the handshake takes its consent path, and declining keeps the old daemon (`ProceedOnExisting`).

So: v0.37.0 daemon started, an agent created under it by the v0.37.0 TUI, that TUI closed, then the **branch** TUI attached and the restart was **declined**:

```
  release  build: 0.37.0-g1122a58
  branch   build: 0.37.0-g2f2c911
  release daemon up, pid 3988483
  agent under the release daemon: pane 1
  PASS the branch TUI detected the build skew and asked for consent:
      | ⚠  Daemon version mismatch  (1 agent(s) running)
      |    [S] restart daemon and continue   [any other key] keep current daemon
  daemon serving this sandbox now: pid 3988483 -> …/bin/dot-agent-deck-0.37.0
  PASS the RELEASE daemon survived; this is a genuine mixed pair
  PASS the branch CLI dispatched an orchestration onto the release daemon
  PASS DELEGATE still routes: Read .dot-agent-deck/worker-task-worker.md for your task.
  PASS WORK-DONE hook still arrives: Worker worker has completed their task. …
  PASS STATUS hooks still arrive: {"status":"Working","active_tool":{"name":"MIXEDTOOL"}}
  daemon serving at the end: pid 3988483 (expected 3988483)
=== MIXED PAIR result: ALL-PASS ===
```

Delegate routes and both hook classes arrive across the version boundary. The daemon pid and `/proc/<pid>/exe` are unchanged end to end, so this is the release daemon throughout, not a silently upgraded one.

### The defect, and the fix, on a live daemon

The same probe against each daemon in turn — the partial dispatch, sampled every 0.3 s. **v0.37.0:**

```
[OLD] tick=13 registry-agents-in-worktree=2 live-tee-procs=2 worktree=GONE
…  (steady through tick=30)
[OLD] FINAL live tee: 3836415 tee …/r0.txt; 3836422 tee …/r1.txt
[OLD] FINAL worktree: GONE
```

Two orphaned children, still running, rooted in a directory that no longer exists. **This branch:**

```
[NEW] tick=1  registry-agents-in-worktree=0 live-tee-procs=0 worktree=present
[NEW] tick=2  registry-agents-in-worktree=0 live-tee-procs=0 worktree=GONE
…  (0 / 0 at every tick through tick=30)
[NEW] FINAL live tee:
```

Nothing orphaned at any sample, and the worktree still reclaimed — safely, because nothing occupies it.

## Tests

Reproduce-first: the failing test came before the fix, and pre-fix RED named the real symptom —

```
a failed dispatch must leave no live agents behind; found 2 orphan(s):
  [(Some("orchestrator"), Some("…/repo-dispatch-partial-unit")),
   (Some("worker-one"),   Some("…/repo-dispatch-partial-unit"))]
PROBE: worktree_still_in_use=true worktree_exists=false
```

Fast tier, unix-gated (Windows CI runs this tier, and a bare `cat` role there would turn "a *later* role failed" into "the *first* role failed" — the test would still pass, for none of the reasons it exists):

| Test | What it pins |
| --- | --- |
| `dispatch::a_partial_orchestration_dispatch_leaves_no_orphans_and_no_deleted_cwd` | The reproducer at the user's altitude: 3 roles, role 2 unresolvable. No live agents, no role-map entries, no orchestrator marker; the tree is reclaimed *because* nothing is in it. |
| `dispatch::a_rollback_with_nothing_started_still_reclaims_the_worktree_and_branch` | **Control.** The first role fails, so nothing ever started — the guard must not fire, and the worktree and branch must still go. Without this, "the tree survived" is indistinguishable from "the guard fires on every rollback", which would wedge every name after a typo. |
| `dispatch::the_rollback_leaves_a_worktree_that_a_live_agent_is_rooted_in` | The guard at its own seam, with a real live PTY child rooted in a real worktree: `Retained { live: 1 }`, tree present, branch present, **registry entry preserved**. Plus the control that the very same call reclaims it once the agent is gone. |
| `spawn::a_partial_orchestration_spawn_leaves_nothing_running_or_registered` | #454's test, whose assertion **inverts by design**. |

Each fix proved load-bearing by reverting it alone: neutering the `worktree_still_in_use` guard reds the guard test; putting the `?` back in the role loop reds the reproducer with the same 2 orphans.

**On the inverted #454 test.** #454 pinned mid-loop registration by asserting the surviving role of a partial failure was still in `pane_role_map`, and said in as many words what it did not assert — *"role 0's child is still running afterwards and the caller gets no handle with which to close it … tracked separately."* #600 is that separate track and it decided the orphan is the defect, so the end state flips from "one role registered" to "none" and the property flips from *ordering* to *atomicity*. #454's ordering remains load-bearing and is still stated at the `register_orchestration_role` call site. What is genuinely no longer observable from that test is inside-the-loop versus after-the-loop registration — under atomicity both end with an empty map; the success-path consequence keeps its coverage in `orchestration/dispatch/001`. The doc comment records all of this rather than quietly rewriting history.

**Why no new L2 test.** This is a daemon-internal failure path with no new user-visible surface, so rule 4's PTY/reel bar does not apply. The fast-tier tests above spawn real PTY children into real git worktrees at the `handle_dispatch` seam — a stricter and far more precise assertion than a grid capture — and the live-daemon probe above covers the user-visible end state end to end.

## Gates

- `cargo fmt --check` and `cargo clippy --workspace --all-targets --features e2e -- -D warnings` — clean before every commit.
- `cargo test-fast` — **2946/2946**.
- `cargo test-e2e` — **8671/8671**. (One earlier run flaked on `idle_worker_008_closed_orchestrator_pane_id_reuse_receives_nothing`; it passes alone, passes as a suite twice, and the clean full re-run above is green. Unrelated to this diff — nothing here runs on that path.)
- Changelog fragments: `changelog.d/575.bugfix.md`, `changelog.d/600.bugfix.md`.

https://claude.ai/code/session_018cjdX65dgiopiDydofDn8x


prageethw (approved):
Thorough fix for both coupled defects. All-or-nothing orchestration spawn is the right call (a partial handle gives the caller nothing actionable, per the three reasons listed), and the rollback guard (agents_rooted_in_worktree, sharing its predicate with the tab-close path so the two can't drift) correctly retains the worktree/branch/registry-entry together when live agents are still rooted there. Test suite includes the deliberately-inverted #454 assertion with a clear explanation, plus a live-agent control test. Cross-version verification (mixed pair + live-daemon probe showing the orphan reproduced on v0.37.0 and fixed on this branch) is convincing. CI green.

PR #646, opened 2026-08-22
fix(delegate): recover a clear=true respawn instead of dropping the task (#584, #606)

### LINKED ISSUE #584: A dispatched orchestration's delegate still fails: the clear=true respawn never emits SessionStart, so the identity gate refuses the task pointer (NoLiveTarget)
An orchestration started with `dot-agent-deck dispatch --orchestration` still cannot delegate to its worker roles, while the same orchestration started with `Ctrl+N` can. The **routing** half of this was fixed by #466 (merged `b0cf872`, shipped in v0.36.1) and is genuinely fixed — the failure has moved one stage later, into the delegate's `clear=true` respawn.

Reported against v0.36.1 with a daemon started from that build (binary mtime `19:07:48`, daemon start `19:08:27`), so this is not a stale-daemon artifact.

## Observed, with the reporter's own control

Both traces are from one `~/.local/state/dot-agent-deck/deck.log`, eleven minutes apart, both `ClaudeCode` workers, both `role=coder`.

**Dispatched orchestration — fails.** Pane `sched-dispatch-issue-755-18-r1`:

```
00:07:51.638  Received delegate signal  pane_id=sched-dispatch-issue-755-17-r0  targets=["coder"]
00:07:51.710  delegate: respawned worker agent for clear=true; waiting for SessionStart on hook
              broadcast  role=coder  pane_id=…-18-r1  new_agent_id=35  timeout_secs=30
00:07:51.716  Received event  event_type=SessionEnd  pane_id=…-18-r1
              …no SessionStart arrives…
00:08:21.712  delegate: SessionStart wait timed out; writing prompt via fallback path
00:08:21.713  delegate: readiness signal handled; holding the task prompt for the post-respawn
              readiness buffer  observed=false  buffer_ms=1000
00:08:22.713  WARN delegate: identity gate refused the task pointer; nothing written
              pane_id=…-18-r1  role=coder  expected_agent_id=Some("35")  outcome=NoLiveTarget
```

Reproduced on the next delegate at `00:14:14` with `new_agent_id=36`, identical shape.

**Interactive `Ctrl+N` orchestration — works.** Pane `19`, same daemon, same agent type:

```
00:19:01.915  Received delegate signal  pane_id=18  targets=["coder"]
00:19:01.970  delegate: respawned worker agent for clear=true; waiting for SessionStart
              role=coder  pane_id=19  new_agent_id=51  timeout_secs=30
00:19:06.868  Received event  event_type=SessionStart  pane_id=19     ← 4.9s
00:19:06.868  delegate: readiness signal handled  observed=true  buffer_ms=1000
00:19:08.033  delegate: worker emitted an event after delivery; no silence notice
```

## What this is not

It is **not** the defect #466 fixed. That one was `delegate from unknown pane` — dropped at the first routing check, because `pane_role_map` / `orchestrator_pane_ids` / `pane_orchestration_map` were populated only by the `AttachRequest::StartAgent` handler that the TUI path uses. In the traces above routing resolves end to end: the delegate signal is received, the sender pane is known, the role resolves to a target pane, and the respawn is issued. #466's fix is present and working.

The user-visible symptom is unchanged, which is why it reads as a regression: the orchestrator delegates, is told nothing is wrong, and waits for a `work-done` that never comes.

## Narrowing already done

- **Hooks are wired correctly in the dispatched worktree.** The same pane emitted a normal `SessionStart` on its *initial* spawn at `00:06:22.304`. Only the respawn fails to announce itself, which points at the relaunch parameters rather than at the worktree's hook installation.
- **Not agent-type-specific.** Both the failing and the working worker are `ClaudeCode`.
- **Not the pane-id string shape.** #466 established that `handle_delegate` is shape-agnostic (a `HashMap` lookup on a `String`), and routing demonstrably succeeds here.

## Not established

Why the respawned agent emits no `SessionStart` on the dispatch path. The obvious candidate — that the two spawn paths preserve different relaunch parameters (command, cwd, environment, hook wiring) for a pane, and the respawn replays the dispatch path's version — is a hypothesis from reading the traces, not a verified cause. It should be reproduced before it is believed.

## Why this shipped

#466's e2e coverage (`tests/e2e_dispatcher_mode.rs`) uses **`cat` as the worker stand-in** — see the module comment at `:234` ("`cat` echoes whatever is written to it"). That proves the delegate pointer is routed and written, which is exactly the half #466 fixed. A `cat` pane cannot exhibit a real agent being respawned with `clear=true` and failing to emit `SessionStart`, so the remaining half was invisible to it.

This is the stand-in trap CLAUDE.md rule 4 and `.claude/skills/reproduce-first` both warn about. Any fix for this issue needs a **real agent** worker on a cheap model, not `cat`.

## Related

- **#466** (merged) — fixed the routing half; its PR body is the best description of how the two spawn paths diverge.
- **#249** (closed, PRD) — "Delegate prompt silently dropped on `clear=true` respawn (Claude/OpenCode workers)". Same symptom family, previously fixed; worth reading for whether this is a regression of it on a path it did not cover.
- **#225** (closed) — the Codex variant of the same respawn race.
- **#243** (open) — retire the load-bearing 30s `SESSION_START_WAIT_TIMEOUT` and 1000ms `DELEGATE_READINESS_BUFFER`, deliberately scoped out of #225. The failing trace pays exactly that 30s timeout, so #243 is adjacent: it would change how long this takes to fail, not whether it fails.
- **#222** (open PRD) — dispatch reimplements orchestration spawning as a separate daemon-side primitive instead of reusing `TabManager::open_orchestration_tab`. #466 unified only the registration. **If the fix for this issue turns out to require genuine parity between the two paths, it belongs in #222 rather than in a competing PRD.**
- **#523** (open) — unify the "who is the orchestrator" rule across the three places that decide it.

## Acceptance

A dispatched orchestration's orchestrator can delegate to a worker role and the worker visibly receives and acts on the task, proven by a test that drives a **real agent** through the genuine spawn → delegate → respawn → work path, with the interactive `Ctrl+N` orchestration as a control in the same test.

### LINKED ISSUE #606: Delegating to a role just after closing its pane permanently breaks that role (respawn NotFound)

vfarcic:
Closes #584. Closes #606.

Both issues land in `dispatch_one_owned`'s `clear = true` arm, which is why they are fixed together: the delegate **destroys the current worker before it has anywhere to deliver the task pointer**, and then treats "no live worker at delivery time" as a terminal, silent drop. `delegate` has already exited 0, so the orchestrator is told nothing is wrong and waits for a `work-done` that can never arrive. They are two doors into that one seam.

### #606 — the close/respawn race (confirmed as reported)

`close_agent` removes the pane's registry entry **before** spending its termination grace, so for up to `AGENT_TERMINATE_GRACE` (3 s) a pane that is being closed has no record at all. A `clear = true` delegate landing in that window got `NotFound`, surfaced `⚠ respawn failed for role …` into the orchestrator pane, and left the role unreachable for the rest of the session. The issue's own diagnosis, verified.

### #584 — the reported hypothesis is **refuted**; the real defect is elsewhere

The issue's leading candidate was that the two spawn paths "preserve different relaunch parameters … and the respawn replays the dispatch path's version", and it explicitly asked for that to be reproduced before it was believed. It does not reproduce. Driving one `clear = true` respawn through the daemon's dispatch primitive (`crate::spawn::spawn`) and another through the TUI's `StartAgent` shape, with a recorder stand-in writing its argv, cwd, pane id, hook socket and `$SHELL` to disk on every launch, the two replacements are launched **identically** — and each respawn reproduces its own initial spawn byte for byte. That is now pinned as `orchestration/dispatch/003`. The only difference is PTY geometry, which is per-pane by design and is replayed faithfully from each pane's own last-known size.

What is genuinely broken is what the delegate does when the replacement is **not live**:

* the post-respawn readiness wait was a fixed 30 s deadline and nothing else, so a replacement that died during boot still cost the full window;
* the fallback then wrote into a pane with no agent on it, the identity gate refused with `NoLiveTarget`, and the pointer was discarded with **one `warn!` and nothing else** — no notice, no retry, nothing on the card.

That is exactly the reporter's symptom: `00:08:21` timeout → `00:08:22` `NoLiveTarget` → silence.

**What I could not reproduce, stated plainly:** *why* that particular replacement child died on the reporter's machine. The role command there is `devbox run agent` — a launcher the deck cannot see through — and the same `deck.log` shows that path's replacements taking **67 s and 87 s** to announce themselves on two other occasions, i.e. routinely past the 30 s wait. The proximate trigger looks environment-side (a devbox/nix launch that failed after its predecessor was killed mid-run) and I have no way to reproduce it here. The in-repo defect above is real, is what produces the reported user-visible outcome, and is what this PR fixes; the launcher's own flakiness is not something the deck can fix, only report.

## The fix

**`AgentPtyRegistry::respawn_or_recreate_agent_for_pane`** (new) wraps the existing respawn. On `NotFound` it (1) waits for an in-flight close to release the pane — `spawn_agent` refuses a pane held for cleanup, so creating one underneath a running `StopAgent` is not merely racy — (2) retries the respawn, since the pane may have acquired a record while we waited, and only then (3) creates a fresh agent from a caller-supplied `PaneRecreateIdentity`. Any error other than `NotFound` is returned untouched.

`close_agent`'s ordering is deliberately **not** changed, which is the issue's other suggested direction: `StopAgent` calls `worktree_still_in_use(registry.agent_records(), …)` *after* `close_agent` precisely because the entry is gone by then, so retaining it through the grace would silently stop dispatched worktrees being cleaned up.

**The delegate** then re-registers the orchestration role when a pane was re-created (a completed close takes it with `unregister_pane`), or the delivery lands but the *next* delegate is rejected with `reached no worker for role(s)` — the permanent breakage #606 reports. That needed a state handle inside the detached dispatch task; `handle_delegate_with_state` is the new entry point and `handle_delegate` delegates to it with `None`, so no existing caller changes shape.

**The readiness wait** now also ends the moment the replacement's PTY reaches EOF (`agent_exit_signal`, the agent-scoped sibling of the existing `pane_close_signal`; a `oneshot` rather than a poll because this path is exercised on a paused Tokio clock by `orchestration/delegate/011`). A replacement that is not the pane's live agent is reported into the orchestrator's own pane and releases its commission — the same treatment the respawn-error arm already gave a failure it *could* see. The new notice follows PRD #249 finding B3's precedent for this family: fixed daemon-authored text, single line, the worker's scrubbed `pane_id_env` as the only interpolation, role and error in the log.

## Rule 12 — the answer, explicitly

**No `PROTOCOL_VERSION` bump and no `.breaking.md`.** `daemon_protocol.rs` is untouched; no wire type, field or role-map value type changed, and no field's meaning shifted. The daemon's behaviour changes only in that a pane which used to error now comes back — and a pane's agent id rolling over under the same `pane_id_env` *is* the respawn contract an older TUI already handles. Bugfix → **patch** bump while `0.x`.

**Cross-version manual test, run both directions**, in a sandbox with `HOME`, both sockets, `DOT_AGENT_DECK_STATE_DIR` **and `DOT_AGENT_DECK_LOG`** isolated:

| | v0.36.2 daemon + branch client | branch daemon + v0.36.2 client |
|---|---|---|
| orchestration comes up | ✅ | ✅ |
| `delegate` exits 0 and routes | ✅ | ✅ |
| `clear = true` respawn replaces the worker | ✅ agent 2 → 3 | ✅ agent 2 → 3 |
| `worker-task-<role>.md` written | ✅ | ✅ |
| hooks arrive (`SessionStart`, status) | ✅ | ✅ |
| `work-done` feedback reaches the orchestrator pane | ✅ | ✅ |

Identical on both sides.

## Rule 4 — tier, and the real-agent proof

* **fast tier** — `orchestration/delegate/022` (#606's close race), `orchestration/delegate/023` (#584's dead replacement), `orchestration/dispatch/003` (#584's dispatch-vs-`StartAgent` control). Stand-ins are appropriate here and the catalog says why: what is under test is a daemon-side lifecycle race, a delivery decision, and a comparison of launch parameters — not an agent's behaviour. Both defect tests were **verified red before the fix**, and red again when each half of the fix is reverted one at a time.
* **L2, real agent** — `tests/fixtures/dispatch-orch-real`'s `coder` role flips `clear = false` → `clear = true`, so `orchestration/dispatch/002` (three real interactive Haiku panes, a real orchestrator *deciding* to shell `dot-agent-deck delegate`, a real worker creating a sentinel in the dispatched worktree) now drives the genuine spawn → delegate → **respawn** → work path. With `clear = false` it delegated to an already-running worker and never touched the respawn — which is how #584 shipped green. One extra Haiku cold boot; no new e2e test.

`orchestration/work-done/005` had to change how it forces a respawn failure: it evicted the worker's agent so the respawn would fail `NotFound`, and that is no longer a failure — it is the case #606 asked us to recover from. It now points the role at a nonexistent binary, which is the production hazard its own comment describes, stated literally. The release behaviour under audit is unchanged.

## Also worth knowing (not fixed here)

`delegate_prompt_injection.rs` sets `DOT_AGENT_DECK_SESSION_START_WAIT_MS` in several tests, but the **delegate** path reads a hard-coded `SESSION_START_WAIT_TIMEOUT` const — only the scheduler mirror honours that variable. The settings are inert. They are harmless today (those tests release the wait with a real `SessionStart` or a paused clock), so I left them alone rather than making ten tests suddenly time-sensitive inside this PR; flagging it because the const's own doc comment implies the knob works.

## Gates

`cargo fmt --check`, `cargo clippy --workspace --all-targets --features e2e -- -D warnings` before every commit; `cargo test-fast` green (3027/3027); `cargo test-e2e` run before opening.


prageethw (approved):
Careful fix for both issues at their shared seam. Appreciate that #584's leading hypothesis (divergent relaunch parameters) was actually tested and refuted rather than assumed, with the real defect (dead-replacement detection) identified instead. The recreate-pane recovery path's ordering (wait for in-flight close via pane_close_in_flight, retry respawn, only then recreate) and the oneshot-based agent_exit_signal (avoiding a poll that would race the paused-clock test) are both well-reasoned around real concurrency hazards. Good that orchestration/dispatch/002 was flipped to clear=true so the real-agent e2e path actually exercises the respawn this touches, rather than staying green by accident. CI green, cross-version test run both directions.

Issue #534, opened 2026-08-13
write_settings is a non-atomic, unlocked truncate-then-write — the defect that creates the malformed settings.json #516/#522 refuse


prageethw:
`write_settings` opens `~/.claude/settings.json` with truncate and writes in place — `std::fs::write` (`src/hooks_manage.rs:43-49`). There is no temp-file + rename, no advisory lock, and no in-process mutex over a file that **two deck processes and a human editor all write**.

## Why this matters more than it looks

This is the defect that **creates** the malformed file [#516](https://github.com/vfarcic/dot-agent-deck/issues/516) and [#522](https://github.com/vfarcic/dot-agent-deck/issues/522) defend against. Those two issues stop the deck destroying a file it cannot parse; this one is a way that file becomes unparseable in the first place. Fixing the refusal without fixing the write leaves the cause in place and only catches the consequence on the *next* run.

## Three exposures

- **Torn read.** Claude Code reading during the write window sees a truncated or empty file. The failure surfaces as Claude's, but the deck opens the window.
- **Torn write / lost update.** Read-modify-write with no lock. Two decks starting concurrently, or a deck racing a human's editor save, and the last writer wins outright. This is not hypothetical — running two orchestrations on one machine is a normal working mode.
- **A crash mid-write leaves a partial file**, which is exactly the input #516/#522 exist to refuse.

For two decks running the *same* binary path the merged result is identical, so the practical risk concentrates on the interrupted-write and human-editor cases. Note that concurrent decks are now frequently *different* paths, which makes genuine lost updates likelier than before, not less.

## The fix already exists in a sibling module

`codex_hooks_manage.rs` fixed precisely this and documents it as its own findings #1/M-2:

- `write_atomic` — temp file in the same directory + `rename(2)` (`src/codex_hooks_manage.rs:264-275`)
- guarded by `static INSTALL_LOCK: Mutex<()>` (`:87`)

Reusing that pattern is the whole fix. [#382](https://github.com/vfarcic/dot-agent-deck/issues/382) already proposes extracting `write_atomic` into a shared helper, which would make this a call-site change rather than a second implementation.

## Provenance

Found in a security audit of the hook-rule identification work now offered as [#506](https://github.com/vfarcic/dot-agent-deck/pull/506), classified **MEDIUM / pre-existing** and deliberately deferred so that PR could stay scoped. Filing it so the finding does not evaporate with the scratch file it was written in.

PR #651, opened 2026-08-22
fix(codex): preserve destination file mode when publishing hook config

### LINKED ISSUE #382: Preserve destination file mode in codex_hooks_manage::write_atomic (and extract the shared helper)

vfarcic:
Closes #382.

## The defect

`codex_hooks_manage::write_atomic` published through `File::create`, which applies `0666 & !umask`, and then renamed that file over the destination. Three call sites went through it: two writing `hooks.json`, and one writing the user's `~/.codex/config.toml` — model choice, auth references, hook-trust records, and anything hand-written there. A config the user had deliberately kept at 0600 came back wider, silently, the first time the deck installed its hooks or recorded trust.

Measured on this box (umask 002): a 0600 destination came back **0664 — group-writable**. With the fix reverted, three tests fail with `left: 436, right: 384`; with it in place all three pass.

Codex credentials are untouched: `auth.json` goes through a separate path that already forces 0600, and this PR does not go near it.

## The fix

The publish reads the destination's existing mode and reapplies it, falling back to `0o600` when the destination does not yet exist — so a file the user restricted keeps exactly the mode they chose, and a file the deck creates is owner-only regardless of umask. `#[cfg(unix)]`, exactly the shape #360 applied to the Devin adapter.

## The extraction

#360 fixed the identical defect in `devin_hooks_manage`, which had been copied verbatim from this module — the copy is why the fix never reached here. Rather than carry a third copy, `write_atomic` and `build_command` now live once in a new `src/agent_hook_config.rs` and both adapters call it. (`shell_quote_if_needed`, the third name the issue mentions, was already shared via `platform::paths` — nothing to do there.)

**`hooks_manage` deliberately keeps its own `write_atomic`.** The issue warned that the Claude adapter defines helpers by the same names; it is not a third copy of this one. It takes `dest` alone and publishes through `create_new`, so a leftover temp file cannot be a symlink the write follows out of the directory (#534) — hardening the other two adapters have not been reviewed for. Folding it in would either drop that or push it onto them unreviewed, so it stays where it is, and both module headers say why.

The extraction is behaviour-preserving. The one difference worth naming: the temp-file *name* fallback for a `dest` with no UTF-8 file name was spelled `"hooks.json"` in one adapter and `"config.json"` in the other, and is now `"config"`. That branch is unreachable in any case where the subsequent `rename` could succeed, and it is documented at the helper.

## Tests

`src/codex_hooks_manage.rs::tests::install_and_trust_never_widen_config_permissions` mirrors `devin_hooks_manage::tests::install_never_widens_config_permissions`, and covers all three call sites: `hooks.json` across install and uninstall, and `config.toml` across the trust edit — including that the trust edit still preserves the user's unrelated bytes, so the mode assertion is not passing over a clobbered file. Three more unit tests sit at the shared helper itself (mode preservation for a new file, a 0600 file and a 0644 file; the rename leaving no stray temp; the command quoting).

- `cargo fmt --check`, `cargo clippy --workspace --all-targets --features e2e -- -D warnings`: clean.
- `cargo xtask linkage-check`: ok.
- `cargo test-fast`: 3053 passed.
- `cargo test-e2e`: the Codex trust/hooks tier the issue asks to re-verify passes in full (127 tests across `codex_hooks_safety`, `e2e_codex_hooks`, `codex_adapter`, `codex_hook_ingestion`, `hook_rule_identification`). The full tier reported 4 failures + 1 timeout out of 8778, all in the real-agent LLM suite; four pass on rerun in isolation, and the fifth (`orchestration_dispatch_002_every_real_agent_role_comes_alive`) **fails on the branch point `caef5b6` too**, with a different symptom each run — pre-existing PTY flake, not this change.

**Did this change the TUI↔daemon protocol/handler contract? No.** This is a local filesystem publish inside the hook *installer*: no wire shape moved, no field changed meaning, and hook delivery, the daemon handlers and the orchestration paths are untouched. **No `PROTOCOL_VERSION` bump and no `.breaking.md` fragment** — `changelog.d/382.bugfix.md` (patch bump) is the right classification.

https://claude.ai/code/session_01HuBg86NwNX5QoJmdVBHGSz


prageethw (approved):
Reviewed the diff: extracts `write_atomic`/`build_command` into a new shared `src/agent_hook_config.rs` used by both `codex_hooks_manage.rs` and `devin_hooks_manage.rs`, correctly reapplying the destination file's existing mode (or `0600` for new files) before the atomic rename, matching the #360 fix already applied to the Devin adapter. `hooks_manage.rs` (Claude) deliberately keeps its own `write_atomic` using `create_new` for the #534 symlink hardening, documented in both module headers. New unit + integration tests cover mode preservation across install/trust/uninstall and confirm the trust edit doesn't clobber unrelated `config.toml` content. All CI checks green.

PR #652, opened 2026-08-22
fix(worktree): render reclaim paths injectively so two directories never alias

### LINKED ISSUE #578: `to_string_lossy` aliases distinct byte-exact worktree paths at the delete-decision surface
## Two different directories, one identical line, at the surface that decides deletion

On Unix, `path_from_bytes` (`src/worktree_reclaim.rs:197`) deliberately preserves arbitrary bytes via `OsStr::from_bytes`. But every human render path converts through `Path::to_string_lossy()`, which collapses **every invalid UTF-8 sequence to `U+FFFD`**. That conversion is not reversible, so distinct byte-exact paths can produce byte-identical display strings.

Meanwhile the removal itself acts on the distinct, byte-exact path values.

**So the operator reads one line and the command acts on another directory — and cannot tell from the output which.**

## Concrete case

Create two clean, merged worktrees whose final path components are the raw Unix bytes `candidate-\xff` and `candidate-\xfe`.

- Both `to_string_lossy()` calls yield `candidate-\u{fffd}`.
- `format_reclaim_human` (`src/worktree_reclaim.rs:620`) therefore prints **two visually identical bullets** in its pending section (`:629`, `:637`, `:648`): `  - /repo/candidate-\u{fffd}`.
- A following `worktree reclaim --yes` removes `/repo/candidate-\xff` and `/repo/candidate-\xfe` — two different directories.

The pending list exists precisely so the operator can decide what the `--yes` will remove. Here it cannot answer that question.

## Expected

At a surface whose entire purpose is a delete decision, two distinct directories must not render identically. Options: escape non-UTF-8 bytes reversibly (e.g. `\xNN`) rather than lossily, or disambiguate colliding display strings, or refuse to offer a reclaim for a path that cannot be rendered unambiguously.

## Note on scope

This is distinct from control-character sanitisation of the same output. That concerns characters that pass through and *rewrite* the display (ESC, CR, bidi); this concerns a conversion that *collapses* distinct inputs to the same output — a different mechanism needing a different fix. Sanitising after `to_string_lossy()` neither causes nor fixes this, because the information is already gone by then.

## Why this is filed here

Verified present on `main` at `e6e6dfea`: `path_from_bytes` at `src/worktree_reclaim.rs:197`/`:213`, and the lossy render at `:629`/`:637`/`:648`, are upstream code. It was found in a fork (`prageethw/dot-agent-deck`, issue #276) while reviewing an adjacent change, and was deliberately kept out of that PR's scope. Nothing about it is fork-specific. Reporting it here rather than only fixing it downstream, so the fix lands where every user of this project benefits.

vfarcic:
Fixes #578.

## The bug

`worktree reclaim`'s pending list rendered every path through `Path::to_string_lossy()`, which maps **every** invalid UTF-8 sequence to `U+FFFD`. The removal, meanwhile, acts on the distinct byte-exact `PathBuf` that `path_from_bytes` deliberately preserved. So two different directories printed as one identical line, and the operator reading the list could not tell which one a following `--yes` would delete — at the one surface that exists to answer exactly that question.

Reproduced before touching the fix, exactly as the issue describes. Two clean, merged, foreign worktrees named `candidate-\xff` and `candidate-\xfe`:

```
2 worktree(s) reclaimable pending confirmation (kept for now):
  - /var/tmp/…/candidate-<U+FFFD>
  - /var/tmp/…/candidate-<U+FFFD>
Run `dot-agent-deck worktree reclaim --yes` to remove them.
```

Both bullets byte-identical, ending `…, 99, 97, 110, 100, 105, 100, 97, 116, 101, 45, 239, 191, 189` — `239 191 189` being U+FFFD.

## Which shape, and why

The issue offered three: escape reversibly, disambiguate colliding strings, or refuse to offer the reclaim. **This takes the first — a reversible escape.**

- **Refusing** was ruled out first: it would regress `worktree/reclaim/008`, which exists because a non-UTF-8-named worktree *is* reclaimable (Greptile P1 on #427). Trading a rendering defect for a functional one is the wrong direction, and it would leave such a worktree permanently unreclaimable rather than merely hard to read.
- **Disambiguating** colliding display strings tells the operator that two lines differ but not *which is which*, so it does not restore the decision — and it is non-local: whether a line gets a suffix depends on the rest of the batch, so the same directory renders differently from one run to the next.
- **Escaping** is total, local, and stable: every path renders to exactly one string, that string identifies it, and it does not depend on what else is in the list.

The escape is `Path`'s own `Debug`, not a hand-rolled one. std's `escape_debug` gives `\xNN` for an invalid byte **and** `\\` for a literal backslash — and that second half is what makes it injective rather than merely different. A `\xNN`-only escape leaves a directory genuinely named `candidate-\xFF` colliding with the one holding raw byte `0xFF`: the collision is relocated, not fixed, and the output looks exactly as correct. That variant is covered by its own test (below), which is the only test that catches it.

Using std also means one code path on every platform — no `cfg` split here that could drift out of step with `path_from_bytes`'s.

Real output now, on a fixture holding all four cases side by side:

```
Kept:
  - "/var/tmp/…/candidate-\\xFF"     <- directory literally named  candidate-\xFF
  - "/var/tmp/…/candidate-plain"
  - "/var/tmp/…/candidate-\xFE"      <- raw byte 0xFE
  - "/var/tmp/…/candidate-\xFF"      <- raw byte 0xFF
```

All four distinct. Ordinary paths are unchanged apart from the surrounding quotes, which are themselves useful here: they delimit the path, so leading/trailing whitespace in a name is visible at the point of deciding to delete it.

## What changed

One helper, `display_path`, at the four human render sites: `format_reclaim_human`'s pending/removed/kept bullets (`:629`/`:637`/`:648` in the issue) and `format_list_human`'s PATH column. `worktree list` is where the verdicts are read before `reclaim` runs, so leaving its column aliased would mean the two halves of one decision could not be matched up.

## Scope

**Control-character sanitisation is not addressed here, and this does not close it.** The issue is explicit that it is a different mechanism — characters that pass through and *rewrite* the display, versus a conversion that *collapses* distinct inputs — and nothing was added for it. `escape_debug` does incidentally escape control characters as a side effect of the escape chosen for injectivity; that is noted in the code as incidental precisely so nobody later reads this as having answered that question.

**`worktree list --json`'s `path` field is deliberately left lossy**, with the reasoning recorded on `serialize_path_lossy`. It has the same aliasing, but it is a versioned machine surface, `reclaim` (the delete decision) has no `--json` at all, and a consumer needing byte-exactness wants the bytes rather than a human escape — so the right answer there is an additive field plus a `SCHEMA_VERSION` decision, not a silent change to what the existing field means. Worth a follow-up issue rather than a quiet ride-along here.

## Tests

Reproduce-first: the test was written and observed failing for the issue's reason before any fix existed.

- **`worktree/reclaim/009`** (new, `tests/worktree_reclaim.rs`) — the issue's reproduction at the user surface: the real binary, two real worktrees named `candidate-\xff`/`candidate-\xfe`, a bare `reclaim`, asserting the two pending bullets differ. It compares the subprocess's **raw stdout bytes**, never `String::from_utf8_lossy` — comparing lossy strings would apply the very conversion under test, so a correct rendering could be reported as colliding purely because the harness collapsed it. Carries a fixture precondition that both names exist on disk with exactly the intended bytes, and a control that a bare `reclaim` leaves both directories in place, so the report is a decision pending rather than a post-mortem. Asserts only that the lines *differ*, not the escape's syntax — so it accepts all three shapes the issue allows rather than pinning this implementation's choice. Catalogued.
- **Five unit tests** in `src/worktree_reclaim.rs` for the parts a worktree fixture cannot cheaply reach: the literal-backslash injectivity edge; the `Removed:`/`Kept:` sections; the `worktree list` PATH column (including that escaping adds no tab, so the row stays seven fields); and that an ordinary ASCII path still reads as itself.

Each assertion was shown to be load-bearing by reverting the fix one variant at a time:

| `display_path` body | result |
|---|---|
| `to_string_lossy` (the bug) | `009` red; 3 of 5 unit tests red |
| `\xNN`-only escape, backslash untouched | everything green **except** the literal-backslash test |
| `format!("{path:?}")` | all green |

The middle row is the point: the half-fix looks correct and passes every test but one.

## Gates

- `cargo fmt --check` — clean
- `cargo test-fast` — 3055/3055
- `cargo test-e2e` — 8778/8780. The two failures are `shell_activity_005`/`006` in `e2e_shell_activity_real_agent`, real-agent prompt-adherence flakes that the test's own panic message identifies as such ("the model never issued the Bash tool call at all … This is a harness flake … rerun rather than reading this as BADGE REGRESSION"). Both pass on an isolated rerun. Neither touches this code.
- `cargo xtask linkage-check` — ok (558 catalog ids, 9 rules)
- `scripts/windows-cross-check.sh` — no errors, and no new warning from `src/`; the two new unit-test helpers are `#[cfg(unix)]` so they are not dead code on Windows

Rule 12 does not apply: no daemon, protocol, orchestration or hook code is touched. macOS could not be cross-checked locally (no `rust-std` for the apple targets on this box) — but macOS is Unix, takes the same branches as Linux, and `display_path` has no `cfg` at all; `build-macos` covers it.

https://claude.ai/code/session_01FRTcKzB7pQ78wpsUaCv48p


prageethw (approved):
Verified the fix: `display_path` (src/worktree_reclaim.rs) renders paths via `Path`'s own `Debug` (`format!("{path:?}")`), which is injective — it escapes both invalid bytes (`\xNN`) and literal backslashes (`\\`), avoiding the relocated-collision half-fix the PR itself calls out and tests for — applied consistently at all 4 human-render sites. The lossy `worktree list --json` `path` field is deliberately left unchanged with a documented rationale. New e2e test `worktree/reclaim/009` (comparing raw stdout bytes, not lossy-converted strings) and 5 unit tests cover the injectivity edge case, TSV-column safety, and ordinary-path readability. All CI checks green.

PR #653, opened 2026-08-22
fix(palette): move the waiting status off yellow and guard its contrast (#579)

### LINKED ISSUE #579: `STATUS_WAITING` yellow fails WCAG AA on light terminals everywhere it is used
## Summary

`palette::STATUS_WAITING` (yellow) fails WCAG AA badly on a light terminal background, everywhere it is used. Measured on a real deck: **~1.70:1** plain, and **~1.07:1** on terminals that render bold ANSI colours as their bright variant (`#FFFF00` on white) — against a 4.5:1 AA minimum. 1.07:1 is close to the worst pairing possible.

```rust
// src/palette.rs:96
pub const STATUS_WAITING: Color = Color::Yellow;
// src/palette.rs:151
SessionStatus::WaitingForInput => STATUS_WAITING,
```

## Where it bites

Because it is reached through `status_color(SessionStatus::WaitingForInput)`, it feeds every status surface at once — the dashboard card's status badge and border, the sidebar role list's `Needs Input` label, and orchestration tab labels. Anything that maps a status to a colour inherits it.

## Evidence

Captured from a real deck driven by genuine hook events, with `tmux capture-pane -e -p` preserving the SGR codes, then rasterized against a white background. The frame shows the sidebar `Needs Input` label and a yellow tab label both barely legible on white, while the `Error` red beside them reads cleanly.

## Expected

A `WaitingForInput` colour that meets WCAG AA against both light and dark terminal backgrounds. This is the exact class of defect the light-terminal-compatibility work exists to eliminate.

## Why no gate catches it

L1 snapshot tests assert the *colour constant*, not its rendered contrast against a background, so the palette can regress to an unreadable pairing with the suite fully green. Any fix is worth pairing with a contrast assertion so this cannot silently return.

## Why this is filed here

Verified present on `main` at `e6e6dfea` — the constant and its `WaitingForInput` mapping are quoted above from upstream. It is pre-existing and app-wide, not caused by any one feature. It was measured in a fork (`prageethw/dot-agent-deck`, issue #312) during visual validation of an unrelated tab-strip change; fixing it there would have meant changing a global palette constant from inside a narrow styling PR, which was rejected deliberately. Nothing about it is fork-specific. Reporting it here rather than only fixing it downstream, so the fix lands where every user of this project benefits.

vfarcic:
Fixes #579.

## What was wrong

`palette::STATUS_WAITING` was `Color::Yellow`, the one named ANSI slot that all but disappears on a light terminal. Measured on a real deck rasterised against white: **1.70:1** for the plain slot, **1.07:1** once a terminal renders the badge's `Modifier::BOLD` as the bright variant — against WCAG AA's 4.5:1.

Every status surface resolves through `status_color`, so that single constant carried the deck card's `Needs Input` badge and border, the embedded pane's border, the stats bar's waiting segment and the orchestration tab labels. The fix is at the constant; no call site is patched.

## The colour, and why it is closer to forced than to taste

Contrast against white and against black pull in opposite directions. Since WCAG contrast is `(L_hi + 0.05) / (L_lo + 0.05)`, a foreground that must clear AA on **both** has a hard ceiling at the point where the two are equal — `L = sqrt(0.0525) - 0.05 ≈ 0.1791`, giving **4.58:1**. That bounds *every* colour, not just the ANSI sixteen, and it sits barely above the bar. Resolved through the xterm palette:

| slot    | L      | plain·white | plain·black | bright·white | bright·black |
|---------|--------|-------------|-------------|--------------|--------------|
| Red     | 0.1298 |     5.84    |     3.60    |     4.00     |     5.25     |
| Green   | 0.4366 |     2.16    |     9.73    |     1.37     |    15.30     |
| Yellow  | 0.5664 |   **1.70**  |    12.33    |   **1.07**   |    19.56     |
| Blue    | 0.0617 |     9.40    |     2.23    |     4.74     |     4.43     |
| Magenta | 0.1739 |     4.69    |     4.48    |     3.14     |     6.70     |
| Cyan    | 0.4807 |     1.98    |    10.61    |     1.25     |    16.75     |

**Magenta.** Its luminance is within 3% of the theoretical optimum and no other free slot is close. A terminal ships a background and a palette together and tunes one to the other, so on a theme-matched terminal magenta clears AA both ways (4.69:1 base-on-white, 6.70:1 bright-on-black). In the mismatched configurations — including the bold-as-bright case that produced the 1.07:1 reading — it still holds 3.14:1 and 4.48:1, above WCAG SC 1.4.11's 3:1 floor for non-text UI components, which is the criterion that actually applies to a border glyph and a status chip.

Red measures as well and is `STATUS_ERROR`. The magenta slot was free because issue #442 retired it as the selection accent (selection is `Color::Reset` now), so nothing else claims it: green/blue/red/dark-gray statuses, cyan focus, `Reset` selection all stay distinct.

## The guard (the part that keeps it from coming back)

New `theme/contrast/002` in `tests/palette_contrast.rs` computes WCAG contrast from whatever the palette currently holds, across two backgrounds × two slot renderings. The two theme-matched pairings must clear 4.5:1; the two mismatched ones must clear 3:1.

This is the gap the issue names. Every existing theme test asserts colour *identity* — `theme/palette/001` reads a rendered border cell and compares it to a constant — and a constant compares equal to itself however unreadable it is. No test ever put a foreground and a background in the same equation, which is why the suite was green throughout.

A paired non-spec unit-guard pins the arithmetic: the 21:1 and 1:1 endpoints, then reproduces the issue's own 1.70:1 and 1.07:1 yellow measurements from the reference xterm palette (which is what ties the model to the reported defect rather than to a guess), then asserts the floors still reject `Color::Yellow` — so a floor that had drifted below the defect cannot look identical to one that holds.

**Verified by reverting the constant**: with `STATUS_WAITING` back on Yellow, `contrast_002` fails with `renders at 1.70:1 on a light terminal (base slot on white), below the 4.5:1 floor`.

## Scope note

Green measures 2.16:1 and cyan 1.98:1 against white — the same class of weakness, deliberately not asserted here. Fixing those means re-picking colours that are already spoken for, which is a separate decision from unbreaking the one role that was measured as unreadable.

## Also updated

Identity assertions moved in step (`theme/palette/001`–`004`, `tabs/orchestration/009`, `ui::tests::test_status_style`), plus `docs/orchestration.md` and `docs/workspace-modes.md`, which named the colour, and several comments still describing Magenta as the retired selection accent.

Presentation only — no daemon, protocol, hook or orchestration change, so no `PROTOCOL_VERSION` bump and no `.breaking.md` (CLAUDE.md rule 12). L1 only per rule 4; a palette constant needs no PTY-attached L2.

## Gates

- `cargo fmt --check` — clean
- `cargo test-fast` — **3051/3051 pass**
- `cargo xtask linkage-check` — ok (559 catalog ids)
- `cargo test-e2e` — 8772/8776. Four failures, none related:
  - `codex_wrap_001`, `form_006`, `shell_activity_006` each **pass on isolated rerun** (flaky under full-tier load).
  - `orchestration_dispatch_002` fails **identically on `origin/main`** (`caef5b6`, checked out and rerun to confirm) with the same `NO PANE — never spawned at all` symptom. Pre-existing, and a real-agent dispatch test cannot be reached by a palette constant.

https://claude.ai/code/session_01Ta9p75mAQsWQVzfUifB4C4


prageethw (approved):
Verified the math and the diff: the WCAG contrast ceiling for a single color legible on both a white and black terminal background is 4.58:1, and Magenta is the closest free ANSI slot to that ceiling (4.69:1/3.14:1 on white, 4.48:1/6.70:1 on black vs. Yellow's reported 1.70:1/1.07:1). The change is confined to the `STATUS_WAITING` constant (`src/palette.rs`) plus updated identity assertions and docs. New `tests/palette_contrast.rs` computes actual WCAG contrast from the live constant (rather than asserting color identity, which is why the original defect shipped unnoticed) and is proven to reject `Color::Yellow`. All CI checks green.

PR #654, opened 2026-08-23
fix(e2e): run OpenCode coverage on the host's own agent auth, not metered OpenRouter

vfarcic:
## Problem

The real-agent e2e tier baked **host-specific auth assumptions into test constants**, so it only ran properly on a machine configured like the one that wrote it.

Two OpenCode tests hardcoded `openrouter/openai/gpt-4o-mini` — in one case not even via the shared constant, just an inline string in a second file. That billed metered OpenRouter credit for coverage a maintainer's existing subscription already pays for, and skipped both tests whenever that balance ran dry. The skip reason named missing *credentials*, which were present the whole time.

Separately, `codex_test_model`'s documentation asserted the **exact reverse of reality**, which cost a long debugging session.

## Changes

**1. OpenCode coverage runs on the host's own auth.** `openai/gpt-5.4-mini` names a *provider*, not an auth mode, so it resolves against whatever `opencode auth` holds — a ChatGPT-subscription (oauth) credential or an API key. Verified end to end on a subscription host: both tests run and pass, no OpenRouter involved.

**2. `opencode_test_model()` + `DOT_AGENT_DECK_OPENCODE_TEST_MODEL`**, mirroring the existing Codex accessor. A host that can only reach OpenRouter overrides it instead of editing two files. Also collapses the duplicate — one call site previously ignored the shared constant entirely.

**3. Corrected the Codex model documentation.** It claimed the `codex-*` family was served *only* to ChatGPT-subscription credentials, and that an **API key** needed the override. Measured 2026-08-23, it is the other way round:

```
$ codex exec --model gpt-5.1-codex-mini        # ChatGPT-subscription auth
400 invalid_request_error: The 'gpt-5.1-codex-mini' model is not supported
when using Codex with a ChatGPT account
```

So a subscription host fails the availability probe and **all four real-agent Codex tests SKIP** — silently, reading as a pass. The failure message sent readers hunting for an API key that was not the problem; it now names both auth modes and points at `codex login status`.

## What is deliberately *not* changed

`CODEX_TEST_MODEL_DEFAULT` stays `gpt-5.1-codex-mini`.

Overriding it (`DOT_AGENT_DECK_CODEX_TEST_MODEL=gpt-5.4-mini`, verified reachable) makes those four tests **run but not pass** — they then fail on a separate defect unrelated to auth or model choice. Converting four silent skips into four red tests is not an improvement while that is outstanding, and `cargo test-e2e` is the pre-PR gate.

**Flipping that default is a one-line follow-up** once the separate defect lands. Everything else here stands on its own.

## Verification

- `cargo fmt --check` — clean
- `cargo test-fast` — 3049/3049
- `cargo test-e2e` — the two OpenCode tests pass on subscription auth with no OpenRouter credit

One unrelated e2e failure exists on this branch, `e2e_dispatcher_mode::orchestration_dispatch_002_every_real_agent_role_comes_alive`. **It fails identically on clean `main`** (verified by stashing this branch's changes and rerunning: same 313s, same assertion), so it is pre-existing and not introduced here.

## Rule 12

Touches no daemon, protocol, orchestration or hook code — `tests/` only. **No `PROTOCOL_VERSION` bump, no `.breaking.md` fragment.**

## Notes for the reviewer

- **Verification hazard worth knowing about.** Running the e2e tier leaks orphaned `dot-agent-deck wrap --agent codex` processes; 35 had accumulated on the dev box, the oldest 8½ days, pinning 385 temp roots / 14.2 GB that `cargo xtask clean-e2e-tmp` will never reap while their PIDs live. The stale state they leave behind produced several wrong conclusions while investigating this. Tracked separately.
- **`DOT_AGENT_DECK_REQUIRE_REAL_E2E=1`** turns every runtime skip into a loud failure with its reason. Worth using to confirm a run genuinely exercised your auth rather than skipping past it — without it, `SKIP:` goes to stderr and is invisible in a green run.

### DIFF
--- a/changelog.d/654.misc.md
+++ b/changelog.d/654.misc.md
@@ -1,0 +1,5 @@
+## Real-agent OpenCode tests run on whatever agent auth the machine already has
+
+The end-to-end suite used to pin its OpenCode coverage to a specific OpenRouter model, so those tests only ran on a machine holding OpenRouter credit — and quietly skipped everywhere else, reporting missing *credentials* even when credentials were present. They now name a provider rather than a billing route, so they resolve against whatever `opencode auth` holds, including a ChatGPT subscription, and a contributor no longer needs a second paid account to exercise them. `DOT_AGENT_DECK_OPENCODE_TEST_MODEL` overrides the choice where a different provider is wanted.
+
+The Codex equivalent is documented correctly for the first time. The guidance had it backwards — it said the `codex-*` models were reachable only with a ChatGPT subscription and that an API key needed the override, when a subscription is in fact what cannot reach them. A subscription host therefore failed the availability probe and skipped every real-agent Codex test silently, which reads as a pass. The skip now explains which authentication is in use and what to set instead.
--- a/tests/common/mod.rs
+++ b/tests/common/mod.rs
@@ -2722,8 +2722,8 @@ pub fn check_opencode_available() -> Result<(), String> {
 }
 
 /// Compiled-in default cheap model for Codex availability probes and real-agent
-/// e2e coverage. Correct for a ChatGPT-subscription (oauth) `~/.codex/auth.json`,
-/// which is what most dev boxes here log in with.
+/// e2e coverage. Reachable by an **API-key** `~/.codex/auth.json`; a
+/// ChatGPT-subscription (oauth) host must override it — see [`codex_test_model`].
 const CODEX_TEST_MODEL_DEFAULT: &str = "gpt-5.1-codex-mini";
 
 /// Env var that overrides [`codex_test_model`] on a host whose Codex credentials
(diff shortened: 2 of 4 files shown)


prageethw (approved):
Verified the diff against the PR-head branch: `tests/common/mod.rs:2798-2821` adds `opencode_test_model()` mirroring the existing `codex_test_model()` accessor pattern, and both `tests/e2e_delegate_respawn_readiness.rs:260` and `tests/e2e_delegate_work_done_chain.rs:314` now route through it (the latter previously inlined the OpenRouter model string directly, bypassing the shared constant). `CODEX_TEST_MODEL_DEFAULT` is confirmed unchanged at `tests/common/mod.rs:2727` — only its doc comment and the `check_codex_available` error message were corrected. Scope is `tests/` only, no protocol/daemon code touched, CI is fully green (8/8 checks).

Issue #514, opened 2026-08-11
ci: narrow release.yml's docs job from `secrets: inherit` to an explicit RELEASE_TOKEN map


vfarcic:
Raised as F6 in [@prageethw](https://github.com/prageethw)'s review of #504. Pre-existing, non-blocking, and deliberately **not** fixed in that PR because it sits on the release path and the last step cannot be verified without triggering a real publish.

## The problem

`release.yml`'s `docs` job calls `docs-publish.yml` via `uses:` and forwards credentials with `secrets: inherit`:

```yaml
  docs:
    needs: [prepare, finalize]
    uses: ./.github/workflows/docs-publish.yml
    secrets: inherit
```

`inherit` forwards **every** repository secret, so the docs-publishing job runs holding `HOMEBREW_TAP_TOKEN` and `SCOOP_BUCKET_TOKEN` — cross-repo write credentials for the Homebrew tap and the Scoop bucket — neither of which it needs. `gh secret list` shows exactly three secrets on this repository, so those two are the entire excess.

Verified by reading `docs-publish.yml`: it references only `secrets.RELEASE_TOKEN` (the guard step and the checkout `token:`) and `secrets.GITHUB_TOKEN` (GHCR login). The latter is provided automatically to a called workflow and needs no mapping.

## The fix

Two coupled edits — an explicitly-passed secret must be declared by the callee:

```yaml
# .github/workflows/docs-publish.yml
on:
  workflow_call:
    secrets:
      RELEASE_TOKEN:
        required: true
```

```yaml
# .github/workflows/release.yml
    secrets:
      RELEASE_TOKEN: ${{ secrets.RELEASE_TOKEN }}
```

A `required: true` declaration is a strict improvement on the runtime guard for the caller path: forgetting to forward the secret then fails at workflow parse time rather than at step 1 of the callee.

## Why it was deferred rather than done in #504

`docs-publish.yml` is reached two ways — `workflow_call` from `release.yml`, and `workflow_dispatch` for a standalone `/publish-docs`. Adding an `on.workflow_call.secrets` declaration governs the *call* path; the dispatch path is expected to keep reading the repository secret exactly as it does today. That expectation is almost certainly right and is the standard dual-trigger pattern, but GitHub's documentation does not state it, and it cannot be confirmed without an actual run — and the only run available pushes a chart-bump commit to `main` and publishes an image to GHCR.

[...]

Issue #324, opened 2026-08-01
Harden latent shell templating in docs-publish.yml and Taskfile.yml


vfarcic:
## Problem

A security audit during #310 found direct interpolation of GitHub/Task contexts into shell bodies in two files. **Neither is exploitable today** — this is latent-hazard hardening, not an open hole.

### `.github/workflows/docs-publish.yml`

Interpolates `inputs.image_tag`, `inputs.push_latest`, `steps.tag.outputs.*` and `github.actor` directly into `run:` scripts at roughly lines 43, 60-61, 92, 97-98, 118-121, 131-135.

It is invoked by `release.yml` with `secrets: inherit`, so it holds the publishing tokens.

**Why it is safe today:** its only in-repo `workflow_call` is from `release.yml`, whose version output is now SemVer-validated at the `prepare` job before it can reach here; and its own `workflow_dispatch` declares no inputs, deriving a hex SHA plus a constant origin/boolean.

### `Taskfile.yml`

The `homebrew-formula` / `scoop-manifest` vars interpolate `{{.VERSION}}` / `{{.NAME}}` into `sh:` subshells — e.g. line 58, `echo "{{.VERSION}}" | sed 's/^v//'`.

**Why it is safe today:** the only automation callers pass the now-constrained version and constant channel names, and direct local `task` invocation already grants shell execution to whoever runs it.

## Why fix it anyway

Both are safe **by data constraint, not by construction**. The guarantee is "no untrusted value currently reaches this sink" — which holds only until someone adds a `workflow_dispatch` input, a new caller, or a new variable. That is a property nobody will re-derive when making an unrelated change, and one of the two sinks holds `HOMEBREW_TAP_TOKEN` / `SCOOP_BUCKET_TOKEN`.

`release.yml` had the same shape and *was* reachable — a crafted tag name or dispatch input executed commands in a privileged job. That was fixed in #310 by rebinding all 13 interpolations across 8 steps through `env:` blocks and referencing them as quoted shell variables. These two files should get the same treatment.

## Proposed fix

Bind every GitHub/Task context through an `env:` block (or an explicit quoted argv element for go-task) and reference it as a quoted shell variable, rather than interpolating it into the script text. Sweep both files rather than patching only the lines listed above.

## Discovered via

Security audit of [#310](https://github.com/vfarcic/dot-agent-deck/pull/310) (issue #250).

PR #658, opened 2026-08-23
ci(docs-publish): verify the yq download instead of trusting sudo wget (#515)

### LINKED ISSUE #515: ci: verify the yq download in docs-publish.yml instead of trusting `sudo wget`
Raised as a minor pre-existing finding in [@prageethw](https://github.com/prageethw)'s review of #504, and filed rather than fixed there to keep that PR's diff to the change under review.

## The problem

`.github/workflows/docs-publish.yml` installs `yq` by downloading a release binary straight into `/usr/local/bin` and marking it executable, with no integrity check:

```yaml
      - name: Install yq
        run: |
          sudo wget -qO /usr/local/bin/yq https://github.com/mikefarah/yq/releases/download/v4.52.5/yq_linux_amd64
          sudo chmod +x /usr/local/bin/yq
```

Two properties make this worth closing rather than tolerating:

- **Nothing pins the artifact's content.** The version tag is pinned, but a release asset is mutable — a tag can be re-pointed and an asset re-uploaded. `-q` also suppresses errors, so a truncated or HTML error-page download produces a file that fails later at `yq -i` rather than at the download.
- **The job holds the admin PAT.** `RELEASE_TOKEN` is in scope for the whole job (the checkout uses it, and the remote it pushes to is authenticated with it), and `yq` runs *after* that checkout, two steps before `git push origin main`. So the blast radius of a substituted binary is a commit of the attacker's choosing pushed to a protected `main` under an admin identity — plus whatever `secrets: inherit` forwards (#514).

Not a live exposure: this is a first-party workflow pulling from the upstream project's own release URL. It is the kind of supply-chain shortcut that is cheap to remove and awkward to reason about later.

## Options, roughly in order of preference

1. **Verify a checksum.** yq publishes `checksums` and `checksums_hashes_order` with each release; download the binary and the checksum file and confirm with `sha256sum -c` before `chmod +x`. Pins content, not just a tag name. Renovate will not bump the digest automatically, so the version and the checksum have to move together — a comment saying so belongs next to it.
2. **Drop `yq` entirely.** The workflow makes three trivial edits — `.image.tag`, `.version`, `.appVersion` in two small YAML files. Since a Rust toolchain is not present in this job, the natural alternatives are `python3 -c` with `pyyaml` (present on `ubuntu-latest`) or a targeted `sed`. This removes the download, the `sudo`, and the version-bump maintenance at once, at the cost of hand-rolled YAML editing.
3. **Use an action that pins by digest**, e.g. a `yq` setup action referenced by commit SHA. Trades one supply chain for another, but a SHA-pinned action is content-addressed where the current URL is not.
4. **Add it to `devbox.json`** and run the job through devbox, which puts it under the lock file and Renovate's existing update flow. Heaviest change; only worth it if other tools follow.

Option 1 is the smallest honest fix; option 2 is the one that removes the problem rather than checking it. Also drop `-q` (or use `--no-verbose`) either way, so a failed download is loud.

## Scope note

`docs-publish.yml` is on the release path and is reached both from `release.yml`'s `docs` job and standalone via `workflow_dispatch`. Whatever lands here should be validated with `gh workflow run docs-publish.yml --ref <branch>` before a tag depends on it — the canary described in `docs/develop/governance.md` step 5.

vfarcic:
Closes #515.

`docs-publish.yml` installed `yq` with `sudo wget -qO /usr/local/bin/yq …` + `sudo chmod +x`, pinning a version *tag* and nothing else. This replaces that with a content-pinned, verified, unprivileged install.

## Which option, and why

**Option 1 (verify a checksum)** — the issue's "smallest honest fix" — but in a stronger form than the issue proposed: **the SHA-256 is committed to this repo** rather than downloaded from the release's `checksums` asset.

That deviation is the one judgement call in this PR, so to be explicit about it: the issue's stated threat is that *"a release asset is mutable — a tag can be re-pointed and an asset re-uploaded."* `checksums` is an asset of that same release, published by the same pipeline, signed by nothing. Fetching it at runtime verifies **the transfer** but not **the content** — anyone who can re-upload the binary can re-upload its checksum in the same motion, and the check passes. Only a digest committed here pins the bytes, which is what option 1 claims to do ("pins content, not just a tag name"). The runtime download of `checksums`/`checksums_hashes_order` would have added a second network dependency on the release path in exchange for no coverage against the named threat.

The `checksums` file is still useful — as a *second source for a human refreshing the pin* — so the step comment records how to read it (its columns are ordered by `checksums_hashes_order`, where SHA-256 is line 18, i.e. field 19 of the row counting the filename as field 1). I verified the pinned digest both ways: `sha256sum yq_linux_amd64` and that column agree on `75d893a0…49fa9`.

**Why not option 2 (drop `yq`).** In fairness it was viable — I checked, and all three edits are single-line substitutions that `sed` could make. I did not take it because of this PR's hard constraint: *do not change what the workflow produces.* Keeping `yq` makes byte-identity a property of construction (same binary, same expressions) rather than something a reviewer has to re-derive — and something that would need re-deriving again on any future change to those two YAML files' shape. `python3` + `pyyaml` was the worse sub-option regardless: it round-trips the whole document and would drop the existing `"v0.37.1"` quoting and re-emit both files wholesale.

**Why not 3 or 4.** A SHA-pinned setup action (3) is content-addressed for the *action*, but the action still downloads `yq` by its own logic — it moves the unverified fetch out of sight into a third party running inside a job that holds the admin PAT. Devbox (4) means provisioning devbox + nix on the runner for one binary in a job that uses neither; disproportionate, and only worth it if other tools follow.

## What the step does now

```
curl --fail --location --silent --show-error \
     --retry 3 --retry-delay 2 --retry-connrefused  →  $RUNNER_TEMP/yq
sha256sum --check --strict                          →  verify against the committed digest
install -m 0755                                     →  $RUNNER_TEMP/bin/yq   (no sudo)
echo "$RUNNER_TEMP/bin" >> $GITHUB_PATH
```

Four things change beyond the checksum:

- **`-q` is gone.** `--silent --show-error` suppresses the progress meter only; `--fail` makes an HTTP error a non-zero exit. Verified: a 404 now exits **22**, prints `curl: (22) The requested URL returned error: 404`, and **leaves no output file at all** — so there is nothing for a later step to mark executable. Previously the failure surfaced downstream at `yq -i` with no indication it was a download problem.
- **A substituted body fails the checksum**, loudly. Verified against an HTML error page: `FAILED`, `WARNING: 1 computed checksum did NOT match`, exit 1.
- **`sudo` is gone entirely**, and nothing is written into a system path. The binary is verified *before* it is installed anywhere, and it lands in a job-local dir put on `PATH` via `$GITHUB_PATH`. An unverified download can no longer reach a root-owned location.
- **`--retry` is there to *preserve* behaviour, not to add a feature** — and it is the one addition here I want a reviewer to look at twice. `wget` retries by default (`--tries` defaults to 20); bare `curl` does not, so a straight swap would have been a quiet robustness *regression*. That matters at this specific line because the step runs **after** "Build and push docs image", so a transient blip strands the job in exactly the half-done state this file's other comments keep warning about — image already in GHCR, `main`'s chart pointer stale (the v0.35.9 shape). curl retries 5xx/408/429 and a refused connection; a 404 stays fatal.

## Output is unchanged

The three edits (`.image.tag`, `.version`, `.appVersion`) are untouched and run under the same `yq` v4.52.5. I ran them against the real `site/helm/values.yaml` and `Chart.yaml` — each is a surgical single-line change that preserves surrounding formatting and quoting exactly (`tag: "v0.37.1"` → `tag: "main-abc1234"`; no reflow, no requoting, no reordering). The resulting chart commit is byte-identical.

## Renovate

Worth knowing before someone "fixes" this: **nothing bumps `yq` today, and nothing did before this PR either.** Renovate has no manager that reads a URL out of a `run:` block, and none of `renovate.json`'s `customManagers` matches this one. So this PR does not *introduce* manual maintenance — it makes the existing manual pin explicit and adds a second value that must move with it. The step comment says so, including the warning that a regex manager taught to bump only the version would produce a PR that fails the checksum check. I deliberately did not add one.

## Verification

- **`actionlint` 1.7.12 + `shellcheck` 0.11.0** (via `nix shell`; the repo pins neither, so `grep -rn actionlint` finds nothing): **exit 0** on the new file, and exit 0 on the `HEAD` version too — so this is clean, not merely no-worse. shellcheck matters here since the step is new shell.
- **Rule 2** — `cargo fmt --check` clean; `cargo clippy --workspace --all-targets --features e2e -- -D warnings` clean. No-ops for a YAML change, but the gate is unconditional.
- **Rule 5** — `cargo test-fast` clean. **`cargo test-e2e` was not run, deliberately**: no Rust, daemon, TUI or harness code is touched, so it has nothing to exercise here, and rule 5 bounds it to the pre-PR step for a reason.
- **Rule 12 — does not apply.** CI workflow only; no daemon, protocol, orchestration or hook code, so there is no TUI↔daemon contract to break, no `PROTOCOL_VERSION` bump and no cross-version run. Stated rather than left silent.
- Fragment: `changelog.d/515.misc.md` (`misc` = "Internal improvements, CI/CD, tooling").

## The canary is yours to fire

`docs-publish.yml` can only really be proven by running it, and the canary is the way — `gh workflow run docs-publish.yml` (`docs/develop/governance.md` step 5). **I have not triggered it**: it pushes a commit to `main` under the admin PAT, which is not mine to do from a feature branch. Note `workflow_dispatch` runs the workflow from the **default branch**, so verifying *this* change needs `--ref agent/dispatch-issue-515` (the standalone dispatch path has the secret in scope, so it exercises the full push). Your call whether to fire it before or after merge.

## Scope

Untouched on purpose: `secrets: inherit` (#514) and the shell templating in this same file (#324).


prageethw (approved):
Independently verified the security claim rather than trusting the diff: fetched the actual v4.52.5 release's checksums/checksums_hashes_order assets from GitHub and confirmed field 19 of the yq_linux_amd64 row matches the digest committed at the new "Install yq" step. Went further and downloaded the real yq_linux_amd64 binary directly and computed its SHA-256 myself — exact match. This is a genuine content pin, not a cosmetic check. Confirmed step ordering via the full workflow file at PR head: "Install yq" runs after "Build and push docs image" and before git push origin main under the RELEASE_TOKEN admin PAT, matching the stated threat model. The replacement script verifies before installing, never touches a root-owned path, removes sudo entirely, and set -euo pipefail ensures the checksum failure aborts the job. No injection surface. CI green.

PR #659, opened 2026-08-23
chore(greptile): review every PR once instead of re-reviewing a subset

vfarcic:
## Problem

Greptile stopped reviewing on **2026-08-18** and has produced nothing since.

Measured 2026-08-23 across every PR in the window: it reviewed consistently from 08-12 through **PR #597 at 13:58Z**, then **zero comments and zero check-runs** on all 15+ PRs afterwards. The dashboard reads exactly **300 used**.

## Cause

`triggerOnUpdates: true` re-reviews on every push. **94 PRs** opened 1–18 August at **~3.4 commits each** ≈ **320 review triggers** against a **300** cap.

So the setting was not buying deep coverage of everything. It bought deep coverage of the first ~88 PRs and **nothing at all** for the remainder, with the cutoff falling arbitrarily wherever that month's volume landed.

## Change

`triggerOnUpdates: false` — review each PR once, when opened.

At ~124 PRs/month against a 300 cap, **every PR gets reviewed and the quota never runs dry**, with room to grow.

**What it costs:** the re-review after you push fixes. That is the smaller loss — `dismiss_stale_reviews_on_push: true` already voids the human approval on every push, so a person re-reads regardless, and initial diffs are where most findings live.

## Why CLAUDE.md changes too

`greptile.json` is JSON and cannot carry a comment, so the rationale has nowhere else to live — without it, someone reasonably flips the flag back.

Rule 8 also needed correcting: it stated Greptile "re-reviews on each new push", which is no longer true and would have you waiting for a review that never arrives.

The failure mode is worth recording on its own: **an exhausted quota creates no check-run at all**, so from inside the repo it is indistinguishable from the app being uninstalled, and nothing anywhere says which. That cost a full investigation tonight before the dashboard settled it. Rule 8 now says to check the quota before concluding the integration is broken.

## Verification

Nothing to run — config and docs only. `cargo fmt --check` clean.

**This PR will not itself be reviewed by Greptile**, since the quota is still exhausted until reset. That is expected, not a symptom.

## Rule 12

No daemon, protocol, orchestration or hook code. **No `PROTOCOL_VERSION` bump, no `.breaking.md` fragment.**

## Separable

The `greptile.json` line and the CLAUDE.md edits are independent — if you want the flag without the documentation, or vice versa, either half can be dropped.


prageethw (approved):
Checked that `greptile.json` flips `triggerOnUpdates` to `false`, `CLAUDE.md` rule 8 is updated to match the new once-per-PR review behavior and documents the quota-exhaustion failure mode, and a changelog fragment is included; no code paths are touched and CI is green.

Issue #170, opened 2026-06-16
Spawned panes inherit login-shell PATH + configurable agent command


vfarcic:
## PRD: Login-shell PATH for spawned panes + configurable agent command

**Problem**: The daemon resolves bare pane commands (e.g. `claude`, `opencode`) against its own process PATH, which lacks `~/.local/bin` when the daemon is launched without a login profile (e.g. over SSH). So normal dashboard panes, scheduled-task fires, and the hardcoded schedule-authoring agent all fail with portable-pty's "Unable to spawn … not found in PATH". Multi-word `devbox run …` commands escape only because they're shell-wrapped and `devbox` happens to be on the daemon's PATH.

**Solution**: Capture the user's login-shell PATH once at daemon startup and set it into the daemon's own environment, so every spawned pane (and the agents' own shell-outs) inherit it. Separately, replace the hardcoded `claude` authoring agent with a user-configurable agent command/picker, defaulting to the existing new-pane default.

**Detailed PRD**: See [prds/170-login-shell-path-and-agent-command.md](https://github.com/vfarcic/dot-agent-deck/blob/main/prds/170-login-shell-path-and-agent-command.md)

**Priority**: High

PR #660, opened 2026-08-23
fix(ui): drop the Enter hint when the new-pane form will refuse it (#589)

### LINKED ISSUE #589: New-pane form footer promises Enter: submit in the name-collision state, where Enter is silently refused

vfarcic:
Closes #589.

## The defect

In the new-pane form's name-collision state the action row correctly drops `[Submit]` and shows `! This name is already in use by a live orchestration.`, while the footer one line below it went on reading:

```
  Tab: switch  ◀▶: mode  Enter: submit  Esc: cancel
```

Pressing Enter there hits the collision guard, which returns `Action::Continue` with no feedback of any kind. Two lines of the same modal, one directly above the other, disagreed about whether submitting was available.

Reproduced exactly before any fix — this is the rendered buffer the new test captured on the red run, which matches the screenshot in the issue:

```
│  Name:                                                    │
│                                                           │
│  ! This name is already in use by a live orchestration.   │
│                                                           │
│ [Cancel]                                                  │
│  Tab: switch  ◀▶: mode  Enter: submit  Esc: cancel        │
```

## Cause

`new_pane_form_footer_hint` chose between `Enter: submit` and `Enter: next` from `name_submits` alone — "is Name the field where Enter submits?" — which cannot see a collision. The action row was answering a *different* question (`form.name_collision()`) a couple of hundred lines away. Two independent notions of "can submit", so they drifted.

## The fix

`new_pane_form_footer_hint` now takes `submit_refused`, and the caller passes **the same `name_collision` local it already computed to drop `[Submit]`** — so the footer and the button row read one value, produced by the one method (`NewPaneFormState::name_collision`) the Enter guard itself calls. They cannot disagree again without the button row being wrong too.

In that state the hint drops the `Enter:` clause entirely rather than naming a refused key, matching how the action row removes `[Submit]` rather than dimming it:

```
  Tab: switch  ◀▶: mode  Esc: cancel
```

No other state's wording changes. In particular a collision does **not** mute Enter on a field where Enter only advances focus — nothing is refused there, so `Enter: next` stays honest — and the `schedule_locked` branch (`Enter: confirm  Esc: cancel`, PRD #170 finding 6) is untouched.

## Tests

Followed `reproduce-first`: the failing test came first, was confirmed red *for the user's reason* (the render above), and went green on the fix.

Per that skill's bias toward extending over duplicating, the render coverage went into the existing spec test **`orchestration/identity/005`**, which already pins this exact state's user-visible truthfulness (`[Submit]` gone). It now also renders the form **those keystrokes actually produced** and asserts the footer omits `Enter: submit`. That detail matters: the existing `render_new_pane_orchestration_name_collision_to_buffer` seam opens focused on *Mode*, where `Enter: next` is honest, so it does not reach the defect — only focus on Name does.

Controls, so each assertion is attributable rather than vacuous:

- typing a free name over the empty one restores **both** `[Submit]` and the `Enter: submit` promise — proving the hint was not just dropped wholesale;
- `new_pane_form_footer_hint(true, false, true, false) == new_pane_form_footer_hint(true, false, false, false)` — proving the collision changes nothing on a field where Enter advances focus.

The unit test beside it was extended with the refusal case and renamed to `footer_hint_names_only_the_enter_action_the_form_will_honour`, since "switches to submit when name focused" no longer describes what it covers. `tests/CATALOG.md` and the `/// Scenario:` comment (rule 7) are updated to match.

Rule 4: this is a footer hint string, so L1 (`TestBackend`) is the right tier and no PTY-attached L2 test accompanies it.

## Rule 12 — not applicable

No daemon, TUI↔daemon protocol, orchestration or hook code is touched. The diff is confined to `render_new_pane_form`'s footer wording and its tests, so there is no wire-shape change, no semantic change behind a stable wire, and no cross-version contract to re-verify. `PROTOCOL_VERSION` is unchanged and no `.breaking.md` fragment is warranted — `changelog.d/589.bugfix.md` is a `bugfix`, which bumps patch.

## Gates

- `cargo fmt --check` — clean
- `cargo test-fast` — 3049 passed, 0 failed
- `cargo xtask linkage-check` — ok (557 catalog ids, 485 annotations, 9 rules)
- `cargo test-e2e` — **not fully green; two real-agent tests still fail for environmental reasons.** Full run: 6 failed + 1 timed out, all of them real-agent/LLM tests. Re-running that set passed 4 of 7; re-running the remaining 3 serially passed another (`dispatch_014`). The two that stayed red are:
  - `orchestration_dispatch_002_every_real_agent_role_comes_alive` — `orchestrator: NO PANE — never spawned at all`.
  - `dispatch_015_three_real_claude_seeds_are_genuinely_confirmed` — 2 of 3 real Claude panes confirmed their seed exactly; the third sat on an `● Error` card.

  Both are agent-spawn / live-LLM outcomes. The cause is visible in the process table rather than inferred: a **second agent session was running its own full `cargo nextest run --workspace --features e2e` from a sibling worktree throughout**, at one point executing `dispatch_015_three_real_claude_seeds` concurrently with this run — the same test, same machine, same Claude credentials — with 211 `dot-agent-deck` processes alive and a 15-minute load average of 24.5.

  This change cannot reach either test. `new_pane_form_footer_hint` has exactly one caller; its result is only pushed as a rendered `Line`; modal geometry measures the chip row and the warning copy but never the footer (`desired_w = chip_row_w.max(warning_w).saturating_add(4).max(56)`); and every non-collision return value is byte-identical to before. Neither test ever enters the collision state, so `render_new_pane_form` emits the same bytes for them with and without this diff. I did not run an `origin/main` baseline to confirm the two were already failing — on a box this loaded it would not have been a clean comparison — so that is stated as a limit rather than glossed over.

## Note for the reviewer: the issue's code pointers are stale

Verified against `origin/main` while re-anchoring. `new_pane_form_hint()` at `src/ui.rs:16805` does not exist — the function is `new_pane_form_footer_hint` and had moved; the guard cited at `src/ui.rs:7851` is now `resolve_seed_command`, unrelated. The defect itself was real and reproduced as filed; only the line references had rotted.


prageethw (approved):
Checked `src/ui.rs`'s `new_pane_form_footer_hint` and its single call site in `render_new_pane_form`: the footer now consumes the same `name_collision` boolean already used to drop `[Submit]` from the action row, so the two can't disagree again, and the new `orchestration/identity/005` assertions render the actual Name-focused state (not just the Mode-focused seam) to confirm `Enter: submit` is dropped while a free name restores it. CI is green.

Issue #397, opened 2026-08-07
macOS: `wrap` process does not exit after its child dies, leaving the wrapper behind at PID 1


vfarcic:
Follow-up from #396, where this was found, documented in the test, and deliberately left unfixed rather than block a Linux-verified leak fix.

## Symptom

On macOS a `dot-agent-deck wrap` process does **not** exit after its child has been reaped. On Linux it exits in well under a second.

Observed via the `wrap_max_lifetime_backstop_ends_an_unsignalled_wrapper_and_its_child` probe in `tests/wrap_io.rs`, which arms the wrapper's max-lifetime backstop (`DOT_AGENT_DECK_TEST_MAX_LIFETIME_SECS=1`) against a child that ignores SIGTERM:

```
the backstop killed the child but the WRAPPER never exited, so it would still be
left behind at PID 1.
```

Three consecutive `build-macos` runs held past **60 s** with the child already dead. Reproduced both while the test held the outer PTY master open and after dropping it between waits, so the held-open terminal is **not** the cause — that was my first hypothesis and it was wrong.

## Why it matters

#396 gave `wrap` the orphan and max-lifetime backstops that `daemon serve` already had, after three `wrap --agent codex` stubs were found alive for **three days** from a worktree that had since been deleted — two of them spinning `while [ ! -e "$WRAP_START" ]; do sleep 0.01; done` at roughly 100 wakeups/second.

The backstop works correctly on macOS: it fires, forwards SIGTERM, and escalates to SIGKILL, so the **child** — the expensive half, and the part that burned CPU — does die. But the wrapper process itself may remain. **The leak fix is therefore complete on Linux and partial on macOS.**

In the deck's normal flow this does not leak: stopping an agent force-kills the whole process group. It matters for the orphaned case the backstop exists to cover.

## What is already ruled out

The `run_wrap_pty` teardown is arithmetically bounded on paper:

- the stdin pump thread is spawned detached and never joined;
- `out_pipe_thread` / `err_pipe_thread` are `None` when all three descriptors are a tty, so their joins do not run;
- the output-drain waits are bounded (300 ms settle → `AGENT_TERMINATE_GRACE` → 500 ms), each followed by a group kill;
- `child.wait()` reaps an already-dead child.

[...]

PR #661, opened 2026-08-23
fix(wrap): bound the wrapped agent child independently of its wrapper (#657)

### LINKED ISSUE #657: e2e runs leak wrapped Codex processes that pin their temp roots forever — 35 orphans / 14.2 GB accumulated
e2e runs leak wrapped Codex processes that pin their temp roots forever — 35 orphans / 14.2 GB accumulated

## Summary

`cargo test-e2e` leaves `dot-agent-deck wrap --agent codex -- codex …` wrappers and their Codex children alive after the test process exits. Because `cargo xtask clean-e2e-tmp` **never** reaps a root whose owning PID is live — at any age, by design (rule 14) — every orphan pins its temp root permanently.

Measured on the dev box 2026-08-23 before cleanup:

- **35 orphaned processes**, all with `(deleted)` working directories under `/var/tmp/dad-e2e-1000/`
- oldest alive **8 days 13 hours** (started Aug 14)
- `/var/tmp/dad-e2e-1000` at **15 GB**
- after killing them, `clean-e2e-tmp` immediately classified **385 directories / 14.2 GB** as `dead-pid` and reapable

So the disk was not merely uncollected — it was *uncollectable*, and would have stayed so indefinitely.

## Not historical

Two of the 35 were spawned by a test run **that same night**, on current `main`. This reproduces now; it is not one old accident.

## Why the existing backstop misses them

`DOT_AGENT_DECK_TEST_MAX_LIFETIME_SECS` (300s) is meant to bound exactly this, and `wrap_max_lifetime_backstop_ends_an_unsignalled_wrapper_and_its_child` passes — so the mechanism works in isolation but is not reaching these children. The likely reason is the one rule 14 already documents for the daemon: the wrapper `setsid`s out of the test's process group, so process-group teardown does not reach it, and nothing else bounds its lifetime.

Both the `/bin/sh -c … wrap --agent codex --` shim and the wrapper binary survive, along with the Codex child under them.

## Why it matters beyond disk

The stale state these leave behind **actively produces misleading evidence**. While investigating #656 I drew three wrong conclusions from it:

- a leftover orphan's `/proc/<pid>/environ` read as if it belonged to the running test, which produced a false "the wrapper doesn't propagate the hook socket"
- two "no `hooks.json` / no trust records" readings came from temp roots left half-cleaned, which produced a false "Codex regression, unfixable"

Anyone verifying an e2e-touching PR runs the tier and mints more of them, so this degrades the reliability of the tier's own diagnosis.

## Suggested direction

1. Ensure the harness reaps wrapped agent children on teardown, not only the direct child — the wrapper detaches, so killing the test's process group is insufficient.
2. Confirm whether `DOT_AGENT_DECK_TEST_MAX_LIFETIME_SECS` is actually armed for wrapped Codex spawns in the harness; if it is, find why it does not fire; if it is not, arm it.
3. Consider making `clean-e2e-tmp` report live-PID-pinned roots more prominently — 385 pinned directories accrued with no signal. It already lists survivors with a reason, but a root pinned by a **days-old** process is a much stronger smell than one pinned by a live run and could be called out as such.

Reaping the 35 by hand (`kill -TERM`, 2 needed `-KILL`) was enough to unblock the reaper; no code change was needed to recover the space.

Related: #322 (the reaper), #461 (PID-liveness rules), #656 (the investigation this obstructed).

vfarcic:
Fixes #657.

## What leaks, and why the existing backstop misses it

The issue's second comment has this measured already, so only the consequence matters here: `DOT_AGENT_DECK_TEST_MAX_LIFETIME_SECS` **does** reach the wrapper and `arm_wrap_self_defense` **does** end it inside its window. The cap lives in the wrapper, though, and the wrapper bounds its child only *transitively* — by a reap loop calling `kill_pid_group`, which needs the wrapper alive to call it. `SIGKILL` is not catchable, so every path that ends a wrapper without letting it reap (the deck's own escalation past `AGENT_TERMINATE_GRACE`, a registry `force_kill_and_wait`, an OOM kill, a nextest timeout) strands the child. And `child_pre_exec` `setsid`s that child into its own session, so from that moment *nothing above it can signal its group at all*.

## Direction chosen: (1), give the child its own bound

Of the three the issue lists:

- **(2) `DOT_AGENT_DECK_EXIT_WHEN_ORPHANED=1` on wrapped spawns** — declined. The issue already notes it narrows the window rather than closing it, and it buys that narrowing by making a wrapper exit on a parent change, which is a new way for a test that restarts or re-parents its daemon to lose its agent. Paying a new flake mode for a partial fix is the wrong trade when a hard bound is available.
- **(3) harness sweep at teardown, matched by `CODEX_HOME`** — declined as the primary fix, though it is the right *shape*. It only runs when teardown runs, and the population that actually leaks is precisely the one whose teardown did **not** run: nextest `SIGKILL`s, panics-with-abort, and a Ctrl-C'd developer run all skip `Drop`. It would fix the tidy cases and miss the ones that made 385 directories.
- **(1) a bound the child owns** — taken. It is the only one that holds when nothing above the child is alive to hold it.

The `setsid` in `child_pre_exec` is untouched, as the issue asks: the wrapper needs the child in its own group so `killpg` targets the child and its descendants and nothing else. This adds a second, independent holder of that same kill rather than trading the grouping away.

### How

`arm_child_group_backstop` forks a small reaper at spawn time, on both wrap paths. It:

- **`setsid`s out of the wrapper's process group first.** The deck tears a wrapper down with `killpg(wrapper_pgid, …)`, which would otherwise take the reaper down alongside the very death it exists to survive. It is double-forked so it reparents to init rather than sitting as a zombie under a wrapper that never waits for it.
- **closes every inherited descriptor**, so it cannot hold the inner PTY master or the child's stdin pipe open and suppress the EOF/`SIGHUP` that might otherwise end the child on its own.
- **polls the child's group and exits the moment it is gone** — the normal case, within one 250 ms tick of a clean teardown.
- at the deadline, if the group is still there, **walks the same `SIGTERM` → `WRAP_TERMINATE_GRACE` → `SIGKILL`** path the reap loop walks.

It is bounded by deadline + grace and holds nothing, so it cannot become the leak it prevents.

**Why the child alone is not enough, and the direct child alone is not enough.** The observed leak was a *pair* — `node …/codex` plus a native `codex-linux-x64/vendor/…` child under it. A per-process cap (`alarm(2)` in the pre-exec, the smallest possible version of this) would kill the node parent and reparent the native child to init, converting one orphan into one orphan. Only a group-level kill reaches both, which is what the reaper does and what the test asserts.

**Gated exactly like `arm_wrap_self_defense`**, on `DOT_AGENT_DECK_TEST_MAX_LIFETIME_SECS`: a production wrapper never sets it, forks nothing, and behaves precisely as before.

**On pid reuse**, since this repo has a firm position on it (`docs/develop/e2e-temp-dirs.md`): the reaper signals only at its deadline and only after re-checking the group is still there, so hitting a recycled pgid would need the group to die and the number to be handed back out inside one 250 ms tick — the whole pid space wrapping in a quarter second, >130k forks/second even at the 32768 default. This *bounds* the window rather than *inferring* reuse, which is the thing that page forbids.

## Interaction with #397

**None, and deliberately so.** #397 is the inverse — on macOS the *wrapper* does not exit after its child dies — and it lives in `run_wrap_pty`'s teardown. This change touches neither the teardown nor the wrapper's exit path; it adds a process that is out of the wrapper's session, holds no descriptor of the wrapper's, and only ever signals the child's group. So it cannot make #397 better or worse.

It does slightly change what #397 costs. Its own test comment records the macOS consequence as "the backstop removes the expensive half of the leak (the child) but may leave the wrapper process itself behind". That stays true, and the expensive half is now covered on a strictly wider set of paths — including the `SIGKILL`ed wrapper, where previously *neither* half was covered. `fork`/`setsid`/`killpg`/`nanosleep`/`clock_gettime` are all POSIX, so the reaper is not Linux-only; the leak was measured on Linux and the test runs wherever the L2 tier does.

**Did this change the TUI↔daemon contract? No — no `PROTOCOL_VERSION` bump, no `.breaking.md`.** Nothing on the wire moved and no field changed meaning. The change is confined to `src/wrap.rs`, is unreachable unless `DOT_AGENT_DECK_TEST_MAX_LIFETIME_SECS` is set, and adds a process that speaks to nobody — it opens no socket, holds no descriptor, and only calls `killpg`. The changelog fragment is `misc` for the same reason: production behaviour is byte-identical.

**The manual test was run anyway**, since this is the wrapper/daemon lifecycle. `v0.37.1` (the previous release, downloaded binary, `server_version: 7`) as the daemon, this branch as the TUI, with `DOT_AGENT_DECK_LOG` isolated alongside the sockets, `HOME` and `DOT_AGENT_DECK_STATE_DIR` — verified afterwards that the real `~/.local/state/dot-agent-deck/deck.log` gained no sandbox daemon line (every `xver` mention in it is inside a `tool_detail=` of this session's own commands) while each sandbox wrote its own 33–36 line log.

Three passes, all **ALL-PASS**:

| pass | daemon | TUI | worker role | result |
|---|---|---|---|---|
| rule-12 required shape | v0.37.1 | branch | bare | delegate routes, work-done arrives, hooks drive the card |
| #657 shape | v0.37.1 | branch | **through `wrap`** | same, plus the wrapper and its `setsid`'d child observed alive with the cap in their env |
| reverse | branch | v0.37.1 | through `wrap` | same |

In each: `dispatch` built the isolated worktree and both roles registered; `delegate --to worker` reached the worker's pty (`Read .dot-agent-deck/worker-task-worker.md for your task.`) — through the wrapper in passes 2 and 3; `work-done` reached the orchestrator's pty; and `SessionStart` + `PreToolUse` hooks drove the worker's card to `{"status":"Working","active_tool":{"name":"XVERTOOL"}}`.

The wrapped passes also gave a **live confirmation of the fix on a real orchestration pane**, not just the unit probe. A `clear = true` respawn left a second wrapper + child behind when the sandbox daemon exited; its reaper (forked 12:17:54) killed the group and exited at the 300 s mark, and all three processes — reaper, leftover wrapper, leftover child — were gone by 12:22:5x. Before this branch that pair would have been a permanent `live-pid` pin.

## Tests

`wrap_child_group_backstop_reaps_a_child_stranded_by_a_sigkilled_wrapper` (`tests/wrap_io.rs`, fast tier, both wrap paths). A child that ignores `SIGTERM`, a grandchild in the child's group that ignores it too, and then the wrapper is **`SIGKILL`ed** so no reap loop can run and its own backstop never fires. The grandchild is the half that matters — it is the shape of the native Codex child, and a fix that only reaps the direct child reparents it to init.

Reproduce-first, measured both ways:

- **without** the fix (call sites commented out): all four probes — child and grandchild, PTY and pipe — survived the full 30 s budget, `child_gone: false, grandchild_gone: false` on both paths;
- **with** it: passes in ~2.5 s per path, 5.1 s total.

`cargo fmt --check` and `cargo clippy --workspace --all-targets --features e2e -- -D warnings` clean. `cargo test-fast`: **3050 passed, 0 failed**.

### `cargo test-e2e`: 8770/8775, and what the 5 were

Run once as rule 5 requires. Every one of the five was chased down; **none is caused by this branch**, and four of them cannot execute a line of it — `wrap` is reachable only for Codex (the sole `IntegrationStrategy::Wrapper` agent) and all four are real-**Claude** tests.

| test | rerun alone | verdict |
|---|---|---|
| `delegate_010_observed_session_start_waits_for_readiness_buffer` | PASS (1.2 s vs 2.8 s failing) | load flake |
| `shell_activity_005_real_claude_bash_child_trips_the_descendant_scan` | PASS (26 s vs 134 s) | load flake |
| `shell_activity_006_real_claude_bash_call_crossing_the_cap_keeps_the_badge_working` | PASS (130 s) | load flake |
| `dispatch_015_three_real_claude_seeds_are_genuinely_confirmed` | PASS (34 s) | load flake |
| `orchestration_dispatch_002_every_real_agent_role_comes_alive` | FAIL | **also fails on `origin/main`** |

`orchestration_dispatch_002` was baselined rather than argued about: checked out `origin/main` (`2de27b2`) in the same worktree and ran it alone — identical failure, same assertion at `tests/e2e_dispatcher_mode.rs:1243` (the `coder` never wrote the sentinel within its 300 s window), same 313 s. Its fixture is three real Haiku Claude panes; no Codex, no wrapper.

### Orphan check after the run

The `ppid=1` sweep from the issue found exactly one Codex orphan on the box, started **03:23**, hours before this branch existed, with the `(deleted)` cwd and `pgrp == sid == own pid` of the #657 signature. The e2e run on this branch minted none.

## Adjacent leak found, NOT fixed here

The same sweep turned up a different long-lived population that this change does not address and should not: `/bin/bash -c source …/shell-snapshots/… && eval '… && sleep infinity'` processes at `ppid=1`, rooted in `/var/tmp/dad-e2e-1000/dad-tests-*` temp roots, the oldest **8.9 days**. These are Claude Code **Bash-tool** children from the real-agent e2e tests (`shell_activity_*` deliberately runs `sleep infinity`), detached by Claude Code into their own session — not `wrap`, not Codex, a different mechanism end to end. They pin their temp roots against `clean-e2e-tmp` exactly the way the Codex children did. Worth its own issue; say the word and I will file it.

## Reviewer notes

- The reaper is a `fork` of the wrapper, so it inherits the wrapper's argv and appears in `ps` as what looks like a second `dot-agent-deck wrap …` at `ppid=1`. Given #657 is partly about stale processes producing misleading evidence, the discriminators are documented in the code and in `docs/develop/e2e-temp-dirs.md`, and both were measured on a live pair: a reaper's `/proc/<pid>/fd` is **empty** where a real wrapper holds nine, and its `pgid == sid != own pid` where a stray has `pgid == sid == own pid`.
- `fork` in a threaded process is the one thing here worth a careful read. Both child arms touch only async-signal-safe libc (`setsid`, `signal`, `close`, `clock_gettime`, `nanosleep`, `killpg`, `_exit`) — no allocation, no locks, no `tracing`, no Rust destructor. The reaper carries its own `CLOCK_MONOTONIC` clock for that reason rather than using `Instant`.
- The handoff `waitpid` cannot block (the intermediate only `setsid`s, forks and `_exit`s, all unconditional) and retries on `EINTR`, because `SignalGuard` deliberately installs its handlers without `SA_RESTART`.
- Rule 3: no new file. Rule 6: the single test was rerun alone before each tier.

https://claude.ai/code/session_01PGDSYU4uoFfGMJNAJAq1Nb


vfarcic:
## The control arm reproduced the leak live, minutes apart, on the same box

Worth recording, because it is a clean A/B rather than an argument.

The cross-version matrix ran the same `SIGKILL`-the-wrapper probe three times. In the two passes where **this branch's** binary was the wrapper, the leftover wrapper and its child were both gone at the 300 s mark, killed by the reaper. In the **reverse** pass the wrapper was the released `v0.37.1` binary — no reaper — and its child is still running as I write this:

```
PID     PPID  PGID     SID      ETIMES  STARTED
2744651 1     2744651  2744651  394     Sun Aug 23 12:19:15 2026
```

`ppid=1`, `pgid == sid == own pid`, past the 300 s cap and climbing: exactly the signature the issue measured on the four Codex children.

And its open descriptors say *why* it never noticed its wrapper die:

```
0 -> /dev/pts/17
1 -> /dev/pts/17
2 -> /dev/pts/17
3 -> /dev/ptmx      <-- the inner PTY MASTER
```

The child inherited a copy of its own PTY **master** across the spawn. Closing the wrapper's copy therefore hangs up nothing — there is still a reader on the master, so the slave never returns `EIO`, and the one signal that would have let a well-behaved child exit on its own never fires. A `tee` did this; a Codex process, which does not exit on stdin EOF anyway, had no chance.

That is also the reason the reaper closes every inherited descriptor before it starts polling rather than only tidying up. Had it kept them it would have pinned the master open for the whole cap window and become the very thing above.

(That stray is from my own sandbox under `dot-agent-deck-xver-657` and has been cleaned up.)

prageethw (approved):
Checked `src/wrap.rs`'s `arm_child_group_backstop`/`child_group_backstop_main` (double-fork reaper, async-signal-safe body restricted to raw libc calls, group kill via `killpg` with SIGTERM→grace→SIGKILL) and the new `wrap_child_group_backstop_reaps_a_child_stranded_by_a_sigkilled_wrapper` test, which SIGKILLs the wrapper directly (bypassing `SignalGuard`) and confirms a SIGTERM-ignoring child *and* grandchild are both still reaped on both wrap paths. The whole mechanism is gated on the test-only `DOT_AGENT_DECK_TEST_MAX_LIFETIME_SECS` env var, so production wrap forks nothing extra and is behaviorally unchanged. CI is green.

Issue #488, opened 2026-08-10
devin_live_001 is red on main: real Devin turn emits Thinking, then no further event within 45s


vfarcic:
Sibling of #460 (three other e2e tests consistently red on `main`) and #351 (load-induced PTY-wait flakes). This one is neither: it is a **real-agent** test against Devin's cloud service, so it has its own failure mode and belongs in its own issue.

## The failure

```
devin_live_001_real_interactive_turn_drives_the_card_live   (tests/common/mod.rs:4133)
```

Surfaced during the pre-release e2e gate for v0.35.10.

```
no broadcast AgentEvent matched the predicate within 45s; observed events: [
    AgentEvent {
        session_id: "creative-curtain",
        agent_type: Devin,
        event_type: Thinking,
        ...
        user_prompt: Some(
            "List the files in this directory with the exec tool, then reply with the exact filename that starts with devin_live_sentinel.",
        ),
        pane_id: Some("1"),
        agent_id: Some("1"),
    },
]
```

## What this tells us

The plumbing works. A real Devin session was created, the directive prompt was delivered, and a `Thinking` event came back over the broadcast channel with the right `pane_id`/`agent_id`. What never arrived is any **subsequent** event — no tool start, no completion — within the 45s window. The agent went quiet after acknowledging the turn.

## Reproduces in isolation, but the cause is likely not local

Per rule 6 it was rerun alone and failed again (55.6s, vs 91.7s in the full-suite run). That rules out the #351 load-contention pattern, but it does **not** establish a defect in this repo: a cloud agent that accepts a turn and then produces nothing for 45s looks the same from here whether the cause is service latency, a quota/availability problem, an upstream behaviour change in how Devin reports tool use, or a genuine regression in our event mapping.

## Worth triaging along these lines

- **Is 45s simply too short for Devin?** `696994a` ("budget devin/live/001's waits and grant it a timeout window") already tuned this once, which suggests the budget has been marginal before. Devin is a remote, queued service — its cold-start is not comparable to a local `claude` process.
- **Did Devin's event surface change?** The test's predicate expects something after `Thinking`. If Devin's native hooks stopped emitting the event we key on, the mapping in the wrapper needs updating rather than the timeout.
[...]

Issue #463, opened 2026-08-09
Retire the `NameCwd` orchestration-identity fallback once pre-v0.35.0 clients are dropped


vfarcic:
## Follow-up from #156

A **compatibility** decision, not a correctness one. Filing so the trigger and the cost are recorded; there is no reason to act on it yet.

### Background

PRD #140 (`cb307ca`, released v0.35.0) replaced the `(name, cwd)` orchestration routing identity with a per-tab instance token, keeping the old tuple as a fallback variant (`src/state.rs:405`):

```rust
pub enum OrchestrationIdentity {
    Instance { id: String, name: String },   // per-tab token
    NameCwd  { name: String, cwd: String },  // legacy fallback
}
```

`NameCwd` is constructed only when a client sends no `orchestration_id` (`src/daemon_protocol.rs:1359`). Both in-tree producers mint one (`src/tab.rs:857`, `src/spawn.rs:499`), so the fallback is reachable **only in a mixed-version pairing with a pre-v0.35.0 build**:

- **old TUI + new daemon** — panes get `NameCwd`, so two tabs from that old TUI in one directory still cross-deliver;
- **new TUI + old daemon** — the old daemon ignores the field and runs pre-#140 routing.

Locally that window is mostly closed by the build-version handshake (`ensure_compatible_daemon_or_die`, `src/build_version_handshake.rs:155`), which restarts a mismatched daemon — silently with no agents running, with TTY consent otherwise. What remains reachable is `HandshakeOutcome::ProceedOnExisting` (the user declines the restart to avoid stranding live agents, PRD #161 D4) and the remote-daemon path in `src/connect.rs`.

In every one of those cases behaviour degrades to exactly the pre-#140 semantics: correct across directories and across differently-named orchestrations, ambiguous only in the same-name-same-directory case that was always ambiguous. That is documented in `changelog.d/140.breaking.md`, so it is a known contract, not a silent misroute.

### Why removing it is not free

Dropping the variant converts graceful degradation into a hard refusal, and the degradation path matters: a token-less pane would end up with **no** identity at all, and per the `None == None` degeneracy in `delegate_targets` that means one merged routing group rather than a clean failure. So removal has to be paired with the daemon actively rejecting token-less orchestration panes, not merely deleting the arm.

Per CLAUDE.md rule 12 this would be a genuine contract change, stricter than #140's:

[...]

PR #665, opened 2026-08-23
fix(dispatch): keep a role card's name across a clear=true respawn (#663)

### LINKED ISSUE #663: orchestration_dispatch_002 is red on main: a dispatched orchestration's worker panes never register their role, so the orchestrator's delegate goes nowhere
`orchestration_dispatch_002` is red on `main`: a dispatched orchestration's worker panes never register their role, so the orchestrator's delegate goes nowhere

## Summary

`e2e_dispatcher_mode::orchestration_dispatch_002_every_real_agent_role_comes_alive` **fails deterministically on `main`**, in isolation, at `-j 1`. It is **not** a load flake — it fails identically alone and in a full run, taking ~313s each time.

Verified 2026-08-23 by stashing all local changes and running against clean `main`: same failure, same duration.

There is currently **no tracking issue** for it.

## The symptom, and why it looks healthy

The test's own message states the stakes:

> the dispatched orchestration's `coder` never did the delegated work within 300s … A dispatched orchestration whose orchestrator cannot reach its own workers looks completely healthy (panes up, cards named, tab live) and gets nothing done: the orchestrator reports the worker is working and then waits forever for a work-done that cannot arrive.

The diagnostic dump reports:

```
- orchestrator: NO PANE — never spawned at all
- coder:        NO PANE — never spawned at all
- reviewer:     NO PANE — never spawned at all
Orchestrator pane sched-dispatch-real-probe-0-r0, coder pane sched-dispatch-real-probe-1-r1.
```

…while the rendered grid in the same dump plainly shows the panes alive.

## The lead: worker panes carry a UUID instead of their role name

In the final grid, the orchestrator card is named correctly but the worker card is not:

```
┌ 1 ClaudeCode · orchestrator ───────── ● Idle ┐
┌ 2 ClaudeCode · 64c085a2-4f… ───────── ● Idle ┐     <-- should be `coder`
```

Observed across two independent runs with different UUIDs (`64c085a2-4f…`, `48eb3c8b-3b…`), so it is systematic, not a one-off.

That matches the "NO PANE" lines exactly: the worker pane exists and is running an agent, but it never registered under its **role**, so the role map has no `coder` entry and `delegate --to coder` cannot resolve it. The orchestrator visibly *did* delegate — its card shows the tool call:

```
Bash — dot-agent-deck delegate --to coder --task "Create a file …
```

and then both panes sit `Idle` forever. The right-hand pane header also reads `orchestrator — disconnected`.

## Why this may be a product bug, not a test bug

The failure mode the assertion describes — a delegate that resolves at delegate time, reports success, and delivers nothing — is the same class as **#524** (`delegate_targets` matching a stale pane) and **#309** (a rejected delegate exiting 0). If worker panes in a *dispatched* orchestration genuinely never register their role, then dispatched orchestrations are broken for real users in the way the test says: everything looks fine and no work happens.

Worth establishing early which it is:

1. Does the worker pane register a role at all in the **dispatch** path, versus the interactive `Ctrl+N` orchestration path that `orchestration_dispatch_001` covers?
2. Is the UUID display name a symptom of the same missing registration, or a separate naming fallback? `NameCwd` / orchestration-identity fallbacks are discussed in **#463**.
3. Does `delegate --to coder` return success to the orchestrator despite resolving nothing? If so that is #309/#524 territory and worth cross-referencing.

## Reproducing

```bash
cargo test-e2e -j 1 orchestration_dispatch_002
```

Takes ~313s. Uses real Claude Haiku via the Claude Max subscription (no metered API cost).

## Note on the surrounding tier

Two other e2e tests were red in the same run and are **not** this bug:

- `devin_live_001` — tracked in #488
- `dispatch_015_three_real_claude_seeds_are_genuinely_confirmed` — a genuine load flake; **passes in isolation** (32s at `-j 1` vs 156s failing in the full run). Filed separately.

This one is the only deterministic failure of the three.

vfarcic:
Closes #663.

## Product bug or test bug?

**Both — and the product half is real.** The issue's headline ("workers never register their role") turned out to be an artifact of a lying diagnostic; the delegate resolved `coder` correctly on the very first try. But a genuine, user-visible defect was sitting underneath it, and it is exactly the one the issue spotted as "the lead": **a dispatched orchestration's role card loses its role name the moment the team does its first piece of work.**

## What actually happens, measured

Reproduced against clean `main` (317s, `-j 1`, in isolation) with `DOT_AGENT_DECK_LOG` + `RUST_LOG=dot_agent_deck=debug`. The daemon log settles all three questions the issue asked:

```
16:09:08.439  Received delegate signal pane_id=sched-…-0-r0 targets=["coder"]
16:09:08.502  delegate: respawned worker agent for clear=true; waiting for SessionStart … new_agent_id=5
16:09:09.129  Received event session_id=4f1a85cb… event_type=SessionEnd   pane_id=…-1-r1
16:09:09.907  Received event session_id=c70493f1… event_type=SessionStart pane_id=…-1-r1
   (…and then nothing from the coder pane, ever)
```

1. **Does the worker register a role on the dispatch path?** Yes. `spawn.rs` has registered it since #454/#600, and the log shows the delegate resolving `coder` and respawning it. Nothing to fix here.
2. **Is the UUID card name the same bug or a separate one?** **Separate, and real** — see below. Nothing to do with `NameCwd` / #463.
3. **Does `delegate` report success while resolving nothing?** No — it resolved, routed and respawned. Not #309/#524 territory.

### 1. The product bug — the card name (`src/state.rs`)

The `clear = true` respawn SIGTERMs the worker. Its `SessionEnd` reaches `AppState::apply_event`, which removes the named session and restores a placeholder — and the placeholder was born nameless. The replacement's `SessionStart` is a fresh generation whose `inherited_display_name` can only reach back to that placeholder, so it inherits the `None` and the card renders the new agent's session id: `ClaudeCode · c70493f1-13…`.

The friendly name describes the **pane** — its role, or a scheduled task's title — not the conversation that just ended. That is already the rule on the supersession path (PRD #127 finding #2); the session-end path did not follow it. It does now, which also means a card keeps its name across an agent's own `/clear` restart.

**Why only a dispatched orchestration shows it.** Every other way of opening one — the `Ctrl+N` form, a restored session, a reconnect — *also* seeds the TUI-side `ui.pane_display_names` mirror that `render_dashboard` falls back to, so the name is repainted from there. `ui::surface_one_orchestration`, the tab a `dot-agent-deck dispatch --orchestration` builds, does not seed it. (That parity gap is now masked rather than closed. It is presentation-only and affects nothing else that works on this path today — `ui.pane_metadata` is not seeded either, so a dispatched tab is not in the session snapshot regardless — so I left it out of a release-blocking PR rather than threading `&mut UiState` through two more signatures. Happy to file it as a follow-up.)

### 2. The test defect — the assertion that actually fired

The failure the test *reported* — the coder never doing the work — was configuration. The harness pins `DOT_AGENT_DECK_DELEGATE_READINESS_BUFFER_MS=0`, because most e2e workers are stand-ins that accept input the instant they exist. PRD #249's own measurement is that a **real** agent's task pointer is *lost at `0`* and *delivered-and-submitted at `1000`*: `SessionStart` means "a session exists", not "the TUI interprets `\r` as submit".

`orchestration/dispatch/002` became a real-agent `clear = true` scenario when PR #646 flipped `dispatch-orch-real`'s `coder`, and it inherited the pin. `orchestration/delegate/014` and `/015`, the other two real respawn scenarios, already opt back in to `1000`. The log above shows the cost: the pointer was written ~400ms after `SessionStart`, into a Claude still painting its banner, and dropped — the coder card ends the run with an empty `Prmt:` and `Tools: 0`.

This is **not** a relaxed assertion. The test still requires the coder to genuinely create the sentinel file; it just runs the readiness value the product actually ships. **With it, the test passes in 43s instead of failing at 317s.**

### 3. The diagnostic that sent the issue the wrong way

`WORK_WAIT` is 300s. The harness's leaked-daemon backstop is *also* 300s. So on a red run the daemon self-exited at `16:13:50` — ~25s **before** the assertion fired at `16:14:14`:

```
16:08:50  test max-lifetime backstop armed (test-only safety net) secs=300
16:13:50  daemon test max-lifetime reached; initiating graceful shutdown
```

`role_diagnostics` then queried a dead socket, got no records, and printed `NO PANE — never spawned at all` for all three roles beside a grid showing all three alive — and `orchestrator — disconnected` in the pane header. That is the evidence issue #663 was filed on. The test now raises the backstop past its own budget (900s; the test runs ~43s), and `role_diagnostics` says outright when the registry came back completely empty, so a dead daemon can never again masquerade as a spawn failure.

**No `PROTOCOL_VERSION` bump, and no `.breaking.md`.** Nothing on the wire moved and no field changed meaning. `AppState::apply_event` is per-process (the TUI and the daemon each keep their own), `display_name` is presentation, and the only signature change (`insert_placeholder_session` now returns the id it minted) is internal Rust. Bump policy: **patch** (`bugfix`).

**Cross-version run done**, with `DOT_AGENT_DECK_LOG` isolated alongside `HOME`, `DOT_AGENT_DECK_STATE_DIR` and both sockets, in a disk-backed sandbox:

- daemon = **v0.36.2** (`build_version 0.36.2-g4bb47a5`, protocol 7), started from the released binary and **never restarted** (one `Daemon listening` line in the sandbox log; the pid is still the old binary's at teardown);
- TUI/CLI = this branch, attached via `DOT_AGENT_DECK_BUILD_ID_OVERRIDE` so the handshake matched rather than recycling the daemon.

Results, driven through a real PTY:

| flow | result |
|---|---|
| branch TUI opens the orchestration on the old daemon | tab built, both cards named `orchestrator` / `worker` |
| branch `delegate --to worker` | exit 0; worker PTY received `Read .dot-agent-deck/worker-task-worker.md for your task.` |
| branch `agent-event --type running` | worker card moved to `● Thinking` |
| branch `work-done --task …` | `Worker worker has completed their task…` written into the orchestrator pane |
| old daemon's own #249 silent-worker report | reached the orchestrator pane unchanged |

## Tests

- **`status/supersede/008`** (new, fast tier, `tests/card_supersession.rs`) — the cheap test that would have caught the registration/naming gap, per rule 4. A spawn-time placeholder, a first generation, its `SessionEnd`, then a replacement `SessionStart` under a different agent id: one card must remain and must still be named. **Verified red before the fix** (`left: None, right: Some("morning-digest")`) and green after.
- **`orchestration/dispatch/002`** gains a **post-delegation** card assertion. The existing card check runs *before* the delegation, so it could never see what a `clear = true` respawn does; the UUID was only ever visible in a failure dump. Now it is an assertion. Strengthened, not relaxed.
- `tests/CATALOG.md` updated for both.

## Gates

- `cargo fmt --check` ✅
- `cargo test-fast` — **3063 passed, 0 failed** ✅
- `cargo test-e2e` (full tier) — **8784 passed**; `orchestration_dispatch_002` **passed**. Four others were red; all four are pre-existing or load, none related to this change, and each was re-verified on this branch:

  | test | status |
  |---|---|
  | `devin_live_001` | pre-existing, tracked in **#488** (named in the issue) |
  | `dispatch_015_three_real_claude_seeds_are_genuinely_confirmed` | known load flake — **passes in isolation on this branch, 33s** |
  | `pi_live_002_native_seeded_orchestration_delegates_live` | load — **passes in isolation on this branch, 21s** |
  | `chain_smoke_pi_002_worker_receives_delegate_and_signals_work_done` | timed out under load — **passes in isolation on this branch, 47s** |

  Worth noting for the flake reading: a second agent session was running its own full `cargo test-e2e` on this 16-core box for part of that run.

## Release note

Fragment: `changelog.d/663.bugfix.md`.

https://claude.ai/code/session_01RbNMpzyuNrhch4wmJYtRTh


prageethw (approved):
Approving. Read the whole diff against the current tree at `d66f78e`.

**The product fix is right, and for the reason stated.** `display_name` is a property of the pane, not of the conversation, and the `SessionEnd` restorer was the one path that did not follow the rule PRD #127 finding #2 already established for supersession. Two things I checked specifically:

- **The name cannot get stuck.** `apply_event`'s metadata refresh (`DISPLAY_NAME_METADATA_KEY`, further down) runs unconditionally and takes precedence over the inherited value, so a genuinely new named occupant of the pane still wins; the new code only fills a hole that was previously filled with nothing.
- **There is no second path that drops it.** The `inherited_display_name` retire loop is gated on `claims_generation`, which is `event.event_type != EventType::SessionEnd`, so a `SessionEnd` never reaches it — the restorer really is the only place the name could be lost, and returning the minted id from `insert_placeholder_session` keeps the key format in one place rather than restating it.

Rule 12: agreed on no `PROTOCOL_VERSION` bump — `AppState` is per-process, `display_name` is presentation, and the only signature change is internal Rust. The cross-version run against the v0.36.2 daemon with an isolated `DOT_AGENT_DECK_LOG` is the right shape.

Test coverage matches the fix at both altitudes: `status/supersede/008` pins the state machine deterministically, and the post-delegation card re-assertion in `orchestration/dispatch/002` is the half that can actually see the regression — the pre-delegation check passed throughout. The `DOT_AGENT_DECK_DELEGATE_READINESS_BUFFER_MS=1000` opt-out of the harness pin is a correction, not a relaxation: the sentinel assertion is unchanged, and `orchestration/delegate/014` and `/015` already run the production value for the same reason. The `role_diagnostics` empty-registry note is worth having on its own — a dead daemon reading as `NO PANE — never spawned at all` is exactly the diagnostic that misdirected #663.

Two non-blocking notes, neither of which I'd hold the PR for:

1. **Please do file the `ui.pane_display_names` parity follow-up.** You call it out in the description as masked rather than closed, and I agree it doesn't belong in a release-blocking PR — but `ui::surface_one_orchestration` not seeding the mirror that every other orchestration path seeds is the thing that made this bug visible on exactly one surface, and it will make the next one visible on exactly one surface too. A tracked issue costs nothing and stops the asymmetry being rediscovered from the symptom again.
2. **`tests/e2e_dispatcher_mode.rs`, the `DOT_AGENT_DECK_TEST_MAX_LIFETIME_SECS` comment** — "900s still bounds a leak (the test itself runs ~320s)" reads as the current runtime, but ~320s is the *failing* run; the description measures the fixed test at 43s. Sizing the backstop against the worst case (the 300s `WORK_WAIT` plus boots) is the right call, so this is purely wording — something like "worst case ~320s; it passes in ~43s" would keep the next reader from concluding the test still takes five minutes.

Issue #666, opened 2026-08-23
A spawn-time prompt whose payload lands during the agent's boot is unrecoverable: both payload attempts fall inside a ~1.65 s window


vfarcic:
## Summary

A spawn-time prompt whose payload is written while the agent is still booting is **unrecoverable**, because the only two attempts that carry the payload both fall inside the agent's startup window and every later attempt is a bare submit probe. Found while fixing #664, whose test this is the residual cause of.

`attempt_writes_payload` is `attempt <= MAX_PAYLOAD_SUBMISSIONS` (2) and nothing else — it has no notion of whether the agent that will read those bytes had even started yet.

## The measured window

From `scheduler/dispatch/015`'s delivery log (the diagnostics added in #664's fix), one run, three panes, same deck, same instant:

```
16:36:23.489  written  beta   attempt=1     <- consumed by the launcher, which then execs claude
16:36:23.528  written  alpha  attempt=1
16:36:25.160  written  gamma  attempt=2
16:36:25.177  CONFIRMED gamma attempt=2     <- 17 ms later: claude was already up
16:36:25.184  written  alpha  attempt=2     <- claude not up yet
16:36:25.658  CONFIRMED beta  attempt=2
16:36:26.185  re-submitting alpha attempt=2 (submit probe, empty payload)
16:36:28.338  ... attempt=3
16:36:32.490  ... attempt=4
16:36:40.641  ... attempt=5
16:36:55.794  ... attempt=6
16:37:10.946  ... attempt=7
16:37:20.376  WARN abandoning alpha attempts=8
```

Attempt 2 lands **~1.65 s** after the launcher `exec`s the agent — `unconfirmed_retry_delay(1)` (500 ms) plus the confirmation wait. A warm `claude` reaches its input handling in roughly that time, so:

- boot under ~1.65 s → attempt 2's payload is accepted and confirmed within ~20 ms (`beta`, `gamma` above);
- boot over ~1.65 s → the payload is gone, and attempts 3-8 spend the next **55 seconds** pressing Enter on an empty input box.

The 1.65 s figure is the entire tolerance. Measured pass rate of the three-pane test tracks machine load exactly as that predicts: 5/5 green at load average ≤ 4 on 16 cores, and roughly a coin flip per pane above ~20.

## Why the payload does not survive in the tty buffer

[...]

Issue #415, opened 2026-08-07
delegate_prompt_injection's PTY/daemon tests starve each other: 6-7/40 fail in parallel, 40/40 pass at -j 1


vfarcic:
Found while porting the #402 fix (#414). This is a **separate defect** from #402's tick alignment, and it is the larger one.

## Measurement

On **unmodified `main`**, same machine, back to back, with the machine otherwise idle (load 2.93 on 14 cores):

| Configuration | Result |
|---|---|
| Full 1533-test suite | **1** failure |
| `-E 'binary(delegate_prompt_injection)'`, parallel | **6-7 of 40** fail |
| `-E 'binary(delegate_prompt_injection)'`, `-j 1` | **40 / 40 pass** |

Running the binary **alone is worse** than running it inside the full suite. That rules out external load as the cause, and it is why this went misdiagnosed: the intuitive experiment (isolate the flaky binary) makes it look *more* broken, not less.

## Mechanism

In the full suite these 40 tests are spread among 1500 others, so few run concurrently. Isolate them and all 40 pack onto the same cores at once — each spawning a PTY, a daemon, and a `cat` stand-in agent — and they starve **each other** against 2-second wall-clock deadlines. The competing load is the tests themselves.

Every failure lands at ~2.02-2.11s, i.e. the deadline, and the assertions report *empty* snapshots rather than wrong ones — nothing arrived in time, as opposed to something arriving incorrectly.

## Why this keeps costing real time

`tests/delegate_prompt_injection.rs` has **no `cfg(feature = "e2e")` gate**, so it runs in the **fast tier** — which has **no `--retries`**, while the informational e2e job gets `--retries 2`. So the tier with zero tolerance is running PTY-and-daemon-spawning tests that do not degrade gracefully under contention. That is the structural asymmetry #402 already flagged, now with a measured mechanism behind it.

Concrete cost so far: #350 was filed as a possible rendering regression when it was a non-deterministic snapshot; #402 spent a review cycle proving a red macOS job was unrelated to its diff; and #364 records the same pattern a third time.

## Suggested direction

A per-binary concurrency cap in `.config/nextest.toml` is likely the cheapest real fix — it addresses the class rather than individual deadlines, and unlike raising timeouts it does not slow the suite when tests pass. nextest supports `test-threads` on a filter expression.

Worth weighing against, or combining with:
- Giving the fast tier a retry policy, or moving these specific tests to a tier that has one.
[...]

PR #667, opened 2026-08-23
test(dispatch): stop /015 spending half its delivery budget on a gate it cannot satisfy (#664)

### LINKED ISSUE #664: dispatch_015_three_real_claude_seeds_are_genuinely_confirmed fails under full-tier parallelism but passes in isolation
`dispatch_015_three_real_claude_seeds_are_genuinely_confirmed` fails under full-tier parallelism but passes in isolation

## Summary

`e2e_prompt_delivery_confirmation::dispatch_015_three_real_claude_seeds_are_genuinely_confirmed` fails in a full `cargo test-e2e` run and **passes cleanly when run alone**.

Measured 2026-08-23 on `main` (`e8c443a`), back to back:

| Run | Result | Duration |
|---|---|---|
| full tier (default parallelism) | **FAIL** | 155.9s |
| `cargo test-e2e -j 1 dispatch_015` | **PASS** | 32.2s |

A 4.8× slowdown before failing is the signature of starvation rather than a logic defect: the test spins up **three** real Claude seeds concurrently and waits for each to confirm, so it is unusually sensitive to how much CPU and PTY bandwidth the rest of the tier leaves it.

## Why file it separately

This is an instance of the class described in **#351** ("cluster of PTY-wait timeouts under full-parallel load — 5-6 tests, different file each run") and **#415** ("6-7/40 fail in parallel, 40/40 pass at `-j 1`"), and it belongs to the strategy question in **#502**. But neither #351 nor #415 names this test, and searching the tracker for `dispatch_015` returns nothing — so the concrete reproduction was not recorded anywhere.

Filing it so the timings above are attached to a specific test rather than to a general description, and so it can be closed together with whatever #502 settles on.

## Possible directions

- The three-seed fan-out is the expensive part; if the assertions do not actually require all three to be in flight simultaneously, staging them would remove the contention without weakening the test.
- If simultaneity **is** the point (three concurrent seeds is plausibly the property under test), then it wants an explicit concurrency bound rather than a longer timeout — a deadline stretched to accommodate a loaded machine stops detecting the regression it exists to catch.
- Worth checking against #395 (`find_in_grid` never retries) and #413 (card snapshots read the wall clock at render time), which are the other two wall-clock-sensitivity issues in this tier.

## Reproducing

```bash
cargo test-e2e                             # fails, ~156s
cargo test-e2e -j 1 dispatch_015           # passes, ~32s
```

Uses real Claude Haiku via the Claude Max subscription (no metered API cost).

## Context

Found while gating the v0.37.2 release. Two other tests were red in the same run and are unrelated: `devin_live_001` (#488) and `orchestration_dispatch_002`, which unlike this one fails **deterministically** in isolation and is filed separately.

vfarcic:
Closes #664.

## What was wrong

`dispatch_015_three_real_claude_seeds_are_genuinely_confirmed` fails in a full `cargo test-e2e` run and passes alone. Reproduced on this branch's base (`e8c443a`), same machine, back to back:

| Run | Result | Duration |
|---|---|---|
| full tier, default parallelism | **FAIL** | 174.1s |
| `cargo test-e2e dispatch_015` | **PASS** | 32.2s |

The interesting part is not the slowdown, it is what the failing run *rendered*. Two of the three panes came back `confirmed_exact=None` **with their cards in `Error`**, while the third had genuinely worked:

```
┌ 2 ClaudeCode · dispatch-r… ● Working ┐   beta   — confirmed, Bash — ls seed-confirm-beta-8c42.txt
┌ 3 ClaudeCode · dispatch-rea… ● Error ┐   gamma  — confirmed_exact=None
┌ 4 ClaudeCode · dispatch-rea… ● Error ┐   alpha  — confirmed_exact=None
```

`Error` there is not "still booting". It is `spawn::abandon_spawn_prompt`'s delivery notice, published when a spawn-time prompt exhausts `prompt_delivery::AUTOMATIC_PROMPT_DEADLINE` — after which the daemon **stops retrying permanently**.

**That is why the issue's "do not simply raise the deadline" is not merely a principle here, it is arithmetic.** The deadline is 60 s, absolute, captured in `spawn::deliver` *before* the readiness wait (`src/spawn.rs:1096`). The test's own wait is already `wait_until(Duration::from_secs(150))` — 2.5× that — and nothing confirms a delivery that has already been abandoned. A bigger number only changes how long the tier takes to go red.

## Where the 60 seconds actually went

I added the deck log to the test's diagnostics before doing anything else, and it answered the question outright:

```
16:17:28.342  prompt written to pane … pane_id="…alpha-2" attempt=1
16:17:29.999  prompt written to pane … pane_id="…alpha-2" attempt=2
16:17:33.153  prompt delivery unconfirmed; re-submitting … attempt=3
16:17:37.306  … attempt=4
16:17:45.458  … attempt=5
16:17:58.191  WARN prompt delivery unconfirmed at the deadline; abandoning … attempts=6
```

Two things fall out of those timestamps:

1. **`beta` confirmed 1.7 s after its write and `gamma` 2.8 s after its own.** This was never three uniformly slow cold boots — it is one pane losing a race the other two won.
2. **`alpha` was abandoned 29.9 s after its first write**, not 60 s. So ~30.1 s of the deadline had already been spent *before* the write.

That 30.1 s is the readiness gate, and in this scenario it **cannot** be satisfied:

- `state::wait_for_session_start` deliberately **skips** a `wrapper_fork`-origin `SessionStart` (`session_start_means_ready`) and holds out for the agent's *native* one;
- /015's bootstrap launcher announces exactly that launcher-origin event, then blocks on `read` and only `exec`s Claude once the write arrives;
- so Claude cannot emit a native `SessionStart` until the gate has already timed out.

The test was paying the full production 30 s for an event that, by its own construction, could not arrive — and paying it out of the same 60 s budget the real agent then had to boot and confirm inside. Its deterministic sibling `scheduler/dispatch/014` already pins that gate to 3 s and says why; /015 never did.

## The fix

**1. Pin the readiness gate (`tests/e2e_prompt_delivery_confirmation.rs`).** `DOT_AGENT_DECK_SESSION_START_WAIT_MS = READINESS_GATE_MS` (3 s, the constant /014 already uses). This is the substantive change: it returns ~27 s of the delivery budget to the retry window the real agent actually gets, which is what production spends it on — a native `SessionStart` normally releases the gate in milliseconds. It changes no deadline, no production constant and no assertion; `AUTOMATIC_PROMPT_DEADLINE` is untouched, and the override is clamped by `src/spawn.rs` so it can only ever *shorten* the wait.

**2. Bound the concurrency (`.config/nextest.toml`)** — the shape the issue asks for when simultaneity is the property, which it is here (see below):

```toml
[[profile.default.overrides]]
filter = 'test(dispatch_015_three_real_claude_seeds_are_genuinely_confirmed)'
threads-required = "num-test-threads"
priority = 50
retries = 1
slow-timeout = { period = "60s", terminate-after = 4 }
```

- `threads-required` reserves the whole pool, so nothing else in the tier runs alongside — the mechanism #415 and #502 both already name for their class B ("nextest test-groups or resource tokens for the PTY-contending suites"), applied to one test rather than to the tier.
- `priority` is load-bearing with it, not decoration: reached mid-run, an exclusive test makes nextest drain every running test before it can take the pool, and this tier's tail holds 250–320 s tests. Taken at run start the reservation is free. Verified — the test reports as `(1/8787)` and nothing else completes during it.
- `retries = 1` is the one piece I would not argue hard for, and it is the piece with the most measured justification: with the exclusive reservation in place and **nothing else in the tier running**, the test still failed once, because three sibling agent worktrees on this box were mid-`rustc` and the 15-minute load average was 58 on 16 cores. `threads-required` structurally cannot see load that is not the tier's. A genuine regression fails both attempts, nextest reports a recovered run as FLAKY so a degrading machine stays visible, and this test is already in the flaky-tolerant pre-PR real-agent tier (rule 4) that runs in no CI job. It is doing real work rather than sitting idle — the sampling below caught runs that failed once and recovered — but it is also the piece I would drop first if you would rather not have it, since it is the only part of this PR that does not make the test itself better.
- `slow-timeout` is widened for the reason the five existing overrides in that file were, and is explicitly **not** a deadline extension: the 150 s assertion budget is untouched. The *failing* path measured 174.1 s against the default 3 × 60 s kill — a 6 s margin — and being killed there costs exactly the per-pane diagnostics this investigation ran on.

**3. Keep the evidence (`tests/e2e_prompt_delivery_confirmation.rs`).** The deck log is captured (`DOT_AGENT_DECK_LOG`, as /014 already did) and its delivery-lifecycle lines — `written` / `re-submitting` / `confirmed` / **`abandoning`** / `stopped`, each with its `delivery_id` and attempt count — print with the assertion. Previously the failure read only as `confirmed_exact=None`, which cannot distinguish "the retry path regressed" (the regression this test exists to catch) from "the machine missed the window". A recurrence should now be readable from the panic instead of reconstructed.

## Which of the issue's two shapes applies

**Simultaneity is the property, so the second one.** The catalog entry is titled "Three **concurrent** real interactive Claude dispatches…"; the module doc says the real scenario "repeats the reported three-dispatch Claude Code startup race"; and the deterministic sibling /014 exists to cover the concurrent race with stand-ins, delegating "real-agent boot behavior" to /015. Staging /015's seeds would leave that half covered by nothing — and would not obviously have worked either, since it cuts only the test's *own* peak load while each seed's window is still shared with everything else.

## Results

| Run | dispatch_015 | Tier |
|---|---|---|
| base `e8c443a`, full tier | **FAIL** 174.1s | 386.4s, 3 failed |
| override only, full tier (load avg 58) | **FAIL** 152.8s, as `1/8787` with the pool reserved | 496.2s, 2 failed |
| **this branch, full tier** | **PASS 6.5s**, first attempt, no retry consumed | 357.7s |

6.5 s is not a typo and not a weakened test: 27 of the old 32.2 s standalone seconds were the unsatisfiable readiness gate. Every assertion still runs against three real interactive Haiku agents, and the exclusive reservation now costs the tier ~6 s at run start instead of ~35 s.

Four other tests were red in that tier run — `mode_live_002` (a cursor-paint race, green in both earlier runs today), `shell_activity_006` (the documented ~1-in-3 model-precondition canary, #502), a `delegate_015` timeout, and `orchestration_dispatch_002`, which was red on the base run too and is filed separately. None of them touch anything this PR changes, and the box was carrying heavy external load throughout (below).

## Honest reliability picture, and the residual

I sampled `dispatch_015` on its own repeatedly rather than declaring victory on one green tier, and the result is worth stating plainly:

| Load average (16 cores) | Result |
|---|---|
| ≤ 4 | **5 / 5 pass**, 5.6–6.3 s each |
| ~18–42 | intermittent: some pass, some recover on the retry, some fail both attempts |

That is not residual noise, it has a specific cause, and the diagnostics this PR adds are what found it. Only attempts 1 and 2 of any delivery carry the payload (`MAX_PAYLOAD_SUBMISSIONS`); attempt 1 is eaten by the swallowing launcher by construction; and attempt 2 lands **~1.65 s** after the launcher `exec`s Claude. Every attempt after that is a bare submit probe. So:

- Claude reaches its input handling inside ~1.65 s → attempt 2 confirms within ~20 ms (measured repeatedly: `gamma` at 17 ms, `beta` at 12 ms).
- Claude takes longer → the payload is destroyed rather than delayed, because the raw-mode switch (`tcsetattr` with `TCSAFLUSH`) discards pending input, and the delivery then spends its remaining ~55 s pressing Enter on an empty box before abandoning.

That 1.65 s is the entire tolerance, which is why the test tracks machine load so sharply — and why neither deadline in play, the test's 150 s or production's 60 s, was ever the constraint.

**I have not fixed that here, deliberately.** It changes prompt-delivery retry semantics on a path #424 tuned carefully (a blind third payload write is what produced the `seedseed` accumulation), it wants its own review, and v0.37.2 is being held. Filed as **#666** with the measurement and a suggested direction: arm one additional payload attempt from a **post-write native `SessionStart`** — positive evidence that the agent started after the write, and therefore the one case where re-writing is correct rather than duplicative — instead of from attempt count. The deck already observes that event to bind the generation; it just does not use it for this.

So what this PR delivers: the test is ~5× faster, no longer throws away half its delivery budget, no longer competes with the tier, says what went wrong when it fails, and is green on a machine doing normal work. What it does not deliver is immunity on a machine at 2–3× oversubscription; #666 is what would give that.

## Scope

Local, per the issue. No global change to the tier's parallelism — that is #502's question, and `threads-required` is a per-test declaration that composes, so whatever #502 lands can subsume this override rather than fight it. Worth noting there: the *general* lesson from this one is not "reserve threads", it is "check whether the test is paying a production wait it cannot satisfy" — that was worth 27 seconds and the whole failure here, and nothing about it is specific to /015.

## Verification

- `cargo fmt --check` — clean
- `cargo xtask linkage-check` — clean
- `cargo test-fast` — 3062 passed
- full `cargo test-e2e` — the real gate here, since a `-j 1` pass proves nothing. Numbers above. Note the box carried substantial *external* load throughout (sibling agent worktrees building; 15-minute load averages between 28 and 80 on 16 cores), which is why the tier wall clocks are not comparable run to run — and also why the run that mattered is a fairly demanding one to have passed.

## Rule 12

Test and test-configuration only; nothing under `src/` changed. No wire shape moved and no field changed meaning, so `PROTOCOL_VERSION` is untouched and no `.breaking.md` fragment is warranted. The one production knob the test now sets, `DOT_AGENT_DECK_SESSION_START_WAIT_MS`, is an existing documented test override that `src/spawn.rs` clamps to `[100 ms, production default]`.

https://claude.ai/code/session_01JXcRhVKiJMzoq2pZdzqb3o


prageethw (approved):
Approving. Read the whole diff against `271e83d`.

**The diagnosis is the valuable part and it holds up.** I checked the claim the rest of the change rests on: `session_start_wait_timeout()` in `src/spawn.rs` clamps `DOT_AGENT_DECK_SESSION_START_WAIT_MS` to `[SESSION_START_WAIT_MIN, SESSION_START_WAIT_TIMEOUT]`, so the pin can only ever *shorten* the gate — it cannot buy the test time, which is what makes this a budget correction rather than a deadline extension. And the scenario genuinely cannot satisfy the gate: the launcher announces a `wrapper_fork`-origin `SessionStart`, `wait_for_session_start` skips exactly that origin, and the launcher only `exec`s Claude after the write it is blocked reading. Spending 30 of a 60s `AUTOMATIC_PROMPT_DEADLINE` on an event that cannot arrive is a defect in the test's own setup, and `scheduler/dispatch/014` already pinned the same constant for the same reason. No assertion moved.

Also checked the mechanical bits:

- **The nextest override is not shadowed.** No earlier `[[profile.default.overrides]]` block in `.config/nextest.toml` has a filter matching `dispatch_015_…`, so `threads-required`, `priority`, `retries` and `slow-timeout` all land, and the per-test `retries = 1` correctly overrides the profile-level `retries = 0`. No other override sets `priority`, so 50 really is scheduled first.
- **The five markers in `delivery_log_evidence` all match live source.** Each of `prompt written to pane; provisional`, `… unconfirmed; re-submitting`, `… confirmed by the agent`, `… unconfirmed at the deadline; abandoning` and `… stopped without confirmation` is present in `src/prompt_delivery.rs` today, so the function will not silently print `<no delivery lifecycle lines>`.

`retries = 1` is the one thing I'd normally push back on, and the justification answers it: this test is already in the flaky-tolerant pre-PR real-agent tier and runs in no CI job, a genuine regression fails both attempts, and nextest surfaces a recovered run as FLAKY rather than absorbing it. The measured load-average-58 case is a fair thing to concede a second attempt to given `threads-required` structurally cannot bound load that isn't the tier's.

One non-blocking note: `delivery_log_evidence`'s `MARKERS` re-states three literals that `delivery_log_states` already carries, so a wording change in `src/prompt_delivery.rs` now degrades two functions silently and in different ways — the state map just stops classifying, and the evidence dump prints its empty placeholder at precisely the moment you need it. Hoisting the five strings to file-level consts that both read (or, better, exporting them from `prompt_delivery.rs`) would make the drift a compile-time concern instead of an invisible one. Not worth blocking on, but this diff is the one that doubles the exposure.

Issue #670, opened 2026-08-24
display_name from event.metadata reaches the dashboard render unsanitized and unclamped


prageethw:
## Problem

`AppState::apply_event` (`src/state.rs`, around line 5671 on current `main`) ingests `display_name` from `event.metadata[DISPLAY_NAME_METADATA_KEY]` with only an `.filter(|n| !n.is_empty())` guard:

```rust
if let Some(name) = event
    .metadata
    .get(DISPLAY_NAME_METADATA_KEY)
    .filter(|n| !n.is_empty())
{
    session.display_name = Some(name.clone());
}
```

No sanitization (control characters, Unicode bidi-override characters like U+202E) and no length clamp are applied before this value is stored and later rendered on the dashboard card. Any process able to post to the hook socket (any agent on the deck) can set an arbitrary, unbounded, unsanitized string here — including terminal control sequences and bidi overrides that can visually reorder or hide adjacent rendered text, or an unbounded string that could affect card layout.

The only thing currently preventing exploitation is `ratatui`'s own incidental grapheme handling at render time — a third-party crate behavior this project doesn't own or test, not a guarantee.

## Fix (already applied on a fork)

Sanitize (escape control/bidi characters) and length-clamp `display_name` at ingest, the same way other untrusted metadata fields arriving through this path are handled. Reference implementation, full review, and audit trail: https://github.com/prageethw/dot-agent-deck/pull/558 (issue https://github.com/prageethw/dot-agent-deck/issues/410).

Note the fork's fix has a known, deliberate scope limit: it covers `SessionState.display_name` (this ingest path / the dashboard card's fallback source) but not `ui.display_names` (the card's *preferred* source, hydrated separately from `AgentRecord.display_name` on a rename/spawn path) — that's tracked as a separate fork-side follow-up (https://github.com/prageethw/dot-agent-deck/issues/562) since it's a different code path with its own audit needed.

## Why this is filed here rather than fixed on the fork alone

Verified against `upstream/main`: the exact ingest code (`event.metadata.get(DISPLAY_NAME_METADATA_KEY).filter(|n| !n.is_empty())`, `src/state.rs`) is present, unchanged from what the fork had before its own fix — this is a defect in shared/unmodified upstream code, not fork-specific behavior, so per this fork's contribution policy it's filed upstream at discovery time.

PR #675, opened 2026-08-24
fix(dispatch): ask the unit shape per unit, and surface a stale worktree base (#674)

### LINKED ISSUE #674: Dispatch queue skills ask the unit shape once per batch (and pr-review-queue never asks), and neither ensures the worktree is cut from a current base
## Problem

Two independent gaps in the dispatch queue skills, both found while dispatching a batch of three issues on 2026-08-24. Filing together because they share a cause: the guidance is written for a single unit, while the skills' own recommended path is a batch of 2–3.

### 1. The shape is asked once per batch, not once per unit

`.claude/skills/issue-queue/SKILL.md` step 7 asks which shape the runner wants in the singular — *"ask which shape they want"* — and then says *"Reuse the answer for later dispatches in the same conversation; re-ask when the runner changes it or a unit is clearly different in kind."* The escape hatch is real but buried at the end of that sentence, and it frames per-unit asking as a later exception rather than the default.

That is wrong by construction for the path the skill most expects. Step 5 explicitly recommends dispatching **2–3** units, and a queue of 2–3 issues routinely mixes kinds. Measured on the 2026-08-24 batch: #669 (an `lstat` guard, roughly ten lines in one function, with a reference implementation already on a fork) and #668 (an audit of every harness spawn path #661 does not reach, plus a reaping mechanism and coverage) are not the same kind of work. Asked as one question the runner got one answer for all three; asked per issue they chose `--single` for #669 and #670 and `--orchestration` for #668. The batch-level question produced an answer the runner did not actually want, which is exactly the outcome step 7 exists to prevent.

`.claude/skills/pr-review-queue/SKILL.md` has the stronger version of the same bug: **it never asks at all.** The file contains zero occurrences of `--list-targets`, has no shape step among its ten `##` headings, and hardcodes the flag in its dispatch template at line 213:

```bash
dot-agent-deck dispatch <name> --single --task-file '.dot-agent-deck/<task-slug>.md'
```

`docs/dispatcher-mode.md:41` is where the singular framing originates, and it is the product's contract rather than skill guidance: *"Which one is your call, not the agent's, so it asks you before the first unit rather than guessing."* "Before the first unit" is what both skills are faithfully implementing.

### 2. Nothing ensures the dispatched worktree is cut from a current base

`dot-agent-deck dispatch --help` confirms there is **no** base or branch option: the worktree is cut from whatever `HEAD` is at dispatch time. So the freshness of the runner's checkout is the only lever, and neither skill pulls it.

`issue-queue` step 0 is explicit in the other direction:

> Do **not** `git pull` to fix this. The runner may have local work, and this skill has no business moving their branch.

That rule is correct **for what it governs** — it is about *verification*, and verifying against `origin/main` instead of pulling is the right answer there. But it silently leaves the *base* stale, and nothing downstream revisits it.

Measured on the same batch: local `main` sat at `f3752e1`, six commits behind `origin/main` at `c701932`. All three worktrees were cut at `f3752e1`. One of those six commits changed `src/state.rs` by 58 lines — the exact file #670 targets, at the exact ingest site it is about. Selection was still sound because every claim had been verified against `origin/main`, but all three units started on a base that was already wrong, and each has to discover and fix that itself.

## Proposed fix

- **`issue-queue` step 7**: ask the shape **per unit**, not per batch, and present the runner with enough per-issue scope to answer. Keep reuse as the explicit *fallback* for a homogeneous batch rather than the default.
- **`pr-review-queue`**: add the shape step it has no version of, and remove the hardcoded `--single` from the line 213 template.
- **Base freshness**: add a step, in both skills, that surfaces a stale checkout **before** dispatching and asks the runner what to do about it — since `dispatch` cannot take a base and the fix means moving their branch, this must be their explicit decision, never a unilateral `git pull`. Take care to word this so it reconciles with step 0 rather than appearing to contradict it: step 0 governs verification, the new step governs the worktree base.
- **`docs/dispatcher-mode.md:41`**: reconsider "before the first unit". This is the product contract, so it is the widest part of the change and worth deciding deliberately rather than editing to match the skills.

## Notes

Both skills are project-local (no `dot-ai-` prefix), so they are ours to edit under CLAUDE.md rule 13 and will not be clobbered by a skill sync.

vfarcic:
> [!IMPORTANT]
> **Update 2026-09-03 — part 2's SKILL-STEP half is superseded and has been removed from this PR. Its CODE half stands unchanged.**
>
> #761 merged on 2026-08-31 and reversed the base-staleness policy in the same two skill files: its step 0 **fast-forwards** the checkout and reports it, where this PR's step 5b / step 2b surfaced the staleness and asked, never pulling. #761 landed first — the reverse of what [my 2026-08-30 comment](https://github.com/vfarcic/dot-agent-deck/pull/675#issuecomment-5471551812) assumed — and left an explicit instruction for this merge in the file it modified: keep what #674 says about `dispatch` naming the base in its own success line, *"and drop its instruction to surface and ask, which is what this step replaces."*
>
> So the merge of `origin/main` (`5a0c4dc`) **dropped** step 5b (`/issue-queue`), step 2b (`/pr-review-queue`) and the step 0 addendum below, and renamed the surviving `Step 2c` → `Step 2b`. Read the two subsections below — *"Nothing ensured the worktree base was current"* and *"Reconciling with step 0, not contradicting it"* — as the history of why the code half exists, **not** as a description of the current diff.
>
> **Unaffected, and still the substance of this PR:** part 1 (the per-unit shape question) in full, part 3 (`docs/dispatcher-mode.md` + `src/ui.rs`) in full, and *"`dispatch` now names the base it cut from"* in full — `describe_dispatch_base`, the `cut from` clause and its three unit tests, which #761 depends on and does not duplicate. The five non-skill files are byte-identical to the twice-approved diff; the diff went +257/−14 → +203/−14, shrinking only by superseded prose.
>
> The gate results quoted at the bottom of this body are from the original runs. Current results are in the re-request comment below.

---

Fixes #674

Two gaps in the dispatch queue skills, filed together because they share a cause: the guidance was written for a single unit while the skills' own recommended path is a batch of two or three.

## 1. The shape is asked per unit

`/issue-queue` asked in the singular and told the runner to reuse the answer, with per-unit asking buried as a late exception. `/pr-review-queue` never asked at all — no shape step among its headings, zero occurrences of `--list-targets`, and a hardcoded `--single` in its dispatch template.

Both now run `--list-targets` once (its answer describes the repo, not the unit) and put the shape question once per unit, as **one prompt with a line per unit** rather than N round trips — with enough per-unit scope next to it to answer. One answer still covers a homogeneous batch, as the runner's answer rather than an assumption.

**Also fixed, not in the issue:** `/issue-queue`'s own step 8 template hardcoded `--single` too, quietly undoing its step 7 for anyone who copied the line. Both templates now read `(--single | --orchestration '<orchestration>')`, with a note that the flag is the runner's answer and a template is the one place it cannot be stored.

## 2. Nothing ensured the worktree base was current

`dispatch` takes no base or branch option — `create_worktree` runs `git worktree add <dir> -b <branch>` with **no start-point**, which git resolves to `HEAD`. Both skills now read the branch `HEAD` is on *and* its distance from `origin/main` before dispatching, show the runner, and ask. Neither pulls.

The consequences differ per skill and each says its own:

- **`/issue-queue`** — the batch on 2026-08-24 had local `main` six commits behind `origin/main`, all three units cut there. One missed commit (`a2a5953`) rewrote 58 lines of `AppState::apply_event`, the function #670 was dispatched to fix, moving the ingest line that issue names from `src/state.rs:5671` to `5707`.
- **`/pr-review-queue`** — a stale base does **not** corrupt the verdict: `/verify-pr`'s `setup.sh` builds its `../<repo>-pr-<n>` checkout from a fresh `refs/pull/<n>/head` fetch and merges a freshly-fetched `origin/<default>`. What it corrupts is the *instructions* — the copy of `CLAUDE.md` whose gates the unit runs and the copy of `.claude/skills/verify-pr/` it executes. Measured over the 30 days to 2026-08-24: **21** non-merge commits touched `CLAUDE.md`, **13** touched `.claude/skills/verify-pr/`.

Both also carry a hazard the issue does not: **`HEAD` is not necessarily `main`.** Dispatching from a pane on a feature branch cuts every unit from that branch. Verified rather than inferred — `git worktree add <dir> -b probe` with `HEAD` on a feature branch put `probe` at the feature commit.

### Reconciling with step 0, not contradicting it

`/issue-queue` step 0 forbids `git pull`, and that stands unchanged. What it gains is a sentence saying it governs *verification*, which is all it ever governed: verifying against `origin/main` needs no pull, so refusing one there costs nothing. The new step governs the *worktree base*, where the same constraint yields a different answer — because fixing it means moving the runner's branch, it has to be the runner's explicit decision, so the step surfaces and asks and never pulls.

### `dispatch` now names the base it cut from

Added after review discussion: the skill check above is **pre-flight** — it exists so the runner can update the branch *before* three worktrees are cut — but nothing recorded where a unit actually started. `dispatch`'s success line now ends `…, cut from main at c701932`, naming the branch the caller was standing on and the commit it resolved to.

The two are complementary, not redundant, and both skills now say so: the pre-flight check is the only thing that can *change* the answer, and the dispatch line is the record the report step quotes instead of recomputing.

Deliberately best-effort. The probe runs before the worktree exists (the state `git worktree add` resolves its absent start-point against), and a repository that cannot answer drops the clause rather than failing a dispatch that has already created the worktree and spawned the agent. Both skills state the consequence explicitly: **a success line with no `cut from` clause means the probe failed, not that the base is fine.** A detached checkout reports `detached HEAD at <sha>` rather than a branch named `HEAD`, which `rev-parse --abbrev-ref` would otherwise hand you.

Three new unit tests cover the branch case, the detached case, and the not-a-repo degradation. The branch test creates its branch explicitly rather than trusting `git init` — `init.defaultBranch` is ambient user config, so asserting on `main` would pass or fail with the machine.

**What was deliberately left out:** a `--base` flag, and a warn-or-refuse when `HEAD` is behind its upstream. Both are preventive rather than descriptive, and both are design questions with real trade-offs — a fetch inside the dispatch path means network I/O and offline behaviour; a `--base` flag adds a way to cut a unit from a tree the user has not checked out, against dispatcher-mode's "your last commit" model. Neither is asked for by #674 and neither has evidence behind it yet.

## 3. `docs/dispatcher-mode.md:41` — the deliberate decision, and why it grew

**The contract was in four places, not one.** `docs/dispatcher-mode.md:41` is a faithful description of `DISPATCHER_SEED_PROMPT` (`src/ui.rs`), the seed compiled into every dispatcher pane, which said *"before the FIRST dispatch of a session"* and *"Reuse the answer for later dispatches in the same conversation rather than asking again."* The issue does not record this, and it changes the decision: editing only the doc would document a promise the product does not keep, and both queue skills run **inside a pane holding that seed**, so a skill saying "ask per unit" would contradict text sitting in the same context. @vfarcic confirmed moving all four together.

**The change, and the argument for it.** The load-bearing half of line 41 is *"your call, not the agent's"*. *"before the first unit"* is a mechanism clause, and on a mixed batch it delivers the opposite of the promise — units 2..N get a shape the user never chose, which is precisely "the agent's call". The doc's own two examples are what settle it: *"work on these three features"* → a team per feature, *"verify these three PRs"* → one agent each. Both establish that the shape follows from **what the unit does**, and both are homogeneous, which is exactly why asking once looked sufficient. The page now says it asks once per unit, keeps both examples, and adds that three of a kind take one answer while a mixed batch is where a single question goes wrong.

The cost is one extra line in one prompt. The measured cost of not doing it is an agent started in a shape the runner did not want.

**Also in that file, and also user-facing:** the page said a unit's copy is made from *"your last commit"*. It now says *"your last commit **on the branch you are currently on**"*, and that there is no way to point a unit anywhere else — the user-facing half of part 2, in the one place a user would look.

**`src/ui.rs`.** Two sentences of the seed, plus a new unit test pinning the per-unit framing in both directions. The pin was **mutation-checked**, not assumed: reinstating `Reuse the answer` in the seed fails `dispatcher_seed_asks_the_shape_once_per_unit` with the expected message. The existing `dispatcher_seed_teaches_mechanics_not_work_methodology` pin still passes untouched.

## Gates

- `cargo test-fast` — **3067 passed, 0 skipped** (3 new). Not inert for the skill half either: `xtask/linkage-check` reads files under `.claude/skills/`.
- `cargo test-e2e` — **run twice**, once per commit, because the diff is not Markdown-only. Neither run is clean and neither failure tracks this change; detail below rather than a "known flakes" hand-wave.

### The e2e failures, and why none of them is this change

**Run 2** (with `src/dispatch.rs`): 8792 tests, **8791 passed, 1 failed** — `codex_wrap_001_synthetic_jsonl_reaches_dashboard`, which **passes in isolation** (15.1s, against 28.0s under full-tier load). It exercises the codex wrapper's JSONL ingestion, which this PR does not touch. All three of run 1's failures passed in this run.

**Run 1** (before `src/dispatch.rs`): 8789 tests, 3 failed, all real-agent:

| Test | Full tier | Rerun in isolation |
|---|---|---|
| `dispatch_013_orchestration_surfaces_and_delegates` | FAIL | **PASS** |
| `shell_activity_006_real_claude_bash_call_crossing_the_cap_keeps_the_badge_working` | FAIL | **PASS** |
| `devin_live_001_real_interactive_turn_drives_the_card_live` | FAIL | FAIL |

`shell_activity_006` diagnosed itself in its own panic — *"the model never issued the Bash tool call at all. This is a harness flake (prompt adherence), NOT a badge regression"*.

`devin_live_001` failed twice, so I did not stop at "real agent, must be flaky". At that point `src/ui.rs` was the only compiled file in the diff, so I reverted **just that file** to `origin/main`, rebuilt, and reran: it **still failed** — and at a different point (`did not see "[Command Mode Ctrl+D]" within 10s`, 26.7s) than on the branch (`no broadcast AgentEvent matched the predicate within 45s`, 71.5s). Three runs, three different failure points, identical with and without the change: pre-existing, real-service, not caused here. The tree was restored to `HEAD` afterwards and verified clean. It then passed in run 2.

The overall picture is a tier with load-sensitive flakes, not a regression: across two full runs the failing set was disjoint, and every failure passed on an isolated rerun except the one proven pre-existing.

Corroborating for the compiled changes specifically: `tests/e2e_dispatcher_mode.rs` — the test that actually exercises the seed constant — passed in both runs, and `dispatch_013`, the real-agent dispatch test, passed in run 2 with the new base probe in place.

**Triggered by path, and answered: no contract change. No `PROTOCOL_VERSION` bump, no `.breaking.md`.** `scan.sh` classifies `src/dispatch.rs` as `PROTOCOL`, so `RULE_12_TRIGGERED=true` — correctly, and it deserves a real answer rather than a wave.

The answer is that **the dispatch result message never crosses the TUI↔daemon wire.** `DispatchResult` has no reference anywhere outside `src/dispatch.rs`; the daemon consumes it in-process and delivers the text by `pty_registry.write_to_pane_and_submit(&signal.pane_id, &result.message)` (`src/daemon.rs:1862`) — it is typed into the calling pane's PTY, not serialised into a response. The CLI has already returned by then and never sees it. So there is no wire field whose shape could move and none whose meaning could shift behind a stable wire, which is the specific hazard rule 12 exists for. An older TUI against a newer daemon sees one longer sentence of prose in a pane.

The seed change is the same story by a different route: it reaches a pane as ordinary pane input via `process_pending_seed_prompts`, not as a typed protocol field.

**I did not run the manual cross-version sandbox**, and I am flagging that as a judgment rather than burying it: the procedure exists to catch a semantic break behind a stable wire, and there is no wire here to be behind. If you would rather see it run before merging, say so and I will.

Fragment filed as `changelog.d/674.bugfix.md` (patch bump).

## Rule 4 — TUI test tier

**No tier required, and worth stating because the trigger fires anyway.** `scan.sh` is path-based, so `src/ui.rs` sets `RULE_4_TRIGGERED=true` (`src/dispatch.rs` is `PROTOCOL`, not `UI`) — but the change there is a text constant handed to a spawned agent plus a unit test. No pane, status, prompt, focus, layout, mode, keybinding or embedded-pane behaviour changes, and no snapshot moved (3064 fast-tier tests, zero `.snap` failures). The one L1 test that covers the constant is extended.

## For the reviewer

`.claude/**` classifies as `EXEC_ON_CLONE` — these files are instructions agents read and act on, so they are worth reading in full rather than skimmed. The two skills are project-local (no `dot-ai-` prefix), so CLAUDE.md rule 13 permits editing them and a skill sync will not clobber the change. Rule 10 respected throughout: every new prose paragraph is a single unwrapped line, and no existing paragraph was reflowed.


prageethw (dismissed):
Traced describe_dispatch_base in src/dispatch.rs: it's captured before create_worktree runs (matching git's actual start-point resolution), correctly distinguishes detached HEAD from a named branch, and degrades to None rather than failing the dispatch when the probe can't answer. Confirmed run_capture_args returns Err on non-zero exit / spawn failure so .ok()? correctly yields None for a non-repo directory, matching the test's expectation. The per-unit shape question fix for issue-queue/pr-review-queue is narrowly scoped and correctly closes the gap where pr-review-queue previously hardcoded --single. Three new unit tests cover branch/detached-HEAD/non-repo cases against the actual helpers, and the doc/skill/changelog prose stays consistent with the code (including reconciling with the existing 'never git pull' rule). All CI checks green. LGTM.

vfarcic:
Merged current `main` into the branch to clear `mergeable=CONFLICTING`. Merge rather than rebase, so the two reviewed commits (`dd7b7d7`, `5af428e`) keep their SHAs and the prior reading still maps onto the history.

**What conflicted:** exactly one hunk, in `src/dispatch.rs`. `src/issue_dispatch_run.rs` and `src/ui.rs` auto-merged. The two sides never actually disagreed — they just landed on adjacent lines of the same `create_worktree` invocation:

- this branch (#674) inserted a base probe immediately *before* that call;
- `main` (#683 / #425) gave the call a fifth `Creator` argument so the worktree ownership marker is written at creation time.

**Resolution — both kept, neither reworded:**

```rust
// #674: read before the worktree exists, which is the state
// `git worktree add` resolves its (absent) start-point against
let base = describe_dispatch_base(&clone_dir).await;

match create_worktree(
    &clone_dir, &paths.worktree_dir, &paths.branch, false,
    Creator::dispatch(name),   // #425: writes the ownership marker
).await
```

The ordering #674 requires is intact: the probe still runs before the worktree is created. Verified after merging that all **8** `create_worktree` call sites carry a `Creator`, and that the `cut from` clause and its three unit tests survived.

**Runtime evidence both behaviours coexist**, from real e2e agent history — a pre-merge build vs. this one:

```
...in /var/tmp/.../.tmpM9h0T5-dispatch-probe-unit
...in /var/tmp/.../.tmp3uStiD-dispatch-probe-unit, cut from master at 191f7f7
```

**No scope creep:** the diff against `main` is +267/−22 — byte-identical to this PR's diffstat before the merge.

**Gates:** `cargo fmt --check` and `cargo clippy --workspace --all-targets --features e2e -- -D warnings` clean; `cargo test-fast` 3196/3196; `cargo test-e2e` run twice. Being straight about the e2e tier: neither full run was all-green, but the two runs' failures were **disjoint** (13 scheduler/devin tests, then a different 2), every one passed on rerun, and none was in the dispatch path — the load-sensitive flakiness rule 5 describes. The 165 dispatch/worktree tests passed as a set, including the real-agent ones that genuinely spawn into a worktree.

All four required checks green on `a3c0543` (`build`, `build-macos`, `build-windows`, `security`), plus the other six.

Note this push dismissed the existing approval under `dismiss_stale_reviews_on_push` — unavoidable, since a conflicted PR cannot merge. Not merging and not re-requesting review; that call is @prageethw's.

vfarcic:
Re-requesting review — @prageethw, your approval here was **dismissed mechanically, not substantively**.

You approved at 2026-08-25T13:15:14Z, and the dismissal commit was \`a3c0543\` (\"Merge origin/main into agent/dispatch-issue-674\") — a merge that brought `main` in and changed nothing in this PR's own diff. `dismiss_stale_reviews_on_push: true` cannot tell that apart from new work, so the approval was voided. No substantive commit landed after your review.

State now: all CI green, no unresolved review threads, mergeable. Auto-merge is armed, so it lands the moment you re-approve — nothing else is outstanding.

I did not merge it directly. The admin bypass would have let me, and CLAUDE.md rule 8 is explicit that doing so is how the gate decays into ceremony.

prageethw (dismissed):
Re-confirming prior review: the only commit since my last approval is a no-op merge of origin/main — this PR's own diff (dispatch.rs, issue_dispatch_run.rs, ui.rs, skill docs, changelog) is unchanged from what I already reviewed and approved. LGTM, all CI green.

vfarcic:
**Relationship to #761 — recording it here so it isn't rediscovered at merge time.**

#761 (`agent/dispatch-issue-760`) rewrites Step 0 of the same two skill files to the opposite policy: this PR's Step 5b / Step 2b say *surface the stale base and ask, never `git pull` on your own*; #761 says *fast-forward it and say you did, no prompt, no question*.

They conflict only trivially — two blocks of 9 and 8 lines, both the trailing reporting bullet. The rest auto-merges, so whichever lands second leaves both policies sitting in the same file with no conflict marker.

#761 is written expecting this PR to land first and says so, so **nothing needs changing here** — the reconciliation belongs on #761, where the details are recorded.

vfarcic:
@prageethw — re-requesting review. **Your two approvals were dismissed mechanically, by merge-`main` pushes, not by any change to the diff you read.**

- **2026-08-25T13:15:14Z** — dismissed by `a3c0543`, a merge of `origin/main`.
- **2026-08-27T12:56:31Z** — dismissed by `5a0c4dc`, the merge of `origin/main` in this push.

The PR went `mergeable=CONFLICTING` again because `main` moved, and clearing that requires a push, which voids the approval. **The only new commit is `5a0c4dc`, the conflict resolution.** No substantive work landed after either review.

**You can re-approve by checking the resolution rather than re-reading the PR** — here is exactly what it did.

## The five properties you enumerated on 2026-08-25 all still hold

The five non-skill files are **byte-identical** to the diff you approved — verified by comparing the `+`/`-` payload of the pre-merge and post-merge diffs per file, not by eye:

| file | changed lines | vs. your review |
| --- | --- | --- |
| `src/dispatch.rs` | 131 | identical |
| `src/issue_dispatch_run.rs` | 2 | identical |
| `src/ui.rs` | 33 | identical |
| `docs/dispatcher-mode.md` | 8 | identical |
| `changelog.d/674.bugfix.md` | 9 | identical |

Re-checked in the **merged** tree, not just in the diff:

1. **`describe_dispatch_base` captured before `create_worktree`** — `src/dispatch.rs:372`, still immediately above the `create_worktree(` call at `:374`, which now carries `main`'s fifth `Creator::dispatch(name)` argument.
2. **Detached HEAD distinguished from a named branch** — the `if head == "HEAD"` arm returning `detached HEAD at {sha}` survives.
3. **Degrades to `None` rather than failing the dispatch** — both `.ok()?` early-returns and the empty-string guard intact; the clause is *appended* (`Some(base) => format!("{opened}, cut from {base}"), None => opened`), so a failed probe still yields the pre-#674 sentence with no dangling `cut from `.
4. **`run_capture_args` returns `Err` on non-zero exit** — unchanged at `src/issue_dispatch_run.rs:1234`, so `.ok()?` still yields `None` for a non-repo directory.
5. **The three unit tests** — `dispatch_base_names_the_branch_and_commit_it_was_cut_from`, `dispatch_base_says_detached_rather_than_naming_a_branch_head`, `dispatch_base_is_absent_rather_than_empty_outside_a_repo`, all present and passing.

## Two marked conflicts, both resolved to `main`'s side

Both were the trailing base-reporting bullet — `issue-queue` step 9 and `pr-review-queue` step 8. `main`'s version already carries this branch's contribution (it quotes `…, cut from main at c701932` and reads a missing clause as a failed probe) **and** adds the distance from `origin/main` that its own step 0 establishes, which this branch's wording could not name. `main`'s side is strictly the better of the two.

## The part that needs your judgment: I removed prose you approved

**#761 merged on 2026-08-31 — after your review — and reversed the base-staleness policy in the same two skill files.** Its step 0 fast-forwards the checkout and reports it; this branch's step 5b / step 2b surfaced the staleness and asked, explicitly never pulling. Those blocks **auto-merged with no conflict marker**, so the merged files carried both policies at once — the outcome [my 2026-08-30 comment](https://github.com/vfarcic/dot-agent-deck/pull/675#issuecomment-5471551812) predicted, except **#761 landed first rather than second**, which is the reverse of what that comment assumed.

#761 anticipated this merge and left the instruction in the file it modified (`issue-queue` step 0, currently on `main`):

> **If a later step in this file also checks the base, this one supersedes the deciding half of it.** Issue #674 added such a step, written when surfacing was the policy; keep what it says about `dispatch` naming the base in its own success line […] and drop its instruction to surface and ask, which is what this step replaces.

Followed exactly:

- **dropped** step 5b (`issue-queue`), step 2b (`pr-review-queue`), and the step 0 addendum that was anchored to a "do not `git pull`" rule #761 deleted;
- **kept** the retained half, which already lives in `main`'s step 9 / step 8 bullet;
- **renamed** the surviving `Step 2c` → `Step 2b`, a label the deletion freed.

Retaining step 5b would have restored precisely the rule #761 warns against restoring, in a file that would then have instructed the agent both to fast-forward without asking and to never pull on its own.

**Everything the PR is named for is untouched.** The per-unit shape question stands in full — and note that `main`'s own #705 subsection already says *"(Issue #674 is the change that made this explicit in the step above…)"* while the step above it still carries the pre-#674 per-session wording. `main` is currently self-contradictory there, and this merge is what resolves it. The `cut from` reporting code also stands in full; #761 depends on it and does not duplicate it.

Net effect on the diff: **+257/−14 → +203/−14**. It only shrank, and only by superseded prose.

## Gates

- `cargo clippy --workspace --all-targets --features e2e,e2e-live -- -D warnings` — clean, exit 0, all four flags.
- `cargo test-fast` — **3959/3959 passed**.
- This PR's own four unit tests — `dispatch_base_*` (3) + `dispatcher_seed_asks_the_shape_once_per_unit` — pass.
- Lane 1 e2e, dispatch scope — `binary(e2e_issue_dispatch) or binary(e2e_issue_dispatch_live)`: **227/227 passed**.
- CI `e2e-deterministic` is **red, and it is not this PR**: `orchestration_remit_001`–`007` in `tests/e2e_orchestration_remit.rs`, i.e. issue #818, which has been red on every branch since #789. `main` at `0153758` — the exact commit merged here — fails identically, and #837's fix touches only that file. Not a required check.
- Lane 2 (credentialed, `binary(e2e_dispatcher_mode) or binary(e2e_issue_dispatch_real)`) — **running locally at the time of writing; I will report the result in a follow-up rather than imply it passed.** Nothing in CI runs it (rule 5).

One thing I did **not** change and want to flag rather than fix quietly: `main`'s step 0 paragraph quoted above still reads as an instruction to a future merger, and after this merge it has been carried out. It is inert rather than wrong, it remains useful to anyone on an older build, and editing #761's rationale is not this PR's business — but it is the one loose end I noticed.

Not merging. Auto-merge not armed either — the removed prose is a real deviation from what you approved, so this should land on a deliberate approval rather than an armed one.

vfarcic:
**Lane 2 result, as promised above — it passed, and it genuinely ran.**

```
cargo test-e2e-live --exclude dot-agent-deck-desktop \
  -E 'binary(e2e_dispatcher_mode) or binary(e2e_issue_dispatch_real)'

Starting 224 tests across 2 binaries (132 binaries skipped)
Summary [63.942s] 224 tests run: 224 passed (1 slow), 0 skipped
```

**Checked that this is not a false green.** Rule 5 warns that an unsatisfiable preflight prints `SKIP:` and *returns normally*, so nextest counts it as passed. There are **zero** `SKIP:` lines in the run — the only matches for skip-ish text are five harness unit-test *names* containing the word "preflight" (`preflight_passes_when_free_space_is_above_the_threshold` and friends), in both binaries.

The eight non-harness tests all ran with real-agent durations rather than instant returns:

| test | time |
| --- | --- |
| `orchestration_dispatch_001_tab_surfaces_with_role_cards` | 63.8s |
| `orchestration_dispatch_002_every_real_agent_role_comes_alive` | 31.8s |
| `new_pane_016_dispatcher_opens_dashboard_card_with_real_agent` | 25.6s |
| `dispatch_013_orchestration_surfaces_and_delegates` (`e2e_issue_dispatch_real`) | — |
| `orchestration_dispatch_004_list_targets_marks_the_declared_default` | — |
| `dispatch_close_001` / `002` / `003` | — |

That covers the dispatcher-mode surface this PR changes, including `--list-targets` and the orchestration path, against real agents. Nothing in CI runs any of it (rule 5), so this is the only place it gets exercised.

No change to the diff — this is a test report, not a push, so your review request stands unaffected.

prageethw (approved):
Re-confirmed after the two merge commits since my last approval (37a6ac29, 5a0c4dcc): both are pure conflict resolutions merging origin/main, not new substance. The 5a0c4dc merge removed the pr-review-queue/issue-queue "surface and ask" prose that #761 superseded (per #761's own left-behind instruction) — the five non-skill files (dispatch.rs, issue_dispatch_run.rs, ui.rs, issue_dispatch_run.rs, docs/dispatcher-mode.md) and the per-unit shape logic remain byte-identical to what I already reviewed. All checks green except e2e-deterministic, whose sole failure (orchestration_remit_001_start_role_compaction_reasserts_remit) is the known remit-test race being fixed separately in #837 — unrelated to this PR's diff. LGTM.

Issue #127, opened 2026-05-25
PRD: Cron-scheduled prompt dispatch (general scheduler primitive)


vfarcic:
## PRD: Cron-scheduled prompt dispatch

**Problem**: Users want to run prompts on a cadence (e.g. "every weekday 09:00, generate a Barcelona weather + GitHub issues digest") and have the result land in the deck where they can read it after a notification. The deck has no scheduler — every periodic task is manual.

**Solution**: Add a small in-process scheduler with two orthogonal primitives: a **cron primitive** (registers a callback on a cron expression) and a **spawn primitive** (`(working_dir, prompt) → Tab/Card handle`, branches on the target dir's `.dot-agent-deck.toml` between orchestration tab and single-agent card). A scheduled task is "a cron whose callback calls spawn once." Issue-dispatch (#120) becomes a layer on top — a cron callback that calls spawn N times after enumerating issues. The scheduler PRD does not pre-bake N-spawn, worktrees, or GitHub-specific logic; those live in #120.

**Detailed PRD**: See [prds/127-cron-scheduled-prompt-dispatch.md](https://github.com/vfarcic/dot-agent-deck/blob/main/prds/127-cron-scheduled-prompt-dispatch.md)

**Depends on**: #126 (notifications — used for failure surfacing)
**Prerequisite for**: #120 (issue-dispatch — composes the primitives here with GitHub-specific logic)

**Priority**: Medium

PR #676, opened 2026-08-24
fix(fsperm): refuse a symlinked target in ensure_owner_only_dir (#669)

### LINKED ISSUE #669: ensure_owner_only_dir (Unix) follows a symlink at the target path instead of refusing it
## Problem

`ensure_owner_only_dir` in `src/platform/fsperm/unix.rs:50` is meant to guarantee a directory exists with owner-only (`0700`) permissions — it's the enforcement point for the state dir, lock root, and other security-sensitive paths a same-uid attacker could plant a symlink at ahead of the daemon.

The current implementation does not check whether `dir` itself is a symlink before creating/chmod-ing it. If a same-uid attacker plants a symlink at the target path before this function runs, `DirBuilder::create` (via its `EEXIST` fallback checking `path.is_dir()`, which follows symlinks) and the subsequent `set_permissions` both follow the symlink and tighten permissions on whatever real directory it points at — an arbitrary directory of the attacker's choosing, not `dir`.

## Fix (already applied on a fork)

Add an `lstat`/`symlink_metadata` check before `DirBuilder::create`/`chmod`, refusing with an `io::Error` if `dir` is a symlink rather than following it. This narrows the window considerably (there's a residual TOCTOU between the `lstat` and the `mkdir`/`chmod` that a race-free `O_NOFOLLOW`+`fchmod` idiom would close more fully — noted as a known remaining gap, not a full fix).

Reference implementation and full review/audit trail: https://github.com/prageethw/dot-agent-deck/pull/552 (issue https://github.com/prageethw/dot-agent-deck/issues/491).

## Why this is filed here rather than fixed on the fork alone

Verified byte-identical against `upstream/main` at the same line (`src/platform/fsperm/unix.rs:50`, `pub fn ensure_owner_only_dir`) — the defect is present in upstream code the fork has not otherwise diverged on, so per this fork's own contribution policy this is filed upstream at discovery time rather than carried only as a fork-side fix.

## Scope note

Two related gaps are tracked as fork follow-ups, not included in this report since they haven't been independently verified against upstream yet: a Windows-side parity gap (the Windows `ensure_owner_only_dir` may have the same class of issue), and `create_owner_only_dir`'s apparent similar unguarded shape on both platforms. See https://github.com/prageethw/dot-agent-deck/issues/560 for the fork-side tracking if useful context.

vfarcic:
Fixes #669.

## The defect, measured

`ensure_owner_only_dir` is the enforcement point that guarantees the deck's security-sensitive directories exist at owner-only `0700` — the state dir (`daemon_attach.rs:136`, `platform/detach/unix.rs:27`) and the daemon lock root (`daemon.rs:498`). It never checked what was already sitting at the path, and **both** of its steps follow symlinks:

1. `mkdir(2)` returns `EEXIST` for a symlink;
2. stdlib's `recursive(true)` fallback swallows that `EEXIST` because its `path.is_dir()` test *resolves* the link;
3. the `set_permissions` that follows then chmods the link's **target**.

So a same-uid attacker who plants a symlink at the state-dir path before the daemon first starts redirects the whole function onto a directory of their choosing.

I did not take this on trust — I planted the symlink against the code as it shipped and read back the result:

```
result=Ok(()) victim_mode=0o700
```

The call reported success, and a `0755` directory entirely outside the deck's tree came back `0700`.

## The fix

`lstat` (`symlink_metadata`) before the `mkdir`/`chmod`, refusing a symlink with a named `io::ErrorKind::InvalidInput` that carries the path, the reason and the remedy. Any *other* stat failure is refused too rather than assumed safe — if we cannot vouch for what is at the path, we do not chmod it.

### This narrows the window. It does not close it.

Stated plainly because it would be easy to read this as a closed race and it is not: a residual TOCTOU remains between the `symlink_metadata` and the `mkdir`/`chmod`, and an attacker who wins that race still redirects the chmod. The race-free idiom (`O_NOFOLLOW|O_DIRECTORY` open + `fchmod` on the descriptor, so no second name resolution exists) would close it more fully and is **deliberately not used here**: a read-open is strictly more privileged than `chmod(2)` by path, so it would refuse to repair a pre-existing `0000` directory that the current code fixes. That trade seemed wrong to make silently inside a bugfix; the reasoning is recorded in the doc comment so a future change starts from it rather than rediscovering it. Happy to file it as a follow-up if you'd prefer it tracked rather than only documented.

The guard covers the **final component only** — the one this function creates and chmods. A symlinked *ancestor* is out of scope and says so in the doc comment. (`platform::detach` already opens `daemon.log` with `O_NOFOLLOW`, so the same discipline exists one level down; the doc comment cross-references it.)

### The refusal is unconditional, and that has a user-visible cost

Worth calling out explicitly for review, because it is the one behaviour change a user could hit: an operator who **deliberately** keeps `~/.local/state/dot-agent-deck` as a symlink to another disk will now get a startup refusal. The remedy is in the error text — point `DOT_AGENT_DECK_STATE_DIR` / `DOT_AGENT_DECK_LOCK_DIR` at the real directory — and the changelog fragment says so in user terms.

There is no discriminating alternative. The threat model is an attacker running as *you*, so an ownership or mode check on the link's target compares our own uid against our own uid and separates nothing. Only the final component is affected, so a symlinked parent (`~/.local/state` itself) keeps working.

## The sibling `create_owner_only_dir` — checked, and it does not share the exposure

Asked and answered rather than assumed. It **never chmods**: `DirBuilder`'s mode reaches only a directory the call itself created, and with a symlink at the target `mkdir(2)` returns `EEXIST` and it creates nothing — so there is no permission-tightening to redirect. That is pinned by a test (`create_owner_only_dir_never_tightens_through_a_symlink`), which passed both before and after this change, and recorded in its doc comment next to the PRD #127 S2 rationale for why it deliberately leaves an existing directory's mode alone.

What it *does* share is that a planted symlink silently redirects where the caller's config write lands. That is a path-redirection question about the write, not about this seam's mode policy — a different class, and outside #669's scope (the issue scopes it out too).

## Rule 12 — cross-version contract: **unchanged**

Answering explicitly rather than leaving it implied. **The TUI↔daemon protocol/handler contract did not change**, in shape or in meaning:

- The diff touches three files: `src/platform/fsperm/unix.rs`, `src/platform/fsperm/mod.rs`, `changelog.d/669.bugfix.md`. No protocol type, handler, orchestration path or hook is in it.
- No `PROTOCOL_VERSION` bump, and no `.breaking.md` fragment: nothing on the wire moved, and no field's meaning shifted behind a stable wire.
- The change is a local filesystem guard on a startup path. An old daemon and a new TUI (or the reverse) interoperate exactly as before — the guard's only effect is whether *this process* refuses to create *its own* state dir.

Classified `bugfix`, so a patch bump.

## Tests (rule 4: not a user-visible TUI change, so a unit test is the tier)

Four unit tests in `platform::fsperm::unix`, written **before** the fix and observed red for the reported reason — the refusal cases returned `Ok(())` / a bare `AlreadyExists`, not a refusal:

| Test | What it holds |
|---|---|
| `ensure_owner_only_dir_refuses_a_symlinked_target` | the reproduction: refusal, `InvalidInput`, victim keeps `0755`, and the symlink is left a symlink |
| `ensure_owner_only_dir_refuses_a_dangling_symlink` | the same refusal where the old code gave a bare `EEXIST` naming neither path nor reason |
| `ensure_owner_only_dir_still_creates_and_repairs_real_dirs` | **the control** — fresh nested create at `0700`, loose-dir repair, idempotence. Without it a guard that refused *everything* would look like a fix |
| `create_owner_only_dir_never_tightens_through_a_symlink` | the sibling's non-exposure, mechanized rather than asserted in prose |

`SITE_AUDIT` in `fsperm/mod.rs` is updated so the machine-checked Unix/Windows audit stays truthful — including recording that the symlink refusal has **no Windows counterpart yet**: `SetNamedSecurityInfoW` resolves by name and follows a reparse point, so the same shape is plausible there and is *unverified*, not known-absent. What narrows it is that creating a reparse point needs Developer Mode or elevation (`docs/develop/windows-symlinks.md`) and that no Windows binaries are released. Left as the issue's scope note has it — a separate, unverified follow-up — rather than fixed blind on a platform I cannot test here.

## Gates

All run on the rebased branch (`main` moved under me mid-work; rebased onto c701932 and re-ran everything).

- `cargo fmt --check` — clean
- `cargo clippy --workspace --all-targets --features e2e -- -D warnings` — clean (all three flags)
- `cargo test-fast` — **3067/3067 pass**
- `cargo xtask linkage-check` — ok (559 catalog ids, 487 annotations, 9 rules)
- `cargo test-e2e` — **8791/8792 pass**, one flake (below)

### The e2e failures, run down rather than waved through

The first full e2e run, on the pre-rebase base, came back 8784/8791 with **7** failures — every one a real-agent test. I did not assume flakiness:

- **6 of the 7 passed individually** (`card_stats_005`, `devin_live_001`, `dispatch_013`, `dispatch_015`, `shell_activity_005`, `shell_activity_006`) — real-agent contention under full-tier parallel load, the flaky-tolerant tier rule 5 keeps out of CI.
- **`orchestration_dispatch_002` failed twice**, so I ran a control instead of guessing: reverted *only* `src/platform/fsperm/` back to the branch base and re-ran it. It failed **identically** — same assertion at `e2e_dispatcher_mode.rs:1243`, same ~315s — proving it was not attributable to this change. It turned out to be already fixed on `main` by a2a5953 (`fix(dispatch): keep a role card's name across a clear=true respawn`), which post-dated my base: **after rebasing, it passes.**

The rebased full run leaves exactly one failure, `shell_activity_005`, which passes in isolation in 28s against 145s under full-tier load — the same contention flake as the rest.

## Not done, and deliberately so

- **The race-free `O_NOFOLLOW`+`fchmod` idiom.** Reasoned about above and recorded in the doc comment; say the word and I'll file it as a follow-up issue rather than leaving it documented-only.
- **Windows parity.** `SetNamedSecurityInfoW` resolves by name and plausibly follows a reparse point, but I cannot test Windows from here, so it is recorded in `SITE_AUDIT` as *unverified* — not silently claimed enforced, and not fixed blind. This matches the issue's own scope note.

https://claude.ai/code/session_017XDCV3CLusnDiWZnVuPn3Q


prageethw (approved):
Traced the fix carefully: the `lstat`/`symlink_metadata` guard sits before both the `mkdir` and `chmod` calls, correctly refuses a symlinked final path component (including dangling symlinks) without following it, and the sibling `create_owner_only_dir" is correctly non-exposed since `mkdir(2)` never follows a symlink target. Tests cover the symlink-refusal, dangling-symlink, and control (normal create/repair still works) cases, plus pin the sibling's non-exposure. The residual lstat→mkdir/chmod TOCTOU is disclosed with a reasoned tradeoff against the O_NOFOLLOW+fchmod alternative. LGTM.

Issue #359, opened 2026-08-03
Decide a repo-wide policy for East-Asian-ambiguous-width characters in width-constrained render regions


vfarcic:
## Problem

`·` (U+00B7 MIDDLE DOT) is classified **East Asian Ambiguous** by Unicode (UAX #11) — a character that existed in both legacy East Asian encodings (Shift-JIS, Big5, GB, EUC), where it was drawn two columns wide, and Western ones, where it is one column. Unicode defers the choice to the renderer, so `unicode-width` exposes two answers: `UnicodeWidthStr::width()` → **1**, `width_cjk()` → **2**.

We use `·` as a separator in width-constrained render regions and measure it with `width()`. On a terminal configured to treat ambiguous-width characters as double-width — a common option, off by default in typical Western setups and frequently switched on by anyone who regularly reads CJK text — the terminal consumes 2 cells where we assumed 1.

## Why measuring conservatively is not sufficient on its own

The obvious patch is to measure with `width_cjk()` so we degrade one column earlier. That alone does **not** fix it: **ratatui also lays out the `Line` using normal width.** Being pessimistic in our own fit check only changes which string we pick — whichever string we pick still contains `·`, is still positioned by ratatui on normal-width arithmetic, and still shifts on such a terminal. The disagreement is between ratatui's model and the terminal, not between us and ratatui.

## Affected sites

- **`card_stats_border_label`** (`src/ui.rs:4698-4699`) — the two compact rungs, `" {last} · {tools} tools "` and `" {last} · {tools} "`. These render as a **right-aligned bottom border title**, so a one-cell shift moves the label toward the `┘` corner glyph and can paint over it. This is the worst case, because the degradation ladder exists specifically to guarantee the label never overruns the border.
- **The card's top title** (`src/ui.rs:14536-14537`) — `format!(" · {name} ")` / `format!(" · {id_display} ")`. Pre-existing. Left-aligned, so a shift pushes into the border dashes rather than over a corner: benign in practice, but the same latent defect.
- **A dialog string** (`src/ui.rs:13016`).

Exposure for the bottom rungs is narrow cards only — wide cards render `" Last: 2m  Tools: 14 "`, which contains no `·`.

## Candidate policies — pick one, apply repo-wide

1. **Avoid ambiguous-width characters in width-constrained regions.** Replace `·` with an unambiguously single-column ASCII separator (a comma reads well). Simplest and robust; costs the typographic `·`.
[...]

Issue #357, opened 2026-08-03
truncate_styled_segments budgets chars, not display columns, for the card top title


vfarcic:
## Problem

`truncate_styled_segments` (`src/ui.rs:4711`) budgets **Unicode scalar values** (`char`s), but its caller passes a budget measured in **terminal columns**. It renders the session card's *top* title — card number, identity/display name, and the registry-coloured agent-type badge — which is a column-measured region.

Consequence: a display name or session id containing wide characters under-counts, so the string is returned longer than the budget, and ratatui then bare-clips it at the right edge **with no ellipsis**.

This is the same defect that PRD #339 fixed for the card's `Dir:` line, where both a reviewer and an auditor found it independently. It was left unfixed for titles because PRD #339 did not touch them.

## Failure scenario

A card whose display name is `项目目录管理` (6 scalars, **12 display columns**) in a title region with, say, an 8-column budget: `truncate_styled_segments` counts 6 ≤ 8, returns the name unchanged, and the terminal clips it mid-glyph with no `…` to signal truncation. Same class of failure for emoji and other wide characters.

Conversely, a combining sequence (`e` + U+0301) is 2 scalars but 1 column, so a tight budget truncates a string that would have fit.

## Why this is now a mechanical fix

The display-width pattern exists **twice** in the tree and either can be followed:

- `src/tab_layout.rs::truncate_to_cap`
- `truncate_with_ellipsis` (`src/ui.rs:4641`), converted by PRD #339

Both accumulate `UnicodeWidthChar::width(c).unwrap_or(0)` and stop before the first character that would exceed the budget. `unicode-width` is already a direct dependency (`Cargo.toml:70`), so no new dependency is needed.

One genuine complication: this function operates on **styled segments** (coloured badge + name), so the width accumulator has to carry across segment boundaries while preserving each segment's own style. It is not a copy-paste of the single-string version.

## Known limitation to carry over

The `truncate_with_ellipsis` fix documents that per-character iteration still splits **grapheme clusters** (emoji ZWJ sequences, base-plus-combining-mark), because measuring clusters requires `unicode-segmentation`, which is deliberately not a dependency. Total width stays correct either way, so the worst case is a cosmetically odd final glyph, never an overrun. The same note should apply here.

## Severity

[...]

PR #677, opened 2026-08-24
fix(state): sanitize and clamp display_name at hook ingest (#670)

### LINKED ISSUE #670: display_name from event.metadata reaches the dashboard render unsanitized and unclamped

vfarcic:
Fixes #670.

## The defect

`AppState::apply_event` ingested `display_name` from `event.metadata` behind a single `.filter(|n| !n.is_empty())` guard and stored it verbatim. The hook socket is reachable by any process on the box — on a deck that means every agent running on it — so that string is fully producer-controlled, and it is drawn straight into a dashboard card title.

Measured against this branch's own test with the fix reverted, this is not theoretical: a `U+202E` RIGHT-TO-LEFT OVERRIDE planted in the name **reached buffer cell (30, 9)**, i.e. it survives into the frame that gets flushed to the real terminal, where it reorders the text around it — on a stacked deck, the neighbouring agents' card titles. ESC, NUL, CR/LF and DEL had nothing stopping them either, and the value was unbounded.

`ratatui`'s incidental grapheme handling is not a defence: it is third-party behaviour this project neither owns nor tests, and the measurement above shows it does not filter the bidi case at all.

## The fix

New `src/untrusted_text.rs` holds **one** implementation of the filter this project already needed in two places:

- `strip_control_and_bidi(s, keep_newlines)` — drops `char::is_control()` (C0, DEL, C1 including U+0085) plus the Unicode bidi formatting/override codepoints (`U+202A..=U+202E`, `U+2066..=U+2069`, `U+200E`, `U+200F`, `U+061C`), which are category `Cf` and so are *not* caught by `is_control`.
- `sanitize_display_name(raw) -> Option<String>` — strip, trim, then clamp to `agent_pty::DISPLAY_NAME_MAX_LEN` via the existing `prompt_delivery::truncate_on_char_boundary`. `None` when nothing usable survives, so the session keeps the name it had rather than losing its title to a malformed frame.

Reuse rather than a third copy, per the issue's instruction:

- The clamp is `truncate_on_char_boundary` — the same truncator `render_session_card` already applies to the equally producer-controlled `session_id` (issue #574), so a clamped name is marked with the same trailing `…` a shortened id already shows.
- The ceiling is the daemon's own `DISPLAY_NAME_MAX_LEN` (128 bytes), so a name arriving over the hook socket cannot be longer than one arriving over the attach socket through `is_valid_display_name`.
- `build_version_handshake::sanitize_for_prompt` and its private `is_bidi_format_char` **were** the second copy; they now delegate to the shared module and the duplicate is deleted. No behaviour change there — the filter is byte-identical, and that module's existing tests still pass unmodified.

Two deliberate choices worth reviewing:

- **Strip, not escape.** These land in width-constrained regions; `\x1b` escaped to four visible characters would spend the title budget on the attacker's behalf. `keybindings::sanitize_for_terminal` escapes because it feeds a one-off stderr diagnostic where showing what was there is the point.
- **Repair, not reject.** `is_valid_display_name` rejects the whole name, correctly, because it is validating a request a user just typed and can retype. This is scrubbing one field of an event whose other contents we still want to apply, so it repairs. The printable payload of an ESC sequence therefore survives as ordinary text (`\x1b[31m` → `[31m`) — stripping that too would mean parsing ANSI and would eat legitimate names containing brackets.

Sanitized at **ingest**, not at the render seam, because the value is stored: it lives on the session and is inherited by a session that supersedes it, so scrubbing at the one place that draws it today would leave the stored value hostile for every later reader.

## Rule 12 — did this change the TUI↔daemon contract?

**No**, neither structurally nor semantically. `PROTOCOL_VERSION` stays at 7 and there is no `changelog.d/670.breaking.md`.

1. **The wire shape did not move.** No field added, removed or retyped. `AgentEvent.metadata` is the same `HashMap<String, String>`.
2. **The semantic case does not apply either, because `display_name` is not a wire field at all.** `SessionState.display_name` is in-process only: `SessionState::live_snapshot()` (`src/state.rs:337`) does not copy it into `SessionSnapshot`, which is the sole wire carrier of live session state. Nothing the daemon emits derives from it — `daemon_protocol`'s role/label strings read `AgentRecord.display_name`, a different field on a different path.
3. **Only the consumer changed.** No producer emits differently, so an old and a new build exchange byte-identical traffic; the difference is entirely in what one of them does with a string after receiving it.

### Cross-version manual test (run, not reasoned)

`v0.37.2` release binary as the daemon, this branch's build as the TUI/CLI. Sockets, `HOME`, `XDG_*`, state, config **and `DOT_AGENT_DECK_LOG`** all redirected into a scratch sandbox, so nothing touched the real daemon or interleaved into the real `deck.log`.

- **Attach handshake** — old daemon `{"server_version":7,"build_version":"0.37.2-gf3752e1","daemon_version":"0.37.2"}`; new TUI attached and the old daemon stayed up and untouched for the whole run.
- **`StartAgent`** — a real `bash` pane started *under the old daemon* from the new TUI's new-pane form; the pane came up live (`bash-5.3$`).
- **`ListAgents`** — new binary's `daemon status` against the old daemon returned the managed agent with its pane, agent id, label and cwd.
- **Hooks** — raw `AgentEvent`s posted onto the old daemon's hook socket by a *third* process were broadcast to the new TUI and rendered; the reported `WaitingForInput` reached the card as `Needs Input`.
- **Delegate** — `delegate --to developer` from the new CLI round-tripped to the old daemon and came back with the old daemon's own structured refusal (`the daemon holds no orchestration role for pane 1, so this delegate was routed nowhere`), so the command path is intact across the boundary. A full delegate *route* needs a live orchestration with real agents on real models; **not run**, and called out rather than implied.
- **The fix, in the live cross-version deck** — the hostile name rendered as
  `1 ClaudeCode · [31m dispatch-670 sweepμμμ…   Needs Input`
  with the neighbouring `clean-neighbour` card untouched: no ESC, no NUL, no DEL, no CR/LF, no bidi override, and the 400-character tail clamped.

One mechanic worth knowing if you re-run this: PRD #103's build-version handshake sees two different build ids, finds the old daemon has no managed agents, and **SIGTERMs it to spawn its own** — silently converting the run into a same-version one. First attempt hit exactly that. The run above sets `DOT_AGENT_DECK_BUILD_ID_OVERRIDE` (`src/build_id.rs`, `debug_assertions`-only, so the release daemon still reports its real id) on the new binary, which makes the handshake pass while the two processes stay genuinely different builds.

## Scope: `ui.display_names` is deliberately not in this PR

The card prefers `ui.display_names` and falls back to `SessionState.display_name`; this PR fixes the fallback source, matching the reference fork's scope (`prageethw/dot-agent-deck#558`, whose own follow-up for this is `prageethw/dot-agent-deck#562`).

That is a judgement, not an oversight, and the two paths are not equally exposed:

- `ui.display_names` is hydrated from `AgentRecord.display_name`, which the daemon filters through `is_valid_display_name` on `StartAgent`/`SetAgentLabel` — already non-empty, ≤128 bytes and free of ASCII control bytes, with failures stored as `None`.
- The hook-socket path this PR fixes had **no gate whatsoever**.

So the residual on the preferred source is bidi-only, on a path reached over the attach socket rather than the hook socket, and closing it means either changing a validation predicate used at ~15 call sites (tab membership names, role names, cwds) or scrubbing at several `ui.rs` sites — a different audit with a different blast radius. Recorded in the catalog entry's **Does not assert** line. Happy to file it as an upstream follow-up if you'd like it tracked here rather than only on the fork.

Also out of scope by the same reasoning: issue #357 (`truncate_styled_segments` budgets chars, not display columns) and #359 (East-Asian-ambiguous width policy). The clamp here counts bytes, matching every other length bound in the project; the title's *fit* is decided later by the code #357 is about.

## Tests

- **`dashboard/pane/012`** (new, L1, catalogued in `tests/CATALOG.md`) — drives the real `AppState::apply_event` with a name carrying ESC/NUL/CR/LF/DEL/`U+202E` and a second one 400 characters long, then renders four cards through `render_card_grid_to_buffer`, the seam that runs the deck's own `ui.display_names → SessionState.display_name` resolution. Asserts the stored names, then walks **every buffer cell** asserting no control character and none of the literally-listed planted characters reached one. The planted set is written out in the test rather than read from the sanitizer, so the assertion cannot agree with a broken implementation.
- **RED verified on both halves** against the unfixed ingest: the stored-name assertion fails, and with that half removed the render assertion fails with `U+202E reached cell (30, 9)`.
- **Unit tests** in `src/untrusted_text.rs`: every bidi codepoint enumerated individually (a range typo would otherwise leave one override live, and one is all a spoof needs); the clamp swept over 2-, 3- and 4-byte characters at four offsets; and a case pinning that stripping happens *before* clamping, since the other order cuts 128 bytes of control-character padding and then scrubs it to nothing, losing a name that fits.
- Gates, all on the rebased tree: `cargo fmt --check`, `cargo clippy --workspace --all-targets --features e2e -- -D warnings`, `cargo xtask linkage-check` (561 catalog ids), `cargo test-fast` (3069 passed).

### `cargo test-e2e`: 8792/8794, two real-agent failures I do not believe are mine

Reporting this rather than rounding it to green. The two that failed are both in the real-agent class rule 4 calls flaky-tolerant, and **both pass when rerun in isolation** (rule 6):

- `dispatch_015_three_real_claude_seeds_are_genuinely_confirmed` — two of three real Claude panes confirmed their seed; the third's Claude CLI went to `● Error` and never emitted a `UserPromptSubmit` across 8 delivery attempts. The card titles in the failure grid render normally (`ClaudeCode · dispatch-r…`), which is this change's own field passing through unmodified.
- `devin_live_001_real_interactive_turn_drives_the_card_live` — the seed was delivered (the observed `Thinking` event carries the full prompt) and the real Devin agent produced no matching follow-up within 45s. The observed event's `metadata` is `{}` — this path never carries a `display_name` at all.

Neither test reads `SessionState.display_name`; `e2e_prompt_delivery_confirmation`'s only `display_name` reference is `AgentRecord.display_name`, a different field on the attach-socket path this PR does not touch. The run also overlapped a second agent session driving real-agent e2e tests in a sibling worktree, contending for the same API quota.


prageethw (approved):
Verified the sanitize path: `strip_control_and_bidi` correctly drops C0/C1/DEL control chars (covers ESC/NUL/CR/LF) plus the bidi override/format codepoints (U+202A-E, U+2066-9, U+200E/F, U+061C), and `sanitize_display_name` strips before clamping so real content isn't lost behind padding, with the length clamp correctly snapping to a UTF-8 boundary. Good test coverage: per-codepoint unit tests, an L1 render test walking every buffer cell post-ingest, and an explicit strip-before-clamp regression test. The duplicate sanitizer in build_version_handshake being deleted in favor of the shared module is a nice cleanup. Scope exclusions (ui.display_names fallback, daemon_client::strip_control_chars, byte-vs-column width) are disclosed rather than silently left. LGTM.

One non-blocking note: strip_control_and_bidi scans the full raw string before any length bound is applied, so a pathologically large display_name gets fully scanned pre-clamp — low severity given the same-uid hook-socket trust boundary, but a pre-truncation size guard would be cheap insurance.

Issue #679, opened 2026-08-24
clean-e2e-tmp's 600s dead-owner floor derives from a 300s cap that one test legitimately pins at 900s


vfarcic:
`cargo xtask clean-e2e-tmp` reaps a **dead-owner** root once it is **600 s** old, and it picks 600 s as *2× the 300 s orphan cap* — the maximum time a `setsid`'d descendant of a dead test process may still be writing under that root. One test in the suite legitimately pins that cap to **900 s**. Between 600 s and 900 s the derivation no longer holds: `--apply` can delete a root that a still-entitled daemon may write to.

Nothing is on fire. The exposure needs the *test process* to be killed **and** its daemon to still be alive past 600 s, and it is one named test rather than an ambient value. But it is a **deletion** tool whose safety argument is now off by 300 s, so it should be closed deliberately rather than left implicit.

## The two numbers

**900 s**, `tests/e2e_dispatcher_mode.rs:1023`, on `origin/main` today (landed in #665, fixing #663):

```rust
.with_env("DOT_AGENT_DECK_TEST_MAX_LIFETIME_SECS", "900")
```

Its comment gives the reason and it is a good one: `orchestration_dispatch_002`'s own `WORK_WAIT` budget is 300 s and the harness backstop was also 300 s, so on a slow run the daemon self-exited ~25 s *before* the assertion fired. `role_diagnostics` then queried a dead socket and reported `NO PANE — never spawned at all` for all three roles while the same dump rendered them alive. That false diagnostic is what #663 was filed on, and it cost a later branch a stable 313 s red before the fix was rebased in. **This value is correct and should not simply be reverted.**

**600 s**, `xtask/linkage-check/src/clean_tmp.rs:221`:

```rust
const DEAD_PID_MIN_AGE: Duration = Duration::from_secs(600);
```

Its derivation is stated in that constant's doc and again in [`docs/develop/e2e-temp-dirs.md`](https://github.com/vfarcic/dot-agent-deck/blob/main/docs/develop/e2e-temp-dirs.md) under *"The 10-minute floor on dead owners, and why it is not optional"*:

> So after nextest SIGKILLs a test, an orphan can keep writing under the dead test's root for up to that cap (**300s**). […] The old age-only rule never had this problem: its 6h floor was 72× the orphan cap. The 10-minute floor restores that protection at 2× the cap, which is ample […]

"2× the cap" is only 2× while the cap is 300.

## Why #668's clamp does not close it

[...]

Issue #678, opened 2026-08-24
Test-harness env mutation (set_var/remove_var) has an unsound SAFETY argument under multi-threaded call sites


vfarcic:
Found by the security audit of #668. Not a blocker for that PR — recorded there as a stated residual rather than a proof — but the SAFETY argument two test-harness helpers rely on does not establish what it claims, and both share the defect, so they should be fixed together.

## The claim, and why it does not hold

`std::env::set_var` / `remove_var` are `unsafe` in edition 2024 because they race any thread concurrently **reading** the environment, in Rust or in C. That is a data race, not merely a lost update.

Two helpers write the process environment during test setup:

- `tests/common/child_lifetime_bound.rs`'s `arm()` — one `set_var` of `DOT_AGENT_DECK_TEST_MAX_LIFETIME_SECS`, under a `OnceLock` (#668).
- `tests/common/mod.rs`'s `detach_from_any_live_deck()` — `remove_var` of the inherited endpoint variables. Pre-existing; `arm()` inherited its argument.

The argument both rest on is "every gate runs under `cargo nextest run`, which gives each test its own process, so the only threads that can exist are the calling test's own." **One process per test does not imply one thread per process.** Several call sites reach these helpers from inside a multi-threaded Tokio runtime whose workers already exist:

- `tests/delegate_prompt_injection.rs` calls `common::init_test_env()` from the body of a `#[tokio::test(flavor = "multi_thread")]`.
- `common::spawn_inprocess_daemon` and `start_real_server()` call it from inside a runtime.
- `tests/daemon_protocol.rs` and `tests/rehydration.rs` call the standalone `arm()` from async helpers.

`WRAP_BIN_LOCK` in `tests/agent_detection.rs` excludes sibling tests that honour that lock. It does not exclude Tokio workers, library threads, or arbitrary C environment readers. The `OnceLock` in `arm()` is load-bearing for idempotence and decorative for thread safety.

This is adjacent to #245 but is not the same hazard. #245 is about process-global state **colliding between tests** under `cargo test`'s shared-process model — a correctness bug. This is about the unsafe contract being violated **within a single test process** by threads that already exist, which is a soundness bug and is present under nextest too.

## What was already tried, so it is not re-tried

Measured during #668; both rejected, with reasons that still stand:

- **Per-`SpawnOptions` cap** — 15+ sites, partly in production code.
[...]

PR #680, opened 2026-08-24
fix(wrap): close the inherited PTY master that strands wrapped agents (#668)

### LINKED ISSUE #668: e2e leaks stand-in agent processes too — 221 orphaned cat children that #661's wrapper fix does not reach
e2e leaks *stand-in* agent processes too — 221 orphaned `cat` children that #661's wrapper fix does not reach

## Summary

#657 reported that e2e runs strand wrapped Codex children, and #661 fixed that. **The same leak exists for stand-in agents, is roughly 20x larger, and #661's mechanism does not apply to it.**

Measured 2026-08-23 on this dev box, censusing every process whose working directory sat inside a `dad-e2e` root:

| Count | Binary |
|---|---|
| **221** | `coreutils` — the `cat` stand-in agents the harness spawns |
| 9 | `bash` |
| 1 | `node` |
| 1 | `codex` |

**232 total; 222 orphaned to `ppid=1`**, the rest their children. Oldest alive **9.4 days**. Every one held an already-deleted working directory, pinning inodes on top of the temp roots they had been pinning before those were reaped.

## Why #661 does not cover them

#661 bounds the **wrapped** child: `wrap` forks a reaper that `killpg`s the child group when the wrapper dies uncatchably, gated on `DOT_AGENT_DECK_TEST_MAX_LIFETIME_SECS`, implemented in `src/wrap.rs`.

The 221 `cat` processes are **not wrapped**. They are stand-in agents spawned directly as PTY children by the harness, so no part of the `wrap` path runs for them. The escape route is the same in shape — an uncatchable end to the parent leaves a child nothing above it will reap — but it arrives through a different spawn path.

So #661 is correct and worth having; it simply fixes the Codex-shaped instance, which is the one #657 happened to observe, and not the numerically dominant one.

## Why #657's numbers understated this

#657's "35 orphans" came from a scan filtered on `codex` in the command line. Stand-ins are `cat`, so 221 of them were invisible to that filter and to every follow-up check in that issue — including the cleanup that reported "35 TERMed". The disk figures there were still right (the roots were what they were), but the process count was off by ~6x and the *shape* was wrong: this is not primarily a Codex problem.

## Suggested direction

Two options, and the second now looks clearly stronger:

1. **Widen #661's mechanism to the direct-spawn path** as well as the wrapped one. Keeps a single technique, but has to be repeated for every spawn shape that exists now or later.
2. **Give the harness a teardown sweep keyed on the temp root.** This covers every spawn shape at once and does not depend on process-group semantics that `setsid` deliberately breaks. It was option 3 in #657's original list and was passed over then; the census is the argument for revisiting it.

## Why it matters beyond disk

Per rule 14, `cargo xtask clean-e2e-tmp` **never** reaps a root whose owning PID is live, at any age. So each orphan permanently costs its root (~400 MB in the observed set) and no cleanup pass can recover it. Across this session that reached ~25 GB of unreclaimable space before manual reaping.

It also degrades diagnosis: stale roots left half-cleaned by orphans produced three wrong conclusions while investigating #656, all from reading state that looked like evidence.

## Reproducing

After any full `cargo test-e2e`:

```bash
for p in /proc/[0-9]*; do
  cw=$(readlink "$p/cwd" 2>/dev/null)
  case "$cw" in *dad-e2e*) echo "${p#/proc/} $(basename "$(readlink "$p/exe")") ppid=$(awk '/^PPid:/{print $2}' "$p/status")";; esac
done
```

`ppid=1` entries outliving their run are the leak. **Do not kill anything outside `/var/tmp/dad-e2e-*`** — a live deck may be running.

Follows #657 (closed by #661). Related: #322 (the reaper), #461 (PID-liveness rules), #502 (e2e reliability strategy).

vfarcic:
Fixes #668.

## What was wrong

`dot-agent-deck wrap` created the inner PTY with `libc::openpty` and marked neither end close-on-exec, so the agent it wraps inherited the **master** of its own controlling terminal alongside the slave it actually needs (measured live: `fd 3 -> /dev/ptmx`, whose `fdinfo` `tty-index` matched the child's own slave). Because the child holds the master, the master's reference count never reaches zero when everything above it dies, the slave never hangs up, and the child blocks on `read` forever. Field census in the issue: 232 processes, 222 orphaned to `ppid=1`, 221 of them `cat`, the oldest alive 9.4 days, each holding an already-deleted working directory.

This is a capability leak, not only a lifetime bug. A master confers two things a slave does not: reading it drains the terminal — the wrapper's tee is the deck's only view of what the agent produced, and any descendant of the wrapped process, including a third-party CLI it shells out to, could race the wrapper to read output before it is forwarded or audited. Writing it injects bytes into the terminal's input queue, i.e. the child's own stdin — with normal echo on, that reappears in the wrapper-forwarded transcript as if a user had typed it, and can drive interactive prompts with no keystroke at all. The child isn't crossing a Unix privilege boundary (same UID throughout), but the master is ambient authority the wrapped process was never meant to receive, reaching every descendant it spawns. The stranding is the symptom that made this visible; the leaked authority is the reason it's more than a leak fix.

## The fix

- `open_inner_pty` now marks the sole master `FD_CLOEXEC` at creation, before `cmd.spawn()`; every master clone happens only after spawn. The exec'd child retains no descriptor reaching the master side. The pipe path creates no PTY at all, so it needed no change.
- The inner **slave** is now also closed on exec. Smaller claim: a spare slave can't recreate the self-pinning defect (every slave descriptor hangs up together once the last master closes), since it can't retain the master. But an extra slave is still a read/write terminal capability outliving the process's own stdio, so it's worth closing too. Measured before as a fourth `/dev/pts/<n>` at fd 4 beside the intended three; measured after, gone, with the child's controlling terminal intact.
- The audit surfaced and closed three more gaps in the same area: the reaper's fd-close ceiling only scanned `0..1024` (descriptor count doesn't bound descriptor number — demonstrated with a non-`CLOEXEC` fd handed to the wrapper at 5000, invisible to the old scan), now `close_range(2)` on Linux with a bounded-loop fallback; the test harness's lifetime backstop is now armed on the in-process `AgentPtyRegistry` spawn path too (previously only the `TuiDeck`/`DaemonProc` paths pinned `DOT_AGENT_DECK_TEST_MAX_LIFETIME_SECS`, leaving 15 test files' spawns unarmed — 11 fast-tier, 4 e2e-gated); and that same lifetime value is now clamped to `1..=300` wherever it's read, closing a data-loss path (below).

Direct before/after on the production behavior, wrapper `SIGKILL`ed with the lifetime env var deliberately **unset** so neither the harness backstop nor wrapper self-defense can rescue the child — this isolates the shipped, non-test code path:

| | before | after |
|---|---|---|
| child `/dev/ptmx` descriptors | 1 (fd 3) | 0 |
| child `/dev/pts` descriptors | 4 (0/1/2 + spare at fd 4) | 3 (0/1/2 only) |
| after wrapper's `SIGKILL` | survived 30s+, reparented to `ppid=1` | exited in ~1s |

### The data-loss path the clamp closes

Not theoretical hardening. `cargo xtask clean-e2e-tmp` reaps a temp root whose owning test process is dead once the root is 10 minutes old — chosen as 2x the 300s lifetime cap the harness pins, per `docs/develop/e2e-temp-dirs.md`. But the harness only pinned that cap when the variable was *absent*, so any ambient value won, including a longer one, and nothing enforced the ceiling the reaper's margin actually depends on. Demonstrated end to end: with `DOT_AGENT_DECK_TEST_MAX_LIFETIME_SECS=3600` exported, a bounded-looking test process passed cleanly — a contributor in that shell could lose a process, wait 10 minutes, run `clean-e2e-tmp --apply`, and delete a root while a nominally-bounded descendant was still entitled to write there for another 50 minutes. The arming helper now clamps to `1..=300`: a *shorter* ambient value still wins (several tests deliberately pin 120s or 10s), but anything longer, zero, or unparseable is replaced by 300.

That ceiling constrains only what the **test process inherits**, and it is scoped to ambient values on purpose, not universal — `#665` pins `900` explicitly on `orchestration_dispatch_002` (via `with_env`, which is applied after the harness's own pin and so overrides it) so that daemon can outlive its own 300s test budget, and the reaper's floor is 600s. Between 600s and 900s that one test's daemon outlives the floor. Neither half of that gap belongs to this branch — both numbers already sit on `origin/main` independently of this clamp — but an earlier draft of this fix over-claimed a universal ceiling in three doc comments; those are corrected here, and the residual is filed as **#679** with three options and their costs, deliberately left for its own review rather than riding along on an fd-hygiene fix.

## Corrections to the issue's own framing

- The leaking `cat` is not a bare stand-in pane — bare `command = "cat"` panes leaked 0/39 times across 12 clean runs and 27 mid-test `SIGKILL`s, because `portable_pty` already sets `FD_CLOEXEC` on both ends. The leak is specifically the child of `dot-agent-deck wrap` (the synthetic-Codex stand-in).
- The orphans come from the path where teardown *did* run, not where it didn't: a mid-test `SIGKILL` of the test process self-heals (killing the outer PTY hangs up the wrapper's controlling terminal, its `SIGHUP` handler forwards and reaps on the way out), while the tidy shutdown path (`AgentPtyRegistry::shutdown_all()` → `killpg(SIGKILL)` with no grace) is uncatchable and skips the wrapper's own reap loop entirely.
- An orphan does not pin its e2e temp root `live-pid` forever, as several earlier internal notes claimed. `clean-e2e-tmp` derives root ownership from the **test process's** PID embedded in the root's name, not from the orphan, so once the test process is dead the root becomes `dead-pid` and reapable after the 10-minute floor — even with the orphan still holding a deleted working directory under it. The real, proven costs are a persistent unkillable process, retained inodes/disk blocks for as long as it runs, and diagnostic pollution in every later census — not a pinned root.

## What was rejected, and why

- **A harness teardown sweep** keyed on the shared temp-root prefix. Can't reach a production instance (the stranding happens via `Drop for AgentPtyRegistry`, which ships in the real binary), and its kill radius is unverifiable on a box running concurrent e2e suites under a shared parent directory — a one-character prefix mistake would kill a stranger's live panes.
- **Teaching `clean-e2e-tmp` to reap processes.** Widens a tool whose entire safety story is "delete directories under a prefix I proved I own" into one that signals processes based on a `/proc` attribution `docs/develop/e2e-temp-dirs.md` deliberately removed for good reason (two earlier implementations of exactly that deleted live runs' working directories).
- **Arming the lifetime cap via nextest's `[env]`.** The key doesn't exist in nextest's config schema; it's silently ignored.
- **Arming it via `.cargo/config.toml`'s `[env]`.** That also reaches `cargo run`, which would hand every developer a deck that self-terminates at 300 seconds.

No TUI↔daemon contract change. `src/wrap.rs` is the only production file this branch touches; it doesn't reach `src/daemon_protocol.rs`, `src/daemon.rs`, or any hook handler. Both sides speak `PROTOCOL_VERSION = 7`. No bump, no `.breaking.md` — including the semantic-break-behind-a-stable-wire case rule 12 calls out specifically.

The cross-version matrix was still run, three passes, **all green**, because this is wrapper/daemon lifecycle behavior even without a wire change:

| pass | daemon | TUI/CLI | worker | attach / restart / spawn / hooks / delegate / work-done | wrapped child holds no master |
|---|---|---|---|---|---|
| 1 | v0.37.2 | branch | bare | all PASS | n/a |
| 2 | v0.37.2 | branch | wrapped | all PASS | PASS |
| 3 | branch | v0.37.2 | wrapped | all PASS | PASS |

## Final e2e gate and process census (shipped tree, `8456ecd`)

Full tier: 8,886 tests run, 8,885 passed. The one failure, `shell_activity_005_real_claude_bash_child_trips_the_descendant_scan`, failed under full-tier load and passed alone in 28.4s — the same fail-under-load/pass-alone pattern seen on the prior gate. Classified as a real-agent/API/load flake, not a regression from this branch.

Process census under an isolated base, immediately after the full tier vs. the drain that follows (report the drain, not just the snapshot — the immediate wave is teardown in progress, not a leak):

| checkpoint | total processes | orphaned (`ppid=1`) | orphaned stand-ins |
|---|---:|---:|---:|
| immediately after (t+0) | 377 | 162 (all `dot-agent-deck daemon serve`) | **0** |
| t+78s | 200 | 77 (all daemons) | **0** |
| t+100s | 106 | 47 (all daemons) | **0** |
| t+218s | 8 (1 daemon + 7 children) | 0 | **0** |

By t+218s only one daemon remained, deliberately bounded to 900s by `#665`'s `orchestration_dispatch_002` pin, with seven parented children (one `cat`, three `dash`, three real Claude binaries) — all live, parented, holding no `/dev/ptmx`. Zero `ppid=1` stand-in agents at any checkpoint. Contrast with the field census in the issue: 232 processes, 222 orphaned to `ppid=1`, 221 `cat`.

## Scope

This does not fix the real-Claude Bash-tool `sleep infinity` population (Claude Code's own process detachment) — out of scope, as already noted when #661 shipped.

## Follow-ups filed

- **#678** — the test-harness `set_var`/`remove_var` soundness residual (unsafe-in-edition-2024, racy against concurrent threads), covering both the new lifetime-bound arming helper and the pre-existing `detach_from_any_live_deck`.
- **#679** — the reaper's 600s dead-owner floor versus `#665`'s legitimate 900s pin, deliberately left out of this PR since moving a deletion tool's safety margin deserves its own review.


prageethw (approved):
Confirmed the mechanism: open_inner_pty previously created the PTY master/slave via openpty without FD_CLOEXEC, so the wrapped agent inherited the wrapper's own PTY master and stayed blocked on read() forever if the wrapper died uncatchably (SIGKILL/OOM) — matches the field-measured orphan evidence in the PR body. The fix wraps both fds in OwnedFd immediately for RAII cleanup on early-return/panic-unwind, then does a proper F_GETFD/F_SETFD read-modify-write to set CLOEXEC on both master and slave before spawn — kernel-enforced at execve() so it holds across every exit path, not dependent on wrapper cleanup code running. No double-close/use-after-close risk; the reaper's separate fd-closing pass runs post-fork in its own address space. Test coverage is direct and non-vacuous: one test reads the live child's /proc/<pid>/fd to assert zero inherited masters, another behaviorally proves a SIGKILL'd wrapper's child now dies via hangup with the existing backstop disabled to isolate the fix. The three secondary issues fixed in the same branch (fd-close-ceiling range, Instant+Duration overflow, i64 truncation) are each pinned by their own regression test. All required checks green. LGTM.

PR #681, opened 2026-08-24
fix(embedded-pane): build every vt100 parser through one guarded constructor (#363)

### LINKED ISSUE #363: wire_stream_pane passes unvalidated dimensions to vt100::Parser::new on the shipped spawn path
`EmbeddedPaneController::wire_stream_pane` (`src/embedded_pane.rs`, around line 983) passes caller-supplied `rows` / `cols` straight to `vt100::Parser::new` without going through `parser_init_dims`, and it is on the **shipped spawn path** — not a test seam.

## Why the dimensions can be zero

Its two hydration callers do validate. The spawn caller does not: `create_stream_pane` forwards viewport-derived dimensions from the layout helpers in `src/ui.rs`, and `right_column_pane_dims`, `mode_side_pane_dims` and `mode_agent_pane_dims` all end in `saturating_sub(2)` (border allowance). On a sufficiently short or narrow terminal those yield **0**, so a spawn constructs a 0-row / 0-column parser in the real binary.

## Why it is latent rather than a live crash

Two things contain it today:

- The very next frame's `resize_panes_to_layout` resizes the parser to the real rect.
- All live output already runs under `guarded_parser_feed`, which catches and contains the vt100 panic and rebuilds the parser at the same geometry.

So the failure is absorbed. But the *construction* is unguarded, and the containment is incidental rather than designed — it depends on a resize arriving before anything feeds the parser, and on the guard staying in place.

For reference, vt100 0.16.2 does panic on a degenerate grid: `grid.rs:683` (`col_wrap`: `prev_pos.row -= scrolled` underflows when a 1-row grid scrolls on wrap). That is the concrete bug being relied upon not to fire.

## Suggested fix

One `parser_init_dims(rows, cols)` call at the top of `wire_stream_pane`, matching what the hydration callers and (as of PRD #341) every seam constructor already do. `parser_init_dims` clamps to `1..=PTY_RESIZE_DIM_MAX` and is the single definition of valid parser dimensions, so this keeps one source of truth rather than adding a second clamp.

## Provenance

Found while fixing the PRD #341 security-audit finding, which was the *same* class of defect but on the release-exposed test seams (`for_render_seam_with_focused_pane`, `for_scroll_seam_with_focused_pane`, `add_scroll_seam_pane`, `render_command_banner_pane_to_buffer`). Those were fixed in that PR by routing through `parser_init_dims` and, for the seed-byte feed, through `guarded_parser_feed`.

`wire_stream_pane` was deliberately left alone there as out of scope — it is pre-existing and contained, and PRD #341 was about mode indication. Filing it so the shipped-path half does not get lost now that the seam half is closed.

vfarcic:
Closes #363.

`EmbeddedPaneController::wire_stream_pane` passed caller-supplied `rows` / `cols` straight to `vt100::Parser::new`. Its two hydration callers hand over dims the daemon vouched for, but the **spawn** caller does not: `create_stream_pane` forwards viewport-derived dims from `ui.rs`'s layout helpers, and `right_column_pane_dims`, `mode_side_pane_dims` and `mode_agent_pane_dims` all end in a `saturating_sub(2)` border allowance. A short or narrow terminal therefore reached the wiring with a zero axis, on the shipped spawn path.

## The failure is at construction, not at first byte

The issue framed this as latent — contained by the next frame's resize sweep and by `guarded_parser_feed`. Writing the regression test first showed it is worse than that. `vt100` 0.16.2's `Grid::new` computes `scroll_bottom: size.rows - 1`, so a zero-row parser is inconsistent the moment it exists:

```
thread '...wire_stream_pane_never_builds_a_degenerate_parser' panicked at
vt100-0.16.2/src/grid.rs:26:28: attempt to subtract with overflow
```

That is the *unfixed* code, driven through the real `wire_stream_pane`. In a release build the subtraction wraps instead (this repo sets no `[profile]` overrides, so `overflow-checks` is off there), leaving a grid with no rows claiming a scroll region ending at row 65535.

So the containment argument does not hold for this case at all: `guarded_parser_feed` wraps *feeds*, not construction. The only thing absorbing it was the resize sweep landing before any output arrived — incidental, not designed. (`col_wrap`'s `prev_pos.row -= scrolled` underflow, which the issue cites, is a separate feed-time hazard on a *1-row* grid, and that one the feed guard genuinely does contain.)

## Fix: one shared constructor, not a second copy of the check

Per the dispatch note, the guard is extracted rather than duplicated. `new_pane_parser(rows, cols)` applies `parser_init_dims` once and is the single construction point for a pane parser.

## Call-site sweep

`git grep -n "Parser::new" -- src` — every site, and what it needed:

| Site | Status before | Now |
|---|---|---|
| `embedded_pane.rs` `seam_pane`, initial build | **Guarded** — called `parser_init_dims` inline (PRD #341) | Routed through `new_pane_parser` |
| `embedded_pane.rs` `seam_pane`, post-panic rebuild | Guarded transitively (reused the clamped locals) | Routed through `new_pane_parser`; geometry now read back off the parser rather than re-derived |
| `embedded_pane.rs` `wire_stream_pane` | **UNGUARDED — the reported defect** | Routed through `new_pane_parser` |
| `embedded_pane.rs` `process_agent_output_chunk`, post-panic rebuild | Not guarded, but read `p.screen().size()` off a live parser — so it *reproduced* whatever geometry that parser held, propagating a degenerate construction instead of healing it | Routed through `new_pane_parser` |
| `hyperlink.rs:404` | Inside `#[cfg(test)] mod tests`; 24×80 fixture | Left raw — not production |
| `embedded_pane.rs` ×4 in `#[cfg(test)] mod tests` | Two deliberately build a raw 1×10 parser to prove the vt100 feed bug still exists (`wide_char_in_one_row_pane_does_not_crash_the_tui`); two are 24×80 fixtures | Left raw — retargeting the 1×10 pair would gut that regression test |

Outside `src/`: `tests/common/mod.rs` and several `tests/*.rs` construct parsers as harness fixtures at fixed valid sizes. Not shipped, left alone.

To stop a *fifth* site from quietly skipping the guard — the failure mode the dispatch flagged, and the same class of drift as any rule kept in two hand-maintained copies — `parser_is_constructed_in_exactly_one_place` pins the structural invariant: in this file's non-test code there is exactly one `vt100::Parser::new`, and it sits inside `new_pane_parser`. Verified non-vacuous by reintroducing a raw call site: the test fails naming both byte offsets.

Also confirmed while sweeping: `resize_pane_pty` calls `screen_mut().set_size` without clamping, but its one caller (`resize_panes_to_layout`) already skips a pane whose target area has a zero axis, so a degenerate geometry is not reachable that way. Recorded in the `new_pane_parser` docs rather than changed — construction is the contract this fixes.

## Rule 4 — test tier, decided deliberately

**L1**, in `src/embedded_pane.rs`'s unit-test module. What is under test is parser *construction* from caller-supplied dims: the assertion is `screen().size()` immediately after wiring. Nothing in that path needs the spawned binary, a real PTY, a daemon, or an agent — the `AttachConnection` is an in-process `UnixStream::pair()` — so an L2 `e2e_*` test would buy a process spawn and a PTY to observe the same value. This is a defect fix, not a major user-facing feature, so rule 4's PTY-attached / real-agent requirements do not attach.

The L1 test is not a stand-in for the shipped path, though: it calls the real `wire_stream_pane` with the real degenerate inputs, and panics in `grid.rs` against the unfixed code.

Three tests added:

- `wire_stream_pane_never_builds_a_degenerate_parser` — the regression, over `(0,0)`, `(0,80)`, `(24,0)`, `u16::MAX` square, and `PTY_RESIZE_DIM_MAX + 1`.
- `wire_stream_pane_preserves_usable_dimensions` — the guard is a floor and a ceiling, not a rewrite; `(21, 38)` still passes through untouched. This is why the change is a no-op for every non-degenerate spawn.
- `new_pane_parser_clamps_and_passes_through` — both halves of the constructor.

Plus `parser_is_constructed_in_exactly_one_place` above. `include_str!` is compile-time, so it reads no file at runtime (~6ms).

## Rule 12 — cross-version contract: **no movement**

Stating the conclusion explicitly, as asked. The TUI↔daemon protocol and its semantics are **unchanged**:

- No wire shape moved: no frame kind, no request/response variant, no field added, removed or retyped. `PROTOCOL_VERSION` stays as it is.
- No semantic break behind a stable wire: `create_stream_pane` still sends the caller's `rows`/`cols` verbatim in `StartAgentOptions`, so the daemon opens its PTY at exactly the size it did before. The clamp applies only to the **TUI-local** vt100 parser.
- No handler, orchestration path or hook behaviour changed.

So no `changelog.d/363.breaking.md`, and this is a patch-level change. The one touch to `src/daemon_client.rs` is a `#[cfg(all(test, unix))]` constructor, absent from the release library — deliberately not a release-exposed test seam, which was PRD #341's finding.

Since nothing crosses the wire, the cross-version sandbox run has no signal to produce here: an older daemon paired with this TUI sees byte-identical traffic.

## Coordination with #680

No rebase needed. #680 is on `src/wrap.rs` and the e2e harness spawn helpers; this touches neither — the diff is `src/embedded_pane.rs`, a `#[cfg(test)]` helper in `src/daemon_client.rs`, and a changelog fragment.

## Gates

- `cargo fmt --check` — clean
- `cargo test-fast` — 3067 passed, 0 failed
- `cargo test-e2e` — 8792 tests, 8788 passed, **4 failed**, all four green on isolated rerun:
  - `manager_016_wheel_over_dialog_does_not_scroll_side_pane` — 7.5s red in tier, **1.5s green alone**
  - `work_done_004_unsolicited_completion_is_visibly_labelled_in_the_attached_tui` — 13.1s red in tier, **1.1s green alone**
  - `shell_activity_005_real_claude_bash_child_trips_the_descendant_scan` — real-agent tier, **green alone** (27.6s)
  - `shell_activity_006_real_claude_bash_call_crossing_the_cap_keeps_the_badge_working` — real-agent tier, **green alone** (134.5s)

  Reported rather than glossed. The two real-agent ones are the flaky-tolerant tier by design (rule 4). For the two deterministic ones the 5–12× wall-clock gap between tier and isolation points at load, and neither can be reached by this change: with a non-degenerate viewport `new_pane_parser` returns its inputs unchanged (pinned by `wire_stream_pane_preserves_usable_dimensions`), and every pane in the e2e suite is spawned at a real terminal size. Behaviour is identical to `main` for every spawn that was not already broken.

## Changelog

`changelog.d/363.bugfix.md`.

https://claude.ai/code/session_019q7XRAoxHjisPf2zchXbzs


prageethw (changes requested):
The guarded-constructor consolidation (new_pane_parser) is correct and complete — verified the two remaining raw vt100::Parser::new sites outside embedded_pane.rs (src/hyperlink.rs:404, various tests/*.rs) are pre-existing fixed-size test fixtures, not production code. Zero-dimension edge case coverage is solid ((0,0), (0,80), (24,0), u16::MAX, PTY_RESIZE_DIM_MAX+1, plus pass-through).

However, this PR's own CI run has `build-windows` failing (conclusion: FAILURE, confirmed via `gh pr view --json statusCheckRollup`), and it's a real, deterministic bug rather than flakiness: the new `parser_is_constructed_in_exactly_one_place` test does `SRC.split_once("\n#[cfg(test)]\nmod tests {")` against `include_str!("embedded_pane.rs")`, but Windows checks out the file with CRLF line endings, so the literal `\n` separator never matches and the `.expect("test module marker moved; update this guard")` panics — confirmed directly from this PR's CI log (job build-windows, run 32765892144): `thread 'embedded_pane::tests::parser_is_constructed_in_exactly_one_place' panicked ... test module marker moved; update this guard`.

src/embedded_pane.rs, the new structural test (~line 4557): please normalize line endings before the split (e.g. `SRC.replace("\r\n", "\n")`) or use a CRLF-tolerant match, and confirm build-windows goes green. Worth noting the PR body's own Gates section reports cargo test-fast/test-e2e clean but doesn't mention build-windows, so this looks like it postdates local verification rather than being a known/accepted failure.

vfarcic:
Thanks — the diagnosis was exactly right, and `build-windows` is now green on this branch (run [32890034277](https://github.com/vfarcic/dot-agent-deck/actions/runs/32890034277), `build-windows` pass in 4m2s on `b179dbb`; its `cargo nextest run --workspace` step succeeded, so the guard test passes there).

**What changed** (`b179dbb`, `src/embedded_pane.rs`, +7/-1 — the guard test only):

```rust
let src = SRC.replace("\r\n", "\n");
let (prod, _tests) = src
    .split_once("\n#[cfg(test)]\nmod tests {")
```

One detail worth flagging beyond what the review named: the CRLF problem affected **two** markers, not just the split. `helper_end` locates the end of `new_pane_parser` with `.find("\n}\n")`, which would have failed the same way had the split not panicked first. Normalizing once at the top covers both, which is why the fix is a single `replace` rather than a CRLF-tolerant pattern at the split site.

**Verification.** Since no local gate runs Windows, I simulated the guard's logic against a CRLF-converted copy of the file before pushing: normalized, LF and CRLF produce identical results (split ok, 1 hit, helper bounds (20366, 20536), hit inside helper); the old code on the CRLF copy fails to find the marker, i.e. reproduces the panic. Linux gates clean — `cargo fmt --check`, `cargo clippy --workspace --all-targets --features e2e -- -D warnings`, `cargo test-fast` 3067/3067.

No changelog fragment change: this is a test-only fix with no user-visible behaviour change.

You're right that the Gates section postdates this — the failure came from CI on the pushed branch, not from a local run, and `build-windows` isn't reproducible locally by design.

prageethw (approved):
The Windows CRLF split failure in `parser_is_constructed_in_exactly_one_place` (my prior CHANGES_REQUESTED) is fixed in b179dbb via `SRC.replace("\r\n", "\n")` before the split, exactly as requested. `build-windows` is now green and all 9 checks pass.

Issue #319, opened 2026-08-01
Hook ingestion has no line-length or connection bound


vfarcic:
## Problem

`src/daemon.rs:850-864` reads hook connections with `BufRead::lines().next_line()`, which can accumulate an arbitrarily large newline-free payload before it ever resolves to a line. Every accepted owner connection also gets its own spawned task, with no cap on how many can be outstanding at once.

Because every managed agent has same-user access to the hook socket (see #318), a compromised or simply buggy agent can send an unbounded, newline-free payload — or open an unbounded number of connections — and consume daemon memory without ever producing a single parseable event.

## Not introduced or worsened by #284

Found during the surrounding audit sweep for #284 (https://github.com/vfarcic/dot-agent-deck/issues/284, shipped in PR #314); it is not a regression from that work and is unrelated to the retire-predicate logic #284 touches.

## Proposed fix

- Bound the accepted line length on the hook-ingestion read path, rejecting (rather than buffering indefinitely) a connection that exceeds it.
- Cap the number of concurrent accepted connections/spawned tasks per daemon.

## Related

Same trust surface as #318 (daemon does not bind hook-event provenance) — both stem from the same-user socket being handed to every managed agent with no additional verification or bounding on what it's allowed to send.

## References

- Issue #284: https://github.com/vfarcic/dot-agent-deck/issues/284
- PR #314: https://github.com/vfarcic/dot-agent-deck/pull/314
- `src/daemon.rs:850-864`

PR #682, opened 2026-08-24
fix(cli): bound --task-file and stdin reads, refuse non-regular files (#328)

### LINKED ISSUE #328: Bound --task-file and stdin reads; reject non-regular files
Found during the security audit of #303 (auditor finding 3). Not introduced by that change, but #303 makes it materially more load-bearing: the corrected guidance now tells every orchestrator and worker to pass task text via `--task-file` as the default, so this read path goes from an expert escape hatch to the primary input for all delegation.

## The exposure

`src/main.rs:472-480` uses unbounded `read_to_string` for both file and stdin input:

- A very large regular file is read fully into memory.
- `/dev/zero` (or any endless character device) grows the allocation until failure/OOM.
- A FIFO or blocking device makes the CLI hang indefinitely, with no timeout.
- Symlinks are followed, so the path read may not be the path intended.

There is a second unbounded allocation boundary in the daemon: `src/daemon.rs:861-865` uses `lines().next_line()`, which buffers the complete JSON line carrying the task.

Non-UTF-8 input does fail closed with `InvalidData`, and there is no current check-then-read TOCTOU because the implementation performs no separate precheck — but any future metadata check must use the already-open handle rather than re-resolving the path.

## Why it was deferred from #303

The practical severity depends on who can plant a hostile file at a predictable path. Today `.dot-agent-deck/` is `0775` with files at `0664`, so a same-group local user can — which is tracked separately in the coordination-file hygiene issue. Tightening those permissions downgrades this from a security issue to a robustness one. #303 shipped as a guidance-text-only patch on release-urgency grounds, and this needs a size-limit policy decision plus tests, so it was split out rather than rushed.

## Proposed fix

- Impose a documented maximum task size on both file and stdin reads by reading at most `MAX_TASK_BYTES + 1` and rejecting over-limit input before serialization.
- Bound the daemon's hook-message read at `src/daemon.rs:861-865` too.
- For path input, open once and inspect metadata from that handle; reject non-regular files.
- Consider no-follow semantics if hostile local workspaces are in scope.

## Care needed

Rejecting non-regular files risks breaking legitimate `-`/stdin and process-substitution usage — that behaviour needs deliberate test coverage. This touches `src/daemon.rs`, so CLAUDE.md rule 12 applies: answer the TUI↔daemon contract question and run the cross-version manual test before opening the PR.

Verified by reading, not by attempting an OOM or blocking reproduction.

vfarcic:
Closes #328.

## The defects

`resolve_task` in `src/main.rs` read the task text with unbounded `read_to_string` on both branches and never checked what the path pointed at. Two distinct defects, both fixed here:

1. **No size bound**, on either input. A huge or endlessly-growing file was read into memory until the process died.
2. **No file-type check**, so a FIFO or a device node was opened and read. A FIFO with no writer blocked the command **forever** — and the block happened *inside `open`*, before anything got a chance to look at the target — while `/dev/zero` never stopped producing bytes.

This matters more than its size suggests: the delegation protocol the deck writes into every orchestrator's context (`src/orchestrator_context.rs`) tells it to prefer `--task-file` over `--task`, precisely because inline task text is at the mercy of the caller's shell. So this is the input path the dispatch and delegation flows are actively told to use, and it was the one without bounds. All three commands that take it — `delegate`, `work-done`, `dispatch` — go through `resolve_task`.

## The fix

A new `src/bounded_read.rs` holds a general primitive and the one policy built on it:

- **`read_capped(reader, max_bytes, source)`** — reads at `max_bytes + 1`, so input exactly at the limit is accepted while anything past it is *detected without being read*. Refused, never silently truncated.
- **`read_task_input(path, stdin)`** — the `--task-file` policy. `-` keeps no file-type requirement (stdin is a pipe or a terminal by design; the cap is the whole of its protection); a path must be a regular file.

Two properties of the path branch worth calling out:

- **Opened once, judged from the open handle.** The type check reads `File::metadata` — an `fstat` on the descriptor already held — not a second `std::fs::metadata` on the path, so the thing checked and the thing read cannot differ. This is what the issue asked for.
- **The open cannot hang.** On Unix the handle is opened `O_NONBLOCK`. Without it the FIFO hang happens *inside `open`* and no later check can help; the flag is ignored for regular files, the only kind that survives the check.

Symlinks are still followed (no `O_NOFOLLOW`) — reaching a task file through a symlink is ordinary, and since the check applies to the resolved target, a symlink pointing at a FIFO or `/dev/zero` is refused just the same. The issue listed no-follow as a "consider"; it would break a legitimate case to close one already covered.

### Why 1 MiB

Deliberately chosen so no legitimate task can reach it while pathological input still gets refused. A task is prose destined for an agent's prompt: this repository's largest PRD is ~117 KiB and a task file is a task *description*, not a whole PRD; and 1 MiB of prose is roughly 250k tokens, past the context window of the agent the text is being written for. Anything above that is pathological rather than long.

**Reuse checked first**, per the dispatch brief: `src/config.rs::load_features_file` already applies exactly this shape (regular-file check, size cap, `take()`-bounded read) for `.dot-agent-deck.toml`, but inline and with different failure semantics — a bad features file keeps the previous value and warns, it does not error. There was no reusable helper or shared constant to adopt, so this adds the first one rather than a second inline copy. `MAX_FEATURES_CONFIG_BYTES` (64 KiB) is a config-file policy and deliberately stays separate from `MAX_TASK_BYTES`.

## One deliberate behaviour change

`--task-file <(some-command)` — bash process substitution — passes a `/dev/fd/N` **pipe**, not a regular file, so it is now refused. The issue flagged this exact risk. The refusal names the working alternative, and stdin has no file-type requirement:

```
$ dot-agent-deck delegate --to coder --task-file <(generate-task)
Error: task file '/dev/fd/63' is a FIFO; --task-file needs a regular file (for a pipe, a
process substitution, or a terminal, pipe the text in and pass `--task-file -` instead)

$ generate-task | dot-agent-deck delegate --to coder --task-file -      # works
```

`-` and ordinary stdin keep working unchanged, and are covered by tests on both sides of the boundary.

## Rule 12 — cross-version contract: **the TUI↔daemon contract did not change**

Neither structurally nor semantically, so no `PROTOCOL_VERSION` bump and no `.breaking.md` fragment.

- **Structurally**: nothing under `src/daemon.rs`, `src/daemon_protocol.rs` or `src/event.rs` is touched. The diff is entirely upstream of serialization — `resolve_task` either produces the same `String` it produced before, or refuses before any message is constructed.
- **Semantically** (the same-wire/different-meaning case): no field's meaning shifts. The set of messages a new CLI can emit is a strict *subset* of what an old one could — same shapes, fewer accepted inputs.

Verified rather than asserted, in a sandbox with its own sockets, `HOME`, state dir **and `DOT_AGENT_DECK_LOG`** (rule 12's log-isolation point), against the real **v0.37.2** release binary:

1. **Wire bytes are identical.** Captured the `delegate` line from the v0.37.2 CLI and from this branch's CLI for the same `--task-file`, over a stub socket. Byte-identical modulo the timestamp:
   `{"message_type":"delegate","pane_id":"xver-pane","task":"Fix `compute()` in \"src/lib.rs\" for $USER\nsecond 'line' with $HOME & a \\ backslash\n","to":["coder"],"timestamp":…}`
2. **New CLI → v0.37.2 daemon.** Started a v0.37.2 daemon and ran this branch's `delegate --task-file <file>`, `delegate --task-file -`, and `work-done --task-file <file>` against it. The old daemon received, parsed and answered all three (`Received delegate signal pane_id=xver-pane targets=["coder"]`, `Received work-done signal`); the only rejection is its own expected `unknown pane` — there is no orchestration in the sandbox — not a parse or protocol failure. A CLI-side refusal (FIFO) never reaches the daemon at all.

**What I did not run**, stated rather than glossed: the full variant of rule 12's manual test — a TUI attached to the old daemon with a live agent under it, routing a delegate into a worker pane and receiving hooks back. What that adds over the above is the *routed* half, and the branch's own `cargo test-e2e` covers it (`e2e_dispatcher_mode`, `e2e_orchestration_route_isolation`, `idle_worker_detector`, `delegate_prompt_injection`). Given the wire is byte-identical, cross-version routing cannot differ from same-version routing here.

## Rule 4 — test tiers

**No TUI harness work needed.** The only new user-visible surface is a CLI error on stderr, printed by the `delegate` / `work-done` / `dispatch` arms before anything reaches the daemon — no pane, status, layout or hook behaviour changes, so neither an L1 snapshot nor an L2 PTY test applies. Coverage is unit tests in the fast tier, at both levels:

- `src/bounded_read.rs` — one per refusal path: at-the-limit accepted, one-byte-over refused, an endless reader refused, oversized file, missing file, directory, **FIFO**, **symlink to a FIFO**, and `/dev/zero`. The three that guard a *hang* run under a 5s watchdog thread, so a regression fails the tier instead of wedging it.
- `src/main.rs` — four at the `resolve_task` seam every command actually calls, proving it routes into the bounded reader: oversized file, exactly-at-the-cap accepted, oversized stdin, non-regular target.

**The tests were confirmed to fail against the old behaviour.** Reverting `read_task_file` to `std::fs::read_to_string` and rerunning: the FIFO and symlink-to-FIFO cases time out at 5s (the real hang), `/dev/zero` allocates until the watchdog fires, and the directory and oversized cases fail on the missing refusal — 5 of 10 red. Restored, all 10 pass.

## Rule 3 — naming

The new file is `src/bounded_read.rs`, named for what it contains. No issue or milestone number in the filename.

## Related: #319

#319 (hook ingestion has no line-length or connection bound) is the **daemon's** hook socket and stays untouched here — a different code path and a different unit, as the brief asked.

Since it comes up: it **cannot use `read_capped` directly**. That path is `tokio::io::BufReader::lines().next_line().await` over an accepted socket — async, line-oriented, and a socket is never a regular file, so neither `read_capped`'s `std::io::Read` signature nor `read_task_input`'s file-type check applies. What does transfer is the *policy*: cap, then refuse rather than truncate or buffer. Its async equivalent is an `AsyncBufReadExt::take`-bounded read per line (plus the connection cap #319 also asks for), which is a genuinely separate implementation.

## Gates

- `cargo test-fast` — 3077/3077 pass.
- `cargo test-e2e` — green on this change; **reported honestly** rather than as a clean sweep. Two full runs: 8801/8802 and 8800/8802. Every failure was a **real-agent live test** (`devin_live_001_real_interactive_turn_drives_the_card_live`, `card_stats_005_real_agent_card_narrows_without_restructuring`), and each passes on an isolated rerun per rule 6 — 13.5s and 25.1s alone against 51s and 53s under a saturated box, i.e. timeout flakes in the tier rule 5 calls flaky-tolerant and keeps out of CI. Neither test contains a single reference to `--task-file` (grep: 0 matches in both files), and neither exercises `resolve_task`. Nothing else in the tier failed in either run.
- Changelog fragment: `changelog.d/328.bugfix.md`. Classified `bugfix` (patch bump), not `breaking` — per `pyproject.toml`, `breaking` is the cross-process compatibility axis specifically, and this project's definition excludes generic user-facing breakage; the process-substitution change above is the latter, and it is called out prominently in the fragment.

https://claude.ai/code/session_01W8hfzaW7kQBhG22u6HSx67

### DIFF
--- a/changelog.d/328.bugfix.md
+++ b/changelog.d/328.bugfix.md
@@ -1,0 +1,9 @@
+## `--task-file` refuses an oversized or non-regular task file instead of hanging or reading it without limit
+
+`dot-agent-deck delegate`, `work-done`, and `dispatch` read the task text with `--task-file <path>`, or from stdin with `--task-file -`. Neither read had a size bound, and neither checked what the path actually pointed at. A path naming a FIFO with no writer blocked the command **forever** — the block happened inside `open`, before anything got a chance to look at it — and one naming an endless character device such as `/dev/zero`, or a file being appended to, was read into memory until the process ran out of it. Both reads are now capped at 1 MiB and refused, never truncated, past that; a path that is not a regular file is refused on its type before a single byte is read.
+
+The delegation protocol the deck writes into every orchestrator's context tells it to pass task text as a file rather than inline, precisely because `--task` lets the caller's shell mangle backticks, quotes and `$VAR` first. That makes `--task-file` the primary input for every delegation, so it is the one that needed the bound.
+
+1 MiB is far above anything a task can legitimately be: this project's largest PRD is ~117 KiB, and 1 MiB of prose is roughly 250k tokens — past the context window of the agent the text is being written for. Both refusals name what was wrong and what to do about it, rather than failing anonymously.
+
+**One behaviour changes for existing callers:** `--task-file <(some-command)` — bash process substitution — passes a `/dev/fd/N` pipe, not a regular file, so it is now refused. Pipe into stdin instead: `some-command | dot-agent-deck delegate --to coder --task-file -`, which is bounded but has no file-type requirement. Symlinks still work and are still followed; the type check applies to the resolved target, so a symlink pointing at a FIFO or a device is refused just the same.
--- a/src/bounded_read.rs
+++ b/src/bounded_read.rs
@@ -1,0 +1,320 @@
+//! Size-bounded reads for caller-supplied input.
+//!
+//! Issue #328. `std::fs::read_to_string` and `Read::read_to_string` are
+//! unbounded in both directions that matter: they allocate however much the
+//! source produces, and they wait however long it takes to produce it. That is
+//! fine for a file this process wrote; it is not fine for a path or a stream
+//! handed in on the command line, where the two pathological shapes are a file
+//! that is enormous (or growing) and a target that is not a file at all — a
+//! FIFO with no writer blocks forever, `/dev/zero` never ends.
+//!
+//! Two things live here, deliberately separated. [`read_capped`] is the
+//! general primitive — read anything, stop at a byte cap, refuse rather than
+//! truncate — and carries no policy of its own. [`read_task_input`] is the one
+//! policy built on it: the `--task-file` reader, which additionally opens a
+//! path once and refuses anything that is not a regular file, and whose error
+//! wording names that flag because it is the only thing that calls it.
+//!
+//! [`crate::config::load_features_file`] applies the same shape inline with
+//! different failure semantics (a bad `[features]` file keeps the previous
+//! value and warns rather than erroring), which is why it is not expressed in
+//! terms of these functions.
+
+use std::io::Read;
+
+/// Upper bound on the task/summary text accepted from `--task-file <path>` and
+/// `--task-file -`.
+///
+/// A task is prose destined for an agent's prompt, so the cap only has to sit
+/// far enough above "the longest brief anyone would actually write" that no
+/// legitimate input can reach it. 1 MiB clears that by a wide margin in both
+/// directions: this repository's largest PRD is ~117 KiB and a task file is a
+/// task *description*, not a whole PRD; and 1 MiB of prose is roughly 250k
+/// tokens, past the context window of the agents the text is being written
+/// for. Anything above it is pathological rather than long, and gets a clear
+/// refusal instead of an allocation.
+pub const MAX_TASK_BYTES: u64 = 1024 * 1024;
+
+/// Read `reader` to a `String`, refusing input larger than `max_bytes`.
+///
+/// `source` names the input in error messages ("task from stdin", "task file
+/// '…'") and is expected to read as a lowercase noun phrase.
+///
+/// The reader is capped at `max_bytes + 1` rather than `max_bytes` so the two
+/// outcomes stay distinguishable: input exactly at the limit is accepted, and
+/// input past it is *detected* without being read — the extra byte is the
+/// evidence, so an endless stream costs one megabyte and a refusal rather than
+/// the machine's memory. Nothing is ever silently truncated.
+pub fn read_capped(reader: impl Read, max_bytes: u64, source: &str) -> Result<String, String> {
+    let mut buf = String::new();
+    reader
+        .take(max_bytes.saturating_add(1))
+        .read_to_string(&mut buf)
+        .map_err(|e| format!("failed to read {source}: {e}"))?;
+    if buf.len() as u64 > max_bytes {
+        return Err(over_limit(source, max_bytes));
+    }
+    Ok(buf)
+}
+
+/// Read the task/summary text for `--task-file <path>`, or from `stdin` when
+/// `path` is `-`. Both are capped at [`MAX_TASK_BYTES`].
+///
+/// `-` deliberately keeps no file-type requirement — stdin is a pipe or a
+/// terminal by design, and the cap is the whole of its protection. A `path`,
+/// by contrast, must name a **regular file**, which is what the two shapes
+/// this guards against are not: a FIFO with no writer never produces a byte,
+/// and a character device such as `/dev/zero` never stops producing them.
+pub fn read_task_input(path: &str, stdin: impl Read) -> Result<String, String> {
+    if path == "-" {
+        read_capped(stdin, MAX_TASK_BYTES, "task from stdin")
+    } else {
+        read_task_file(path)
+    }
+}
+
+/// The path branch of [`read_task_input`], with two properties beyond
+/// [`read_capped`]:
+///
+/// * **Opened once, judged from the open handle.** The type check reads
+///   `File::metadata` (an `fstat` on the descriptor already held), not a second
+///   `std::fs::metadata` on the path, so there is no window in which the thing
+///   checked and the thing read can differ.
+/// * **The open cannot hang.** On Unix the handle is opened `O_NONBLOCK`,
+///   because a plain `open(2)` of a FIFO with no writer blocks *inside the
+///   open* — before any check could run, which would reproduce the hang this
+///   function exists to prevent. The flag is ignored for regular files, the
+///   only kind that survives the check below.
+///
+/// Symlinks are followed deliberately (no `O_NOFOLLOW`): a task file reached
+/// through a symlink is ordinary, and since the check is applied to the
+/// resolved target, a symlink pointing at `/dev/zero` or a FIFO is refused
+/// just the same. This matches [`crate::config::load_features_file`].
+fn read_task_file(path: &str) -> Result<String, String> {
+    let source = format!("task file '{path}'");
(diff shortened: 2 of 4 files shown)


prageethw (changes requested):
The size-bound and TOCTOU design in src/bounded_read.rs is correct: metadata() is read off the already-open handle (no gap between the type check and the read), the size check uses metadata.len() as a fast pre-check plus a hard take(max_bytes+1) cap during the actual read so growth-during-read is still caught, and symlinks are correctly followed since the check applies to the resolved target. Test coverage of the refusal cases (FIFO, symlink-to-FIFO, char-device, directory, over-by-one) is thorough.

However, this PR's own CI run has `build-windows` failing (conclusion: FAILURE), and it's a real bug rather than flakiness: on Windows, `OpenOptions::read(true).open()` on a directory path fails immediately with `Access is denied (os error 5)` at the open() call itself, so the flow never reaches the `metadata.is_file()` check that produces the friendly 'is a directory' message — `directory_is_refused_as_not_a_regular_file` fails as written, and the shipped Windows UX diverges from what's documented (an unhelpful OS error instead of the intended refusal message).

src/bounded_read.rs (read_task_file, ~line 93-104) and the test at ~line 249-257: please either special-case the Windows ERROR_ACCESS_DENIED/directory case to produce the same friendly message, or relax the test to assert only that the input is refused (not the exact wording) on Windows, and confirm build-windows goes green.

vfarcic:
Thanks — the diagnosis was exactly right, and `build-windows` is now green on this branch (run [32890489000](https://github.com/vfarcic/dot-agent-deck/actions/runs/32890489000), `build-windows` pass in 4m8s on `10c412f`; all four required checks pass).

**I took the first of your two options** — special-case the Windows open failure — rather than relaxing the test. The reason is that the changelog fragment promises refusals that "name what was wrong and what to do about it, rather than failing anonymously", and relaxing the test would have meant that promise quietly stopped being true on one platform while the suite went green. Fixing the behaviour keeps the documentation honest and needs no fragment correction.

`src/bounded_read.rs` (+32/-5, no other files):

```rust
let file = match options.open(path) {
    Ok(file) => file,
    Err(e) => {
        return Err(if std::fs::metadata(path).is_ok_and(|m| m.is_dir()) {
            not_a_regular_file(&source, "a directory")
        } else {
            io_err(e)
        });
    }
};
```

Three things worth flagging, since this touches the TOCTOU property you verified:

1. **It does not reintroduce the TOCTOU.** Consulting the path is only reachable *after* the open has already failed, so no descriptor exists to consult and nothing is read on that branch either way. The only thing a race can change is the wording of an error that is returned regardless — never whether the read happens, or what gets read. The success path still reads `metadata()` off the open handle, untouched.
2. **Only a directory is rewritten, not every access-denied.** `ERROR_ACCESS_DENIED` is also what a genuinely unreadable regular file returns, so the branch is keyed on `is_dir()`, not on the error kind. A missing path or a permission-denied file still reports the OS error — `missing_file_error_names_the_path` covers that and is unaffected.
3. **The message moved into a shared `not_a_regular_file` helper**, mirroring the existing `over_limit`, so the type check and the recovery cannot drift into two almost-identical strings.

**Verification.** No local gate runs Windows, so I checked the new branch on Linux by reproducing its shape: a `0o000` directory makes `open` fail with `EACCES` while `stat` still succeeds — the same situation as Windows' `ERROR_ACCESS_DENIED` — and a throwaway test confirmed that path still yields `is a directory ... --task-file needs a regular file`. I removed that test before committing to keep the diff scoped; the branch's standing coverage is `directory_is_refused_as_not_a_regular_file` itself, which now exercises the recovery on Windows CI and the `metadata` check on Unix. Linux gates clean: `cargo fmt --check`, `cargo clippy --workspace --all-targets --features e2e -- -D warnings`, `cargo test-fast` 3077/3077.

prageethw (approved):
The Windows directory-open bypass I flagged (my prior CHANGES_REQUESTED) is fixed in 10c412f: the Windows open-failure path now special-cases the directory case via a shared `not_a_regular_file()` helper, so the message stays byte-identical cross-platform, and the test wasn't relaxed. `build-windows` is now green and all 9 checks pass.

PR #683, opened 2026-08-24
fix(worktree): write the ownership marker at creation time (#425)

### LINKED ISSUE #425: Write the worktree ownership marker at creation time, so the deck can recognise its own worktrees

vfarcic:
Fixes #425.

## Problem

`src/worktree_reclaim.rs` defined the ownership marker (`OWNER_MARKER_FILENAME`, line 36) and **read** it (line 357) to decide whether a merged, clean worktree may be removed unattended. Nothing in `src/` ever **wrote** it — `git grep -n OWNER_MARKER_FILENAME -- src` returned exactly those two lines.

So the gate was correct but hollow: every worktree the deck created reported `OWNED=no` and could only reach the `ask` verdict, never `remove`. The field evidence in the issue (five live worktrees, `OWNED=no` on all five) is the expected output of that, not a sampling artefact — and a confirmation that is always required stops being a confirmation.

## Fix

The marker is written at creation time, in the success arm of `issue_dispatch_run::create_worktree`.

Marker and gate now live together in a new `src/worktree_owner.rs` (writer, reader, and the git-metadata-dir resolution both sides share). That co-location is the point: the reclaim path deletes directories, so a drift between where the marker is written and where it is looked for is a silent regression back to exactly this bug. `worktree_reclaim::ownership_of` is now a three-line call into it, and the byte-exact `path_from_bytes` / `trim_trailing_newline` helpers moved along with it so writer and reader compute the same path for a non-UTF-8 worktree name.

Four properties, each of which is a way to get this wrong:

- **Location.** `<repo>/.git/worktrees/<name>/dot-agent-deck-owner`, resolved by `git rev-parse --git-dir` from inside the worktree — never in the working tree. Verified directly rather than assumed: a marker there leaves `git status --porcelain` empty, and `git worktree remove` deletes it along with the worktree. An in-tree marker would make every marked worktree permanently dirty, and the gate keeps every dirty worktree — so it would make marked worktrees permanently **un**reclaimable, the gate defeating itself.
- **Best-effort, never fatal.** A failure warns and is dropped; the dispatch still runs. The cost of a missing marker is one `--yes` later, which is the fail-safe direction.
- **Idempotent.** One whole-file write, no append.
- **Says what created it.** The document records the creation path (`dispatch` / `issue-dispatch`) and its subject (the dispatch name, or `<task>#<issue>`), plus deck version, timestamp, pid and branch — per the issue's "record a claimant, not a bare claim". The gate stays an **existence** check and never parses that content, so a future format change cannot silently reclassify every existing deck-created worktree as foreign.

## Creation-path sweep

`create_worktree` is the **only** `git worktree add` in `src/` (`git grep -n 'worktree", "add'` → two lines, both inside it), so both deck creation paths were fixed by one change:

| Creation path | Entry point | Status |
| --- | --- | --- |
| `dispatch <name>` — single agent | `dispatch::handle_dispatch` → `create_worktree` | **changed** (marker written) |
| `dispatch <name> --orchestration <n>` — the orchestration spawn | same call; one worktree, shared by every role | **changed** (same site) |
| Issue-dispatch fire flow — one worktree per issue | `issue_dispatch_run::dispatch_one_issue` → `create_worktree` | **changed** (marker written) |
| Reclaim / re-create | `worktree_reclaim` only ever *removes*; it has no creation path | n/a |

Two `git worktree add` call sites outside `src/` were found and deliberately **not** changed — `.claude/skills/verify-pr/setup.sh` and `.claude/skills/dot-ai-worktree-prd/create.sh`. Those worktrees are created by an agent workflow, not by the deck, so marking them would be the false positive this whole mechanism exists to prevent. (`dot-ai-worktree-prd` is additionally a vendored mirror — CLAUDE.md rule 13.)

`cargo xtask clean-e2e-tmp` was not touched, as instructed.

## Worktrees that already exist: not adopted

**They stay foreign and keep asking**, including the ones on this machine right now. Reasoning, which is the same shape as rule 14's:

- The marker is an ownership **claim** consumed by a path that **deletes directories**, so the dangerous direction is the false positive.
- On disk, "a worktree the deck created last week" and "a worktree a human or an orchestrator created" are indistinguishable. Anything that adopted the first would adopt the second.
- The cost of not adopting is bounded and visible — the `ask` verdict and an explicit `--yes` on a named path. The cost of adopting wrongly is an unattended removal of somebody else's work.

The same rule decides the concurrency case in code: `WorktreeCreation::AlreadyClaimed` means the directory was already on disk when our add ran, so *this* process did not create it, and that arm leaves the marker alone. Nothing is lost — a concurrent dispatch that genuinely created it writes its own marker from its own `Created` arm. This is pinned by a test, not just by a comment.

An "adopt this worktree" verb (option 2 in the issue's comment) is deliberately out of scope here; it is an operator assertion and belongs in its own change, next to its own confirmation surface.

## Rule 12 — cross-version contract: **unchanged**

The TUI↔daemon contract did not move, in either sense:

- **Wire shape:** no change. `src/daemon_protocol.rs` is untouched, `PROTOCOL_VERSION` is unchanged, and no protocol type gained, lost or re-typed a field. The diff is `dispatch.rs`, `issue_dispatch_run.rs`, `worktree_reclaim.rs`, the new `worktree_owner.rs`, plus tests and docs.
- **Semantic break behind a stable wire:** none. The marker is a local file in a git admin dir; it never crosses a process boundary, is never serialized into a request or a response, and no existing field's meaning changes. `worktree list|reclaim` is a CLI verb that shells out to `git`/`gh` synchronously with no daemon involvement at all.
- **Mixed versions degrade safely in both directions.** An old daemon creating a worktree writes no marker, so a new `worktree reclaim` reads it as foreign → `ask`. A new daemon creating one writes a marker an old `reclaim` already knows how to read (#422 shipped the reader). Neither direction can produce a *wrong removal*; the worst case is an extra confirmation.

Given no wire type, handler signature or hook payload is touched, the sandboxed previous-release-daemon run was judged not applicable rather than skipped; there is no contract for it to exercise. Bump policy: **patch** (`.bugfix` fragment), per `docs/develop/versioning.md`.

## Rule 4 — TUI tier: none needed

No user-visible TUI surface changes: no pane, status, prompt, focus, layout, mode, embedded pane or hook delivery is touched, and the TUI renders none of this. The only user-visible change is in the `worktree list` / `reclaim` CLI report — the `OWNED` column now reads `yes` for deck-created worktrees, which is the fix — and that surface is covered at its own tier by the fast real-binary-subprocess integration suite (below). No L1 snapshot and no L2 PTY test is warranted.

## Tests

- **`worktree/reclaim/010`** (`tests/worktree_reclaim.rs`) — the required end-to-end: a worktree created through the **real** creation path (calling production `create_worktree`, never the fixture's `mark_owned`) reads `owned: true` / `verdict: "remove"` from `worktree list --json`. A hand-made `git worktree add` sibling, identical in every other respect, is the **control** at `owned: false` / `verdict: "ask"` — without it, `owned` could be `true` for everything and the test would still pass.
- **`create_worktree_marks_the_worktree_as_deck_owned_without_dirtying_it`** — the marker lands in the metadata dir and not under the worktree; `git status --porcelain` stays empty (the self-defeating failure mode); re-marking replaces rather than appends.
- **`create_worktree_never_marks_a_worktree_it_did_not_create`** — the `AlreadyClaimed` arm leaves a foreign worktree unmarked.

**Negative control run:** with the one `write_marker_best_effort` call commented out, `worktree/reclaim/010` fails with exactly the reported symptom —

```
{"owned":false,"verdict":"ask","reason":"reclaimable: PR is merged and the tree is clean,
 but the deck cannot prove it created this worktree", ...}
```

— and the unit test fails alongside it. Restored, all three pass.

## Docs

New [`docs/develop/worktree-ownership.md`](docs/develop/worktree-ownership.md) (maintainer-facing, excluded from the site per rule 11, linked from `CONTRIBUTING.md`): where the marker lives and why not in the working tree, which paths write it, why best-effort and idempotent, why the gate never parses the content, and the never-adopt rule with its reasoning — the same "proven or asserted, never inferred" discipline `docs/develop/e2e-temp-dirs.md` records for `clean-e2e-tmp`.

Changelog fragment: `changelog.d/425.bugfix.md`.

## Gates

- `cargo fmt --check` — clean
- `cargo test-fast` — 3066 passed, 0 failed
- `cargo test-e2e` — 8791 passed, 0 failed (7 slow)
- `cargo xtask linkage-check` — ok (catalog entry for `worktree/reclaim/010` added)

https://claude.ai/code/session_01Vf8oAucb2pPLc47WhatFJf


prageethw (approved):
Confirmed the marker write happens in create_worktree's Created arm only (src/issue_dispatch_run.rs), immediately after git worktree add succeeds and before returning to any caller — critically it is never written on the AlreadyClaimed arm, which correctly prevents a process from falsely claiming a worktree it didn't create. The write is best-effort and idempotent (whole-file write, no append), and while it isn't OS-atomic with git worktree add, it happens synchronously before any caller observes success and a freshly-created worktree can't yet satisfy the MERGED+clean reclaim gate anyway, so the residual window has no exploitable consequence — the correct practical answer given git has no combined primitive here. Two unit tests (including a negative test that the marker is never set on a worktree this process didn't create) plus an e2e test with a hand-made-worktree control and a confirmed fails-without-the-fix check. Docs and changelog present. CI green. LGTM.

Issue #688, opened 2026-08-25
Guard src/ unit tests against reaching a live deck: scrubbing inherited endpoints does not stop a child resolving one


vfarcic:
Found while fixing #666. One instance was fixed there (`scheduler/dispatch/016` case G); this issue is the **guard** that stops the next one being written.

## The gap the existing mechanism does not cover

`tests/harness_isolation.rs` states the rule and `tests/common/mod.rs` enforces it: a test process run from inside a deck pane inherits `DOT_AGENT_DECK_SOCKET` / `_ATTACH_SOCKET` / `_PANE_ID` / `_AGENT_ID`, anything it spawns inherits them too, and its hooks then post into the developer's **live** dashboard — a card appears under a fixture pane id and vanishes again.

Two things that enforcement does not reach:

1. **It lives under `tests/`.** The lib target's own `#[cfg(test)]` unit tests do not link `tests/common/`, so nothing scrubbed those four variables for any of the **61** `src/**` unit tests that spawn a real pane (`agent_pty.rs` 39, `daemon.rs` 9, `spawn.rs` 7, `state.rs` 3, `dispatch.rs` 1, `ui.rs` 1, `daemon_protocol.rs` 1).

2. **Scrubbing the parent's environment is not sufficient, and this is the important half.** `agent_pty::spawn` already `env_remove`s all four variables from every child, so a spawned child never *inherits* an endpoint. It can still **resolve** one: `platform::paths::socket_path()` falls back to `$XDG_RUNTIME_DIR/dot-agent-deck.sock`, which is the developer's live daemon.

Measured during #666: running the offending fixture **with** a full four-variable scrub applied still produced **3 foreign `SessionStart`s in 8 runs**. Clearing the parent's environment cannot reach a path the child computes for itself.

## How the one real instance arose

The leak did not come from a *command*. Every command any `src/` unit test spawns is `/bin/sh`, `/usr/bin/true`, `/bin/cat`, `sleep`, a `sh -c 'exit ${VAR:+1}'` probe, or a deliberately-nonexistent path — none of which emits anything.

[...]

Issue #685, opened 2026-08-24
A duplicated prompt turn is reported as a clean confirmation in 3 of 4 repetition shapes


vfarcic:
Measured while auditing #666's payload rearm. **Not introduced by #666** — this is true on `main` today, without the rearm. Filed so the measurement is on the record rather than living in a scratch design note.

## What

When an automatic prompt's payload ends up in the agent's input box more than once and is submitted as a single turn, the deck usually cannot tell that apart from a clean confirmation, and says nothing.

`prompt_submission_accumulated` — the detector that recognises a doubled submission, terminates the delivery and logs a `warn` — is consulted **only after `prompt_submission_matches` says no**. Shape 3 of the matcher accepts N newline-separated copies as **confirmation**, and the hook's 200-byte `USER_PROMPT_MAX_LEN` truncation collapses every repetition of a long prompt onto the same prefix as a single copy.

Measured by running the shipped functions over both repetition shapes at 2 and 3 copies, short and >200-byte prompts, each through the hook's truncation:

| prompt | separator-free (`seedseed`) | newline-separated (`seed\nseed`) |
|---|---|---|
| short (< `USER_PROMPT_MAX_LEN`) | `Accumulated` → terminal + **warn** | `Confirmed` → terminal, **silent** |
| long (> `USER_PROMPT_MAX_LEN`) | `Confirmed` → terminal, **silent** | `Confirmed` → terminal, **silent** |

Three copies behave identically to two in every cell.

**So the warn fires in one of four shapes.** A `dispatch` or issue-dispatch prompt is routinely well over 200 bytes and therefore gets it in **none** of them: a duplicated turn is reported to the operator as an ordinary successful confirmation.

## Why this is not "widen the detector"

Both halves are deliberate and are pinned by existing tests. Do not reverse them:

- `repeated_copies_of_the_seed_count_as_delivered` asserts `seed\nseed` *and* `seed\nseed\nseed` **match**. Shape 3 exists so a delivery stops retyping into an agent that is demonstrably already working.
- `bare_concatenation_is_accumulation_and_not_confirmation` closes with `!prompt_submission_accumulated(seed, "{seed}\n{seed}")` — "the newline-separated shape belongs to the matcher, not here".
- `prompt_submission_matches`'s own residual section already withdraws the copy bound for long prompts.
- Both reviewers on #424 explicitly refused to widen the confirmation matcher.

[...]

Issue #684, opened 2026-08-24
A TUI-owned prompt into a launcher-wrapped pane is dropped when the real agent announces itself


vfarcic:
Found while fixing #666, and it is the reason #666's fix ships on the daemon-owned delivery paths only.

## Symptom

A `[[modes]]` `seed_prompt`, or an orchestration role's opening prompt, delivered into a pane whose command is a launcher/wrapper — `devbox run claude …`, a bootstrap script, anything that emits a `wrapper_fork`-origin `SessionStart` before `exec`ing the real agent — is **abandoned the moment the real agent emits its own native `SessionStart`**.

The user sees `Seed prompt not delivered (the agent's conversation changed); abandoned`, and the delivery log records `delivery-target-changed`. The prompt is discarded with the pane healthy and idle.

## Affected population

Both **TUI-owned** delivery paths: `ui::process_pending_seed_prompts` and `ui::deliver_orchestrator_prompt`, on any pane that declares a launcher handoff before the deck writes. That is the same launcher class #424 §3 names as the most fragile, and the same population #666 is about.

The **daemon-owned** paths — `dispatch`, the scheduler, issue-dispatch — are **not** affected. They go through `crate::state::latch_generation`, which correctly refuses to bind a `wrapper_fork` start and permits exactly one handoff to the genuine generation.

## Code path

1. **`src/state.rs:5583-5636`** (`AppState::apply_event`). A `wrapper_fork` start on a pane with no generation takes the `None => true` advance branch, so `pane_hook_session` becomes the **wrapper's** session id. The `launcher_origin_start` guard only stops such a start from *replacing* a live conversation; it does not stop it from *establishing* one.
2. Three TUI sites then read that value with **no genuine-start discriminator**: the seed path's `PromptDelivery` capture literal (`src/ui.rs:3559`), `bind_delivery_generation` (`src/ui.rs:4040`), and `bind_generation_before_retry` (`src/ui.rs:4120`). The delivery binds the launcher's session id.
3. **`src/state.rs:5625-5633`.** When the genuine start arrives it is an announcement naming a different generation, so `note_generation_closed` fires and `pane_generation_closures` goes 0 → 1.
4. **`src/ui.rs:3979-3984`** (`delivery_target_changed`) returns `true` on **either** the closure-count check (`pane_generation_closures(pane_id) > delivery.closures_at_write`, stamped before the RPC) **or** the `expected_session_id` comparison.
[...]

Issue #637, opened 2026-08-21
unconfirmed_retry_delay's first bare-CR retry (500ms) can race a genuine, slower confirmation


prageethw:
## Bug: `unconfirmed_retry_delay`'s first bare-CR retry (500ms) can race a genuine, slower confirmation

`src/prompt_delivery.rs:451-456` (verified byte-identical on `upstream/main` at the time of filing):

```rust
pub fn unconfirmed_retry_delay(attempts: u32) -> std::time::Duration {
    const BASE_MS: u64 = 500;
    const CAP: std::time::Duration = std::time::Duration::from_secs(15);
    let shift = attempts.saturating_sub(1).min(8);
    std::time::Duration::from_millis(BASE_MS.saturating_mul(1u64 << shift)).min(CAP)
}
```

## Reproduction / expected vs actual behavior

A prompt delivered to a `ConfirmationCapability::Reports` producer (one that will eventually confirm via a genuine hook/status event) schedules a bare-CR retry using this schedule: 0.5s, 1s, 2s, 4s, 8s, then a 15s cap. The doc comment above the function explains the *cap* is sized to cover "the whole `AUTOMATIC_PROMPT_DEADLINE` window ... in single-digit attempts" — but nothing sizes the *first* attempt against how long a genuine confirmation can actually take to arrive.

Measured against real Codex (a wrapper-strategy agent whose confirmation arrives via a native hook, not synchronously): a genuine TEXT confirmation can legitimately take up to ~8.45s to arrive after the prompt is delivered. Under the schedule above, a bare-CR retry fires at the 500ms mark — **7.95s before** the genuine confirmation would have landed — resubmitting into the pane while the real confirmation is still in flight.

**Expected**: no retry should fire while a genuine confirmation for this producer could plausibly still be in flight.
**Actual**: the first retry fires at 500ms regardless of the producer's realistic confirmation latency.

## Fix (already shipped on a fork, offered here since upstream still carries the defect)

Raise the first attempt's delay to 10s (comfortable margin past the measured 8.45s worst case), keeping the existing 15s cap for subsequent attempts:

```rust
pub fn unconfirmed_retry_delay(attempts: u32) -> std::time::Duration {
    if attempts <= 1 {
        std::time::Duration::from_secs(10)
    } else {
        std::time::Duration::from_secs(15)
    }
}
```

[...]

PR #689, opened 2026-08-25
fix(dispatch): recover a spawn-time payload destroyed during the agent's boot (#666)

### LINKED ISSUE #666: A spawn-time prompt whose payload lands during the agent's boot is unrecoverable: both payload attempts fall inside a ~1.65 s window

vfarcic:
Fixes #666.

## The defect

`attempt_writes_payload` was `attempt <= MAX_PAYLOAD_SUBMISSIONS` (2) **and nothing else** — it had no notion of whether the agent that would read those bytes had started yet. Both payload-carrying attempts therefore fell inside the agent's boot window, and every attempt after that was a bare submit probe with an empty payload.

So a prompt whose payload was written during boot was **unrecoverable**. From the measured log in #666: attempt 2 lands ~1.65 s after the launcher `exec`s the agent; two panes confirmed within ~20 ms, and the third had its payload destroyed, then spent 55 seconds pressing Enter on an empty input box before being abandoned at attempt 8.

The bytes do not survive in the tty buffer: an agent TUI's raw-mode switch uses `tcsetattr` with `TCSAFLUSH`, which **discards** pending input. A payload written into the gap between `exec` and raw-mode setup is destroyed, not delayed — and nothing downstream could distinguish that from a payload sitting unsubmitted in an input box.

## The fix

One **additional** payload-carrying attempt per delivery, **armed** rather than scheduled. The two blind payload writes keep their exact present meaning, and a *blind* third rewrite stays refused forever — which is what preserves the `seedseed` guarantee from #424/#570.

Arming requires the conjunction of six facts, all necessary:

| | fact | rules out |
|---|---|---|
| **S** | **Standing** — before the write, something the deck can vouch for said what occupies this pane (a `wrapper_fork` launcher-handoff declaration, or the deck's own spawn record) | an unauthenticated post-write producer claim on a pane the deck cannot vouch for |
| **U** | **Unbound at the write** — the delivery had no bound hook generation | a delivery whose agent had already announced a conversation; the bytes are in its box and a rewrite would duplicate |
| **G** | **Genuine start** — a `SessionStart` not carrying `WRAPPER_FORK_SESSION_START_ORIGIN` | a launcher's own boot-provenance event, emitted before the agent exists |
| **I** | **Identity** — same pane, exact non-optional `agent_id` | the daemon's synthetic card-surfacing start, a sibling pane, a rebound successor |
| **W** | **After the write** — strictly after the ordering watermark | everything already visible when we wrote |
| **T** | **The producer's start precedes its first prompt** | **Codex and OpenCode** — see below |

Then a 500 ms `REARM_READINESS_BUFFER` measured from the arming `SessionStart` before the armed attempt may carry the payload — a session existing is not the same as a terminal ready to accept a newline, and without this the fix would reproduce its own bug one level up.

### Fact T is a deliberate departure from the issue's suggested direction

#666 proposed arming on any post-write native `SessionStart`. That is correct for Claude Code and **wrong for two shipped agents**. On Codex, codex-cli 0.145.0 posts its native `SessionStart` when the first *turn* starts — measured in #243 at T+29.999 s, with the injected prompt's own `UserPromptSubmit` at T+30.004 s. On OpenCode, `session.created` fires ~16 ms *after* the prompt is accepted, and a 35 s idle cold boot emits nothing at all (#146).

On both, "a native `SessionStart` arrived after our write" is a **consequence of a prompt we already delivered successfully**. Arming there would write a third copy of a dispatch task into an agent that has already accepted the first — a genuine duplicate *turn*, arguably worse than the corrupted `seedseed` that `MAX_PAYLOAD_SUBMISSIONS = 2` exists to prevent.

`agent_start_precedes_first_prompt` is therefore exhaustively matched, so a new `AgentType` must answer rather than inherit a default. `Devin` answers `false` because nobody has measured it — an unverified `true` is the unsafe direction. This is deliberately the same predicate **#243 half 2** proposes as an explicit `AgentSpec` field; when that lands, this function's body becomes a registry lookup and no call site moves.

### The declared type may only withhold, never grant

A round of review caught the first implementation violating **#424 F4**: standing had been collapsed to a `bool`, so fact T was decided from the *arming event's declared* `AgentType`. A pane the deck itself spawned as Codex could be armed by an event merely claiming to be Claude.

Fact S now records **which agent type the deck believed occupied the pane before the write**. The declaration must **equal** that belief, and T is asked of the belief:

| case | believed | declared | result |
|---|---|---|---|
| A | Claude (spawn record) | Claude | arm |
| C | Claude (spawn record) | Codex | refuse |
| G | Codex (spawn record) | Claude | refuse |
| H | OpenCode (spawn record) | Claude | refuse |
| D | Claude (pre-write `wrapper_fork`) | Claude | arm |

Both directions are load-bearing: C fails if you trust the spawn record alone, G/H fail if you trust the declaration alone. Precedence gives the deck-spawn record priority over the launcher declaration (it is the half no producer can assert), and the first declaration wins in both stores so a producer cannot *walk* the believed type to whatever it needs the post-write declaration to match.

## Scope: which delivery paths this fixes

**Fixed — daemon-owned:** `dispatch`, the scheduler, issue-dispatch.

**Not fixed — TUI-owned** (`ui::process_pending_seed_prompts`, `ui::deliver_orchestrator_prompt`): both call sites pass a permanently-disarmed `AgentStartRearm::default()`, so their behaviour is byte-for-byte unchanged.

The reason is a **prior** defect, filed as **#684**: a `wrapper_fork` start *establishes* the pane's generation, three `ui.rs` sites read it with no genuine-start discriminator, and the real agent's genuine start is then counted as a conversation closure — so `delivery_target_changed` removes the delivery outright. Arming a delivery that no longer exists changes nothing, so #684 has to land first. When it does, enabling the rearm on both TUI paths is a two-argument deletion.

## Rule 12 — the contract does not move

**Wire shape: no.** `PROTOCOL_VERSION` stays at **7**; `git diff origin/main -- src/daemon_protocol.rs` is empty. No frame, field, enum variant or handler changed. Every new item is in-process Rust. The armed attempt is an ordinary `WriteAndSubmit` under a fresh `…#a<n>` id, which the daemon's ledger fingerprints as new work exactly as `#a2` is today.

**Semantic: no.** The guarantee "at most two payload submissions reach a pane" becomes "at most three, and the third only on positive evidence the first two were destroyed". No peer enforces or reads the other's attempt count, no handler branches on it, and the safety property the constant backs — never blindly append a payload to bytes we already put in that box — is enforced daemon-side by `user_typed_since_writing_payload` plus the writer-held backstop, both unchanged and in force.

**Cross-version manual test: run, and it passes.** A `v0.37.2` daemon with two live agents under it, branch TUI attached against it (skew correctly reported, daemon never restarted, confirmed still the old binary at the end). The delegate routed (`Received delegate signal pane_id=1 targets=["coder"]`, task file written, pointer rendered). The work-done hook arrived and rendered in the orchestrator pane. Status hooks moved the card to `Needs Input`. Zero `ERROR` lines. Sandbox fully isolated including `DOT_AGENT_DECK_LOG`.

→ `changelog.d/666.bugfix.md`, patch bump. Not `.breaking.md`.

## Tests

| id | tier | what |
|---|---|---|
| `scheduler/dispatch/016` A–H | L1 | Eight cases pinning the arming contract at the detached daemon seam, asserted on bytes reaching a real `/bin/cat`. **D** is load-bearing: `AgentType::from_command("devbox run claude …")` is `None`, so an implementation satisfying A but not D would ship green and fix nothing in the field. **G/H** pin that a declared type cannot grant. |
| `scheduler/dispatch/014` | L2 synthetic | New fifth pane whose two-stage launcher swallows **both** payload attempts — the case #666 names as having no coverage anywhere. |
| `scheduler/dispatch/015` | L2 real | All three real interactive Haiku panes now swallow two payloads. This is the rule 4 bar: a live interactive agent visibly receiving and submitting a seed whose first two writes were destroyed. |
| `src/prompt_delivery.rs` | unit | The blind-refusal contract, the 499/500 ms buffer pair, one-shot, the six necessity cases, and `agent_start_precedes_first_prompt` exhaustively — with #243's and #146's measurements quoted in the assert messages, so a future contributor flipping one has to argue with the evidence. |

`/015` asserts the recovered payload lands on **some** attempt > 2 rather than exactly 3: attempt 3 is unreachable by construction in that fixture (the launcher `exec`s Claude only after attempt 2's write, and attempt 3 fires 1.000 s later, while real Claude needs 1–3 s). Verified by measurement, including with the buffer set to 0. `/014` keeps the exact `attempt=3` assertion, where the stand-in makes it deterministic.

`/016` also gained a **structural** attempt boundary replacing a timing heuristic, plus detachment from any live deck — see #688.

## Effect on the other open issues in `src/prompt_delivery.rs`

| issue | verdict |
|---|---|
| **#637** first retry can race a genuine confirmation | touches, helps slightly. `unconfirmed_retry_delay` unmodified; the rearm's buffer is measured from the arming start, never from the write, so it can only add delay. |
| **#526** confirm by digest not prefix | helps / adds motivation. The rearm inherits #526's ordering-not-causality residual and adds one more consumer of the fix that closes it. |
| **#527** `delivery_id` not byte-bounded | touches trivially. The armed attempt rides an ordinary `…#a<n>` id, no new class. |
| **#528** unify the three implementations | helps. All new policy is pure data in one module, and each path gains the same edits at points it already has. |
| **#529** the 10 s fallback writes with no buffer | does not touch; complementary. `REARM_READINESS_BUFFER` is a deliberately separate constant guarding a different write against a different signal — **do not merge them**. |
| **#530** `dispatch --single` exits 0 early | does not touch. Raises the fraction that eventually confirms, making #530 more useful. |
| **#544** payload concatenates with a user draft | touches; does not widen. The armed write takes `user_typed_since_writing_payload` and the writer-held backstop unchanged. |
| **#547** ambiguous partial write retains its record | does not touch. The armed write pushes its `PayloadRecordRelease` holder in the same arm attempt 2 does. |
| **#243** pre-prompt readiness signal | does not touch; lays a stepping stone (fact T is #243 half 2's predicate written down honestly). |
| **#570 / #424** | preserved. Generation binding, the ordering watermark, identity checks and both user-draft guards unchanged; the blind third rewrite stays refused. |

## Accepted residuals, recorded rather than papered over

- **Standing is not uniformly unforgeable.** The deck-spawn half is (`spawn_agent_type`'s only writers are the spawn/respawn paths). The **launcher-handoff half is producer-assertable** — it establishes *not-post-hoc*, which is an ordering property, not producer authentication. Measured baseline: the attach socket has no peer-credential check and `KIND_STREAM_IN` frames are forwarded raw to the pane's PTY writer, so a same-user process **already** has unlimited arbitrary bytes into any pane. Forging the arming evidence buys strictly less — one write, once, of the deck's *own* prompt text. This is a new consumer of **#543**, not a new class.
- **Two or three copies, and the detector is quieter than assumed.** The rearm inherits #526's ordering-vs-causality gap; in the rarer sub-case the box can end up with three copies rather than two. Measured: `prompt_submission_accumulated`'s warn fires in **one of four** repetition shapes, and never for a >200-byte prompt, because every repetition truncates onto the same 200-byte prefix. Filed as **#685**. Deliberately not "fixed" by widening the matcher — shape 3 is pinned by two existing tests and both #424 reviewers refused to widen it.

## Follow-ups filed

- **#684** — a TUI-owned prompt into a launcher-wrapped pane is dropped when the real agent announces itself. Blocks arming the TUI paths.
- **#685** — a duplicated prompt turn is reported as a clean confirmation in 3 of 4 repetition shapes.
- **#688** — guard `src/` unit tests against reaching a live deck: scrubbing inherited endpoints does not stop a child *resolving* one via `$XDG_RUNTIME_DIR`.

## Gates

- `cargo fmt --check` — clean
- `cargo test-fast` — **3072/3072**
- `cargo xtask list-tests`, `cargo xtask linkage-check` — clean
- `cargo test-e2e` — 8794/8797. Three failures, all real-agent and none in the prompt-delivery surface: `devin_live_001` (#488), `shell_activity_006` (#502), and `dispatch_013`. `dispatch_013` is **not** a regression from this branch — production code is byte-identical between the run where it passed and the run where it failed (the only intervening commit is entirely inside `mod tests`); the box was at load average 60 on 16 cores with two concurrent orchestrations. Under investigation, tracked separately.
- Cross-version manual test — passed (above).


prageethw (approved):
Traced the AgentStartRearm arming conjunction: it requires six independently-necessary facts (pre-write standing the deck can vouch for, no bound generation at last write, a genuine non-wrapper_fork start, exact pane+agent_id identity, strict post-write ordering, and a producer whose native SessionStart precedes its first prompt — explicitly false for Codex/OpenCode to avoid arming on a start that's itself a consequence of the delivery). Confirmed a declared AgentType can only withhold eligibility, never grant it, closing the F4 hole noted in the PR's own review history. The 500ms readiness buffer avoids reproducing the same boot-timing bug one level up, and the fix is correctly scoped to the daemon-owned path (TUI-owned paths left disarmed pending #684, stated not hidden). Wire protocol unaffected, PROTOCOL_VERSION correctly left unchanged. Test coverage is thorough: RED-before-fix L1 suite (scheduler/dispatch/016 A-H) covering the arming conjunction cases, an L2 synthetic 5-pane concurrent test, and an L2 real-Claude 3-pane test. The residual duplicate-turn risk from event misordering is disclosed and bounded, tracked as #685 rather than hidden. All CI checks green. LGTM.

Issue #305, opened 2026-07-31
feature req: cross agent hooks within same worktree


prageethw:
Use case: review request among cross-agents

It would be cool if Claude could ask Codex to review code and the Codex pane displays that it is working on Claude's code request, and vice versa.

1. Need a hook
2. Need a way to capture visible working action in Codex in the above case after Claude asks to act on a code review

vfarcic:
Thanks for the request! Some of this may already work today, so let me lay out what ships now — and then ask what's still missing for you, because one phrase in your description could mean two very different features.

## Cross-agent review requests already work, via orchestrations

An orchestration is a tab with one agent per role, where one role coordinates and hands work to the others. The roles can be *different agents*, which is your Claude→Codex case:

```toml
[[orchestrations]]
name = "review"

[[orchestrations.roles]]
name = "orchestrator"
command = "claude"
start = true

[[orchestrations.roles]]
name = "reviewer"
command = "codex --sandbox workspace-write --ask-for-approval never -c \"sandbox_workspace_write.network_access=true\""
description = "Reviews code changes for correctness, style, and edge cases"
prompt_template = "Review the change. Report findings only — do not modify code."
```

Claude runs `dot-agent-deck delegate --to reviewer --task "..."`, the task lands in the Codex pane, Codex works, and `dot-agent-deck work-done` routes a summary back into Claude's pane. There's no hook involved — the daemon routes it directly, which is also why it survives detaching and reattaching the TUI. Full docs: https://agent-deck.devopstoolkit.ai/docs/orchestration/

Also worth saying: nothing here privileges Claude. If you want Codex to be the one asking, move `start = true` to the codex role.

## Two Codex setup requirements that may be biting you

Codex landed only three days before you filed this, and it has two gotchas that produce *exactly* the symptoms you describe:

1. **Codex as a role or worker needs sandbox network access.** `dot-agent-deck delegate` and `work-done` reach the daemon over a local socket, and Codex's `workspace-write` sandbox blocks that by default — so those commands **fail silently** and the pipeline never moves. Hence the `-c "sandbox_workspace_write.network_access=true"` in the config above. Details: https://agent-deck.devopstoolkit.ai/docs/troubleshooting#codex-as-a-role-or-worker-allow-sandbox-network-access

[...]

Issue #230, opened 2026-07-27
devbox run -- codex command  not picked as agents , only direct codex


prageethw:
devbox run -- claude picked up, though.

prageethw:
It happens to be that it is not picked up until the first command is executed, closing now

PR #691, opened 2026-08-25
feat(config): let a role or mode declare the agent its command launches (#308)

### LINKED ISSUE #308: Role config cannot declare its agent, so a launcher command (`devbox run -- codex`) shows "No agent" until the first task

vfarcic:
Closes #308.

## The problem

`AgentType::from_command` identifies an agent by the **basename** of its command, looking through a conservative allow-list of known launchers (`sudo`, `env`, `sh -c`, …). A command that starts its agent through something else — `devbox run -- codex`, `mise exec -- codex`, `nix develop -c codex`, a `make` target, a bespoke `run-codex.sh` — resolves to `None`, so the pane got no badge and no status: the card read literally **No agent**.

For **Codex** this was more than a label. Failed detection also means **no wrapper**, and Codex emits no pre-prompt `SessionStart` — its native one is posted when the first *turn* starts. The wrapper's synthetic fork-time `SessionStart` was the only thing that could identify the pane before its first prompt, and failed detection is exactly what removed it. So a launcher-started Codex pane sat blank from launch until the first task was delegated, then silently started working. Claude never showed the symptom because it announces itself at launch — which is why the same `devbox run` wrapper looked fine for one role and broken for the next.

Reported twice by the same user, three days apart (#230, #305).

## The change

Roles and modes take an optional `agent` key naming what the command actually launches:

```toml
[[orchestrations.roles]]
name = "reviewer"
command = "devbox run -- codex --sandbox workspace-write"
agent = "codex"
```

Extending the launcher allow-list was deliberately **not** the approach: it is an arms race that never closes and cannot help a bespoke `run_codex.sh` at all. Substring-matching the command is worse — a script named `test-agent` would be missed, and a `codex-notes` script running Claude would false-positive.

## Precedence — three sources of truth

| source | what it is | drives badge | drives launch shape (wrap) |
|---|---|---|---|
| **declared** | the new `agent` key, re-read from disk at every spawn and every delegate | yes | **yes** |
| **derived** | `AgentType::from_command(command)` | yes | yes |
| **learned** | a hook event via `set_agent_type` | yes | **never** |

**declared > derived > frozen `spawn_agent_type` (fallback only).** Learned never participates in the launch decision.

Why declared outranks derived without reopening PRD #225 finding 1: that finding made the respawn seam derive the wrap decision from the command it is actually launching, consulting the stored identity only when the command implies no type — correct **because `spawn_agent_type` is frozen**, captured at a previous spawn and potentially stale against an edited command. A config declaration is a different kind of value: re-read from the same file, in the same pass, as the command beside it, so it cannot be stale. `codex/spawn/007` and `008` guard exactly that distinction and pass with their assertions untouched.

## The invariant this change had to not break

`src/agent_pty.rs:1750-1766` documents PRD #225 Defect 2: a hook-learned `Some(Codex)` was replayed into `SpawnOptions::agent_type` on respawn, and the same pane silently came back up wrapped — its exec line mutating after the first delegate. **A value recorded for display must never change how a pane launches.** An auditor traced this independently and confirmed no hook-supplied route into the exec decision: `set_agent_type` writes only `RunningAgent::agent_type`, never `spawn_agent_type` or `PaneRecreateIdentity`, and the single production constructor of the latter takes command and declaration from the same freshly-loaded role object.

### The badge could not come from `SessionState.agent_type`

The obvious implementation — seed the declaration into the pane's session state — was tried and is **wrong**. Three readiness gates (`src/ui.rs:3403`, `:3549`, `:4615`) spell "the agent has started" as `agent_type != AgentType::None`; each, on a non-`None` type, sets `ready_since = now` and writes after the ~500 ms spawn buffer instead of waiting for the real `SessionStart` (or its 10 s fallback). Seeding would fire the orchestrator prompt into a `devbox` that has not started Codex yet — losing it, **for precisely the population this key exists to serve**, and in the same class as PRD #225 Defect 1.

So the declared badge lives in `UiState::pane_declared_agent`, read **only** by `render_session_card` and only while the observed type is still `AgentType::None`. Nothing that decides timing sees it. That is the same observed-vs-declared split #225 made one layer down. An agent that has identified itself always wins, so a wrong declaration self-corrects rather than masking the real agent.

## Cross-version contract (rule 12): the contract did NOT move

**No `PROTOCOL_VERSION` bump, no `.breaking.md`.**

- `src/daemon_protocol.rs` has **no diff** across the whole branch.
- `StartAgent.agent_type: Option<AgentType>` already existed and its meaning is unchanged — "which AI agent this pane runs, as resolved by the spawning TUI". Only the TUI-side *source* widened, from `from_command` alone to declaration-then-`from_command`.
- The daemon's use is untouched: `agent_pty::spawn` already preferred `opts.agent_type` over deriving from the command (PRD #20 finding #19), which `codex/spawn/006` pins.
- One newly *reachable* wire value — an unrecognised declaration puts `Some(AgentType::None)` on the wire, which `from_command` could never produce. Safe: that variant carries `#[serde(other)]`, i.e. it is the universal decode catch-all every protocol version already has, and every reader treats it as "no agent", which is what the declaration means.

**Verified by the manual cross-version run**, not only by analysis: a `v0.37.2` daemon with this branch's TUI, in a sandbox with `HOME`, both sockets, the state dir **and `DOT_AGENT_DECK_LOG`** isolated. The declaration was honoured at spawn (dashboard rendered `Codex · worker`; the old daemon's process tree showed `wrap --agent codex -- ./worker.sh`), delegate routing survived, status hooks (`Thinking`, `WaitingForInput`) survived, and `work-done` retired the delegation and notified the orchestrator — twice.

**Known and accepted limitation, exercised deliberately in that run:** an old daemon paired with a new TUI honours the declaration at spawn but **not** on its own `clear = true` delegate respawn, because that half lives in the daemon's `state.rs`. Editing the role from `agent = "codex"` to `agent = "claude"` and re-delegating, the `v0.37.2` daemon relaunched through its frozen `wrap --agent codex -- ./worker.sh`, ignoring the edit. The task still routed, status and work-done still arrived, nothing was corrupted. That is a feature gap, not a compatibility break.

## Config compatibility

An existing config with no `agent` key behaves **exactly** as before, verified four ways: `#[serde(default)]` on both fields with no `deny_unknown_fields`; a unit test that asserts a keyless role's resolved type equals `AgentType::from_command(...)` **verbatim** rather than a hand-written expected value, so it cannot pass if the fall-through ever stops being the derivation; every changed seam reading `declared.or_else(<the exact previous expression>)`; and the pre-existing `codex/spawn/001`–`004` still green.

`agent = ""` / whitespace resolves to *unset* (fall through to the command), not to an explicit `AgentType::None` — otherwise "clearing" the key by emptying it would silently strip the wrapper off a perfectly inferable `codex …` command.

## Security hardening (from the audit) — note the blast radius

An auditor found that `dot-agent-deck validate` interpolated the raw, project-controlled `agent` string into stderr unescaped and unbounded. `.dot-agent-deck.toml` travels with a repository, so it arrives from clones, contributor branches and PR checkouts. A repo shipping `agent = """not-an-agent\n[error] 'trusted': validation passed"""` could forge a validation line that reads as if the deck authored it; ANSI could repaint the terminal; an overlong value could flood a CI log. Same class as issue #576 (`session_warnings`).

Fixed at `impl Display for ValidationIssue` — the single output seam — so **every** `scope` and `message` is covered by construction rather than one call site being patched. Bounded *before* escaping (cutting afterwards could sever a `\u{1b}` and emit `\u{1`), escaping rather than stripping (a value that silently loses characters reads as a *different* value), reusing `build_version_handshake::is_bidi_format_char` rather than respelling the codepoint set. Verified against the real binary with the auditor's exact payload, not only in unit tests.

**Blast radius, deliberate:** this hardens **pre-existing** fields too — mode names, orchestration names, role names, regex patterns. Output for a well-formed config is byte-identical (both helpers return `Cow::Borrowed` unchanged when there is nothing to do). Registry matching still happens on the **raw** name before any sanitizing, pinned by a test proving `claude` wrapped in RLO+ESC stays unknown while bare `claude` stays known.

## Tests

| id | tier | what it pins |
|---|---|---|
| `codex/spawn/009` | L2 PTY-attached | a declared launcher role wraps **and** the card reads `Codex` with no task delegated and no hook — the user-visible fix |
| `codex/spawn/010` | fast PTY registry | declared beats derived; survives `clear = true` re-create; a conflicting learned observation changes neither badge nor exec line |
| `codex/spawn/011` | L2 PTY-attached | the same for a `[[modes]]` agent pane (shell-injection seam) |
| `codex/spawn/012` | L2 real-agent | real Codex behind a bespoke `run-codex.sh` badges before the first prompt |
| `dashboard/pane/012` | L1 | the render fallback, **and** that an identified agent always wins over a declaration |
| `codex/spawn/007`, `008` | fast PTY registry | unchanged assertions — the #225 guards and the config-compat control |
| `mode_activation_failure_unwinds_every_pane_registration` | fast, in-`src` | a failed `open_mode_tab` leaves none of the six registrations behind — including the placeholder session, the ghost card |
| `mode_activation_failure_with_unclosable_pane_preserves_every_registration` | fast, in-`src` | a pane whose `close_pane` failed is still live, so all six survive; flipping the guard back to an unconditional purge fails this test on its first assertion |
| `repeated_mode_activation_failures_do_not_accumulate_pane_state` | fast, in-`src` | two failed activations leave the maps exactly as zero activations do — the original leak symptom, and what catches a rollback covering five of six maps |
| `a_parse_error_cannot_smuggle_escapes_out_of_the_config`, `an_ordinary_syntax_error_keeps_its_multi_line_frame` | fast, in-`src` (`src/project_config.rs`) | both drive the **real** `load_project_config` against a real file, so the ESC has to survive `toml`'s own rendering to be worth escaping; the second pins that a genuine syntax error keeps its multi-line frame |
| `multiline_escaper_*` ×4 | fast, in-`src` (`src/config_validation.rs`) | the escaper itself: control/C1/DEL/bidi neutralised, the frame's own newlines kept, the whole diagnostic bounded, a clean frame returned untouched |

The five rollback and parse-error entries are plain `#[test]`s in `src/`, with no `#[spec]` id and no `tests/CATALOG.md` entry — the established convention for this seam, matching the `spawn_failed_mode_does_not_persist_last_command` sibling they sit beside.

`cargo test-fast`: **3087 passed, 0 failed** on this head, up from the 3078 this section previously quoted. `cargo fmt --check` and `cargo clippy --workspace --all-targets --features e2e -- -D warnings` clean; `cargo xtask linkage-check` is untouched, since none of the tests added since carries a `#[spec]` id. `cargo test-e2e` on this head (`8fb030f`, recorded with `DOT_AGENT_DECK_RECORD=1`): **8815 tests run — 8812 passed (2 slow, 1 flaky), 3 failed, 0 skipped**. The +9 over the 8806 this section previously quoted are the in-`src` unit tests added since (four `multiline_escaper_*`, two `ProjectConfigError` formatting, three abandoned-pane rollback). **All three failures are real-agent tests, and every one was verified green in isolation:**

- `devin_live_001_real_interactive_turn_drives_the_card_live` — passed alone, 13.7s.
- `shell_activity_006_real_claude_bash_call_crossing_the_cap_keeps_the_badge_working` — passed alone, 137.7s. Its own harness printed `PRECONDITION NOT MET: … This is a harness flake (prompt adherence), NOT a badge regression: rerun rather than reading this as BADGE WRONG.`
- `card_stats_005_real_agent_card_narrows_without_restructuring` — the one that initially reproduced, then cleared by measurement. An `origin/main` control (`14e246e`, separate worktree) **passed** in 26.2s, and three sequential isolated runs on `8fb030f` **passed** in 24.5s / 24.6s / 24.7s — **4/4 green** once the machine was quiet. `tests/e2e_card_layout.rs` is **byte-identical** between `origin/main` and `8fb030f`, so the test did not move under this branch. The original failure coincided with a *second* full e2e suite from another worktree saturating the same 16-core box — load **75.73**, **52** concurrent real Haiku CLIs, **175** `dot-agent-deck` processes — and its shape (only `SessionStart` observed, the follow-on event never arriving inside 120s) is what CPU/API starvation produces on this test. The 24.5–26.2s spread across all four clean runs also proves it was not silently skipping: `skip_unless!(check_claude_available())` would have returned in milliseconds, so a real interactive Haiku agent genuinely ran each time. **The honest limit:** the failure was **not** reproduced under deliberate load, so contention is strongly supported but not *proven* as the cause.

Five real-Codex tests — `codex/spawn/012`, `codex_live_001`, `codex_worker_001`, `codex_hooks_001` and `delegate_009` — took the shared runtime-skip path in `tests/common/mod.rs` (`DOT_AGENT_DECK_CODEX_TEST_MODEL` unset → the harness selected `gpt-5.1-codex-mini`, which the available credentials cannot reach). That helper returns early after printing `SKIP:`, so nextest counts all five as **passes** — flagging it explicitly so the 8812 is not read as stronger than it is.

## Changelog fragments

- `changelog.d/308.feature.md` — the feature itself: telling the deck which agent a launcher command starts.
- `changelog.d/308.bugfix.md` — the two hardening entries, one section each: a config that will not parse can no longer repaint the terminal `validate` prints into, and a mode that fails to activate no longer leaves a card behind for a pane that is gone.

The second is `bugfix` rather than `misc` because both of its entries are **directly user-visible**, which is the line `pyproject.toml` draws: `misc` is for what "improves the project but users don't directly see". A terminal repainted by someone else's config, and a dashboard card for a pane that no longer exists, are both things that happen to the user in front of them.

## What a declared type unlocks

A declaration is the only way the deck can know what an agent is **before** that agent has spoken, which unlocks anything that must act at or before spawn. One concrete consumer: **#686** (a silent pane reported as undelivered when it is actually showing a first-run trust prompt) could emit a precise agent-specific message instead of a generic one. #686 did not depend on this and was fixed independently — it landed on `main` as `14e246e` while this PR was in review, and is merged into this branch. Still noted as a possibility, not a coupling.

## Follow-ups from the audit — both now fixed in this PR

Both were recorded here as deferred. Both have since landed on this branch (`207f06d`, `f44223f`), so nothing in this list is still outstanding.

1. **Failed mode activation now unwinds all six pane registrations.** A new `rollback_abandoned_pane` helper (`src/ui.rs`) removes `managed_pane_ids`, the placeholder session (via `remove_sessions_for_pane`, so a session the daemon has already bound to the pane goes too rather than leaving a second ghost card), `pane_display_names`, `pane_names`, `pane_metadata` and `pane_declared_agent`. Five of those leaked before this PR and every one of them is card-visible; the sixth is the map this PR added. The rollback runs **only when `close_pane` succeeded** — if the close failed the pane is still live, so its card, name and metadata stay visible and recoverable and the error is logged instead, matching the line PRD #92 F4 already draws for an explicit `Ctrl+W` close. The mode-**restore** arm now routes through the same helper, so a future per-pane map is added in one place rather than remembered at two call sites; it keeps its unconditional purge because it goes on to substitute a fallback pane for the same saved pane (PRD #69) and gating it would leave two cards for one restored pane — an asymmetry documented in the code.
2. **The TOML parse-error sink is escaped.** Sealed at `ProjectConfigError`'s `Display` rather than at the `eprintln!`, the same shape as the `ValidationIssue` fix above, and it matters because `validate` is only the most obvious sink: `dispatch::list_targets` folds the same error into a message rendered to a *coding agent's* terminal, and the hydration path logs it. Sealing only `main.rs` would have left both raw. Escaping is per-line with the error frame's own newlines preserved (`escape_multiline_for_terminal`, the `keep_newlines` distinction `sanitize_for_prompt` already draws), so a real syntax error stays exactly as readable as before while the quoted config line is escaped. Measured on the real binary: a hostile config's raw ESC bytes go **2 → 0** with the gutter frame intact, an ordinary syntax error's output is **byte-identical** before and after, and a 100 000-character flood renders as 4030 bytes with the true length reported. Known cosmetic limitation, on *hostile* input only: the caret line can drift, because `toml` computes the column from the raw source line and escaping expands it — the caret still points at the right column of the original file, and nothing moves on a clean config.

## Note on `tests/`

`tests/pane_close.rs` gained a single `agent: None,` field initializer — compiler-forced (`E0063`) by adding a field to a `pub` struct constructed by literal. No assertion, `#[spec]` id, `Scenario:` comment or catalog entry moved.


prageethw (dismissed):
Verified the new agent key on [[orchestrations.roles]]/[[modes]] is Option<String> with #[serde(default)] on both structs, so an absent key reproduces prior behavior exactly (pinned by a_role_without_the_key_is_unchanged). An unrecognized name resolves to AgentType::None rather than a silent guess, and surfaces as a dot-agent-deck validate warning rather than a hard error. The declared > derived > frozen spawn_agent_type precedence is carefully reasoned against the prior PRD #225 invariant and pinned by dedicated tests including a real-agent L2 test, and resolve_declared_agent is shared with wrap --agent so the two can't drift apart. The two bundled hardening fixes (terminal-injection escaping in validate output, pane-registration leak on failed mode activation) are directly motivated by this feature's own new code paths, not scope creep. Docs updated in three files plus changelog. CI fully green. LGTM.

One minor non-blocking note: docs/orchestration.md and docs/workspace-modes.md say matching is 'matched exactly and in lower case,' but the tests pin case-sensitive matching ("Codex" vs "codex" are distinct) — the phrasing could be misread as case-insensitive normalization. Worth a follow-up doc tweak, not a blocker.

prageethw (approved):
Fresh re-review at current head (7ee978742), since my prior review was dismissed by a later merge commit. Verified the precedence chain (declared > derived-from-command > frozen spawn_agent_type), the sole PaneRecreateIdentity construction site still populates agent_type from a fresh config read (no regression of PRD #225 Defect 1), the declaration is correctly kept out of readiness-gate timing, and the terminal-injection escaping + pane-registration-leak fixes are sound. Test coverage across unit/L1/L2-synthetic/L2-real tiers is thorough. One non-blocking nit: docs/orchestration.md's "matched exactly and in lower case" phrasing reads ambiguously as case-insensitive, while matching is actually case-sensitive exact-match against the lower-case name — worth a one-line wording tweak, not blocking.

Issue #610, opened 2026-08-19
PRD: Pilot Dosu for repository knowledge and documentation drift


vfarcic:
## PRD: Pilot Dosu for repository knowledge and documentation drift

**Problem**: dot-agent-deck's engineering knowledge is spread across code, PRDs, issues, contributor documentation, and agent skills, while the repository's high change rate makes retrieval and documentation drift increasingly costly. Existing automation covers CI, review, labels, stale issues, and dispatch, but does not provide one evaluated cross-source knowledge layer for maintainers and coding agents.

**Solution**: Run a bounded, reversible Dosu pilot on this public open-source repository, focused on MCP knowledge retrieval, duplicate detection, and reviewed documentation-drift findings. Keep public replies mention-only and leave stale handling, PR sizing, automatic repository write-back, and workflow labels disabled until measured evidence justifies expansion.

**Detailed PRD**: See [prds/610-pilot-dosu-repository-knowledge.md](https://github.com/vfarcic/dot-agent-deck/blob/main/prds/610-pilot-dosu-repository-knowledge.md)

**Priority**: Medium

vfarcic:
Closing alongside PR #612.

The pilot proposal aged past the tree it was written against — #612 was 301 commits behind with a conflict in `src/agent_pty.rs` — and the decision is not to carry it forward.

Not closing as "wrong": if repository knowledge and documentation drift become worth piloting a tool for again, this is cheaper to write fresh against current `main` than to revive. The branch `docs/prd-610-dosu-pilot` is kept, so the original write-up remains readable.

Issue #417, opened 2026-08-08
PRD: Persist the orchestration split ratio across restarts


vfarcic:
## PRD: Persist the orchestration split across restarts

**Depends on #342 merging.** The mechanism this issue extends (`TabManager::orchestration_split_narrow`) does not exist on `main` — it lives only on PR #342's branch. Nothing here is actionable until that lands.

## Problem

PRD #336 adds `Ctrl+l` to toggle the orchestration sidebar/pane-column split between the default 34/66 and a narrower-sidebar 25/75, as a deck-global reading preference. It deliberately does not persist: the split lives entirely in the TUI process, so detaching and starting the TUI again drops you back to 34/66 even though the daemon kept your panes alive and restored them.

Concretely, on the #342 branch:

- `TabManager::new()` hardcodes `orchestration_split_narrow: false` (`src/tab.rs:175`).
- Nothing in `config.rs`, `session.rs`, `state.rs` or `daemon_protocol.rs` reads or writes it — it is neither persisted nor on the daemon wire.
- Restored orchestration tabs seed their `split_narrow` mirror from that fresh `false`, so a restore lands on the default too.

## Why this matters

This is the same argument that moved the split from per-tab to deck-global during #342's review, applied to a different axis. From that discussion:

> sidebar width is a reading preference, not a property of which orchestration happens to be open [...] anyone who prefers the narrow sidebar re-toggles forever.

A preference that resets on every restart has exactly that cost. It bites harder than "resets on next launch" suggests, because this is an always-on daemon app where detach/reattach is a routine daily action rather than a rare cold start — the panes survive, so the reset is visible precisely when the user expected continuity.

This was agreed at the time and simply never filed:

> So I'm inclined to make it global, and to follow up separately with persistence so the choice survives a restart (that part is additive, so it doesn't need to block this).
> — https://github.com/vfarcic/dot-agent-deck/pull/342#issuecomment-5182607713

> Persistence as a separate follow-up sounds right; agreed it shouldn't block this.
> — https://github.com/vfarcic/dot-agent-deck/pull/342#issuecomment-5214592165

## The hard part is already done

[...]

PR #694, opened 2026-08-25
feat(skills): add /prd-queue to dispatch PRD work, composing the task by shape (#693)

### LINKED ISSUE #693: No skill for dispatching PRD work: whether /prd-full belongs in the dispatched instructions is decided ad hoc, and depends on the unit's shape
## Problem

There is a project-local skill for dispatching **issues** (`.claude/skills/issue-queue`) and one for **PRs** (`pr-review-queue`), but **none for PRDs**. So every PRD dispatch composes its task ad hoc, and the decisions that composition has to make are currently made from memory or not at all.

That matters because a PRD dispatch has a decision an issue dispatch does not: **whether `/prd-full` belongs in the dispatched instructions** — and the correct answer depends on the unit's shape.

## The decision the skill has to encode

**`/prd-full` is a single-agent skill.** Its own description: *"Run a PRD end-to-end autonomously — start, iterate until done, create a PR, and wait for its CI + bot reviews to settle before reporting. Stops before merge for manual validation."* It has no notion of roles or delegation, and it does not cover this project's additions: the Telegram notifications, the demo reel (PRD #180), the `DOT_AGENT_DECK_RECORD=1` e2e gate, the test-plan gate, or the tester→coder TDD chain.

So:

- **Dispatching a single agent on a PRD** → `--task "Execute the /prd-full skill for PRD <n>."` is right and complete. The unit has the repo, the PRD and the skill.
- **Dispatching a team on a PRD** → `/prd-full` is the *wrong* instruction. Pointing an orchestrator at a single-agent skill invites it to do the work itself instead of delegating, which defeats the reason a team was dispatched. The orchestrator role's `prompt_template` (`.dot-agent-deck.toml:87–157`) already **is** this project's orchestration-aware expansion of the same lifecycle.

`/prd-full` is a `dot-ai-*` synced mirror, so per CLAUDE.md rule 13 the fix is never to patch it — the composition decision belongs to the dispatcher.

## What a dispatched orchestrator actually receives (verified)

`spawn.rs:776` calls `orchestrator_context::prepare_orchestrator_prompt`, which writes **one file**, `<cwd>/.dot-agent-deck/orchestrator-context.md`, and hands the pane a pointer to it. Measured on the live `#308` unit (24,986 bytes):

| Lines | Content | Source |
|---|---|---|
| 1–71 | the orchestrator workflow, incl. `## Notifications (PRD #126)` at 44 | `.dot-agent-deck.toml:87–157` |
| 72–79 | `## Available agents` | role `description`s |
| 80–130 | `## Delegation protocol` | compiled into the binary |
| 131–142 | `## Important` | ditto |
| 143+ | `## Your task` — the dispatched task verbatim | the `--task-file` |

Two consequences the skill must handle:

1. **There is no formal precedence — only ordering**, template first, task last. If a task says "execute `/prd-full`" while lines 1–71 say "follow steps 1–7", nothing arbitrates but the model. The composed task must therefore **state the precedence explicitly** rather than leave it implied.
2. **Heading levels interleave.** A task written with `##` sections lands with those sections as *peers* of the template's `## Notifications` and `## Delegation protocol`, not nested under `## Your task`. The composed task should either sit a level below or open with an explicit "everything under `## Your task` is this unit's specific job".

Workers combine differently and need no change here: `compose_worker_task_file` (`src/state.rs:2224`) builds `{role_template}\n\n## Task\n\n{task}` plus a work-done footer, per delegation.

## The notification gap, and where it actually is

A **team** does get the notification instructions — they are line 44 of that file. A **`--single`** dispatch gets none, because there is no role template at all.

That is the shape that would run `/prd-full`, and it is the shape most in need of the notification: dispatch is fire-and-forget with **no return edge**, so a unit that finishes with "PR open, checks settled, stop" tells nobody. The template's own selection rule is *"every pause where the human may have walked away"* — a dispatched unit is that case by construction. So the composed task for a single-agent PRD dispatch should carry a notify-on-finish instruction.

## Proposal

Add `.claude/skills/prd-queue/SKILL.md`, a project-local sibling of `issue-queue`, owning:

- listing open PRDs (the `PRD` label — note `issue-queue` step 2 *excludes* that label by construction) and eliminating those already in flight, reusing `issue-queue`'s three-way check (closing refs, PR bodies, dispatch branches/worktrees)
- **asking the shape per unit**, not per batch (see #674)
- composing the task **by shape**: single → `Execute the /prd-full skill for PRD <n>` plus notify-on-finish; team → the role template governs, with the precedence stated explicitly so it does not compete with `/prd-full`
- carrying the stop condition every dispatched unit needs, since nothing watches the pane
- claiming, naming `prd-<n>`, verifying the name is free, dispatching, and reporting worktree paths

It should reuse `issue-queue`'s hard-won safety rules verbatim rather than re-derive them: `--task-file` over `--task`, fresh non-title-derived slug, fence issue/PRD-derived text as untrusted data, re-check state immediately before each dispatch, and never delete a branch to free a name.

## Notes

- Project-local skill (no `dot-ai-` prefix), so it is ours to edit and will not be clobbered by a skill sync — CLAUDE.md rule 13.
- Deliberately **not** a CLAUDE.md rule: CLAUDE.md governs the repository, while this is a dispatcher decision procedure, which is what the queue skills already own.
- `dot-ai-worktree-prd` covers the older non-dispatch worktree flow and is unaffected.

vfarcic:
Closes #693.

Adds `.claude/skills/prd-queue/SKILL.md` — the third sibling of `/issue-queue` and `/pr-review-queue`, for PRDs. Nothing under `src/`, `.dot-agent-deck.toml` or the existing queue skills is touched, and no CLAUDE.md rule is added (#693 explains why this deliberately lives in a skill: CLAUDE.md governs the repository, while this is a dispatcher decision procedure).

## The decision it encodes

`/prd-full` is a **single-agent** skill — *"run a PRD end-to-end autonomously"* — with no notion of roles or delegation.

- **`--single`** → `Execute the /prd-full skill for PRD <n>` is right and complete, plus the notify-on-finish instruction below.
- **`--orchestration`** → `/prd-full` is the *wrong* instruction and the team task says so outright. Pointing an orchestrator at a single-agent skill invites it to run the lifecycle itself instead of delegating, which defeats the reason a team was dispatched. The orchestrator role template (`.dot-agent-deck.toml:87–157`) already **is** this project's orchestration-aware expansion of the same lifecycle.

`/prd-full` is a `dot-ai-*` synced mirror, so per CLAUDE.md rule 13 it is never patched; the composition decision belongs to the dispatcher. The skill therefore asks the shape **per PRD** (#674) and writes a different task document per answer.

## The two mechanics, verified against `origin/main`

1. **No formal precedence, only ordering.** `spawn.rs:776` → `prepare_orchestrator_prompt` (`src/orchestrator_context.rs:186`) writes one file, `<cwd>/.dot-agent-deck/orchestrator-context.md`: role template, available agents, delegation protocol, `## Important`, then `## Your task` with the dispatched task last. Nothing arbitrates between them but the model, so the team task states the precedence explicitly and names the one real conflict rather than gesturing at it.
2. **Heading levels interleave.** Measured on the live `#308` unit's 24,986-byte context file: `## Notifications` at 44, `## Available agents` at 72, `## Delegation protocol` at 80, `## Important` at 131, `## Your task` at 143 — and the task's own `## Goal` at 153, `## Context (untrusted data…)` at 170 and `## Stop condition` at 207, all at that same level. **The rule chosen is demotion** (no task file emits a `##`), and the skill says why: the section whose scope must not float to document level is the untrusted-data fence, and heading level is the only one of the two candidate mechanisms that a reader gets right without reading the prose. The team task also carries a precedence line, but doing a different job — demotion carries *scope*, the precedence block carries *authority* — and the skill says that too rather than quietly picking both.

Workers need no change: `compose_worker_task_file` (`src/state.rs:2224`) wraps each delegated task under its own role template separately, which is also why the team task deliberately omits the gate list — coder and tester already have it.

## The notification asymmetry

A team gets the notification instructions from its role template. A `--single` dispatch has no role template at all, and is the shape most in need of them: dispatch is fire-and-forget with **no return edge**, so a unit ending with "PR open, checks settled, stop" tells nobody. The single task carries the contract from `.dot-agent-deck.toml` — the explicit `chat_id` from `TELEGRAM_CHAT_ID`, the ban on `telegram_get_updates`, fire-and-forget, the `notify-log.md` line — restated inline because those are security requirements that must hold even if the unit never opens the TOML. It maps onto that block's existing vocabulary rather than inventing strings, and explicitly forbids `DONE: merged & closed`, which would be false for a run that stops before merge.

## Carried over, and the one inversion

Reused verbatim from `/issue-queue` rather than re-derived: `--task-file` over `--task`, file-writing tool never a heredoc, a fresh slug never derived from the PRD's title, PRD-derived text fenced as untrusted data with references preferred over pasted contents, verification against `origin/main` after `git fetch`, the three-way in-flight check, claim-by-re-read-then-read-back, check the branch name is free, never delete a branch to free a name, re-check before *each* dispatch.

**The inversion is called out where a reader would otherwise copy the filter across:** `/issue-queue` step 2 emits `select((.labels|map(.name)|index("PRD"))==null)`; this skill selects **on** the label. The skill says the mistake fails silently and tells you to check the `==null` is gone.

Two additions that are specific to PRDs, both measured on 2026-08-25:

- **A `PRD` label is not a PRD.** 34 of the 39 open PRD-labelled issues have a `prds/<n>-*.md` document on `origin/main`; five do not (#610, #417, #239, #193, #183). The whole lifecycle reads that document, and `/worktree-prd`'s `create.sh` aborts without it.
- **The in-flight check matches `agent/dispatch-(prd|issue)-<n>`.** #421 carries the `PRD` label and is live under `agent/dispatch-issue-421`; a `prd-`-only grep returns nothing for it. Its dispatch produced merged PR #464, which landed the *document* and left the PRD unimplemented — so the issue is open and available while the branch name is permanently spent.

It also adds a step 0b for #674's base-freshness gap (surface a stale checkout and ask; never a unilateral `git pull`), worded to reconcile with step 0 rather than contradict it. Doing the same for the two sibling skills stays with #674.

## Gates

- `cargo clippy --workspace --all-targets --features e2e -- -D warnings` — clean. All three flags used.
- `cargo test-fast` — **3063 passed, 0 skipped**. Not inert for a skill edit: `xtask/linkage-check` has tests that read files under `.claude/skills/`.
- `cargo test-e2e` — **deliberately not run.** The final diff is Markdown-only: two new files, `.claude/skills/prd-queue/SKILL.md` and `changelog.d/693.misc.md`, with zero lines changed under `src/`, `tests/` or any build file. Nothing the e2e tier spawns can observe this change, so the most expensive gate in the repo (CLAUDE.md rule 5) has nothing to add here.
- **Rule 12** — no `PROTOCOL_VERSION` bump and no `.breaking.md` fragment. This touches no `src/`, so it cannot move the TUI↔daemon protocol/handler contract on either axis: not the wire shape, and not a same-wire shift of meaning. Stating the conclusion rather than leaving it unaddressed.
- **Rule 4** — no user-visible TUI behaviour changes. No pane, status, prompt, focus, layout, mode, embedded pane or hook-delivery surface moves, so no L1 or L2 test is required. Again, stated rather than left ambiguous.
- **Rule 10** — no hard-wrapped prose; every prose paragraph in both files is one line.
- **Changelog fragment** — `changelog.d/693.misc.md`, via the `dot-ai-changelog-fragment` skill. `misc` per `pyproject.toml`'s own description ("internal improvements, CI/CD, tooling … users don't directly see it"); the breaking question was asked first and answered no, per rule 12 above. **Worth a reviewer's call:** both sibling skills shipped with *no* fragment at all (`e062a33` for `/issue-queue`, `2ad70ca` for `/pr-review-queue`), so this deviates from precedent. It is here because #693's brief asked for one; drop it if you would rather match the siblings.

Branch is cut at `c701932`, one commit behind `origin/main` (`cfe1a25`, a Renovate `open` crate bump) — noted for honesty per the step 0b this PR adds. Markdown-only, so it neither conflicts nor is affected.

https://claude.ai/code/session_01BzRBZgS4DQmQkBUg42JD7F


prageethw (approved):
Reviewed the new .claude/skills/prd-queue/SKILL.md as the third sibling of /issue-queue and /pr-review-queue. The core design decision — composing a different task document for --single vs --orchestration dispatch, since /prd-full is single-agent-only and wrong for a team dispatch — is well-reasoned. Confirmed the selection filter is exactly the inverse of /issue-queue's (select(...index("PRD"))==null), so there's no functional overlap with the existing sibling skills. The skill is careful about injection/safety: task content is always written via file-writing tool (never heredoc/shell redirection), branch/slug names are self-invented from [a-z0-9-] plus the issue number rather than derived from PRD title/body, and PRD-derived text is explicitly fenced as untrusted data both in the composed task and when printed to the runner's terminal. No src/tests changes, changelog fragment present, CI green. LGTM.

Issue #699, opened 2026-08-25
opencode_auto_submits_daemon_injected_prompt fails under load: readiness gate measures output quiet, not input readiness


vfarcic:
`opencode_auto_submits_daemon_injected_prompt` (`tests/e2e_delegate_work_done_chain.rs:355`) fails deterministically once the box is oversubscribed, and passes deterministically when it is not. It is **pre-existing** — not caused by any branch currently in flight.

## Measured, both sides of a branch point

Found during the pre-PR e2e tier for #318/#319 and investigated by building the branch's base commit in a separate worktree and running the single test by filter, interleaved so both samples saw the same machine conditions.

**Quiet box (1-min load 2–9):**

| | result |
|---|---|
| branch ×3 | 3/3 **PASS** (~10–13s each) |
| base ×3 | 3/3 **PASS** (~10s each) |

**Loaded box (64 busy-loops on 16 cores, ~4× oversubscription, matching the ~4.7× of the tier where it first failed):**

| | result |
|---|---|
| branch ×3 | 0/3 — **FAIL** (~103–106s each) |
| base ×3 | 0/3 — **FAIL** (~104–108s each) |

Identical symptom on both sides. The hogs were pure CPU busy-loops with no memory pressure.

## Symptom

The OpenCode splash with an **empty input box and no conversation at all** — not a prompt sitting unsubmitted in the box, and not a user message awaiting a reply:

```
OpenCode did not auto-submit the daemon-injected single-line prompt
(no '4444' reply rendered).
   ┃  Ask anything... "Fix broken tests"
   ┃  Build · GPT-5.4 mini OpenAI
```

## Mechanism

The test's readiness gate is a **quiet-output heuristic, not an input-readiness one**. `wait_until_agent_output_settled` (`tests/common/mod.rs:7060`) returns once the PTY snapshot has not grown for 1500 ms, subject to a 6 s `min_since_first_output` floor and a 30 s cap.

Under load, OpenCode's startup stalls for longer than 1.5 s *while still initialising*. The helper declares settled, `write_to_pane_and_submit` fires into a TUI that has not yet attached its input reader, and the keystrokes are dropped. That is why the failing screen shows an **empty** box rather than an unsubmitted prompt — the bytes never reached a reader, so there was nothing to echo.

[...]

Issue #698, opened 2026-08-25
generation_ownership's pane-only arm ignores pane_handed_over, so the ownership layers disagree in the admitting direction


vfarcic:
Split out of the #318/#319 PR (branch `agent/dispatch-issue-318-319`), where it is **residual 5**. Found by a third-pass reviewer reading the code, then **confirmed by a test** rather than left as an inference.

## The gap

That PR's provenance fix works by making two layers share one retirement rule instead of each writing its own: `token_still_live`, `pane_is_protected` and `generation_ownership`'s keyed arm all apply "live records may speak; retired ones may speak for their pane until `pane_handed_over` or `pane_claimed_by_other`".

`generation_ownership`'s **pane-only** arm — `(Some(pane), None)`, the arm a **token-less** event reaches — is now the single place that does not. It matches *any* record still naming the pane, with no `exited` and no `pane_handed_over` filter.

## Reproduction (no respawn involved)

Pinned by `agent_pty::spawn_tests::a_handed_over_then_reaped_pane_is_unprotected_yet_still_owned`:

1. R1 spawns on pane P and exits naturally; its record lingers unreaped.
2. R2 spawns on P via `StartAgent` (allowed — R1 is exited); the publish loop sets `R1.pane_handed_over = true`.
3. R2 is closed via `StopAgent`; `close_agent` removes only R2. R1 still lingers as `{exited, pane_handed_over}`.

Then `generation_ownership(Some(P), None)` returns **`Owned`** while `pane_is_protected(P)` returns **false**, so a token-less event naming P is classified `Foreign` and applied. The two layers disagree in the admitting direction.

## Why it was left open

- P is functionally defunct — both generations are gone.
- Pre-fix, `manages_pane` (`!exited`) also answered `false` there, so nothing regressed.
- A `Foreign` card is exactly what a same-user peer can already mint for any pane id this daemon never managed, so the incremental exposure is nil. It is one more instance of #601's named remainder, not a new class.

Both `pane_is_protected` docs now state this exception rather than denying it; an earlier draft claimed the guard was true "in exactly the cases where the ownership layer would otherwise say `Owned`", which was one case too strong.

## Why it is not a one-line fix

[...]

Issue #697, opened 2026-08-25
Foreign session and pane growth on the hook socket is unbounded


vfarcic:
Split out of the #318/#319 PR (branch `agent/dispatch-issue-318-319`), where it is **residual 3** — assessed and declined for that PR rather than deferred silently.

## The gap

#319 bounds the hook socket's **transport**: `MAX_HOOK_LINE_BYTES` (64 KiB) per line, `MAX_HOOK_CONNECTIONS` (128) concurrent, with a compile-time assertion holding the reader's own buffers under the 16 MiB `MAX_FRAME_LEN`. That is a real bound and it is the one #319 asked for.

It does **not** bound retained state. Token-less foreign `SessionStart` events are deliberately admitted (the #601 / #543 external-agent compat path). Each fresh pane id auto-registers and each fresh `session_id` inserts another `SessionState` retaining the event and its strings. There is no global cap on sessions or on foreign panes, and one admitted connection can stream unlimited sub-64-KiB lines. So a same-user peer can grow daemon memory with a stream of *valid, bounded* lines carrying unique ids.

The #319 changelog fragment and PR body were corrected to scope the 8 MiB figure to the reader's buffers precisely because of this — neither claims the hook socket can no longer exhaust the daemon.

## Why a cheap cap was not available

The eviction *policy* is the design problem, not the mechanism:

1. Sessions for managed panes must never be evicted — that deletes a live card.
2. By the time `AppState::apply_event` reaches its `or_insert_with`, a flood's session is **indistinguishable** from a real external agent's. An LRU would therefore silently delete genuine foreign cards under attack — the exact compat path #601 and #543 say not to disturb.
3. `apply_event` is also the **TUI's** path (a TUI installs no ownership oracle), so a cap there changes TUI behaviour too, not just the daemon's.
4. Capping `sessions` alone would not cap the growth: `managed_pane_ids` gains an entry per foreign `SessionStart` on the same path, and pane auto-registration is a second retained-state channel.

## Threat model

Same-user, as with #318 — the socket is owner-only. A same-user process already has code execution, so this is a hardening/DoS-surface issue rather than a privilege boundary. It is filed because this repository explicitly treats same-user peers as a DoS boundary elsewhere (see #527 for the same shape on the attach socket's delivery ledger).

## Suggested direction

[...]

Issue #696, opened 2026-08-25
Hook-socket verbs other than AgentEvent are unauthorized by pane (GetSeed first)


vfarcic:
The hook socket's line and connection bounds cover every verb (they run before `DaemonMessage` parsing), but the #318 provenance guard covers `AgentEvent` only — `admit` runs on that branch alone. `Delegate`, `Dispatch`, `WorkDone`, `GetSeed` and `ListTargets` are handled ahead of it and take their `pane_id` from the payload verbatim, so any same-user process that can reach the socket can still act as a pane it does not own. **`GetSeed` is the sharpest and should be fixed first:** `take_pending_seed_native` atomically removes the pane's pending prompt *and* marks it natively delivered, so one forged request both steals potentially sensitive prompt text and suppresses the intended delivery — the pane never receives its seed and the daemon believes it did. `Delegate` and `Dispatch` can route work or create a whole worktree plus an agent attributed to another pane; `WorkDone` can retire another pane's outstanding delegation and its idle watch; `ListTargets` is read-only but leaks the pane's cwd. Pane ids are readily available from the deterministic attach endpoint's unauthenticated `ListAgents`. The token machinery generalises with little work — each signal gains an optional `agent_token` and the same derive-don't-compare treatment `AgentEvent` now gets — but doing it inside #318 would have been unreviewed scope creep, since both issues scope provenance to hook events. These are pre-existing broken-access-control paths, not regressions from #318.

Split out of the #318/#319 PR (branch `agent/dispatch-issue-318-319`), where it is named as residual 2. Found by the security audit of that work; ranked first among that PR's residuals.

PR #700, opened 2026-08-25
Bound the hook socket, and bind a hook event to the pane entitled to send it

### LINKED ISSUE #318: Daemon does not bind hook-event provenance: any same-user process can drive another pane
## Problem

`dot-agent-deck agent-event` reads the caller-controlled environment variables `DOT_AGENT_DECK_PANE_ID` and `DOT_AGENT_DECK_AGENT_ID` (`src/main.rs:704-748`) and forwards them as-is. The daemon neither verifies that pair against its own registry nor derives it independently from the connection (`src/daemon.rs:926-984`, `src/event.rs:451-546`).

Concretely: one agent process, or any other same-user process, can run

```
DOT_AGENT_DECK_PANE_ID=<other-pane> DOT_AGENT_DECK_AGENT_ID=<fake> dot-agent-deck agent-event --type running
```

and, given an ordinary current timestamp, retire another pane's card out from under it.

## What is already correct — do not over-rate the severity

- The hook endpoint is **owner-only**. On Unix the socket binds atomically at mode `0600` (`src/platform/ipc/unix.rs:121-130`); on Windows it uses a current-user-only DACL.
- Foreign OS users are excluded entirely — this is not remotely or cross-user exploitable as-is.
- The actual exposure is that **every managed agent runs as the owner and is handed the socket** — so the trust boundary that matters here is agent-to-agent (or agent-to-any-other-owner-process) within the same account, not user-to-user.

## Inherited, not introduced by #284

This predates #284 (https://github.com/vfarcic/dot-agent-deck/issues/284, shipped in PR #314). Before #284, the same process could already forge a `SessionStart` via `dot-agent-deck hook` or raw JSON on the socket, and that path was — and remains — intentionally timestamp-unguarded. #284's non-`SessionStart` retire path makes this reach more ergonomic (a plain `agent-event --type running` now suffices where a full `SessionStart` payload was needed before), but it does not create a new trust boundary.

## Remote transports do not currently increase severity

- SSH remote connect starts the TUI/daemon on the authenticated remote account and does not expose this socket over the network (`src/connect.rs:637-645`).
- Kubernetes connect is currently rejected as unsupported.

## MUST-REASSESS TRIGGER

**Re-evaluate this issue before any Kubernetes or network hook transport ships.** That is the condition under which this stops being a local-only, same-user concern and becomes a real cross-boundary exposure.

## Rule-12 warning for whoever implements this

Adding daemon-side provenance verification is plausibly a **semantic break behind a stable wire** — an older agent binary may not satisfy a newer daemon's provenance check even though the wire shape is unchanged. Expect to run the CLAUDE.md rule 12 cross-version manual test, and possibly a minor version bump; see [`docs/develop/versioning.md`](https://github.com/vfarcic/dot-agent-deck/blob/main/docs/develop/versioning.md).

## References

- Issue #284: https://github.com/vfarcic/dot-agent-deck/issues/284
- PR #314: https://github.com/vfarcic/dot-agent-deck/pull/314
- `src/main.rs:704-748`, `src/daemon.rs:926-984`, `src/event.rs:451-546`
- `src/platform/ipc/unix.rs:121-130`, `src/connect.rs:637-645`

vfarcic:
# Bound the hook socket, and bind a hook event to the pane entitled to send it

Closes #318. Fixes the transport half of #319.

Two defects on the same socket, fixed together because they are the same sentence: **what the daemon will read**, and **whose word it will take for it**. The bounded transport is #319; the provenance binding is #318.

This branch is **stacked on PR #677** (`agent/dispatch-issue-670`), not on `main`. #677 sanitizes `display_name` at hook ingest in `src/state.rs` — the same event, one layer over. Editing both in parallel would have produced a conflict in the ingest path with no way to review either change on its own, so this one sits on top. Merge #677 first.

---

### #319 — the reader was unbounded, in three dimensions

`run_hook_loop` iterated `BufReader::lines()`, which accumulates a newline-free payload without limit — a peer that writes bytes and never a `\n` grows daemon memory until the process dies, taking every pane's agent with it — and it spawned one task per accepted connection with no cap on how many were outstanding. Fixing either alone leaves the other open.

The third dimension is not in the issue text and was found by the security audit of the first implementation: **a cap without a deadline converts a memory-exhaustion bug into a function-denial bug.** The connection permit is held for the whole connection, so 128 peers that connect and send *nothing* — or trickle bytes forever under the line cap — deny every later hook, `GetSeed`, `Delegate`, `Dispatch`, `WorkDone` and `ListTargets` connection, with no memory pressure and no natural recovery. Reply writes had no deadline either, so a peer that stopped reading pinned a permit from the other direction once its receive buffer filled.

| Bound | Value | Why |
|---|---|---|
| `MAX_HOOK_LINE_BYTES` | **64 KiB** | An ordinary event measures **300–900 bytes** — a 36-byte `session_id`, a pane id bounded by `PANE_ID_ENV_MAX_LEN`, a `tool_detail` already truncated to 120 chars, a `user_prompt` already truncated to 200, and a 64-hex-char token. The one field with no producer-side cap is `metadata["bash_command"]`, the untruncated command behind a `ToolStart`; a heredoc-bearing one-liner is the realistic tail and a few KiB is a large one. 64 KiB is ~70× a typical event and an order of magnitude past the worst legitimate `bash_command`. Deliberately **not** the 16 MiB `MAX_FRAME_LEN`: that frame carries PTY scrollback, a hook line carries a small structured record. Over-limit **closes the connection**; the bytes are never truncated and re-parsed, because a padded prefix must not smuggle a short valid event in behind it. |
| `MAX_HOOK_CONNECTIONS` | **128** | Hook connections are short-lived by construction (connect, write one line, exit), so a real fleet's steady state is a handful. The long-lived ones are the reply-bearing verbs on the same socket, and a `Dispatch` holds its connection across a `git worktree` add plus a spawn — seconds. 128 leaves room for every role of several concurrent orchestrations to be mid-dispatch while a whole fleet's hooks land. Excess is **closed promptly, never queued**: queueing converts a connection flood into unbounded pending state, which is the exhaustion the cap exists to prevent. |
| `HOOK_LINE_TIMEOUT` | **30 s** | On the WHOLE `next_line()` call, not on one `read`. A per-read idle timeout resets on every byte moved, so it never fires against a peer dribbling one byte every 29 s — the same shape the byte cap already refuses to be beaten by. Every legitimate producer writes its line in a single `write_all` immediately after connect and then half-closes; the genuinely long-lived case (a `Dispatch` doing a worktree add plus a spawn) is spent inside the daemon's handler and is **not measured here at all**. So 30 s is some three orders of magnitude above the only legitimate wait there is, which leaves room for a loaded machine while bounding a wedged permit to half a minute rather than forever. It is not meant to be tight: an attacker who reconnects every 30 s can still hold the cap, and no achievable value changes that. What it buys is that the wedge cannot outlive the attacker. |
| `HOOK_REPLY_WRITE_TIMEOUT` | **10 s** | `Delegate` / `GetSeed` / `ListTargets` write a few hundred bytes to a local socket, which completes in microseconds unless the peer stopped reading. Shorter than the read deadline because there is even less reason to be slow — the daemon already has the bytes. A failed or timed-out reply now also **closes** the connection instead of looping back to read, which used to hold the permit for another full deadline for no possible benefit. |
| `MAX_LOGGED_HOOK_LINE_BYTES` / `MAX_RAW_PAYLOAD_LOGS_PER_CONNECTION` | **512-byte payload prefix / 5 per connection** | Every malformed line was logged verbatim up to 64 KiB, so a line loop grew `deck.log` without bound. The cap keeps a **512-byte prefix and then appends a short `…[truncated, N bytes total]` marker**, so the logged value is deliberately a little over 512 rather than a whole line: that is over half of even the largest ordinary event (300–900 bytes measured), and the head of a payload is where the diagnostic value is, at 128× less cost in the pathological case. Five payloads is enough to diagnose a misconfigured hook (which repeats the *same* payload) and too few to be an amplifier; the one-line "further payloads will not be logged" summary is emitted from **both** logging branches, so a budget spent on unrecognized `event_type`s no longer goes silent. The excess-connection warning is rate-limited to one line per 10 s carrying a suppressed count. |

`MAX_HOOK_LINE_BYTES * MAX_HOOK_CONNECTIONS <= MAX_FRAME_LEN` is held by a `const _: () = assert!(…)`, so widening either constant past the budget fails the build rather than a test.

### #318 — being able to connect was authority to speak for any pane

`dot-agent-deck agent-event` reads `DOT_AGENT_DECK_PANE_ID` from its own environment and forwards it; the daemon neither verified that against its registry nor derived it from the connection. Any same-user process could run `DOT_AGENT_DECK_PANE_ID=<someone-elses-pane> dot-agent-deck agent-event --type running` and drive another pane's card. The socket is already owner-only (`0600` / current-user DACL), so file permissions are not the answer.

The daemon now mints a **256-bit CSPRNG capability token** per agent at spawn, injects it into the child's environment as `DOT_AGENT_DECK_AGENT_TOKEN` (with the same strip-then-inject the pane id already gets, so a deck pane never inherits its parent's token), and on ingest **resolves the token to its own `(pane_id, agent_id)` and uses that**. The payload's claim is *overwritten*, never compared — a valid token for pane A on an event naming pane B binds to A, so B is never driven. "Compare and trust" gets that wrong the moment the comparison is skipped, reordered, or made lenient. The claim is used only to make the daemon *stricter*.

Fail-closed scope is **protected** panes only: a token-less or unknown-token event naming a pane this daemon still holds — live, or retired with no successor, or mid-reservation/close/respawn — is **refused**; the same event naming a pane it does not hold still registers a foreign card exactly as before. That asymmetry is deliberate and is #601's named remainder.

**The token reaches exactly two places: the spawned child's environment, and the `StartAgent` reply to the peer that spawned it.** There is no request that fetches it afterwards — see "What the audit changed" below.

Four lifecycle properties the audit forced into the code rather than into the prose:

- **Provenance asks the admission layer's question, not the routing layer's.** On PTY EOF `pump_reader` only sets `exited`; the record lingers until something reaps it, which for a pane nobody closes is never. The refusal guard used to ask `manages_pane` — "is an agent running here" — which answered *no* for such a pane, so a token-less event naming it was classified `Foreign`, kept its payload-claimed pane, and was then **positively accepted** by the older `generation_ownership` layer, because a retired generation keeps its pane until another claims it. It is now `pane_is_protected`, which is true for exactly the panes that layer would accept for, plus the reservation / cleanup / respawn holds. Token resolution is tied to the same rule (`!pane_handed_over && !pane_claimed_by_other`, i.e. `generation_ownership`'s own retirement rule), so the agent's own final `Idle`/`SessionEnd` racing its PTY EOF still binds, while a replayed token cannot bind once a successor has taken the pane. See "What the re-audit changed" below — the first attempt at this got it backwards.
- **A respawn keeps the pane.** `respawn_agent_for_pane` removes the outgoing record first and can then spend the full 3 s termination grace before reserving the pane again. Across that window the pane had no record, reservation or cleanup hold, so a token-less event naming it was classified foreign and broadcast — a window whose width is an attacker-visible constant. A new `respawn_holds` map covers it and blocks nothing (deliberately separate from `cleanup_holds`, which *does* refuse a spawn). Its value is a per-pane **reference count**: as a set, any guard's drop removed the single entry, so two overlapping respawns had the loser's early `NotFound` return clear the winner's still-live hold and re-open the window.
- **Pane-less events consult the registry.** The rule "a process that manages no panes is watching external agents and takes their pane-less events" was evaluated as `managed_pane_ids.is_empty()` — but an ordinary `StartAgent` deliberately does not add to that set, so a daemon running a whole fleet still read as "manages no panes". It now asks the ownership oracle, with `Unknown` denying (this is a branch where the *absence* of a claim grants).
- **The capability field never reaches the log, in any JSON spelling, on either branch.** Two branches logged the original raw line after `admit` had stripped the typed value, which defeated `AgentToken`'s redacted `Debug` entirely, and a log file is a longer-lived and easier source than `/proc/<pid>/environ`. They still redact **differently, because they hold different things**, but neither is now a spelling the other catches. The unknown-`event_type` branch has a payload that **did** decode, and a textual scan is provably wrong there: JSON permits escaped member names, `"agent_\u0074oken"` decodes to the key `agent_token`, so serde populates the field and `admit` honours and strips it as a real capability — while the raw line contains no matching substring. That branch re-renders the payload from the parsed JSON, projected onto an allowlist of the event's own members, which drops every spelling of the token field *and* a token smuggled under a member the event does not have. The malformed branch has only bytes (it is reached *because* the line did not parse), so it keeps a textual scan — but the scan now walks JSON string literals and asks serde what each **decodes to**, instead of matching one spelling. That closes what the third-pass audit called the steerable branch: a line can be valid JSON and still fail `AgentEvent` decoding (a duplicate typed member, a member of the wrong type), or fail to be JSON at all (a trailing comma), and the *sender* picks which by how they break it — so a field redaction that only worked on one branch was a redaction the sender could opt out of. What neither branch can do is value-scoped, and it is stated rather than implied: a token a peer copied into `session_id`, `tool_detail` or a `metadata` value is indistinguishable from data and is logged.

---

## What the audit changed, and why it is worth reading

The first implementation added an `AttachRequest::GetAgentToken` verb so the test harness could speak for panes the *deck* spawned. It was flagged as a departure at the time and approved on the reasoning that an attach peer already holds a strictly stronger capability (it can `StartAgent` arbitrary commands and type into any PTY, #601), so handing out a status-reporting capability is not a new exposure.

That reasoning is true about **impact** and misses the **mechanism**. `ListAgents` is unauthenticated on the same owner-only socket, so the complete exploit is `ListAgents` → `GetAgentToken(victim_id)` → token-bearing event, which `admit` then derives and accepts for a pane the caller never spawned. The token stops being evidence of *which process sent the event* and becomes a value any same-user process can ask for — which makes #318's central claim false, not merely narrower.

The verb is removed. It had zero production callers. Tests now obtain the token the way a real participant does: a test that spawns its own agent reads it off the `StartAgent` reply, and a pane the deck spawned **publishes its own injected token** from its environment before `exec`ing its stub (the shell form of what `pane_input_007`'s in-pane Python stand-in already did). That is strictly more faithful than a retrieval verb, because it also proves the spawn-time injection reached the child. No bypass, no test-only escape hatch, no debug-build-only variant — all three reopen the hole.

---

## What the re-audit changed, and why the first fix looked closed

A second security audit of the audit-fix commit found two of the seven findings still open. Both are worth recording, because both are the same failure mode: **a test that pins a predicate rather than an outcome.**

**The revocation fix was on the wrong axis.** "A natural exit does not revoke authority" was closed by filtering `!exited` in `resolve_agent_token`, and checked by a unit test asserting `!manages_pane(pane)` and `resolve_agent_token(..).is_none()`. Those are two registry predicates agreeing with each other, and neither is a refusal — in the classifier `!manages_pane` is precisely what *selects* `Foreign`, on which the payload's claimed pane survives, the daemon applies the event, and `generation_ownership` accepts it because a retired generation keeps its pane. So a token-less event naming an exited pane drove its card indefinitely, with the test green and the stale token not even needed.

Worse, revoking on `exited` bought nothing end to end even on its own terms: the agent's own final report, racing its PTY EOF, had its token revoked and then landed anyway down the `Foreign` path. The fix is to make provenance ask the question the layer *behind* it asks — `pane_is_protected` — and to tie token resolution to the same retirement rule, so the two layers share one rule instead of each writing their own, and the disagreement finding 1 exploited is closed rather than moved. (Not a *total* equivalence — see residual 5 for the one corner that remains, and where it is pinned.) The new coverage runs `admit` then `apply_event` and looks at the card, with an anti-vacuity check that the forged event moves the card when provenance is bypassed.

**The log redaction was textual on a branch that had a parsed value.** Scanning for the literal `"agent_token"` is the only thing available on a line that did not parse, and it is provably insufficient on one that did: `{"agent_\u0074oken":"…"}` is a valid spelling of the same member, serde populates the field from it, `admit` honours and strips it, and the scanner sees nothing to redact — so the unknown-`event_type` branch logged a live capability verbatim, with truncation irrelevant at any payload under 512 bytes. The branch that decoded now re-renders from the parsed JSON instead.

Three smaller things landed with them: the respawn hold became a reference count (a shared marker let one holder release another's), the "further payloads will not be logged" summary is emitted from both logging branches rather than one, and the prose that overclaimed — "capped at 512 bytes", "past a whole ordinary event", "F9 fixed", "an event it does not own" — is corrected here and in the changelog.

---

## What the third-pass audit changed

It endorsed the retirement relaxation — revocation at PTY EOF bought no end-to-end denial under the neighbouring layer's behaviour, strict pane protection on top of it would have lost genuine final hooks, and successor reservation closes direct stale-token replay — and found no cross-user escalation and no unsafe Rust. One blocker, and the failure mode is worth recording because it is the *third* variation on the same theme.

**The fix was complete for the question it was asked, and the question had a second door.** The paned guard is right, and it made the paned route to token-less post-exit card-driving impossible. The pane-LESS route was left, and it reaches the same card. Once the last ordinary paned agent exits naturally the registry says two things at once about the same moment: the retired generation still owns and protects its pane (`pane_is_protected`, `token_still_live`), and this daemon manages no panes (`manages_any_pane`, which filters exited records on purpose). Both are correct answers to their own question. Together they meant a peer that sent *nothing at all* — no `pane_id`, no `agent_id`, no token, only the retired card's `session_id` — was admitted by a fallback that exists for external-agent watchers, and drove that card with its paned identity intact. No copied token needed. Deterministic whenever no live, reserved or held sibling keeps the oracle busy.

The previous round's own docs called that asymmetry harmless. It is not harmless; it is *right*, which is a different claim, and the auditor's sentence for it is exact: **the separate `!exited` policy causes finding 3.** So the divergence stays — counting exited records would pin a deck that runs nothing as "manages panes" for the rest of the daemon's life and refuse the external-watcher path permanently — and the collision is paid for at the collision instead. A pane-less event may not take over a session belonging to a pane the registry still protects, asked ahead of both existing grounds so it binds #454's identity exemption too. A pane-less event naming a new session, or an existing pane-less one, is untouched.

Two smaller ones. The **log redaction split was attacker-steerable**: a line can be valid JSON and still fail `AgentEvent` decoding, or fail to be JSON at all, and the sender picks which by how they break it — so writing the capability under an escaped member name put it in the *malformed* branch, which matched one literal spelling and missed it. Both branches now recognise every spelling (the decoded one structurally, the malformed one by decoding each string literal it walks), so there is no branch to steer into. And the **retired-token lifetime is now stated rather than implied**: "bounded by record reaping" reads tighter than it is, and the honest answer is potentially until the daemon exits. See residuals 6 and the log bullet above.

---

## Rule 12 — the cross-version contract

**Wire shape: `PROTOCOL_VERSION` stays at 7.** The token rides `AgentEvent.agent_token` and `AttachResponse.agent_token`, both additive `#[serde(default, skip_serializing_if = "Option::is_none")]` fields of exactly the shape `agent_version` / `schema_version` / `guarded_send` already use. Nothing in the protocol carries `deny_unknown_fields`, so an older peer ignores both; a newer peer tolerates their absence; a producer with no token emits byte-identical JSON. The only thing that moved the shape was the `GetAgentToken` variant (`AttachRequest` is internally tagged with no `#[serde(other)]`), and it is gone. Note also that `admit` **strips** the token from the event on every path before the broadcast, so a token never appears on a `KIND_EVENT` frame at all.

**Semantics: a break behind a stable wire.** An event that used to be accepted — a hook naming a managed pane with no token — is now refused. Recorded in `changelog.d/318.breaking.md`. Per `docs/develop/versioning.md` this is a **minor** bump while in `0.x`.

Two consequences of *not* bumping, stated plainly because the earlier revision leaned on the bump as a mitigation:

- The break is **not announced by a handshake refusal**. A mixed-version TUI and daemon still pair up, and the break is visible only in the daemon log's refusal line, which names the cause and what to do. That is the honest trade: a bump would announce it, at the cost of forcing every user of an unaffected pairing through a daemon recycle for a wire that did not change.
- The earlier rationale that the forced recycle "re-mints tokens for agents stranded token-less by the break" was **empty, and would have been empty with the bump too.** There is no such agent. A daemon owns its children and `AgentPtyRegistry::drop` → `shutdown_all()` kills them, so an agent cannot outlive its daemon and find itself talking to a newer one. The one exception is a daemon killed with `SIGKILL`, which leaves an orphan — and a replacement daemon does not manage that orphan's pane, so its hooks take the unchanged **foreign** path and keep driving a card. `changelog.d/318.breaking.md` has been corrected accordingly; what genuinely breaks is a hook run by a *different, older* `dot-agent-deck` binary than the daemon (a stale absolute path in `~/.claude/settings.json` — re-run `dot-agent-deck hooks install`), or a launcher that scrubs the environment (`env -i`, `sudo` without `-E`).

**The cross-version manual run is done.** A v0.37.2 daemon (`0.37.2-gf3752e1`) with two orchestration panes under it, driven by the branch build (`0.37.2-gf4bac9a`): the handshake pairs (`server_version: 7` both ways), a delegate still routes, and hooks still arrive — status, `PreToolUse`, and work-done. Reversed, with the branch daemon and the v0.37.2 client, the semantic break shows exactly where it should: a token-less event naming a **managed** pane is refused (`reason=no capability token`) and never broadcast, while the same old binary's `SessionStart`/`PreToolUse` for a pane the daemon does not hold still registers and drives a card, and `delegate`/`work-done` — `DaemonMessage`s rather than `AgentEvent`s — are untouched in both directions. Captured off a stub socket, the v0.37.2 and branch `agent-event` payloads are byte-identical apart from the timestamp until a token is present, at which point exactly one field is appended; no `KIND_EVENT` frame ever carries it. The stale-hook case reproduces as described — the card freezes, the daemon names the cause, and re-running `dot-agent-deck hooks install` restores it. `DOT_AGENT_DECK_LOG` was isolated alongside the sockets, `HOME` and `DOT_AGENT_DECK_STATE_DIR`, because the log path resolves separately and two interleaved daemons in one file are genuinely hard to read afterwards. Full transcript, including the daemon's refusal lines, in the run notes.

Two things the run measured that this section did not previously say, neither of which contradicts it:

- **The mixed pairing reasoned about above is one the deck actively tries to eliminate on the next TUI launch.** The PRD #103/#161 build-version handshake compares `DAD_BUILD_ID`, which differs for *any* rebuild, and all three of its branches were exercised: with agents live on a TTY it **prompts**; with **no agents running it recycles the old daemon silently**, with no prompt at all; on a non-TTY it **exits 1** without attaching and without recycling anything. So a mixed pair survives only where a human declines the prompt with agents live. That does not weaken "not announced by a handshake refusal" — it is not a protocol refusal, it is not specific to this change, and it does nothing for a *hook* binary, which is where the break actually bites — but it is the honest picture of how long a mixed pair lasts, and it is the thing a reader of this section is most likely to get wrong.
- **`dot-agent-deck hooks install` clears a stale rule only when that rule shares the installing binary's basename and its path is gone.** That is the ordinary upgrade shape and it comes out clean. When both binaries exist under different basenames — a worktree build alongside an installed release — both rules survive and both fire, so the card recovers but the refusal line keeps appearing once per hook. Worth naming: it saves the next person an investigation into "why is it still refusing after I ran the remedy".

---

## Which issues this touches, honestly

**Resolves**

- **#318** — a hook event is accepted for a managed pane only from a sender holding that pane's token, and the pane/agent are derived from the token rather than the payload.

**Partially addresses** (none of these should be closed)

- **#401** — substantially addressed for panes the daemon manages: agent-reported status can no longer be posted for a pane the sender does not own. Not closed, because #401 contemplates *authenticating the channel*, and a token readable out of another process's environment is not authentication. #401's own option 1 is exactly this trade.
- **#543** — standing is now bound to a daemon-minted token for managed panes. The residual it names survives by design: a producer that can reach the socket can still post an unmarked `SessionStart` for an *unmanaged* pane id. #543's candidate strengthening is peer-PID, considered and not chosen (below).
- **#601** — a forged `SessionStart` can no longer pre-authorize a pane id the daemon manages. It can still insert an *unmanaged* pane id into `managed_pane_ids`, which is the foreign-agent compat path, so `managed_pane_ids` still is not an ownership proof — and nothing in this guard is built on it as though it were. The pane-less fix above removes the one place that mistake had crept back in.

**Leaves untouched**

- **#617**, **#527** — out of scope, unchanged.

---

## The honest limit

The threat model is a **same-user process**, and a same-user process can read another process's environment (`/proc/<pid>/environ` on Linux). So this is **provenance binding plus a raised bar — not authentication against an adversary who already has code execution as this user.** Say it that way and nothing more.

**Peer-PID was considered and not chosen.** `platform::peercred::peer_pid` already exists and would resist a token-reading attacker, but it does not cover wrapper or detached hook runners whose process chain back to the pane's PTY leader is broken, and it puts a live-process-table dependency on the ingest path. It is the candidate strengthening tracked under #543, not a substitute for this.

## Residuals, stated so a reviewer does not have to find them

1. **Pane-less events, post-fix.** They are still classified `Foreign` at provenance admission — a pane-less event names no pane, so there is nothing for the pane-scoped guard to check. Two things changed. The fallback asks the registry, so a daemon that manages any pane no longer admits a pane-less event on the strength of a `session_id` alone. And — third-pass audit finding 3, **corrected here from an earlier draft that got this wrong** — a pane-less event may not take over a session that belongs to a pane the registry still protects, whatever either of those grounds says. The earlier draft described the remaining surface as "a daemon that genuinely manages **no** panes", and that was not the case that mattered: once the last ordinary paned agent exits naturally, its unreaped record keeps the pane protected and its token resolving, while `manages_any_pane` — which filters exited records on purpose, since counting them would pin a deck that runs nothing as "manages panes" forever — answers *unclaimed*, and ordinary `StartAgent` panes are deliberately absent from `managed_pane_ids` too. Both grounds were therefore satisfied for a daemon that was still protecting the pane, and a token-less, pane-less, agent-id-less event naming the retired card's `session_id` drove it: the same outcome the paned guard closes, reached around the side, with no copied token. It is closed at the collision (`AppState::paneless_event_may_take_over`) rather than by widening `manages_any_pane`, so the external-watcher path is narrowed and not disabled. What genuinely remains is a daemon watching external agents whose sessions are pane-less, where taking such events is the whole job, plus issue #454's retained exemption — **a pane-less event whose payload claims the id of a daemon-owned pane-less agent**. Note precisely what that second one is: the id is a *claim in the payload*, not proof about the sender, and any same-user peer can read it from the unauthenticated `ListAgents`. It is kept deliberately (dropping such an agent's `SessionStart` makes `agent_writable` fall back to its `Live` default, so input reaches a target that declared itself view-only), not because the sender is established. That is the #401/#543 residual, narrowed, not closed.
2. **The sibling verbs on this socket are unguarded.** Provenance guards `AgentEvent` **only**. The bounds *do* cover every verb, because they run before `DaemonMessage` parsing — but `Delegate`, `Dispatch`, `WorkDone`, `GetSeed` and `ListTargets` are handled ahead of the `AgentEvent`-only `admit` call and take their `pane_id` from the payload verbatim. **This PR protects hook `AgentEvent`s, not "the hook socket".** Follow-up filed as #696; `GetSeed` is first (see below).
3. **Foreign-session growth is unbounded, and the 8 MiB number does not cover it.** The `const` budget bounds the reader's byte buffers and nothing else. Token-less foreign `SessionStart`s are deliberately admitted, each fresh pane auto-registers and each fresh session id inserts another `SessionState` retaining the event and its strings, with no global session or foreign-pane cap; one admitted connection can stream unlimited sub-64-KiB lines. This is pre-existing semantic-state growth, not a flaw in the bounded reader, and `changelog.d/319.bugfix.md` now says so rather than claiming the socket "can no longer exhaust the daemon".

   **Assessed and declined for this PR rather than deferred silently.** A cheap global cap is not available. The eviction policy *is* the design: sessions for managed panes must never be evicted (that deletes a live card), and by the time `apply_event` reaches its `or_insert_with` a flood's session is indistinguishable from a real external agent's — the exact foreign-agent compat path #601 and #543 tell us not to disturb, so an LRU would silently delete real cards under attack. `apply_event` is also the TUI's path (a TUI installs no oracle), so a cap there changes TUI behavior too. And capping `sessions` alone would not cap the growth: `managed_pane_ids` grows by one entry per foreign `SessionStart` on the same path. Filed as #697, where the eviction argument belongs.

4. **Hook-derived log growth is reduced, not globally bounded.** The per-connection budget of five raw payloads and the 512-byte prefix are real, and together they take a malformed-line loop from 64 KiB per line to a few KiB per connection. They are **not** a bound on how much a peer can make the daemon write. Three gaps, stated rather than implied: the budget is per *connection*, so a peer that reconnects gets a fresh five — which is also how a repeatedly-invoked one-shot hook naturally behaves; every successfully decoded event emits its `Received event` line **before** the raw budget is consulted at all; and wire fields that reach that line, `session_id` and `tool_detail` among them, have no daemon-side length cap beyond the 64 KiB transport line. A real bound needs a global token bucket on hook-derived logging, which is not in this PR. Under the same-user threat model the same process can write to that file directly, which is why this is stated as a residual rather than fixed here — but "bounded" would be an overclaim and `changelog.d/319.bugfix.md` deliberately says only that the old amplification was reduced.

5. **The two layers still disagree in one corner, and it is the one the wording used to deny.** An earlier draft of this PR — and of both `pane_is_protected` docs — claimed the guard is true "in exactly the cases where the ownership layer would otherwise say `Owned`". The third-pass reviewer read that as one case too strong and named a sequence; it is now **confirmed by a test** rather than left as a reading, in `agent_pty::spawn_tests::a_handed_over_then_reaped_pane_is_unprotected_yet_still_owned`. A token-less event reaches `generation_ownership`'s `(Some(pane), None)` arm, which matches **any** record still naming the pane — no `exited` and no `pane_handed_over` filter. So for a pane whose first generation was handed over and lingers unreaped while its successor has since been closed (`StartAgent` twice, then `StopAgent` — no respawn involved), `pane_is_protected` answers `false` while `generation_ownership` answers `Owned`, and a token-less event naming it is classified `Foreign` and applied. **Left as it is in this PR, deliberately.** The pane is functionally defunct with both generations gone; the pre-fix `manages_pane` (`!exited`) also answered `false` there, so nothing regressed; and a `Foreign` card is precisely what an attacker can already mint for any pane id this daemon never managed — making this one more instance of residual 1 / #601's named remainder rather than a new class of exposure. Tightening the pane-only arm to honour `pane_handed_over` would close it and is the obvious candidate, but it sits beside the keyed arm and deserves its own change and its own argument; filed as #698. Both docs now state the exception instead of denying it.
6. **A retired generation's token stays valid potentially until the daemon exits.** A natural exit sets `exited` and removes nothing, so the record — and with it the token — lingers until something reaps it. The reapers are `close_agent`, `respawn_agent_for_pane` and `shutdown_all*`, all deliberate acts on that pane, plus idle shutdown, which fires only when there are no attached clients, no live siblings and no enabled schedules, and only when idle shutdown is on. **An attached TUI, one live sibling, one enabled schedule, or `DOT_AGENT_DECK_IDLE_SHUTDOWN_SECS=0` is therefore enough to keep a copied token valid for the daemon's whole life.** The docs used to say the window is "bounded by record reaping", which reads tighter than it is; `token_still_live` now states this. It is the same window the ownership layer already grants, exploiting it still needs same-user access to the token, and it reaches only that generation's card. An `exited_at` grace was considered and **not** added: it invents an arbitrary deadline for the legitimate final hook this rule exists to preserve, and it would have to move `generation_ownership`, `pane_is_protected` and `token_still_live` together or reopen exactly the disagreement re-audit finding 1 exploited. Narrowing it properly means reaping lingering records — which also leak a PTY master and a reader thread — and that is its own change.

## Follow-up issue: authorize the other hook-socket verbs by pane

> The hook socket's line and connection bounds cover every verb (they run before `DaemonMessage` parsing), but the #318 provenance guard covers `AgentEvent` only — `admit` runs on that branch alone. `Delegate`, `Dispatch`, `WorkDone`, `GetSeed` and `ListTargets` are handled ahead of it and take their `pane_id` from the payload verbatim, so any same-user process that can reach the socket can still act as a pane it does not own. **`GetSeed` is the sharpest and should be fixed first:** `take_pending_seed_native` atomically removes the pane's pending prompt *and* marks it natively delivered, so one forged request both steals potentially sensitive prompt text and suppresses the intended delivery — the pane never receives its seed and the daemon believes it did. `Delegate` and `Dispatch` can route work or create a whole worktree plus an agent attributed to another pane; `WorkDone` can retire another pane's outstanding delegation and its idle watch; `ListTargets` is read-only but leaks the pane's cwd. Pane ids are readily available from the deterministic attach endpoint's unauthenticated `ListAgents`. The token machinery generalises with little work — each signal gains an optional `agent_token` and the same derive-don't-compare treatment `AgentEvent` now gets — but doing it inside #318 would have been unreviewed scope creep, since both issues scope provenance to hook events. These are pre-existing broken-access-control paths, not regressions from #318.

## #527 — the shape transfers, the constant does not

`MAX_HOOK_LINE_BYTES` is a *transport* cap on one line of the **hook** socket. `WriteAndSubmitExtras::delivery_id` arrives on the **attach** socket, whose frames are already capped at `MAX_FRAME_LEN` (16 MiB), so the constant is the wrong axis and the wrong socket. What does transfer is the pattern, and #527 needs it: `MAX_DELIVERY_RESULTS` (8192) caps the ledger's **entry count** but nothing caps each `delivery_id`'s **length**, so 8192 × a multi-MiB id is a real ceiling the daemon will hold. The fix has the same shape as this one — a small explicit per-field byte cap, refused rather than truncated, with the number justified against a real id (a UUID or a `mint_delivery_id` value, i.e. tens of bytes) — and belongs next to the ledger, not in `hook_ingest`. **Not fixed here.**

---

## Tests

New L2 coverage (rule 4 — this touches the spawned binary, the daemon and real hook delivery, so L2 rather than L1 snapshots; every `#[spec]` carries its rule-7 `/// Scenario:` comment):

- `hooks/delivery/008` — an oversized newline-free line is dropped without buffering; a well-formed event on a fresh connection still lands.
- `hooks/delivery/009` — with the cap saturated, the excess connection is closed and an admitted peer still drives its card.
- `hooks/delivery/010` — **new in this round.** Idle peers hold every slot, a legitimate event provably cannot land, and then the daemon reclaims the slots at its read deadline with no peer closing and no restart. Uses `DOT_AGENT_DECK_TEST_HOOK_TIMEOUT_MS`, which is clamped to the production constant and can therefore only make the daemon *stricter*.
- `hooks/provenance/001` — pane A's token on an event naming pane B binds to A; B's card does not move.
- `hooks/provenance/002` — a token-less event naming a managed pane is refused.
- `hooks/provenance/003` — a token-less event naming an unmanaged pane still registers a foreign card (the #601 compat path).
- `hooks/provenance/004` — a hook line spelling the capability member as `agent_\u0074oken`, plus the same secret under a member the event does not have, reaches the daemon's unrecognized-`event_type` log branch: the log names the bad `event_type` and contains neither copy.

New fast-tier integration coverage, `tests/hook_provenance_admission.rs` — **the shape the re-audit asked for.** It runs the daemon's two real steps, `admit` then `AppState::apply_event`, against a real `AgentPtyRegistry`, and asserts on the *card*:

- a token-less event naming a pane whose agent has exited is refused and the card does not move — and the same forged event applied with provenance bypassed **does** move it, so the assertion cannot pass vacuously;
- the exiting agent's own late final event, carrying its token, still lands (the tension the fix must not break);
- a replayed token cannot drive a pane a later generation took;
- a **pane-less**, token-less, agent-id-less event naming the retired card's `session_id` cannot take it over either — third-pass finding 3, the same outcome reached around the side of the pane-scoped guard. Provenance calls it `Foreign` (it claims no pane, so there is nothing for that guard to check), so the refusal has to come from admission; the anti-vacuity clone here swaps in an **empty** registry, where the very same payload does move the card, which pins the registry's continued claim on the pane as the only thing protecting it;
- a pane this daemon never spawned is still the foreign path, so #601's compat path is not narrowed.

Unit coverage for the parts that are pure data, so the security decision table does not need a PTY: the bounded reader (under / exactly-at / over the limit, the padded-prefix smuggling case, EOF with no trailing newline, byte-at-a-time trickle, CRLF, invalid UTF-8), the deadlines on a virtual clock (idle peer, trickling peer, prompt peer unaffected, the knob can only shorten), the full provenance decision table including "a valid token for pane A cannot drive pane B", token stripping on every path, redacted `Debug`, textual log redaction out of a line that does not parse — **including every shape the third-pass audit's steerable branch accepts: an escaped member name in a line that never parses, a duplicate typed member, a typed member of the wrong type, and a capability sent as an array rather than a string; plus the two directions that keep it honest, a *value* whose text reads like the member name being left alone and a nested spelling still being taken out** — structural log redaction of an escaped member name and of a token under an unexpected member, plus a guard pinning the allowlist to `AgentEvent`'s real wire shape, the retirement rule from both directions (a natural exit keeps the token for its own pane; a successor taking the pane revokes it), a lingering exited pane still being protected, the respawn hold including **two overlapping holders releasing independently**, the pane-less admission rule in all four oracle states, and the pane-less collision guard binding #454's identity exemption as well as the fallback.

The real-agent bar is met by the existing e2e rows (`delegate/014`, `delegate/015`, `delegate_work_done_chain_claude`, `opencode_auto_submits_daemon_injected_prompt`, `card_stats/005`), which are the end-to-end proof that a live agent's own native hooks still drive its card through the new guard.

## Platform note

`src/platform/csprng/windows.rs` **could not be compiled locally** — this devbox's toolchain has no `std` for `x86_64-pc-windows-msvc`. Every symbol was verified by hand against the vendored `windows-sys-0.61.2` source (`BCRYPT_ALG_HANDLE = *mut c_void`, `BCRYPTGENRANDOM_FLAGS = u32`, `NTSTATUS = i32`, `BCRYPT_USE_SYSTEM_PREFERRED_RNG = 2u32`, and the `Win32_Security_Cryptography` feature), and independently by the reviewer. **`build-windows` in CI is the first real check of that file.**

---

## Pre-PR e2e tier

`cargo test-e2e` was run twice on `02bcefd` — once with `DOT_AGENT_DECK_RECORD=1` (**9021/9024**) and once without (**9023/9024**). Every failure across both runs was chased to a verdict rather than retried until green:

- `dispatch_005_respects_max_per_run`, `shell_activity_005_real_claude_bash_child_trips_the_descendant_scan` and `delegate_015_real_opencode_worker_acts_on_clear_true_delegate` failed in run 1 and **passed in run 2 under substantially heavier load** (the box went 1.35 → 75.6 on 16 cores, ~4.7x oversubscribed, from another checkout's tier plus 16 concurrent `rustc`). `delegate_015` matters most here because it exercises the `clear = true` respawn path this PR touches with `respawn_holds`: it went from a 180s in-tier timeout to a **93s in-tier pass** while the box was ~5x oversubscribed, which is the run that should have exposed a contending or leaked RAII hold if one existed.
- `opencode_auto_submits_daemon_injected_prompt` failed in run 2 and reproduced in isolation. It is **not caused by this branch** — measured by building the base commit (`a324614`) in its own worktree and running the test interleaved with this branch: **3/3 pass both sides on a quiet box, 0/3 both sides at ~4x oversubscription**, same symptom and same duration. Its readiness gate measures output quiet rather than input readiness, so under load it injects into a TUI that has not yet attached its input reader. Filed as **#699**; this PR is exonerated structurally too (the test file is not in the diff, `wait_until_agent_output_settled` and `write_to_pane_and_submit` are untouched, and the test pins `DOT_AGENT_DECK_SOCKET` at a dead path so no hook admission runs at all).


prageethw (changes requested):
Read the security-critical path in full (hook_ingest.rs's transport-bound and provenance modules, the token minting/revocation in agent_pty.rs, the admit() call ordering in daemon.rs, the wire additions in daemon_protocol.rs). The derive-not-compare binding is correct (verified against the adversarial a_valid_token_for_one_pane_cannot_drive_another test), revocation is correctly tied to the real ownership-retirement rule, log redaction correctly handles escape-based member-name smuggling, and the token is never exposed on ListAgents or broadcast frames. No bypass or logic bug found.

Blocking issue: this PR targets agent/dispatch-issue-670, not main, so no ci.yml run has ever fired for it (only PR Labeler and aarch64-crossbuild have run — checked gh api actions/runs for the head SHA directly). build, build-macos, build-windows, security, semgrep, and sonarqube have never executed. This matters specifically here because the PR body itself flags src/platform/csprng/windows.rs as hand-verified but never compiled, since this devbox's toolchain has no std for x86_64-pc-windows-msvc — build-windows is exactly the check that would turn that into a real verification, and it's never run. The base branch's own PR (#677) already merged to main on 2026-08-25, so retargeting/rebasing this PR onto main should let CI run cleanly. Please retarget so build-windows (and the rest of the suite) actually executes before merge — happy to re-review once it's green.

PR #716, opened 2026-08-26
fix(readiness): stop waiting for a signal only the withheld prompt can cause (#243)

### LINKED ISSUE #243: Pre-prompt readiness signal: retire the load-bearing 30s SESSION_START_WAIT_TIMEOUT and 1000ms DELEGATE_READINESS_BUFFER

vfarcic:
Fixes #243.

The delegate readiness gate waited up to `SESSION_START_WAIT_TIMEOUT` (30 s) for a `SessionStart` before writing a delegated prompt. For two agents that signal is **a consequence of the very prompt the gate is withholding**, so the gate could never fast-path and every such delegate paid the full timeout plus the readiness buffer.

- **Codex** (`Wrapper`): the native `SessionStart` fires at the first *turn*, i.e. after a prompt is submitted.
- **OpenCode** (`Plugin`): no pre-prompt signal exists at all; `session.created` arrives ~16 ms *after* the prompt is accepted.

The discriminator was `hook_install.is_some()` — which correctly means "has native hooks" but was standing in for "will emit a real `SessionStart` before it needs a prompt". Both agents satisfy the first and not the second: one mis-predicate, two victims.

## Measured, before

Captured from an unmodified production daemon log while this very orchestration delegated to its own Codex-backed tester:

```
21:44:11.071  respawned worker for clear=true; waiting for SessionStart   timeout_secs=30
21:44:41.072  SessionStart wait TIMED OUT; writing prompt via fallback    (+30.001s)
21:44:41.072  holding prompt for readiness buffer   observed=false  buffer_ms=1000
21:44:42.751  SessionStart  agent_type=Codex                              (+31.680s)
```

`observed=false` — the gate never saw a signal. The native `SessionStart` lands **1.68 s after the prompt was injected**, so it is caused by the prompt rather than by boot. Three separate delegates in the same log timed out at **30.001 s**, identical to the millisecond.

A controlled A/B on the same pane, swapping only which agent occupied it: **Codex 31.68 s / `observed=false`** against **ClaudeCode 4.20 s / `observed=true`**.

## Measured, after

| path | before | after |
|---|---|---|
| Codex delegate, warm launcher | 31.4 s | **5.6 s** |
| Codex delegate, cold `devbox` | 31.4 s | **20.6 s** — and *delivered* rather than silently lost |
| OpenCode delegate | 30.3 s | **~8.3 s** |

Real-agent e2e, full tier, `0 skipped`: `orchestration/delegate/009` (real Codex) PASS 36.1 s with ready→submit measured at 5.758 s; `/015` (real OpenCode) PASS 40.4 s; `/014` (real Claude, untouched baseline) PASS 15.0 s.

## The fix, in two halves

**Half 1 — a wrapper-side readiness signal.** `dot-agent-deck wrap` now watches the wrapped child's terminal and emits a `SessionStart` carrying a new `session_start_origin`, on either of two observed facts: the child clearing `ICANON`/`ECHO` on the inner PTY (`wrapper_interface_ready`), or output settling for 750 ms (`wrapper_interface_settled`). The latch is per-fact, so a launcher-hosted child emits the weak fact and later the strong one.

**Half 2 — an explicit `AgentSpec::pre_prompt_readiness` field.** `NativeSessionStart` / `WrapperInterfaceReady` / `NoSignal` / `Unknown`, replacing the `hook_install.is_some()` mis-predicate. **An unknown agent keeps today's conservative wait** — "we do not know what this is" is not evidence that skipping is safe. Only a positive `NoSignal` declaration takes the short path. Both the delegate seam and the scheduler's copy read it.

## Three regressions this work introduced, and then fixed

Every one was found by running a **real agent**; every one would have shipped green on the synthetic tier.

1. **The wrapper's weak fact fired inside the launcher.** `devbox run codex-big` prints one banner line then goes silent for **2750–4132 ms** computing its shellenv — before `codex` is exec'd. Output-settled fired 13/13 launcher probes, ~2.0–3.4 s *before* raw mode. With the original zero-buffer treatment the prompt was written into the launcher's cooked-mode discipline and echoed away: **PRD #225 Defect 1, reintroduced through the fix for it**, 3/3 runs.
2. **Raw mode is not input-readiness either.** Launched directly with no launcher, `codex` takes raw mode at fork **+100 ms**, at TUI *init*. So the original premise — "a child reading raw keystrokes has nothing left for a blind interval to cover" — is false on the measured article, and has been retracted everywhere it appeared. The strong fact now selects **which** buffer is paid rather than skipping one.
3. **The `NoSignal` path had the same defect.** Before this issue an OpenCode delegate sat out the 30 s dead wait — which happened to give the replacement 30 s to boot. Removing it left the 1000 ms buffer, derived against a 650 ms *stub*, as the sole cover. `orchestration/delegate/015`, executed for the first time by anyone, failed 2/2 at the shipped default.

## Three buffers, each derived from a real agent

| buffer | value | applies to |
|---|---|---|
| `WRAPPER_INTERFACE_READINESS_BUFFER` | 5000 ms | the strong fact, on a pane this daemon spawned as a wrapper host |
| `NO_SIGNAL_READINESS_BUFFER` | 8000 ms | a declared-`NoSignal` agent |
| `DELEGATE_READINESS_BUFFER` | 1000 ms | everything else — native start, weak fact at window expiry, timeout fallback, any interface fact the launch record cannot vouch for |

`DOT_AGENT_DECK_DELEGATE_READINESS_BUFFER_MS` overrides all three, preserving the #199 operator escape hatch.

**5000 ms** is not a threshold: codex-cli eats keystrokes through its whole TUI init, ending in a teardown-and-repaint at ~607 ms, so the loss window is **not monotonic** (350 ms parked 5/5, 400 ms delivered 5/5, 586/595 ms parked again). Worst requirement under 2× core load was 3601 ms; 5000 is 1.39× that.

**8000 ms** comes from **176 runs** against real `opencode` 1.18.23 with three-way classification. Zero runs parked — every loss was a *swallow*, so a longer interval introduces no second failure mode — and the path **is** monotonic, unlike the wrapper's. The requirement is a single observable event: the instant OpenCode paints `Ask anything...`. Idle 2.5 s, contended ~4.5 s, 4× oversubscribed 12 s. 19/19 delivered at the shipped value.

**Both are mitigations, not bounds**, and say so in their own docs. They are other programs' startup times, they scale with load, and no number measured on one box bounds them on another. The durable answer is an observation that scales — which for OpenCode this work hands PRD #234 an unusually clean target, since the boundary is a single string on a PTY the deck already owns.

The **upgrade window** is `SESSION_START_WAIT_TIMEOUT` by definitional alias rather than a new constant: it introduces no new tunable, and the worst case is precisely today's known-safe behaviour rather than a new failure mode. An 8 s window lost the prompt on a cold `devbox` (fact 2 at 1.1 s, expiry at 9.1 s, write at 14.1 s, strong fact at 15.98 s — bytes drained fused, the CR read as newline-in-input).

## Security

An audit found the interface marker **forgeable** — `src/daemon.rs` accepts a raw `AgentEvent` line with a free-form `metadata` map, reproduced from a bare `python3` with no deck env vars. Releasing the gate was *already* forgeable pre-change via a bare unmarked `SessionStart`; what the first draft newly granted was **suppressing the buffer**. With the zero-buffer branch deleted that escalation no longer exists, and a forgery can only mis-price toward the 1000 ms every other agent already gets. `agent_spawned_as_wrapper_host` reads the **frozen** launch-shape record, never the hook-upgradable badge, so a producer cannot claim to be a wrapper host. Comments claiming the origin is unforgeable have been retracted.

## Rule 12

**No `PROTOCOL_VERSION` bump** — `src/daemon_protocol.rs` is byte-identical to v0.37.2; no field added, removed or retyped; `metadata` is free-form and the `session_start_origin` key pre-existed. Verified live: `daemon hello` reports `server_version: 7` on both builds.

**Semantic break behind a stable wire — yes**, carried by `changelog.d/243.breaking.md`: a wrapper now emits up to three `SessionStart` events where it emitted one. Bump policy while `0.x`: breaking → **minor**.

Cross-version matrix re-run against **v0.37.2**, all four cells, delegate routing and hook arrival confirmed in each:

| daemon | wrapper | pointer | routes | hooks |
|---|---|---|---|---|
| v0.37.2 | v0.37.2 | 34.1 s | ✅ | ✅ |
| v0.37.2 | **branch** | 7.3 s | ✅ | ✅ |
| **branch** | v0.37.2 | 34.1 s — degrades cleanly | ✅ | ✅ |
| **branch** | **branch** | 12.25 s | ✅ | ✅ |

**Old daemon + new wrapper is a genuine regression in that mixed combination**, and the changelog states it rather than waving it past: an old daemon releases on the first interface event — the weak guess — so against a real launcher it writes into the launcher. It takes a deliberate act to reach, since a wrapper is spawned by the daemon that is running it. The suspected double-event generation hazard did **not** reproduce: every wrapper event carries the same wrapper-derived session id, so the second only refreshes the high-water mark; the revocation comes from the agent's own native start with a different id, which is pre-existing and is fixed on this build.

## Scope discipline

- **#528** (unify the three spawn-time delivery implementations) — **helped.** Not unified, but the delegate and scheduler gates now answer "is there anything to wait for" and "how long do we hold" from one shared predicate and one shared resolver instead of two.
- **#529** (the 10 s fallback writes with no buffer) — untouched, slightly helped: the concept it needs, "a readiness fact that does not prove input-readiness still owes a buffer", is now named and implemented daemon-side.
- **#526 / #527 / #530 / #544 / #547 / #637** — same file, neither helped nor blocked. The one adjacency is #666's fact-G rearm, widened from "not the fork marker" to "not any wrapper marker", which is strictly stricter.
- **#424** (the TUI-side 10 s gate) — deliberately **not** absorbed.
- **#146** — its `server.connected` avenue stays refuted and is now recorded in-repo so the next reader of the plugin docs finds the answer.

## Tests

New: `codex/wrap/006`, `orchestration/delegate/024`–`/028`, `agent/readiness/001`–`/003`, plus in-crate unit tests for the buffer resolvers and for `resolve_expired_wait`'s provisional branch. Modified: `/007`–`/012`, `/023`, `/009`, `/015`, and the catalog prose that went stale.

Every guard test was **proved load-bearing** by reverting its guard and showing red. Two rounds of guard tests were caught being *vacuous* — passing whether their guard existed or not — and re-founded rather than nudged. `/024` and `/026` in particular could not be repaired by widening their budgets: with the upgrade window equal to the timeout, "released by the buffer" and "released by nothing" land in the same instant, so their fixtures had to change instead.

Fast tier **3202/3202**. Full e2e tier `8927 tests run, 8922 passed, 0 skipped`; the five failures are two flakes that pass in isolation and three that **reproduce unchanged at the merge base**.

## Known follow-ups

- The wrapper's per-fact latch cannot yet cancel a buffer already in flight, so a launcher-hosted Codex pays the weak fact's window before the strong one releases it.
- `codex/live/001` and `codex/hooks/001` are red on `main` in exactly the shape this issue fixes, because they are the only real-Codex tests that never opt into a readiness buffer.
- `codex/worker/001` asserts on raw LLM-written bytes with no trim or re-read; also red on `main`.

Demo reel for issue #243: https://youtu.be/JfhLX7lqyJE. Watch before merging.


vfarcic:
Demo reel for issue #243: https://youtu.be/JfhLX7lqyJE. Watch before merging.

vfarcic:
Filed the follow-up from the "Known follow-ups" list: **#724** — the wrapper's per-fact latch cannot re-price a buffer already in flight, so a launcher-hosted Codex pays the weak fact's window in full before the strong fact would have released it (the 5.6 s vs 20.6 s gap measured here). The issue carries the constraint that matters: re-pricing must not become a path by which a forged interface marker suppresses a buffer, since that escalation is exactly what deleting the zero-buffer branch closed.

Greptile reviewed this at head (23 files, 0 comments). No open threads, all CI green, awaiting review.

prageethw (approved):
Confirmed the fix resolves the circular-wait: Codex's native SessionStart only fires after the withheld prompt is delivered, so readiness now releases on the wrapper's own PTY-observed interface facts (src/wrap.rs InterfaceWatch) instead, bounded to min(upgrade_window, timeout) so nothing regresses past the pre-fix baseline. Provenance is checked against the daemon's own frozen spawn_agent_type (src/agent_pty.rs) rather than the producer-writable event badge, closing the forgery path an earlier round left open — directly tested by delegate_028 with an explicit control assertion. tests/agent_readiness.rs pins every agent's pre_prompt_readiness via exhaustive match, and delegate_009 was fixed to assert a latency bound instead of passing identically on timeout-vs-signal. Thorough, well-tested fix. CI green.

Issue #250, opened 2026-07-28
Source builds report DAD_VERSION=0.1.0, breaking version negotiation (build.rs has no injection path)


vfarcic:
## Symptom

A binary built from source outside a tagged git checkout reports its version as **`0.1.0`** and its build id as **`0.1.0-unknown`**. Every version-aware code path then misbehaves against it.

Already visible in the wild: [#199](https://github.com/vfarcic/dot-agent-deck/issues/199) lists *"dot-agent-deck: any version (tested on v0.1.0)"* — that is this fallback, not a real release.

## Root cause

`Cargo.toml` carries `version = "0.1.0"` as a **placeholder**; the real version is derived at build time in `build.rs` from git:

- `DAD_VERSION` ← `git describe --tags --abbrev=0`, falling back to `CARGO_PKG_VERSION` (`build.rs:6`)
- `DAD_BUILD_ID` ← `<version>-g<short-sha>[-dirty]`, falling back to `<version>-unknown` (`build.rs:12`, `:99-105`)

Every git call degrades gracefully instead of failing the build — correct in isolation, but it means all of them degrade *together* whenever git metadata is absent: a source tarball, a shallow or `.git`-less clone, `cargo install --git`, a sandboxed build (Nix, distro packaging), or a checkout with no tags fetched.

Critically, `build.rs` emits `cargo:rustc-env=DAD_VERSION={version}` **unconditionally**, so there is currently **no way for a packager or CI job to inject the correct version from outside**. The fallback cannot be overridden.

## Impact

1. **Compatibility-based version-mismatch handling (PRD #161)** — a client claiming `0.1.0` negotiating against a real daemon.
2. **The remote upgrade nudge** — it is newer-only, so a source-built remote looks permanently ancient and prompts for upgrade on every connect.
3. **`remote add` / `--no-install` pre-flight**, which requires a matching version on the remote.
4. **The `DAD_BUILD_ID` daemon-restart handshake** — a constant `-unknown` across genuine rebuilds means an in-place upgrade may not force the daemon restart it is supposed to.

Release builds are unaffected: `release.yml` builds in a tagged checkout, so `git describe` resolves and the real version is embedded.

## Proposed fix

Make the version injectable, preserving current behavior when nothing is injected. Resolution order:

1. Pre-set `DAD_VERSION` / `DAD_BUILD_ID` in the build environment, if present
2. git (`describe` / `rev-parse`) — today’s path, unchanged for release builds and dev checkouts
[...]

Issue #239, opened 2026-07-28
Windows agent teardown: close the post-spawn Job Object adoption window


vfarcic:
## Problem

On Windows, a descendant process spawned by an agent in the window between `CreateProcessW` returning and `AssignProcessToJobObject` running never joins the agent's Job Object, so it survives `TerminateJobObject` and outlives agent closure or daemon shutdown.

`portable_pty::spawn_command` returns an **already-running** child; `AgentProcessGroup::adopt` assigns it to the job only afterwards. Job membership is inherited forward only, so anything created in that gap is permanently outside the job.

Raised by Greptile on #235 (`src/agent_pty.rs:1046`, P1) and independently by the PRD #163 reviewer. It is the sole reason for that PR's 3/5 confidence score, and it was consciously accepted as a documented v1 limitation rather than fixed — see `src/platform/proc/windows.rs` module docs, the "Known v1 limitation" section on `AgentProcessGroup::adopt`, and the "v1 behavior-equivalence" section of `prds/done/163-windows-platform-backends.md`.

## Why it was not fixed in #163

Three of the four ways to close it are unreachable with `portable-pty` 0.8.1:

| Approach | Status |
|---|---|
| `CREATE_SUSPENDED` → assign → resume | Unreachable — creation flags are hard-coded (`EXTENDED_STARTUPINFO_PRESENT \| CREATE_UNICODE_ENVIRONMENT`) and the primary-thread handle is closed inside `spawn_command` before it returns, so there is no thread left to resume. |
| `PROC_THREAD_ATTRIBUTE_JOB_LIST` in `STARTUPINFOEX` | The correct Win32 answer — the process joins the job atomically at creation and the window is *zero*. Unreachable: `portable-pty` builds its attribute list internally with capacity 1, for the pseudoconsole. |
| Outer inherited job | Does not work — to be inherited it would have to contain the daemon itself, so terminating it would kill the daemon and every other agent. Nesting still leaves the escaped descendant in the outer job only. |
| Own the spawn | The only real fix. See options below. |

## Options

1. **Upstream a patch to `portable-pty`** exposing either `CREATE_SUSPENDED` or a job-list hook. Cheapest, and zero ongoing maintenance cost for us. Worth attempting first.
2. **Fork or vendor `portable-pty`** and add the `STARTUPINFOEX` attribute-list entry ourselves.
[...]

Issue #195, opened 2026-06-24
linkage-check: PR #188 forbidden-sleep hits + CI does not run linkage-check


vfarcic:
## linkage-check: forbidden-sleep hits in e2e_delegate_work_done_chain.rs + CI doesn't run linkage-check

Surfaced while working PRD #120 (not introduced by it).

### Two related problems
1. **Pre-existing Decision-21 violations.** `cargo xtask linkage-check` exits non-zero on 4 "forbidden sleep call (Decision 21)" hits in `tests/e2e_delegate_work_done_chain.rs` (lines ~114/169/183/210), introduced by PR #188 (`fe929d2`). Replace the `tokio::time::sleep(...)` calls with proper wait-for-condition polling.
2. **CI does not run `linkage-check`.** `CLAUDE.md` rule 7 states "CI's linkage-check rule 7 fails the build…", but no GitHub workflow invokes `cargo xtask linkage-check` — which is how the PR #188 sleeps slipped in. Either wire linkage-check into CI (so its rules — Scenario comments, Decision-21 sleeps, catalog linkage — are enforced), or correct rule 7's claim.

Low urgency, but it's a real CI-coverage gap.

vfarcic:
Closing — both problems this issue raised are resolved on `main`, and verified as fixed rather than suppressed.

**1. The forbidden-sleep hits are gone.** `cargo xtask linkage-check` now exits **0**: `linkage-check: ok (560 catalog ids, 488 annotations, 100 allowlisted, 9 rules)`.

The fix is the one this issue asked for — the raw `tokio::time::sleep(...)` calls in `tests/e2e_delegate_work_done_chain.rs` were replaced with proper wait-for-condition helpers, not silenced. The file now uses `common::wait_until_agent_output_settled`, `common::wait_for_path_async` and `common::wait_for_rendered_agent_text`, and carries a comment at lines 55-57 naming Decision 21 and those helpers explicitly.

Checked specifically because the run reports 100 allowlisted entries: **`e2e_delegate_work_done_chain.rs` appears in no allowlist.** The Decision-21 rule itself is live at `xtask/linkage-check/src/main.rs:484`.

**2. CI runs linkage-check.** `.github/workflows/ci.yml:221` carries a real step:

```yaml
- run: cargo xtask linkage-check
```

[...]

PR #718, opened 2026-08-26
ci: pin the two dependencies that resolved from mutable refs (#450, #323)

### LINKED ISSUE #323: Pin cross in release.yml and aarch64-crossbuild-check.yml (installed unpinned)
## Problem

Both `.github/workflows/release.yml` and `.github/workflows/aarch64-crossbuild-check.yml` install cross with:

```
cargo install cross --locked
```

with **no version pin**, so CI silently resolves whatever is newest on crates.io (today 0.2.5, the newest of 25 published versions). `--locked` pins cross's *own* dependencies, not the version of cross being installed.

This matters more since #310, which relies on cross forwarding `DAD_VERSION` into the build container so the aarch64 artifact reports the version being released. That behaviour was **measured** working on cross 0.2.5 with real docker builds: with `CROSS_BUILD_ENV_PASSTHROUGH=DAD_VERSION` the container resolved the injected value; without it the container fell back to the checkout's previous git tag, and a real cross build embedded `0.36.0-CROSSPROOF` into a genuine ARM aarch64 ELF while the negative control baked `0.1.0`.

The measurement is only valid for the version measured. An unpinned install means a future cross release can change passthrough semantics and nobody finds out from a diff.

## Mitigations already in place (from #310)

The failure is no longer silent, which is why this is hardening rather than a live defect:

- A second, independent forwarding channel — `CROSS_CONTAINER_OPTS: -e DAD_VERSION` — which forwards below cross's own config layer and was measured sufficient *on its own* with passthrough removed.
- A fail-closed **"Verify the released version is baked into the artifact"** step that greps the finished ARM64 binary and fails the release if the released version is missing.

So a cross regression now breaks the release loudly instead of publishing a mislabelled artifact. Pinning removes the surprise entirely.

## Proposed fix

1. Pin cross to an explicit version in both workflows (`cargo install cross --locked --version X.Y.Z`), and let dependabot or a periodic bump move it deliberately.
2. **Optional, related:** extend the artifact-version verification to the **native** legs (x86_64 linux, macOS, Windows). #310 deliberately scoped it to the cross leg because the native legs have no container boundary to lose the value at, and an untested `grep -a` against Mach-O and PE binaries that can *block a release* is a worse hazard than the coverage it adds. Doing it properly needs tested per-platform equivalents.

## Discovered via

[#310](https://github.com/vfarcic/dot-agent-deck/pull/310) (issue #250), while resolving a P1 review finding about cross-container version propagation.

### LINKED ISSUE #450: Pin the DeterminateSystems Nix actions to SHAs
Follow-up to #391.

The `nix` job in `.github/workflows/ci.yml` uses two third-party actions pinned to mutable major tags:

- `DeterminateSystems/nix-installer-action@v22`
- `DeterminateSystems/magic-nix-cache-action@v14`

The cache action additionally downloads and executes a binary at runtime from `cf.install.determinate.systems` (visible in the job log: `Fetching from https://cf.install.determinate.systems/magic-nix-cache-closure/stable/X64-Linux`) — a host the project did not previously contact.

This matches existing repo convention (`actions/checkout@v7`, `dtolnay/rust-toolchain@v1`), so it is a deliberate-choice question rather than a defect. It is the same class as #323 (unpinned `cross`), and the runtime binary download arguably makes these two a stronger candidate for SHA pins than most.

**Mitigating factors today:** `ci.yml` grants only `contents: read` and `pull-requests: read`, and the job sets `determinate: false`, `use-flakehub: false` and `diagnostic-endpoint: ""`, so it stays on upstream Nix and the GitHub Actions cache with no hosted service and no secret.

For contrast, the flake's own supply chain is already strongly pinned: all four `flake.lock` inputs (`ipetkov/crane`, `numtide/flake-utils`, `NixOS/nixpkgs`, `nix-systems/default`) carry a full rev plus narHash.

vfarcic:
Closes #450. Closes #323.

Two issues, one PR: they are the same defect class in the same place — a CI job resolving a dependency from a reference that can move after review — and they touch overlapping files, so splitting them would only create a conflict.

## #450 — SHA-pin the DeterminateSystems Nix actions

`ci.yml`'s `nix` job used two third-party actions on mutable major tags. A major tag is a ref its owner can repoint at any commit, so what a reviewer read and what CI later executed were only guaranteed to match until someone upstream moved the tag.

| | before | after |
|---|---|---|
| `ci.yml:522` | `DeterminateSystems/nix-installer-action@v22` | `…@ef8a148080ab6020fd15196c2084a2eea5ff2d25 # v22` |
| `ci.yml:547` | `DeterminateSystems/magic-nix-cache-action@v14` | `…@908b263ff629f4cc17666315b7fd3ec127c6244d # v14` |

Style follows the pins already in this repo (`actions/checkout@3d3c42e5… # v7.0.1` and friends in the generated `issue-labeler*.lock.yml`): full SHA, human-readable tag in a trailing comment. That keeps the version readable at a glance and gives Renovate the token it needs to bump the pin.

### How each SHA was resolved

From the upstream repository, twice, by independent routes — not from a search result:

```
$ gh api repos/DeterminateSystems/nix-installer-action/git/ref/tags/v22 \
    --jq '{type: .object.type, sha: .object.sha}'
{"type":"commit","sha":"ef8a148080ab6020fd15196c2084a2eea5ff2d25"}

$ gh api repos/DeterminateSystems/magic-nix-cache-action/git/ref/tags/v14 \
    --jq '{type: .object.type, sha: .object.sha}'
{"type":"commit","sha":"908b263ff629f4cc17666315b7fd3ec127c6244d"}

$ git ls-remote --tags https://github.com/DeterminateSystems/nix-installer-action.git 'refs/tags/v22*'
ef8a148080ab6020fd15196c2084a2eea5ff2d25	refs/tags/v22

$ git ls-remote --tags https://github.com/DeterminateSystems/magic-nix-cache-action.git 'refs/tags/v14*'
908b263ff629f4cc17666315b7fd3ec127c6244d	refs/tags/v14
```

**The tag-object indirection was checked, not assumed.** Both refs report `.object.type == "commit"`, and `ls-remote` returns no `^{}` peeled line for either — so both are *lightweight* tags pointing straight at a commit, and the SHA above is the commit `actions/checkout` will fetch rather than an annotated tag object wrapping it. Both commits were then confirmed to exist and dereference cleanly (`nix-installer-action` → "Update deps, go to node24 (#239)", 2026-03-25; `magic-nix-cache-action` → "Update `detsys-ts`… (#195)", 2026-05-15).

### Does SHA-pinning address the runtime binary download? **No — and that is recorded in the file rather than fixed here.**

#450's second point is that `magic-nix-cache-action` downloads and executes a binary at runtime from `cf.install.determinate.systems/magic-nix-cache-closure/stable/X64-Linux`. Pinning the action **does not** cover that, and I checked that `nix-installer-action` has the same shape of behaviour (it fetches `nix-installer` from `install.determinate.systems`). A commit pin fixes the *fetcher*, not the *fetched* — and the cache action's URL names a `stable` channel, which is itself a mutable ref.

Reading each action's `action.yml` at the pinned SHA, both expose `source-binary` / `source-branch` / `source-pr` / `source-revision` / `source-tag` / `source-url` inputs that *would* pin the payload. I deliberately did **not** set one: picking a value needs its own upstream verification and a measured green run of the `nix` job, which is a separate change from making these two refs immutable. The gap and the mechanism are now written in a comment beside the pin, so the next person does not have to rediscover either.

The job's existing containment is unchanged and still relevant, as #450 notes: `contents: read` + `pull-requests: read` only, with `determinate: false`, `use-flakehub: false`, and both diagnostic endpoints disabled.

## #323 — pin `cross`

Both workflows installed cross with `cargo install cross --locked`. As the issue says, `--locked` reads like a version pin and is not one: it fixes the dependency *graph* of whichever version resolves, while the version still floats to whatever crates.io calls newest.

- `release.yml:289` → `cargo install cross --locked --version 0.2.5`
- `aarch64-crossbuild-check.yml:66` → `cargo install cross --locked --version 0.2.5`

**Why 0.2.5 specifically.** Not "newest today". `release.yml` carries a comment recording that `CROSS_BUILD_ENV_PASSTHROUGH` forwarding of `DAD_VERSION` into the aarch64 container was **"MEASURED, not assumed (cross 0.2.5…)"** — the fix for #250. Pinning to any other version would have silently invalidated that measurement, so the pin is set to the version the measurement describes. 0.2.5 is also in fact still the newest of cross's 25 published versions (released 2023-02-04), so this pin freezes current behaviour rather than changing it.

I updated two comments in `release.yml` that had gone stale in the *other* direction — both previously said the install was **not** version-pinned and used that as part of their reasoning:

- the `MEASURED, not assumed` parenthetical now reads "cross 0.2.5, now the version the install step above pins explicitly" instead of "the version `cargo install cross --locked` resolves to today";
- the `CROSS_CONTAINER_OPTS` block's rationale ("`cargo install cross --locked` is NOT version-pinned") now says the pin means such a change can only arrive with a visible bump, with this channel kept as the second, independent line of defence. The belt-and-braces design from #310 is unchanged — both channels are still set, and the fail-closed "Verify the released version is baked into the artifact" step is untouched.

**Both workflows moved together, as required.** `aarch64-crossbuild-check.yml`'s header asserts it mirrors the release leg — *"same runner, same toolchain action, same `cargo install cross --locked`"*. That wording no longer matched, so it now reads "same version-pinned `cargo install cross`" and names the cross version as part of what is mirrored, with a pointer at the install step. Pinning one side alone would have left the file asserting a sameness that had stopped being true.

## Out of scope, reported for the maintainers to decide

This repo is **mid-migration** on action pinning: the generated `issue-labeler.lock.yml` / `issue-labeler-preview.lock.yml` are fully SHA-pinned, while the hand-maintained workflows are not. Measured on this branch, **15 distinct actions** across the hand-maintained workflows are still on a mutable tag:

```
actions/checkout@v7                      docker/build-push-action@v7
actions/download-artifact@v8             docker/login-action@v4
actions/labeler@v7                       docker/setup-buildx-action@v4
actions/setup-node@v7                    dtolnay/rust-toolchain@v1
actions/stale@v11                        jetify-com/devbox-install-action@v0.15.0
actions/upload-artifact@v7               softprops/action-gh-release@v3
arduino/setup-task@v3.0.0                Swatinem/rust-cache@v2
                                         taiki-e/install-action@v2
```

(reproduce with `grep -rhoE 'uses: [^ ]+@[^ ]+' .github/workflows/*.yml --exclude='*.lock.yml' | grep -vE '@[0-9a-f]{40}$' | sort -u`)

**Not pinned in this PR, on purpose.** Each one needs its own upstream tag→commit verification, and `softprops/action-gh-release` and `docker/login-action` sit on the release/publish path where a wrong pin fails at the worst moment. Doing all 15 would turn a change reviewable in one sitting into a broad one — and the wider policy (pin everything? let Renovate manage digests? which tiers?) is a maintainer call, not something to settle as a side effect of two issues. Worth noting `dtolnay/rust-toolchain@v1` interacts with `scripts/check-pin-lockstep.sh`, which parses those `toolchain:` sites, so that one is not a pure find-and-replace.

## Rule compliance

- **Rule 12 (cross-version contract check):** **no** TUI↔daemon contract change. The diff touches no Rust source, no wire type in `src/daemon_protocol.rs`, no orchestration path and no hook handling — it is three workflow files and two changelog fragments. So `PROTOCOL_VERSION` stands, there is no semantic-break-behind-a-stable-wire case, and no `.breaking.md` fragment applies. The cross-version sandbox run is not applicable: there is no protocol surface for an older daemon and a newer TUI to disagree about.
- **Rule 4 (TUI tests):** no user-visible TUI change — no pane, status, prompt, focus, layout, mode, embedded pane or hook delivery is touched — so no L1 or L2 test is required, and none was added.
- **Rule 3 (filenames):** no new source or test file. The two changelog fragments are named by issue number because towncrier requires that (`changelog.d/<issue>.<type>.md`); that is the tool's convention, not a milestone prefix on a long-lived source file.
- **Rule 9 (experimental flag):** not applicable — no new user-visible surface.
- **Changelog:** `changelog.d/450.misc.md` and `changelog.d/323.misc.md`, one per issue so both get their issue link in the release notes. `misc` per `pyproject.toml` ("Internal improvements, CI/CD, tooling") — CI-only hardening that users do not directly see.

## Gates

| gate | result |
|---|---|
| `cargo clippy --workspace --all-targets --features e2e -- -D warnings` | pass (all three flags, per rule 2, even though this is CI-only) |
| `cargo test-fast` | pass — **3252 tests, 0 failed**. Not inert here: `xtask/linkage-check/src/pin_lockstep.rs` reads `.github/workflows/` |
| `scripts/check-pin-lockstep.sh` | **exit 0** — `ok: Rust toolchain pinned at 1.97.1 on both sides (7 workflow sites)`, `ok: cargo-nextest pinned at 0.9.143 on both sides (3 workflow sites)` |
| `cargo test-e2e` | **not run — deliberately.** The final diff is workflows + changelog fragments only: zero Rust source or test files, so no e2e target can observe it and the tier's ~LLM-API cost would buy no signal. Rule 5 scopes it to the pre-PR step for changes it can actually see. |

**CI is the real proof**, since these changes only take effect where they run — see the verification comment below for the `nix` and `aarch64-crossbuild-check` results.


vfarcic:
## CI verification: blocked by a GitHub Actions outage, **not** by this change

These changes only take effect where they run, so the intended proof is a green `nix` job (#450) and a green `aarch64-crossbuild` job (#323). Neither has reported, and **no Actions run was created for this PR at all**:

```
$ gh api "repos/vfarcic/dot-agent-deck/actions/runs?head_sha=21c59b2…" --jq '.total_count'
0
```

The cause is platform-side. GitHub's status API reports the **Actions component in `major_outage`**, with an incident ("Incident with Actions") opened at **2026-08-26T15:11:58Z**. This PR was created at 15:47:58Z, i.e. inside the incident window.

Three pieces of evidence that this is the outage and not something in the diff:

1. **Zero runs were *created*** — not created-and-failed. A workflow file this branch broke would surface as a `startup_failure` run, and there is no run of any kind for the head SHA.
2. **A different branch shows the same signature.** `agent/dispatch-issue-243` pushed at 15:24:58Z — after the incident opened — and got `startup_failure` for both `Docs` and `PR Labeler`, each with **zero jobs and no retrievable log**. That branch does not share this one's edits.
3. **The one non-Actions check ran normally.** `Greptile Review` is a GitHub App rather than an Actions workflow, and it completed in 2m12s on this PR. So the PR itself is wired up correctly; only the Actions side is unavailable.

**A re-trigger will be needed once Actions recovers.** Because the runs were never created rather than created-and-failed, recovery does not retroactively produce them, and no `synchronize` event will fire without a new push. Two clean options, no empty commit required:

- `gh pr reopen 718` after closing — `ci.yml` lists `reopened` in its `types:`, and `aarch64-crossbuild-check.yml` uses the default `pull_request` types, which include it; or
- `gh workflow run ci.yml --ref agent/dispatch-issue-450-323` and `gh workflow run aarch64-crossbuild-check.yml --ref agent/dispatch-issue-450-323` — both workflows declare `workflow_dispatch` and both exist on the default branch, which is the condition GitHub requires.

I am explicitly **not** reporting this as verified until those two jobs are green. Local gates are all green (see the description); what is missing is exactly the part only CI can supply.

---

## Related finding for the out-of-scope policy question

While diagnosing the above I read the repo's Actions permissions and noticed a setting that bears directly on the "should we pin the other 15?" decision flagged in the description:

```
$ gh api repos/vfarcic/dot-agent-deck/actions/permissions
{"enabled":true,"allowed_actions":"all","sha_pinning_required":false}
```

GitHub now exposes **`sha_pinning_required`** as a native repository-level Actions policy — it can enforce SHA-pinned action references at the platform level rather than by convention and review. It is currently `false` here.

That is worth knowing before anyone hand-pins the remaining 15: flipping this to `true` would make every mutable-tag reference in the hand-maintained workflows fail rather than silently resolve, which turns "did we remember to pin it?" from a review question into a mechanical one — but it would also break those 15 immediately, so the ordering matters (pin first, then enforce). I have **not** changed the setting; it is a maintainer decision and outside this PR's scope. Noting it only because it changes what the wider policy could look like.

prageethw (approved):
Confirmed both dependencies pin to genuinely immutable refs: cross pinned to --version 0.2.5 on crates.io, and DeterminateSystems/nix-installer-action + magic-nix-cache-action pinned to full 40-char commit SHAs (verified not shortened/annotated-tag SHAs) in ci.yml. Scope stays confined to the three workflow files plus changelog fragments — no source touched, no check weakened. CI green.

PR #727, opened 2026-08-28
fix(xtask): derive clean-e2e-tmp's dead-owner floor from the longest lifetime cap tests pin (#679)

### LINKED ISSUE #679: clean-e2e-tmp's 600s dead-owner floor derives from a 300s cap that one test legitimately pins at 900s

vfarcic:
Closes #679.

## The defect

`cargo xtask clean-e2e-tmp` never reaps a root whose owning test process is alive, but a **dead** owner is not proof the tree is unreferenced: the deck's daemon `setsid`s out of the test's process group and can keep writing under that root for as long as its own `DOT_AGENT_DECK_TEST_MAX_LIFETIME_SECS` cap. So a dead-owner root is held for a floor first — and that floor was **600 s**, documented in `DEAD_PID_MIN_AGE` and in `docs/develop/e2e-temp-dirs.md` as *"2× the 300 s orphan cap"*.

It had not been 2× since #665. `orchestration_dispatch_002` pins **900** so its daemon outlives the test's own 300 s work budget (without which the failure dump reports `NO PANE — never spawned at all` for roles the same dump renders alive, #663), and `TuiDeck` applies the builder's `extra_env` **after** its `env_clear`, so that value reaches the child verbatim and #668's ambient `clamped()` never sees it. Between 600 s and 900 s the stated derivation no longer held: `--apply` could hand `remove_dir_all` a directory a live, still-entitled daemon was writing under.

Nothing was on fire — reaching it needs the test process killed **and** its daemon alive past t+600 s **and** `--apply`, which is not the default. But it is a **deletion** tool whose safety argument had gone 300 s short, which is why #679 asked for it to be closed deliberately rather than left implicit.

## What changed

Issue option 1 (raise the floor), plus the half that stops it recurring — option 1's own stated cost was that the constant would then "encode a number no single test explains, so it needs the derivation spelled out where it can be re-checked".

**1. The floor is derived, not literal.** `DEAD_PID_MIN_AGE = MAX_PINNED_ORPHAN_CAP_SECS * 2` (900 → 1800 s). The new constant is documented as *the longest cap any test pins*, which is what the floor is actually entitled to assume.

**2. linkage-check rule 11** fails the build when any `DOT_AGENT_DECK_TEST_MAX_LIFETIME_SECS` pin under `tests/` exceeds that constant. The next cap raise is now a red build naming the number to move rather than a silent re-opening of the same window. It reads two shapes off comment-stripped source:

- the variable adjacent to a literal — `.env(VAR, "900")`, `.with_env(...)`, `(VAR, "300")`, `(VAR.into(), "300".into())`, and the multi-line `.env(` spelling;
- a `const …MAX_LIFETIME…_SECS` binding with a literal value, which is how `WRAP_TEST_MAX_LIFETIME_SECS = "120"` is covered even though `wrap_io.rs`'s nine sites pass it by name.

A cap computed at run time is out of reach of any text scan — like rule 10 this is the belt, not the braces, and it covers every pin that exists today so the *next* one written is the case it is for.

**No per-line opt-out**, deliberately, and this is the one place it departs from rules 8 and 10. Those guard a local choice a reviewer can wave through at the site; this guards a pair of numbers that must move together, so the only correct response to a longer cap is to raise the constant and let the floor follow.

**3. The floor test pins the derivation** (`DEAD_PID_MIN_AGE == MAX_PINNED_ORPHAN_CAP_SECS * 2`) instead of asserting a bare `>= 600 s` — which is exactly what let the two drift.

Docs updated to match: `docs/develop/e2e-temp-dirs.md` (the "30-minute floor" section, the decision table, the `--older-than` note, plus a new subsection on why the two numbers are now checked against each other), `CLAUDE.md` rule 14, and the doc comments in `tests/common/child_lifetime_bound.rs`, `tests/agent_lifetime_bound.rs` and `src/wrap.rs` that pointed at the old derivation or at #679 as open.

### Why not options 2 or 3

**Option 2** (clamp the explicit path, raise the ceiling to 900) would reorder `TuiDeck`'s `extra_env`, which other keys rely on winning, and would weaken the ambient clamp for everyone to legalise one value. **Option 3** as filed (have the reaper read a recorded cap) means trusting state read from the directory it is about to delete. A build-time guard bounds the *written* pins instead, which is the population the floor actually needs bounded.

## Cost

A killed run's roots are held **30 minutes** instead of 10 before `--apply` can take them. Real, paid on the safe side, and small against the case #461 was filed for (280 roots / 6.2 GB, youngest 4h09m) and against the usual reason to reach for the tool — a tmpfs filling over hours, not seconds after a Ctrl-C. Nothing else about the reaper moved: a live owner is still never reaped at any age, `--older-than` still decides only the no-usable-PID roots, and `--ignore-liveness` is still the one operator escape.

## Verification

- `cargo test-fast` — **3359 passed, 0 failed**, including 9 new rule-11 tests and the tightened floor test.
- `xtask-linkage-check` against this tree — `ok (593 catalog ids, 521 annotations, 100 allowlisted, 11 rules)`.
- **Rule 11 proven to fire**, by temporarily lowering `MAX_PINNED_ORPHAN_CAP_SECS` to 100 and re-running against the real checkout: it reported all six genuine sites across both shapes — `e2e_dispatcher_mode.rs:1023` (the 900), `common/mod.rs:673` and `:5596` (tuple/owned literals), `common/mod.rs:70` and `common/child_lifetime_bound.rs:74` and `agent_lifetime_bound.rs:67` (constants) — and nothing else. Restored to 900 afterwards.
- **CLI smoke on a fixture** (dry run, `--root` a scratch dir): a 40m dead-owner root reaps; 20m and 5m ones are kept — the 20m case is precisely the one that would have been deleted under the old 600 s floor — and a live-owner root at 3h is kept.

No `PROTOCOL_VERSION` or wire change: this touches an xtask, its tests, and docs, so CLAUDE.md rule 12's cross-version check does not apply. Changelog fragment: `changelog.d/679.misc.md`.


prageethw (approved):
Verified the fix end to end: `MAX_PINNED_ORPHAN_CAP_SECS` (900s) is now the single source both `DEAD_PID_MIN_AGE` and the new linkage-check rule 11 derive from, closing the 600–900s window issue #679 describes. Rule 11's regex-based scan is well-covered (multi-line `.env()`, tuple/owned shapes, the `WRAP_TEST_MAX_LIFETIME_SECS` const-proxy case, comment-blindness) and the load-bearing assertion (`the_floor_is_derived_from_the_cap_the_rule_enforces`) pins the guard to the same constant it protects. Docs (CLAUDE.md, e2e-temp-dirs.md, the two lifetime-bound test files) are all updated consistently. CI green including Greptile.

PR #728, opened 2026-08-28
fix(remote): strip terminal control and bidi bytes from remote-supplied text before it is printed (#723)

### LINKED ISSUE #723: Remote ssh stderr reaches the terminal unescaped in `remote add` / `upgrade` / `list` / `remove`
Surfaced while reviewing PR #721 (`remote doctor`, PRD #345). The PR body flagged it explicitly as out of that PRD's scope and said it was "worth its own follow-up issue" — this is that issue. Nothing about it is fixed by #721.

## The gap

`SshError`'s `Display` embeds the remote's **raw stderr** verbatim. In `src/remote.rs`:

```
84:  "Could not reach {host}:{port}. … \nDetails: {detail}"
92:  "ssh authentication to {target} failed. … \nDetails: {detail}"
105: #[error("ssh to {target} failed: {detail}")]
807: "Failed to download dot-agent-deck v{version} … \nDetails: {detail}"
```

`detail` is set from `stderr.trim().to_string()` in `classify_ssh_error` (`src/remote.rs:114`) and from `install.stderr.trim()` (`src/remote.rs:947`) — both untrusted bytes chosen by the endpoint being contacted.

Those `Display` strings then reach the user's terminal unescaped through four `eprintln!("{e}")` sites in `src/main.rs`:

| line | command |
|---|---|
| 1268 | `remote add` |
| 1281 | `remote list` |
| 1293 | `remote remove` |
| 1317 | `remote upgrade` |

So a host that answers an ssh connection can write arbitrary terminal control sequences into the operator's terminal — cursor movement, scrollback manipulation, title setting, overwriting earlier output. `remote add` is the sharpest case: it is the *first* contact with an endpoint the operator has not yet decided to trust, and it runs before any registry entry exists.

## Why it is worth fixing rather than tolerating

The repo already has the seam and already applies it elsewhere. `src/untrusted_text.rs::strip_control_and_bidi(s, keep_newlines)` is the existing helper, and `remote doctor`'s own render path (added in #721) escapes its remote-derived text — `src/remote_doctor.rs` checks `c.is_control() && c != '\n'`. So `remote doctor` is safe and `remote add` / `remote upgrade` sitting next to it are not, which is an inconsistency inside one command family rather than a missing capability.

## Suggested shape

Seal it at `Display` rather than at each `eprintln!`, the way `src/project_config.rs:45` records doing for the same class of reason ("Sealed here at `Display` rather than at `main.rs`'s `eprintln!(\"{e}\")`"). Passing `detail` through `strip_control_and_bidi(.., keep_newlines: true)` at construction — in `classify_ssh_error` and at `src/remote.rs:947` — covers all four print sites at once and cannot be forgotten by a fifth caller added later.

Newlines should be kept: the messages deliberately put `Details:` on its own line, and multi-line ssh stderr is genuinely useful diagnostic output.

## Tests

Unit tests in `src/remote.rs` asserting that a `classify_ssh_error` built from stderr containing `\x1b[2J`, `\x1b]0;title\x07` and a bidi override yields a `Display` string with the control bytes stripped and the newlines intact. No e2e tier needed — this is a pure string-transformation boundary.

vfarcic:
Closes #723.

`SshError`'s `detail` was `stderr.trim().to_string()` and its `Display` embedded it verbatim, so the four `eprintln!("{e}")` sites in `src/main.rs` — `remote add`, `list`, `remove`, `upgrade` — wrote whatever a host chose to put on ssh's stderr straight into the operator's terminal. `remote add` is the sharpest case: first contact with an endpoint nobody has decided to trust yet, running before any registry entry exists.

## What changed

`scrub_remote_text` (`src/remote.rs`) wraps the existing `untrusted_text::strip_control_and_bidi(s, keep_newlines: true)`, and every remote-supplied string is passed through it **at the point it enters an error type** — not at the four `eprintln!` sites. That is the constraint the issue records, and the reason `src/project_config.rs` gives for the same decision: one seam covers every sink by construction and cannot be forgotten by a fifth caller. Sanitizing at the print sites would pass the same tests today and rot the first time a `remote` subcommand is added.

Newlines are kept deliberately — these messages put `Details:` on its own line, and multi-line ssh stderr is the useful part of the diagnostic. The trim runs **after** the strip, so a trailing line made of nothing but an escape sequence cannot leave a blank line behind, which `stderr.trim()` alone did.

## Scope: wider than the issue as filed, deliberately

#723 names two constructors — `classify_ssh_error` and the `DownloadFailed` site. Both are fixed. But `remote add` reads the remote's own bytes at **four more** points that print through the same four `eprintln!`s:

| variant | source | reached when |
|---|---|---|
| `UnameFailed` | `uname -s -m` stderr | the arch probe fails — **before any ssh error is classified** |
| `UnsupportedArch` | `uname -s -m` stdout | the probe's output does not name a supported platform |
| `VersionMismatch` | the remote binary's stdout/stderr | the installed version does not match |
| `HooksInstallFailed` | `hooks install` stderr | the hook install fails |

The first of those is reached by a host that connects perfectly well, which makes it *more* available to a hostile remote than the `SshError` path, not less. Fixing only `detail` would have left the sharpest case open while the changelog fragment claimed the opposite, so all of them are scrubbed the same way. Happy to narrow this back to the two named sites if you'd rather they were a separate PR.

**Still untouched, and flagged rather than silently fixed:** `add` and `upgrade` relay the remote `hooks install` command's *successful* stdout with a bare `print!("{}", hooks.stdout)`. That is a passthrough decision (should relaying a subcommand's console output be filtered at all?) rather than an error-path one, and it is outside what #723 argues about. Worth its own issue if you want it closed.

## What did not change

Classification reads the **raw** bytes throughout — the ssh error matcher, `detect_platform`, and the `actual != version` comparison — so nothing about *which* error you get depends on what was stripped. Only the error field is scrubbed. A remote whose output contains nothing unusual prints byte-identically to before.

One cross-effect is recorded in both doc comments: `connect::ssh_error_detail` reads `SshError`'s `detail` for `remote doctor` and escapes it. It now receives a stripped value, so its escape became a second line of defence and the doctor quotes the printable residue (`[2J`) rather than the escaped original (`\u{1b}[2J`) for transport failures. No safety change; the doctor's other producer-controlled fields, including `probe_remote_version`'s own stderr path, escape at their own seams and are untouched.

## Tests

Unit tests in `src/remote.rs`, all against one fixture carrying an ANSI erase (`\x1b[2J`), an OSC title-set (`\x1b]0;pwned\x07`) and a RIGHT-TO-LEFT OVERRIDE, spanning two lines:

- `classify_ssh_error_scrubs_control_bytes_from_every_detail_variant` — every `SshError` variant that carries a `detail` (`ConnectionRefused`, `AuthFailed`, `Other`; `HostKeyVerificationFailed` and `Io` carry none), asserting routing is unchanged, the control bytes are gone, the interior newline survives, and `Details:` keeps its own line.
- `download_failure_scrubs_control_bytes_from_the_remote_stderr` — the `install_and_verify` site.
- `remote_add_scrubs_control_bytes_from_every_remote_derived_error` — the four `RemoteAddError` variants above, driven through `add()` with a scripted `SshExecutor` double.
- `scrub_remote_text_keeps_interior_newlines_and_trims_the_edges` — the strip/trim ordering.

Each was verified to **fail** with the fix reverted, not merely to pass with it.

No e2e tier: this is a pure string-transformation boundary, as the issue says.

## Gates

- `cargo fmt --check` — clean
- `cargo test-fast` — 3354 passed, 0 failed
- `cargo test-e2e` deliberately **not** run (rule 5 bounds it to the pre-PR step and this change adds no e2e coverage)

**Rule 12 does not apply** — this touches neither the daemon, the TUI↔daemon protocol, orchestration, nor hooks. No wire shape moves and no field changes meaning, so `PROTOCOL_VERSION` is unchanged and no `.breaking.md` fragment is warranted; the cross-version manual test has nothing to exercise.

**Rule 4 does not apply** — no user-visible TUI surface changes. The changed text is CLI stderr, covered by the unit tests above.

Changelog: `changelog.d/723.bugfix.md`. `bugfix` because towncrier has no `security` type here and its direct precedent — the 0.38.0 "config that will not parse can no longer repaint the terminal `validate` prints into" entry — landed as one; it also gives the patch bump a security fix should get, where a new type's bump semantics would be undefined. The fragment itself is framed as the hardening it is. Say the word if you'd rather add a `security` type to `pyproject.toml`.


prageethw (approved):
Reviewed the terminal-injection fix. `scrub_remote_text` is applied at construction of every remote-derived error field (ssh classifier's `detail`, arch-probe stdout/stderr, version-check output, hook-install stderr) rather than at the print sites, so a fifth caller can't reopen the gap — and the classifier still matches on raw stderr, so scrubbing can't change routing. Stripping (not escaping) with newlines deliberately preserved is the right call for an error message read for advice rather than displayed as a report; the `remote doctor` policy split is documented and consistent. Test coverage is thorough: the hostile fixture is exercised across every `SshError`/`RemoteAddError` variant that carries remote text, with an explicit terminal-safety assertion helper. CI green including Greptile.

Issue #732, opened 2026-08-28
Durable binary resolver: decide a provenance/trust policy for candidates it pins into agent config


vfarcic:
Follow-up from the security audit of PRD #381's PR (`agent/dispatch-prd-381`). The cheap half of this was taken in that PR; this issue is the part that needs a design decision rather than a patch.

## Context

PRD #381 added `platform::paths::durable_binary_path()`, which decides what absolute path gets written into `~/.claude/settings.json`, `~/.codex/hooks.json`, the OpenCode plugin and Devin's config — files those agents **execute**. Resolution order:

1. `current_exe()` if it is not under `target/{debug,release}`;
2. a. `<home>/.local/bin/<name>` if it exists and is executable;
   b. otherwise the absolute path of `dot-agent-deck` as resolved on `$PATH`;
   c. otherwise refuse.

## What the #381 PR already does

- Rejects relative and empty `$PATH` entries (`is_untrustworthy_path_entry`).
- Rejects a candidate whose **file** is group- or other-writable.
- Requires effective-user executability (`access(X_OK)`), not merely "some execute bit set".
- Never writes a bare command name and never writes a path that does not currently exist.

## What is deliberately left open

The auditor's HIGH finding asks for full provenance validation on every candidate: expected **owner**, **ancestor-directory** writability, and validation of the **canonical target** behind a symlink.

That was not folded into #381 for three reasons, all of which deserve challenging here rather than being treated as settled:

1. An attacker-writable directory on the user's own `$PATH` is already a full compromise of that account, independent of this resolver — the user runs `ls`, `git` and everything else from that same `$PATH`.
2. The baseline this replaced is *worse*: on `current_exe()` failure the old code wrote the **bare name** (issue #536), which resolved through `$PATH` at every hook fire rather than being pinned once.
3. An ancestor-directory writability check has real false-positive risk — `/usr/local/bin` is group-writable by `admin` on stock macOS — and the failure mode here is a **hard refusal to install**, so a false positive breaks a legitimate user outright.

## The symlink question specifically

Step 2a deliberately does **not** canonicalize: if `~/.local/bin/dot-agent-deck` is a symlink into a checkout's `target/debug`, the resolver writes the *symlink spelling*, on the grounds that a stable name the user controls is more durable than the artifact behind it. The auditor's position is that a stable link name does not make its target durable, and that the canonical target should be validated for both artifact-location and trust even if the non-canonical spelling is what gets written.

[...]

Issue #731, opened 2026-08-28
agent_hook_config::write_atomic uses a predictable temp name opened with File::create, so a pre-created symlink is followed


vfarcic:
Found by the security audit of PRD #381's PR (`agent/dispatch-prd-381`). Pre-existing; deliberately left out of that PR because the fix is in shared infrastructure whose callers reach beyond that PR's scope.

## The problem

`src/agent_hook_config.rs:61-80` publishes via a same-directory temp file named deterministically — `.<name>.tmp.<pid>` — opened with `File::create`, which **follows symlinks**.

A writer able to place entries in the Codex or Devin config directory can observe or predict the deck's PID, pre-create that temp name as a symlink, and make the victim process truncate, chmod and overwrite an arbitrary victim-writable target before the final rename.

Affects: Codex `hooks.json`, Codex `config.toml` trust records, Devin `config.json` — every caller of the shared writer.

## Same class, separate sites

The malformed-config backup writes use predictable destinations with plain `std::fs::write` and follow symlinks the same way:

- `src/hooks_manage.rs:87-88`
- `src/codex_hooks_manage.rs:247-248`
- `src/devin_hooks_manage.rs:338-339`

## Suggested fix

- Random same-directory temp name opened with `create_new` / `O_EXCL` (or a securely configured `tempfile`), never opening an existing temp entry, then publish by rename.
- Refuse symlinked backup destinations, or create backups with `O_EXCL` and a collision-safe name.

## Why it was not fixed in the #381 PR

That PR routes OpenCode's plugin write **into** this shared publisher (it was previously the only one of the four writers using a non-atomic `std::fs::write`, which followed a symlinked destination and could leave a partial plugin for OpenCode to load). That is a net improvement — temp-plus-rename removes the destination-truncation and partial-file problems — and it inherits this temp-name predictability, which fixing here resolves for all four writers at once.

## Provenance

Auditor finding MEDIUM-3 from the review of PRD #381's PR. Traced from the code, not exploited.

vfarcic:
Correction to this issue's scope, from implementing the PRD #381 batch.

[...]

Issue #730, opened 2026-08-28
Codex and Devin hook install: unconditional repoint of valid deck rules, and suffix-only ownership deletes sibling handlers


vfarcic:
Two related defects on the same surface, found while reviewing PRD #381 (PR for `agent/dispatch-prd-381`). Neither is introduced by that PR — both are pre-existing — and both were deliberately left out of it to keep the diff scoped.

## 1. `install_impl` repoints a valid-but-different deck rule on every startup

PRD #381 establishes the invariant *"repair only when the target is **positively missing**, never merely because it differs from what I would write."* That invariant is met for Claude (`hooks_manage::command_is_dead_deck`, gated on `try_exists() == Ok(false)`) and for OpenCode (`opencode_manage`, after #381). It is **not** met for Codex or Devin: both `install_impl`s strip every deck-owned rule by exact command suffix and re-add unconditionally, so a deck-owned entry pointing at a *different but still valid* deck install is repointed on each launch.

PRD #381's M4 names Codex explicitly under "stale entries repaired on startup, idempotent", so this is a knowingly-unmet stated invariant rather than an omission.

Post-#381 this **converges** rather than thrashing — every worktree now resolves the same durable path, so the second write is a no-op in the common case. The residual: a user deliberately pointing Codex or Devin at a second, valid checkout gets silently repointed at their durable install.

## 2. Suffix-only ownership deletes a user's sibling handler

`codex_hooks_manage::rule_is_dot_agent_deck` and the Devin equivalent return true if **any** handler in a rule object ends with the deck's command suffix, and the caller then `retain`s the whole rule away — so a user who has one rule containing both a real deck command and their own audit command loses their own handler during automatic reinstall.

Claude already does this correctly: `strip_deck_commands` removes only the matching command and preserves siblings. Codex and Devin should do the same.

A user-authored command that intentionally ends in `hook --agent <agent>` is also indistinguishable from a deck entry. Note the *substring* case is already safe and tested (`codex_hooks_install_001_substring_match_does_not_delete_user_hook`) — a command merely mentioning `dot-agent-deck` is never touched. The collision is the exact suffix plus mixed-rule granularity.

### 2b. The same suffix gates Codex scoped trust

`codex_hooks_manage::deck_owned_entries` uses that same suffix as part of the trust predicate, so a crafted same-file command carrying the suffix can be classified for scoped trust if it appears in the listing during the install-to-trust window.

## Suggested direction

[...]

PR #733, opened 2026-08-28
fix(hooks): resolve a durable binary path before writing it into agent config (PRD #381, closes #536)

### LINKED ISSUE #381: Hook install writes a transient target/ build path into global config, silently breaking all hooks
## PRD: Resolve the hook binary path to a durable location

**Problem**: Hook installation writes `std::env::current_exe()` into persistent global agent config, so running a locally-built binary bakes a transient `target/{debug,release}` path into `~/.claude/settings.json`, `~/.codex/hooks.json` and the OpenCode plugin. When that build directory is cleaned or its worktree pruned, every hook silently breaks for every project. `auto_install()` fires on dashboard startup and prints nothing, so the damage is invisible until a hook fails.

**Solution**: Never write a build-artifact path into global config — resolve to the installed binary instead, self-heal config that already points at a missing binary, and close the test gap that let this ship (the test seam hardcodes a bare `dot-agent-deck`, so the real derivation is never exercised).

**Detailed PRD**: See [prds/381-hook-binary-path-durable-resolution.md](https://github.com/vfarcic/dot-agent-deck/blob/main/prds/381-hook-binary-path-durable-resolution.md)

**Priority**: High

### LINKED ISSUE #536: A current_exe() failure writes a PATH-relative hook command into a file Claude Code executes
When `current_exe()` fails, both `auto_install` and `install` fall back to the bare string `"dot-agent-deck"` (`src/hooks_manage.rs:332-334`, `:369-371`), and `auto_install_to` hardcodes it (`:357`). That writes

```
dot-agent-deck hook
```

— no path — into a file whose commands **Claude Code executes**.

## The consequence

Resolution then follows `$PATH`, so whichever `dot-agent-deck` is earliest on `PATH` at hook time runs, not the deck that wrote the rule. A writable early-`PATH` directory turns this into arbitrary execution on every tool call, for as long as the rule survives.

`current_exe()` rarely fails, so the likelihood is low. The finding is the *direction* of the failure: on losing the ability to identify itself, the code falls back to a **less** specific command, in a file that drives execution. The safe direction is the opposite one.

## Fix

If `current_exe()` fails, skip the install and warn, rather than writing an unqualified command. A hook rule that was never written is a recoverable state; one that resolves through `$PATH` is not obviously wrong to anyone reading the settings file.

## Interaction with the rename work

[#506](https://github.com/vfarcic/dot-agent-deck/pull/506) makes rule identification path-based rather than name-based, which makes an unqualified command anomalous in a second way: it can no longer be matched to the binary that wrote it, so it is also the shape that accumulates.

## Provenance

Security audit finding M-5 (MEDIUM, pre-existing) from the review of the work offered as #506, deferred so that PR could stay scoped. The matcher behaviour was executed directly; the `$PATH` consequence is traced from the code rather than exploited.

vfarcic:
## What this fixes

Hook installation wrote `std::env::current_exe()` into **persistent, user-level, all-projects** configuration. When the running deck is a local build that is a `target/{debug,release}` path — gitignored, removed by `cargo clean`, gone the moment its worktree is pruned. Every hook then fails with `/bin/sh: 1: <path>: not found`, naming a path the user never typed.

PRD #381 records this as observed live on 2026-08-04, across three integrations at once: `~/.claude/settings.json` pointing at a **deleted** worktree's `target/release` build, `~/.codex/hooks.json` at a then-**live** worktree's `target/debug` build (working, but primed to break), and the OpenCode plugin's `BINARY_PATH` at the same deleted path. 21 broken references in total.

Three properties make it a defect rather than a footgun: `auto_install()` is documented as never printing to stdout, so merely launching the dashboard from a local build rewrote global config with no consent; it re-armed on every dev-build run; and the repo already knew better — `remote.rs` invokes `~/.local/bin/dot-agent-deck hooks install` with the comment *"Use the absolute path consistently."*

## The resolver

One function, `platform::paths::durable_binary_path()`, decides what path gets written into another program's config:

1. `current_exe()` if it is **not** inside a `target/debug` or `target/release` directory — an installed binary installing itself is the normal case and keeps working.
2. Otherwise a durable path: `~/.local/bin/<name>` if it exists and this user can execute it; else the absolute path of `dot-agent-deck` as resolved on `$PATH`; else **refuse**.
3. Never a bare command name, never a path that does not currently exist.

The refusal is the load-bearing part: an immediate legible complaint at install time replaces a silent bad write discovered days later by a broken hook.

## Also closes #536

`current_exe()` failure used to fall back to the bare string `dot-agent-deck` — no path — into a file Claude Code executes, so resolution followed `$PATH` and a writable early-`PATH` directory became arbitrary execution on every tool call. #536's principle is that on losing the ability to identify itself, the code must fail toward **more** specific, never less. The resolver now refuses instead.

**Closing it took more than the write side.** The security audit found self-heal could *preserve* a legacy bare pin: `try_exists("dot-agent-deck")` resolves against the process **cwd**, so launching from any directory containing a file of that name made a dead bare pin look alive, and it survived into the rewritten config. A read-side predicate, `pin_is_repairable`, now holds an existing pin to the same invariant a freshly resolved path must satisfy — absolute, present, a regular file, executable by this user, not a build artifact. A non-absolute pin is repairable before any filesystem call, so the cwd is never consulted.

## Decisions worth recording

**Placement vs `binary_name()` (#560/#561, commit `190bd14`).** The resolver lives **beside** `binary_name()` in `src/platform/paths.rs`, not built on it. `binary_name()` answers a different question — *what generated text tells an agent to run the deck* — and may legitimately return a bare name, which is exactly what this resolver must never do. It reuses the module's helpers (`resolves_on_path`, `is_untrustworthy_path_entry`, `path_contains_executable`) rather than routing through `binary_name()`.

**Seven call sites, not the PRD's six.** `devin_hooks_manage.rs`'s `current_binary_path()` has the identical shape and writes a path Devin executes; the PRD's list predates that integration. M2's own wording — *"no remaining `current_exe()` feeding an external config write"* — is not satisfiable without it. `grep current_exe src/` now leaves only spawn/attach paths (`wrap.rs`, `mode_manager.rs`, `daemon_attach.rs`) and the resolver's own test seam.

**CLAUDE.md rule 12 — no `PROTOCOL_VERSION` bump, no `.breaking.md`.** Confirmed by measurement, not argument: a real v0.38.0 daemon served a branch TUI and branch CLI through delegate routing, work-done feedback, native hook status transitions (Thinking → Working → Idle) and raw `AgentEvent` delivery with no degradation. `PROTOCOL_VERSION` is 7 on both sides at source *and* over `daemon hello`, and `src/daemon_protocol.rs` is not in the diff. `changelog.d/381.bugfix.md`, patch bump.

**CLAUDE.md rule 9 — the experimental flag does not apply.** No new pane, field, command, tab, footer or keybinding; the PRD's Out of Scope forbids changing the `hooks install` CLI surface, and nothing here needed a new flag or env var. The only new output is an error on an existing command's failure path.

**#533 / #549 were out of scope and stayed out.** #533's *part 2* (`hooks_manage`'s binary-name-as-a-**primary**-value) is resolved here as a side effect — that literal was `auto_install_to()`'s hardcoded seam, deleted outright. Part 1 (`wrap.rs`) is untouched, and two further sites of the same construct at `mode_manager.rs:205`/`:425` are recorded on #533. #549 is unaffected: `binary_name()` is unchanged.

## Closing the coverage gap

The PRD names why this shipped: `auto_install_to()` hardcoded `let binary_path = "dot-agent-deck".to_string();`, so the tests never executed the broken derivation. That line is gone and the resolver injected.

The two new L2 tests (`hooks/install/004`, `hooks/install/005`) drive the **real** `target/debug` binary through the PTY harness, so `current_exe()` genuinely *is* a build artifact rather than a value a test supplied. They were written first and confirmed RED — 004 reported 24 violations in the exact 10 + 10 + 1 shape of the field report.

Every fix was additionally verified non-vacuous by reverting it and confirming the matching test went red, then byte-comparing the restored source. **No test in this PR passes with the resolver reverted.**

One harness change was required and is load-bearing: the L2 harness passes the **host** `$PATH` through, and this machine has a real `~/.local/bin/dot-agent-deck`, so without intervention every hook-dependent e2e test would have resolved step 2b to the *host's installed deck* rather than the branch build — silently testing stale code, and failing outright on a machine with no installed deck. `tests/common/mod.rs` now seeds the sandbox HOME's `~/.local/bin/<name>` as a symlink to the binary under test, so the tier exercises the real resolution order instead of bypassing it.

## Gates

| gate | result |
|---|---|
| `cargo clippy --workspace --all-targets --features e2e -- -D warnings` | clean (verified against a forced re-lint, not a warm-cache replay) |
| `cargo test-fast` | 3387 passed, 0 skipped (from 3350 on `main`) |
| `cargo xtask linkage-check` | ok |
| `cargo test-e2e` (full tier, `DOT_AGENT_DECK_RECORD=1`) | 9293 run, 9287 passed, 6 failed — **all six passed on individual rerun; none attributable to this branch** |
| rule 12 cross-version run vs v0.38.0 | pass; classification confirmed |

On the six: `devin_live_001`'s message reads like a hook regression, but the captured grid shows Devin still on its first-run welcome banner at the 60s mark with the prompt unconsumed, so no hook was due. `shell_activity_005`/`006` are *positive* evidence for this branch — their dumped event lists contain a delivered `SessionStart` with the right `pane_id`; only the model's Bash call never happened, which their own precondition assert self-diagnoses as prompt-adherence flake. The other three are outside every area this branch touches.

The cross-version run was fully sandboxed, and the operator's four real agent config files were verified **byte-identical** (SHA-256 and `cmp`) before and after — which matters more than usual here, since the subject of this PR is code that rewrites exactly those files.

## Findings deferred, with tracked issues

The review and audit surfaced more than this PR should carry. Each is filed rather than silently accepted:

- **#730** — Codex and Devin `install_impl`s strip and re-add every deck-owned rule unconditionally, so a valid-but-different pin is repointed on each launch; and suffix-only ownership can delete a user's sibling handler co-located in one rule object. Pre-existing. This is a **knowingly unmet** PRD M4 invariant, annotated as such in the PRD's Success Criteria rather than ticked. Worth knowing: the user-visible shape is a **flap**, not a one-time re-pin — two valid installs resolve different durable paths, so the pin oscillates between them.
- **#731** — `agent_hook_config::write_atomic` uses a deterministic `.<name>.tmp.<pid>` opened with `File::create`, which follows a pre-created symlink. This PR routes OpenCode's plugin write *into* that helper (it was the only one of the four writers using a non-atomic `std::fs::write`, which followed a symlinked destination and could leave a partial plugin for OpenCode to load) — a net improvement that inherits the temp-name issue, which #731 then fixes for all three affected writers at once.
- **#732** — candidate provenance policy for the resolver. The cheap half shipped here: reject a candidate whose own file is group- or other-writable, and require effective-user executability (`access(X_OK)`) rather than "any execute bit set". The rest — owner checks, ancestor-directory writability, canonical-target validation — needs a design decision about shared install prefixes that this PR should not make unilaterally, since `/usr/local/bin` is group-writable by `admin` on stock macOS and the failure mode here is a **hard refusal to install**.

## Process fix included

Rule 12's cross-version procedure had a hidden prerequisite that silently produces false passes: with **zero** agents under the old daemon, the newer TUI takes `MismatchAction::SilentRestart`, SIGTERMs it and lazy-spawns its own with no prompt and no output — so every check then passes against a *new* daemon and you report a clean cross-version gate having measured new-against-new. Rule 12's *"with an agent under it"* turns out to be load-bearing, and the consent prompt must be **declined**. Both are now written into `docs/develop/versioning.md`.

Closes #381
Closes #536


prageethw (approved):
Reviewed the durable hook-binary-path resolver. `durable_binary_path`'s resolution order (non-artifact current_exe → ~/.local/bin candidate → $PATH walk → refuse) correctly never falls back to a bare crate name (issue #536's actual vector — a bare name in a hook command is resolved by the agent's own `/bin/sh` $PATH, not the deck's), and the artifact check is component-wise rather than substring so it doesn't misfire on a path merely containing 'target'/'debug'/'release'. The write-mode and access(2)-based executable checks (vs. the old `mode & 0o111` bitmask, which accepted a file this user can't actually execute) are correct fixes for real audit findings, scoped deliberately narrowly with the remaining gap tracked as #732 rather than silently widened. `pin_is_repairable`'s self-heal correctly treats a malformed (bare/relative) pin as always repairable regardless of `try_exists`'s cwd-relative behavior, closing the original #536 gap on the read side too. The Windows basename-matching fix (`binary_names_match`) shows real iteration — caught by this PR's own build-windows run and then made platform-injectable for testability, which is the right response to a defect a Linux dev machine can't reproduce. Extensive unit coverage for every resolution step and edge case (symlinks, group/world-writable, unusable-but-exec-bit-set). CI green across Linux/Windows/macOS.

PR #737, opened 2026-08-29
fix(ci): read flow-style YAML pins in the lockstep guard (#710)

### LINKED ISSUE #710: check-pin-lockstep's toolchain scanner anchors at line start, so flow-style YAML pins are invisible to it but visible to Renovate
Raised by Greptile as a P2 on #707 and deferred there deliberately. It is real, and it is a mismatch against the principle `scripts/check-pin-lockstep.sh` states about itself: *"A pin this script reads but Renovate does not — or the reverse — would be a lockstep between the wrong two things."*

## The gap

`scan_workflow_toolchain` finds sites with a line-start anchor:

```bash
grep -nE '^[[:space:]]*toolchain:' "$f"
```

`renovate.json`'s customManager is **unanchored**:

```json
"matchStrings": ["toolchain:\\s*(?<currentValue>\\d+\\.\\d+\\.\\d+)"]
```

So a flow-style input — `with: { toolchain: 1.98.0 }` — is tracked by Renovate and **invisible to the guard**. If such a site existed and drifted while the block-style sites agreed, the guard would report `ok`.

The `cargo-nextest` scanner does not have this problem: it already matches the token anywhere on a non-comment line (`^[[:space:]]*[^#[:space:]].*cargo-nextest`).

## Why it was not fixed in #707

**Un-anchoring naively makes things worse.** `ci.yml`'s own header comment reads:

```
  # match on `toolchain:` and `tool: cargo-nextest@` anywhere under
```

That line contains the literal `toolchain:` followed by a backtick. An unanchored match picks it up, `trim_value` yields a non-semver, and the guard reports **the repository's own documentation as an unreadable pin** — a false positive on the very file the check exists for. (The nextest scanner dodges this only because its outer grep excludes comment lines.)

A correct fix therefore needs two changes together, not a one-line regex swap:

1. match the token on any **non-comment** line, mirroring the nextest scanner's `^[[:space:]]*[^#[:space:]].*` shape; **and**
2. extract the value with a pattern rather than `${hit#*toolchain:}` + suffix-stripping, because `with: { toolchain: 1.98.0 }` currently trims to `1.98.0 }`, which fails the semver test and would be reported as unreadable even though it is a perfectly good pin Renovate reads fine.

Step 2 also has to stay compatible with the two behaviours added in #707: a quoted-but-valid value must still FAIL (Renovate cannot read it), and a `$`-expansion must still be skipped (`windows-cross-check` echoes `toolchain: $WINDOWS_CROSS_CHECK_TOOLCHAIN` as diagnostics).

## Priority

Low. No workflow in this repository uses flow style for `with:`, and #707 was already carrying substantially more than #648 asked for; a scanner rework belongs in a change that can be reviewed and tested on its own.

## Done when

`with: { toolchain: X.Y.Z }` is read as a pin and compared, `ci.yml`'s header comment is still not read as one, and the quoted-pin and `$`-expansion behaviours from #707 are unchanged — each covered by a test in `xtask/linkage-check/src/pin_lockstep.rs`.

vfarcic:
Closes #710.

## The defect

`scripts/check-pin-lockstep.sh`'s `scan_workflow_toolchain` found its candidate sites with `grep -nE '^[[:space:]]*toolchain:'`. That line-start anchor matches **block-style YAML only**, while `renovate.json`'s customManager is unanchored (`toolchain:\s*(?<currentValue>\d+\.\d+\.\d+)`). So a flow-style mapping — `with: { toolchain: 1.98.0 }` — is a real pin Renovate tracks and bumps, and the guard could not see it at all. Reproduced against `origin/main`'s script with one block-style site agreeing and one flow-style site a whole minor release away:

```
ok: Rust toolchain pinned at 1.97.1 on both sides (1 workflow site(s))
ok: cargo-nextest pinned at 0.9.143 on both sides (2 workflow site(s))
exit=0
```

The same fixture now reports the drift and exits 1.

## The sibling scanners

Both were checked, as the issue asks.

- **`scan_workflow_nextest`** does *not* have the anchor problem — it already matched its token anywhere on a non-comment line — but it has the same class of bug one step later: its value token class ran to the closing brace, so `{tool: cargo-nextest@0.9.143}` came back as `0.9.143}`, failed the semver test, and was reported as an **unreadable pin** rather than compared. Fixed in the same change.
- **`scan_devbox`** is unaffected. It reads `devbox.json` with `grep -oE "\"$name@[^\"]*\""`, which is unanchored already, and JSON has no block/flow distinction or comments.

Both workflow scanners now share one line finder (`pin_lines`) and one value extractor (`pin_value`), so they cannot diverge again.

## The deliberate decision, and why

A flow-style pin is **accepted and compared**, not reported as unreadable. The script's own stated principle is that "a pin this script reads but Renovate does not — or the reverse — would be a lockstep between the wrong two things": flow style is tracked, so it can drift, and comparing pins that can drift is the entire job. Reporting it would be a false alarm claiming Renovate cannot read something it reads perfectly well. This reasoning is written into the script inline, matching that file's style.

Un-anchoring naively makes things worse, which is why this was not a one-line regex swap. `ci.yml`'s own header comment reads ``match on `toolchain:` and `tool: cargo-nextest@` anywhere under .github/workflows``; an unanchored match reads that line as a site and reports **the repository's documentation of this very check** as an unreadable pin. Full-line comments therefore stay excluded — the one deliberate divergence from Renovate, and a cheap one: Renovate ignores that line too, for its own reason (no bare `X.Y.Z` follows the token there). What it costs, if someone ever writes a literal version into a full-line comment, is a pin Renovate bumps and this guard never compares — the harmless direction, and the same `ci.yml` header already says in prose not to write versions there.

#707's two behaviours are unchanged and covered:

- a **quoted-but-valid** value still FAILS, now in flow style too — `{ toolchain: "1.97.1", components: clippy }` stops the value at the `,` with its quotes intact, so they reach the semver test and fail it;
- a **`$`-expansion** is still skipped. That matters more now than it did: `windows-cross-check`'s `echo "toolchain: $WINDOWS_CROSS_CHECK_TOOLCHAIN"` puts the token mid-line, so the anchored scanner could never reach it and the skip was exercised on the nextest side alone.

## No false positives

Output on the repository as it stands is **byte-identical** to `origin/main`'s script — `diff` of both runs is empty: seven toolchain sites, three `cargo-nextest` sites, exit 0. The two mid-line `toolchain:` occurrences that un-anchoring newly exposes (`ci.yml:267`'s `echo` and `ci.yml:586`'s header comment) are handled by the `$` skip and the comment exclusion respectively.

## Tests

Four added to `xtask/linkage-check/src/pin_lockstep.rs` (15 -> 19), in the existing style:

- `a_drifted_flow_style_toolchain_pin_fails` — the exact defect shape (block site agreeing, flow site drifted); also asserts the failure is *not* reported as "unreadable", pinning the accept-and-compare decision;
- `a_drifted_flow_style_nextest_pin_fails` — the sibling scanner's half;
- `agreeing_flow_style_pins_pass` — the false-positive guard; this is what goes red if the value extraction is ever widened back to the rest of the line, since `1.97.1 }` is not a version;
- `a_quoted_flow_style_toolchain_pin_fails` — #707's quoted-pin rule surviving in the spelling that could most easily have broken it.

**Every one of the four was verified to FAIL against `origin/main`'s script** (swapped the script in, ran the tier, swapped it back). The comment on the existing `shell_expansions_and_comments_are_not_pins` was also updated to record that it only became load-bearing for the toolchain half with this change.

## Gates

- `cargo test-fast` — **3354 passed, 0 failed** (3350 before; the 4 new tests confirmed to run, see the `pin_lockstep::*` lines).
- `cargo test-e2e` not run: this change is shell + `xtask` only and touches no `src/`.
- **CLAUDE.md rule 12's cross-version contract check does not apply** — no daemon, TUI<->daemon protocol, orchestration or hook surface is touched, so `PROTOCOL_VERSION` is unchanged and no `.breaking.md` fragment is warranted. Stated explicitly rather than silently skipped.

Changelog fragment: `changelog.d/710.misc.md` (CI/tooling; not user-facing).


prageethw (approved):
Verified: the anchoring bug fix in `pin_value`/`pin_lines` correctly makes flow-style YAML pins (`with: { toolchain: 1.98.0 }`) visible to the guard, matching Renovate's unanchored regex, while the full-line-comment exclusion avoids false-positiving on the script's own documentation. All 4 new tests confirmed to fail against pre-fix script per the PR description; output on current repo state is byte-identical. CI all green (build/build-windows/build-macos/security/nix).

PR #738, opened 2026-08-29
fix(hooks): publish agent hook configs through an exclusive create over an unpredictable temp name (#731)

vfarcic:
Fixes the finding in #731 (auditor MEDIUM-3 from the review of PRD #381's PR).

## The defect

`src/agent_hook_config.rs`'s `write_atomic` — the shared publisher behind Codex's `hooks.json` and `config.toml` trust records and Devin's `config.json` — built its scratch path as `.<name>.tmp.<pid>` and opened it with `std::fs::File::create`.

Both halves were wrong together. The name is fully derivable from the destination's file name and a pid anyone on the box can read, and `File::create` **follows a symlink**. A writer able to add an entry to the agent's config directory could pre-plant that name pointing at any file it could write, and the publish would truncate that target, chmod it, and fill it with the deck's bytes. The `std::fs::rename` afterwards moved the *symlink* onto the destination (rename does not follow one either), so the config the deck was asked to update did not even end up holding the result — nothing was left at the expected location to show what had happened.

This is only exposed where the config directory is not yours alone: a private `~/.codex` on a single-user machine was never at risk; a shared, group-writable or otherwise multi-tenant home was.

## Reproduced before fixing

Per `.claude/skills/reproduce-first`, the failing test came first. Against unfixed `main` the victim file came back holding `[110, 101, 119]` — the deck's own `"new"` bytes, in an attacker-chosen file, with the destination never written:

```
thread '…::write_atomic_does_not_follow_a_symlink_planted_at_the_legacy_temp_path' panicked:
assertion `left == right` failed: the publish followed a planted symlink and overwrote the victim
  left: [110, 101, 119]
 right: [118, 105, 99, 116, 105, 109, 32, 98, 121, 116, 101, 115]
```

## The fix

The scratch file is taken with `create_new` — `O_CREAT|O_EXCL`, which POSIX requires to fail with `EEXIST` when the path names a symbolic link, dangling or not — over an unpredictable name, redrawn on collision across a bounded 16 draws.

The two properties are independent **on purpose**, and the PR keeps them separately tested:

- `O_EXCL` is what makes following impossible. It holds even against a name guessed outright.
- The unpredictable name is the second layer: it means the name cannot be reserved ahead of time at all, and it demotes the remaining attack from a redirected write to a squat that costs one retry.

A collision is never resolved by unlinking whatever holds the name — that would let a squatter steer which entry the deck deletes, and there is no need, since the next draw is a different name.

The random tail is a keyed hash under `std::hash::RandomState`'s keys (seeded once per thread from the OS random source), plus a process-wide counter so no two draws in one run share an input. **Deliberately no new dependency**: `O_EXCL` — not the quality of the tail — is what makes a squatted name unfollowable, so the tail only has to be unguessable to a writer that cannot observe the keys, which a keyed hash already is. (The issue sanctioned either this or a securely-configured `tempfile`; `tempfile` is currently a dev-dependency only, and promoting it to a runtime dep is a supply-chain call that buys nothing the security property needs.)

### Behaviour deliberately preserved

- The Unix branch still copies the destination's current mode (defaulting to 0o600), so a publish never widens a config (#360, #382). Verified by the pre-existing `write_atomic_preserves_the_destination_mode_and_creates_owner_only`, unchanged and green.
- The write is still `sync_all`-ed before the rename, and the rename keeps atomic-replace semantics.
- No temp file is left beside the destination. The pre-existing stray-file assertion is unchanged and green.

### Two things tightened along the way

- **Creation is owner-only on Unix** rather than `0666 & !umask`, so the file is never briefly group- or world-readable in the window between `open` and the `chmod`. The mode the publish *lands* is unchanged — still the destination's own, applied by `fchmod`, which no umask filters.
- **A leak the refactor uncovered**: a failing `write_all` or `sync_all` used to leave the scratch file beside the destination, because only the `rename` had a cleanup path. All three now share one.

## Design decision: succeed, don't fail loudly

The issue text left open whether a squatted temp path should make the publish fail. **It succeeds** — the destination receives the new bytes.

Ablation B below is the argument. With `O_EXCL` but the *predictable* name restored, the planted symlink is correctly not followed, but every one of the 16 attempts hits the same squatted name and the whole publish fails. That converts an unexploitable squat into a reliable denial of the deck's hook installation, triggerable by anyone who can write one entry to the directory. Since the name is now unpredictable and redrawn, a collision is simply not an attacker-controlled condition worth failing on. Exhausting *every* candidate still surfaces as an `AlreadyExists` error rather than a silent write elsewhere, and has its own test.

## Tests

New, in the same file's `mod tests`, matching the existing style (all symlink-specific ones `#[cfg(unix)]` for `build-windows`):

- `write_atomic_does_not_follow_a_symlink_planted_at_the_legacy_temp_path` — the reproduction. Victim untouched, destination receives the new bytes, destination is a real file rather than the renamed symlink.
- `create_temp_at_skips_planted_symlinks_and_lands_on_a_free_name` — the collision path driven deterministically over a fixed candidate list (a live symlink, a dangling one, a free name), since a randomly drawn name cannot be made to collide on purpose. Asserts both squatters are skipped *and* left in place, and that the dangling symlink's target was not created.
- `create_temp_at_reports_alreadyexists_when_every_candidate_is_taken` — exhaustion is an error, not a silent write elsewhere, and does not disturb what holds the names.
- `temp_path_is_unpredictable_and_never_the_legacy_shape` — 64 draws are distinct, none is the legacy `.<name>.tmp.<pid>`, and each stays beside the destination.

All four pre-existing tests in the module are unchanged and green.

**Each half of the fix was ablated and confirmed load-bearing:**

| Ablation | Result |
|---|---|
| A — drop `O_EXCL`, keep the random name | `create_temp_at_skips_planted_symlinks_…` and `create_temp_at_reports_alreadyexists_…` go **red** |
| B — keep `O_EXCL`, restore the predictable name | `temp_path_is_unpredictable_…` and `write_atomic_does_not_follow_…` go **red** |

## Rule 12 — did this change the TUI↔daemon contract?

**No, and no `PROTOCOL_VERSION` bump is needed.**

`write_atomic` is an on-disk file-writing helper, entirely in-process. It carries no wire message, no field, and no daemon handler. It does not change *what* is written, *where* it is written, or the permissions the published file ends up with — only the name and open-mode of the transient scratch file that exists between the write and the rename, which no other process is meant to observe and which is unlinked by the rename. There is no semantic-break-behind-a-stable-wire either: an older and a newer build produce byte-identical destination files.

The changelog fragment is therefore `bugfix`, not `breaking` (`pyproject.toml` scopes `breaking` specifically to the cross-process TUI↔daemon compatibility axis).

## Scope note — the issue's second half is NOT in this PR

#731 also lists a same-class set of *backup* writes that use predictable destinations with plain `std::fs::write` and follow symlinks the same way:

- `src/hooks_manage.rs:87-88`
- `src/codex_hooks_manage.rs:247-248`
- `src/devin_hooks_manage.rs:338-339`

Those are untouched here — a different mechanism (best-effort backup of an unparseable config, in three separate modules) with its own behavioural question about whether a symlinked backup destination should be refused or renamed around. **Please keep #731 open**, or split the backup half into a follow-up, rather than treating this PR as closing it. That is why the commit says `Refs #731` and not `Closes`.

`src/hooks_manage.rs`'s own `write_atomic` is also deliberately untouched: it is not a copy of this one, and it already publishes through `create_new` (#534) — at the fixed name, unlinking a squatter and retrying once, which is safe (`remove_file` does not follow a symlink) but does have the DoS shape described above. The module header comment is updated to describe the two accurately now that they differ.

## Gates

- `cargo fmt --check` — clean
- `cargo xtask linkage-check` — ok (593 catalog ids, 10 rules)
- `cargo test-fast` — **3354 passed, 0 failed**
- `cargo test-e2e` — 9177 tests, 9176 passed. One failure, `dispatch_005_respects_max_per_run`, is **unrelated to this change and a load-contention flake**: every caller of the changed function is in `codex_hooks_manage`/`devin_hooks_manage`, while `dispatch_005` runs a default (Claude Code) task against a `gh` stub and never reaches either module. It failed at 0.916s under a saturated 9177-test parallel run; re-run in isolation it passes (1.04s), and the full `e2e_issue_dispatch` binary passes 91/91.


prageethw (approved):
Solid fix — O_EXCL (via `create_new`) is the real defense against the symlink-follow, and the unpredictable name is a well-reasoned second layer against DoS-by-squat. Tests are thorough (reproduction + ablations for both properties + exhaustion path), cleanup-on-failure gap is fixed too, and the scope note correctly leaves #731's backup-write half open. All gates green including Greptile.

Issue #745, opened 2026-08-29
PRD: an agent overview as the desktop app's landing screen


vfarcic:
**Placeholder PRD** — intent and known constraints only. Detail gets written when work starts.

## Intent

Opening the desktop app should land on an **overview of every agent the daemon is running** — a fleet-at-a-glance screen — rather than dropping straight into a working deck. From there you drill into a specific agent or orchestration.

Written single-daemon-first, but the layout must generalise to several daemons (see #742), because retrofitting a daemon dimension onto a screen designed around one is the expensive version of this.

## Where it stands today

The app opens directly into the control deck: a repo header plus agent tiles with embedded terminals. There is no index, no "what is running right now" summary, and nothing that survives having more agents than fit on screen. A daemon commonly runs 9+ agents, and the current layout renders full terminal panes for each.

## Why this is separate from #742

#742 is about **connecting to several daemons at once** — transport, per-endpoint state, partial failure. This is about **what you see first** and is worth having even with one daemon. They meet in the layout, so this PRD should define the overview so that "grouped by daemon" is a natural extension rather than a rewrite. Either can ship first; if this one does, it must not bake in the assumption that there is exactly one daemon.

## Known constraints

- **Most of the interesting columns do not exist yet.** In live mode `desktop/src/lib/bridge.ts:150-180` hardcodes `model: "Unavailable"`, `tokens: 0`, `cost: 0`, `contextPercent: 0`, `duration: "—"`, `attempt: 1`, `worktree`/`branch` `"Unavailable"`. An overview is exactly the screen that wants those. Either it shows what the daemon genuinely exposes today — status, agent type, cwd, tool counts, live hook events, orchestration membership — or this PRD grows a daemon-side dependency to surface the rest. **That choice should be made deliberately**, because an overview padded with "Unavailable" is worse than a narrower one that is honest.
- **Do not attach a terminal per agent on this screen.** PTY streams are live subscriptions; attaching 9+ at once to render thumbnails burns bandwidth and CPU for content nobody is reading. The overview should use snapshot/status data, with attach happening on drill-in. Note PRD #176 M1.3's multi-pane throughput qualification is still outstanding, so the ceiling here is unmeasured.
- Needs to work when the daemon is unreachable, incompatible, or owns zero agents — the empty state is the first-run experience.
- The existing Mode vs Orchestration tab buckets are the natural grouping; agents in an orchestration should read as a unit, not as N unrelated tiles.

## Related

[...]

Issue #742, opened 2026-08-29
PRD: connect the desktop GUI to multiple daemons simultaneously (fleet view)


vfarcic:
**Placeholder PRD** — intent and known constraints only. The detailed plan gets written when work starts, and this may split into more than one PRD.

## Intent

One desktop window observing and controlling **several daemons at once** — e.g. one local plus two remote servers — rather than one daemon at a time. At the limit this is a company-level control panel: many daemons, each with many agents, visible in one place.

## Why this may be the strongest case for a GUI at all

PRD #176 justifies the desktop app with the agent-communication graph — a *single* deck's DAG. A **cross-daemon fleet view is something the TUI structurally cannot do**, because it runs on one machine and shows one daemon. That is a sharper answer to "why a GUI" than the current flagship.

It is not hypothetical. On 2026-08-29 a single remote host was measured running three distinct long-lived daemons (an installed release plus two from separate dispatch worktrees), alongside ~166 short-lived test-scoped ones from a `cargo test-e2e` run. There is currently no way to see that whole.

## Known constraints

- **Depends on #741** (connect to any daemon anywhere). This is strictly a superset: multi-daemon without remote is just multi-local.
- **`DesktopState` is a singleton** (`desktop/src-tauri/src/lib.rs:997`, a single `.manage(DesktopState::default())`). It becomes a keyed map of bridges, one per endpoint, with connection status tracked per endpoint rather than globally.
- **The UI gains a dimension.** Today the hierarchy is deck → tabs → panes; this adds daemon above it. Worth deciding early whether agents from different daemons ever appear in one view or stay segregated.
- **Nothing new is needed in the protocol** for the observation half — N connections to N sockets, each speaking the existing wire.
- **Partial failure becomes normal.** With one daemon, disconnected is a whole-window state; with several, one endpoint being unreachable, incompatible, or mid-reconnect must not degrade the others. The current single `ConnectionStatus` cannot express that.
- **Build-stamp policy gets harder.** Several daemons will legitimately be different builds — see the stamp discussion in #741.

## Suggested phasing

Observe first, control second. The read-mostly subset (list, attach, watch events, stop) is exactly what already works remotely; anything that mutates project state needs the daemon-side project resolution from #741.

## Related

- #741 (connect to any daemon anywhere) — prerequisite
- PRD #176 (desktop GUI) — assumes a single local daemon
- PRD #174 (cross-project orchestration dispatch) — cross-project edges the fleet view would render

[...]

Issue #633, opened 2026-08-20
PRD: Execution telemetry, outcomes, and cost accounting


vfarcic:
## PRD: Execution telemetry, outcomes, and cost accounting

**Problem**: Users cannot consistently attribute elapsed time, retries, resource use, model spend, interventions, and verified outcomes to a task across heterogeneous agents.

**Solution**: Investigate a provider-neutral execution telemetry and outcome model, including what can be measured reliably, how it should be retained, and whether OpenTelemetry should be an export boundary.

**Detailed PRD on `main` after merge**: [prds/633-execution-telemetry-outcomes-cost-accounting.md](https://github.com/vfarcic/dot-agent-deck/blob/main/prds/633-execution-telemetry-outcomes-cost-accounting.md)

**Detailed PRD on the PR branch**: [prds/633-execution-telemetry-outcomes-cost-accounting.md](https://github.com/vfarcic/dot-agent-deck/blob/docs/candidate-agent-control-plane-prds/prds/633-execution-telemetry-outcomes-cost-accounting.md)

**Priority**: Medium

PR #779, opened 2026-08-30
PRD #745: an agent overview for the desktop app, live against a real daemon

vfarcic:
A fleet overview for the desktop app: every agent a daemon owns, grouped the way the daemon already groups them, described only by things that are actually true, and with **no terminal and no PTY** anywhere on it. Closes the first two iterations of PRD #745 (M1–M12). Refs #745, #801.

33 files, +8868/−157. Suite: **4001 Rust**, **208 frontend**, 60 desktop-crate.

## What it does

Open **Overview** from the rail against a live daemon and you get your fleet: standalone agents first (matching the TUI's always-first dashboard tab), then orchestrations as single units with roles in `role_index` order and the coordinator badged, then mode buckets. Status, agent, last activity, uptime, CLI, active tool, tool count, working directory and last prompt are available; **four are on by default** and a picker lets you choose the rest, remembered per mode across restarts.

Try it: `cd desktop && pnpm tauri dev`, or `pnpm dev` with `?fixture=1&state=crowded` for a fifteen-agent fleet with no daemon.

## The two things worth reviewing carefully

**Demand-driven attach (M7).** This is the substantive behavioural change and it touches code people rely on. Attaching a PTY was never tied to rendering one: `attachAgents` fired from `connect()`, from *every* `desktop://snapshot`, and from `start_daemon`, attaching every agent in the snapshot regardless of what was on screen. Each attach opens a real `AttachStream` and replays full scrollback — so nine agents cost nine sockets and nine replays before anyone looked at anything.

Now: shown terminals are always attached and **never capped** (the deck mounts one per tile, so a cap would kill visible panes); leaving one moves it to a bounded warm set, still attached, so returning costs no replay; overflow evicts least-recently-used with a full client-side teardown; and an **empty shown set flushes warm to zero**, which is what makes the overview's claim true however you arrived at it. `MAX_WARM_TERMINALS = 3`.

**The columns are honest by construction.** `OverviewAgent` is a `Pick<>` of the fields a daemon genuinely reports, and `COLUMN_FIELDS` is `as const satisfies readonly (keyof OverviewAgent)[]` — so naming `model`, `cost`, `tokens` or `contextPercent` on this screen is a **compile error**, not a review catch. Those don't exist in daemon state at all (not on the wire, not in `RunningAgent`, not in `SessionState`); inventing them is #633's discovery work. Live mode also stopped fabricating `attempt: 1` and `branch`.

## Daemon changes — two, both additive, no bump

| | |
|---|---|
| `SessionSnapshot.last_activity_ms` | M9 — the daemon already maintained it as the `ListAgents` tie-breaker; it just never crossed the wire |
| `AgentRecord.spawned_at_ms` | M11 — new, stamped in `spawn_agent`, the source for Uptime |

Both `#[serde(default, skip_serializing_if = "Option::is_none")]`, which `daemon_protocol.rs:11-14` names an explicit do-not-bump case. **`PROTOCOL_VERSION` stays 8**, no `.breaking.md`. `AgentRecord` travels only inside the `ListAgents` pull reply — `BroadcastMsg` carries no variant of it — so there is no daemon→older-client push path where an added field could break a frame decode.

**Uptime supersedes the PRD's original rejection of session duration**, and the reasoning changed rather than being waived: `SessionState.started_at` is *event-derived*, so an agent that has never emitted a hook has none. Spawn time is signal-independent and an observation. It also gives the semantics asked for with no flag — a respawned worker reports its current iteration, an orchestrator nobody restarted reports its whole lifetime, because those are two different records.

## Rule 12 — cross-version test RUN, and it passed

Paired `630b0ab` against **v0.39.0** (`1ea0fe7`) — two real builds, no `BUILD_ID_OVERRIDE`. One `Attach protocol listening` line for the whole run, one daemon pid and build id end to end, and `server_version: 8` read off the **live daemon's wire** via a raw `Hello` frame. Mismatch prompt appeared naming the live agents; declining attached the branch TUI to the old daemon with full state restored. **Delegate routed, `work-done` round-tripped, status hooks repainted the card**, plus a pane spawn and byte-identical `daemon status` output from both binaries. Same-version controls at both ends. Nothing behaved differently.

**Two limits, stated rather than glossed:**
- **Stand-in agents** (`bash --norc --noprofile -i`), not real ones. Routing, hook delivery and card repaint are proven; a real agent under the pairing is not.
- **The desktop leg was not run** — Connect anyway → Overview → confirm Uptime and Last activity are *blank* against a previous-release daemon. No harness drives a real Tauri window (#836).

The run also found a hole in rule 12's own procedure — the daemon's 30s idle window is a *second* cause of the same false pass the rule warns about — now documented in CLAUDE.md.

## #801, fixed twice over

The app refused any daemon whose `git describe` differed, even with the protocol agreeing — and with live agents there was no path forward at all, since **Replace daemon** is correctly disabled then and the env-var override never reaches a Finder-launched `.app`.

1. **Connect anyway** — confirmation-gated, session-scoped, caveat stays visible all session.
2. **Compatibility is now classified by release version, not the git stamp.** Same `major.minor` (while 0.x) connects silently; a differing minor prompts, because under this project's policy a minor bump *is* the declared compatibility break. Parsing fails safe — only positive evidence that both keys parsed and matched reaches a silent connect. **The protocol check remains first and is never overridable.**

## Not in scope, deliberately

Drill-in navigation, a group view, a single-agent view, and promoting the overview to the actual **landing screen** — all iteration 3. **So this does not fully satisfy #745's title**, and the PRD says so plainly: an overview with no exits is worse than no overview, so it ships additively alongside the deck, which is untouched in layout and contents and remains the default.

Multi-daemon (#742) is not built, but the model carries a daemon identity and the overview keys by `(daemonId, agentId)`, so a second daemon becomes a sibling group rather than a rewrite.

## Known and not caused by this PR

- **`e2e-deterministic` is red on `main` itself** (`orchestration_remit_001/007`). Lane 1 is advisory, not one of the four required checks.
- **#807** — a load-sensitive `find_in_grid` race, triaged here as pre-existing (35/35 passes on both branches) with 36 more of the same shape in lane 1.
- **#836** — no browser-level coverage for the desktop web build; jsdom performs no layout, so the scroll region and card alignment are verified by eye only.
- **#804** — the TUI's reconnect `Last:` readout, made fixable by M9 and deliberately deferred: `last_activity` is tie-break ordering evidence, so seeding it could change which session wins.

## One open question for the reviewer

The overview shows 160 characters of every agent's **last prompt** per row, 512 on hover, on a screenshot-friendly screen. Recorded in the PRD as an accepted product choice with both sides argued — the TUI already renders the same value as its `Prmt:` line, but the overview aggregates it fleet-wide, which the TUI never does. Overrulable.


prageethw (approved):
Checked the two areas flagged as substantive: the additive daemon-protocol fields (`last_activity_ms`, `spawned_at_ms`) are proven forward/backward compatible by explicit round-trip tests in both directions, correctly justifying no `PROTOCOL_VERSION` bump; and the demand-driven attach/warm-eviction logic in `bridge.ts` is carefully invariant-commented and matches the PR description. `OverviewAgent`'s `Pick<>`+`as const satisfies` makes column honesty a compile error, not a review catch. No unsafe HTML rendering of daemon-supplied strings (last prompt, etc.). CI is green across the board including Greptile. LGTM.

PR #780, opened 2026-08-31
fix(devbox): provide Tauri's GTK/WebKit stack so both mandated gates run on Linux

### LINKED ISSUE #771: Both mandated gates are unrunnable in a devbox shell on Linux: devbox.json lacks the GTK/WebKit packages
Hit while implementing PRD #740 on 2026-08-30. Not specific to that work — this affects any Linux contributor and has done since `daf94f0`.

## Symptom

In a `devbox shell` on Linux, both gates CLAUDE.md mandates fail:

```
$ cargo clippy --workspace --all-targets --features e2e -- -D warnings
The system library `glib-2.0` required by crate `glib-sys` was not found.

$ cargo test-fast
.../dot_agent_deck_desktop-<hash>: error while loading shared libraries:
libgdk-3.so.0: cannot open shared object file: No such file or directory
```

Rule 2 (clippy, pre-commit) and rule 5 (`cargo test-fast`, per-task) are the two gates every contributor is told to run before every commit. Neither can run.

## Cause

`daf94f0` added `desktop/src-tauri` to the workspace. Both mandated commands carry `--workspace`, so both now select `dot-agent-deck-desktop`, which needs WebKitGTK and GTK. `devbox.json` provides neither — its `packages` list has the Rust toolchain, Node, pnpm and tooling, but no GTK/WebKit.

CI is unaffected and therefore silent about it: `ci.yml`'s `build` job installs the compile set explicitly (lines 172-175).

## The non-obvious part: `apt install` does not fix it

The intuitive workaround fails, and it fails in a way that wastes time:

```
$ sudo apt-get install -y libwebkit2gtk-4.1-dev libgtk-3-dev \
    libayatana-appindicator3-dev librsvg2-dev libxdo-dev
$ pkg-config --exists glib-2.0 && echo OK          # OK
$ find /usr/lib -name "libgdk-3.so*"               # present
$ cargo test-fast                                  # STILL FAILS
```

A devbox shell runs under **nix glibc**, whose loader cache does not exist:

```
$ ldconfig -p
ldconfig: Can not open cache file /nix/store/...-glibc-2.42-67/etc/ld.so.cache
$ ldconfig -p | wc -l
0
```

Zero entries, so system libraries are invisible to the dynamic linker no matter what apt puts on disk. `ldd` on the test binary reports `not found` for all ten of `libgdk-3.so.0`, `libgtk-3.so.0`, `libwebkit2gtk-4.1.so.0`, `libglib-2.0.so.0`, `libgobject-2.0.so.0`, `libpango-1.0.so.0`, `libgdk_pixbuf-2.0.so.0`, `libcairo.so.2`, `libcairo-gobject.so.2`, `libdbus-1.so.3`.

## And the obvious second workaround breaks something else

Exporting `LD_LIBRARY_PATH=/usr/lib/x86_64-linux-gnu` makes both Rust gates pass — and then breaks nix's Node:

```
$ node -e 1
node: symbol lookup error: /nix/store/...-nodejs-slim-24.19.0/bin/node:
undefined symbol: uv_tcp_keepalive_ex
```

It picks up the **system** libuv instead of its own. So `pnpm install`, `pnpm build`, `pnpm test` and any Tauri bundle all fail while that variable is set. The Rust gates and the frontend toolchain cannot both work in one shell that way.

Measured workaround for now: export it for `cargo` invocations only, and unset it for anything touching Node.

## Suggested fix

Add the GTK/WebKit packages to `devbox.json` as **nix** packages, so they resolve through the nix store where nix's own loader can find them. Roughly `webkitgtk_4_1`, `gtk3`, `libayatana-appindicator`, `librsvg`, `xdotool`, plus `pkg-config` if not already implied.

The point of doing it as nix packages rather than documenting an apt command is exactly the trap above: an apt instruction produces libraries the devbox loader cannot see, which looks like the instruction did not work.

## Worth checking alongside

Whether `flake.nix` needs the same, and whether a non-devbox Linux contributor (system toolchain, system glibc) is fine with plain apt — they likely are, since the nix loader is the thing at fault. If so, the docs answer differs by environment and should say so.

Bundling the desktop GUI locally needs more than the compile set, too: the `.deb` bundler additionally wants `patchelf`, `fakeroot`, `file` and `desktop-file-utils` (measured against a real `tauri build`).

vfarcic:
Fixes #771.

## What was broken

`desktop/src-tauri` became a workspace member in `daf94f0`, and both gates CLAUDE.md mandates carry `--workspace`. So `cargo clippy --workspace --all-targets --features e2e -- -D warnings` (rule 2, pre-commit) and `cargo test-fast` (rule 5, per-task) both build `dot-agent-deck-desktop`, which needs GTK 3, WebKitGTK and glib — and `devbox.json` provided none of them. On Linux the first failed at build time, the second at run time. CI never noticed: `ci.yml`'s `build` job installs the compile set with apt itself. The only workaround was `--exclude dot-agent-deck-desktop`, which silently leaves the desktop crate unlinted and untested.

## What this changes

- **`tauri-deps/flake.nix`** (new, mirroring the existing `gcloud/` local-flake precedent) builds the *transitive* pkg-config closure of the same libraries `ci.yml` installs with apt. It ships `.pc` files and nothing else.
- **`devbox.json`** gains `path:tauri-deps#tauri-deps` and `pkg-config@0.29.2`, plus an `env` block pointing `PKG_CONFIG_PATH` at the profile.
- **`scripts/devbox-smoke.sh`** resolves each module through `pkg-config` and fails on any that is missing — CI's `devbox` job is the only job that can see this regress.
- **`docs/develop/desktop-gui.md`** documents the devbox path, the non-devbox apt path, and why mixing them wastes a day.

## Why a flake rather than six more `packages` entries

Two reasons, both measured:

1. **Transitivity.** A devbox package lands in the profile as a bare store path, so `gtk3` — even with its `dev` output selected via devbox's `--outputs` — contributes `gtk+-3.0.pc` and nothing its `Requires:` line names, because a Nix profile does not follow propagated build inputs. Measured directly: the profile held **1** `.pc` file with the packages added that way, and `pkg-config --exists glib-2.0` still failed. Spelling the closure out by hand is ~30 entries that rot on the next nixpkgs bump. The flake gets it from Nix's own pkg-config setup hook — **183** `.pc` files — and cannot drift.
2. **Platform.** `webkitgtk_4_1` and `libayatana-appindicator` exist on Linux only. Restricting a devbox package to Linux requires devbox's *object* form for `packages`, and `devbox add --platform` rewrites the whole array into it — which breaks `scripts/check-pin-lockstep.sh`, whose `scan_devbox` reads the toolchain pins as `"rustc@1.97.1"` string entries and would report `devbox.json pins no rustc at all`. The flake is Linux-only by itself and leaves the array untouched.

## Why no `LD_LIBRARY_PATH`

Issue #771 measured that `LD_LIBRARY_PATH=/usr/lib/x86_64-linux-gnu` fixes both Rust gates and then breaks Nix's `node` (`undefined symbol: uv_tcp_keepalive_ex`), taking `pnpm install` / `build` / `test` with it. Pointing it at a Nix tree instead would be the same class of hazard — `LD_LIBRARY_PATH` outranks `DT_RUNPATH` for every process in the shell.

It is not needed. Each `.pc` names absolute store paths, so the linker sees `-L/nix/store/…-gtk+3-3.24.52/lib`, and Nix's `ld` wrapper turns every in-store `-L` into a matching `-rpath`. Verified on the built test binary:

```
$ readelf -d target/debug/deps/dot_agent_deck_desktop-8fa9f018820c7996 | grep RUNPATH
 Library runpath: [ … /nix/store/…-gtk+3-3.24.52/lib:/nix/store/…-webkitgtk-2.52.6+abi=4.1/lib:
                    /nix/store/…-glib-2.88.3/lib:/nix/store/…-cairo-1.18.4/lib:… ]
```

`node --version` and `pnpm --version` still run inside the shell (see the smoke output below).

## Verification

Reproduced first, in the devbox shell on this branch's worktree, before any change:

```
$ cargo test-fast
error: creating test list failed
Caused by:
  for `dot-agent-deck-desktop`, command `.../dot_agent_deck_desktop-e643c55aa133309b --list --format terse` exited with exit code 127
  --- stderr:
  .../dot_agent_deck_desktop-e643c55aa133309b: error while loading shared libraries:
  libgdk-3.so.0: cannot open shared object file: No such file or directory
```

`cargo clippy` passed at that point *only* because this machine already had the apt packages installed by hand; `pkg-config` was the system one searching `/usr/lib/x86_64-linux-gnu/pkgconfig`. After the change `pkg-config` is Nix's, whose default search path contains only its own store path, so nothing can silently fall back to apt.

The GTK/Tauri crates were then cleaned (`cargo clean -p dot-agent-deck-desktop -p glib-sys -p gtk-sys -p webkit2gtk-sys …`, 1595 files / 2.7 GiB) so the build scripts and the link step re-resolved from scratch. All three gates, run inside the devbox environment:

```
$ cargo fmt --check                                                    # exit 0
$ cargo clippy --workspace --all-targets --features e2e -- -D warnings
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 5m 12s   # exit 0
$ cargo test-fast
     Summary [  40.001s] 3472 tests run: 3472 passed, 0 skipped           # exit 0
```

The desktop crate's 35 tests execute for the first time in that shell (`dot-agent-deck-desktop daemon_bridge::tests::…` etc.), where before the tier could not even list them.

`scripts/devbox-smoke.sh` under `devbox run`, exactly as CI's `devbox` job invokes it:

```
== desktop gui toolchain ==
v24.12.0
10.34.5
== tauri system libraries ==
0.29.2
  glib-2.0                   2.88.3
  gobject-2.0                2.88.3
  gio-2.0                    2.88.3
  gtk+-3.0                   3.24.52
  gdk-3.0                    3.24.52
  gdk-x11-3.0                3.24.52
  webkit2gtk-4.1             2.52.6
  javascriptcoregtk-4.1      2.52.6
  libsoup-3.0                3.6.6
  ayatana-appindicator3-0.1  0.5.92
  dbus-1                     1.16.2
  librsvg-2.0                2.62.3
  libxdo                     4.20260303.1
…
OK: devbox environment is usable
```

`node --version` / `pnpm --version` in that same output are the guard against the `LD_LIBRARY_PATH` regression issue #771 describes.

`scripts/check-pin-lockstep.sh` does not fire — `devbox.json`'s `packages` stays an array of `name@version` strings, so its `scan_devbox` reads it unchanged. It runs against the real checkout from `cargo test-fast` via `xtask/linkage-check/src/pin_lockstep.rs`, which is green above.

### `cargo test-e2e`

```
     Summary [ 263.297s] 9382 tests run: 9381 passed (3 slow), 1 failed, 0 skipped
```

The one failure is `e2e_shell_activity_real_agent shell_activity_005_real_claude_bash_child_trips_the_descendant_scan` — one of the two real-agent canaries `docs/develop/shell-activity-signal.md` describes, which drive a live Claude with a Bash tool and are LLM-variance-prone by construction. It saw `SessionStart` and then no matching `AgentEvent` within its 120s budget. Re-run alone per rule 6:

```
$ cargo test-e2e -j 1 -E 'test(shell_activity_005_real_claude_bash_child_trips_the_descendant_scan)'
        PASS [  26.435s] (1/1) …
```

`cargo test-fast` re-run afterwards on the same quiet machine: **3472 passed, 0 failed**.

<details>
<summary>An earlier e2e run on this branch reported 44 failures — that was machine load, and the two runs together are worth recording</summary>

The first run of this tier reported `9382 tests run: 9338 passed, 42 failed, 2 timed out` in **684s**. The machine was concurrently saturated by another agent session's eight `burner.sh` processes (from `…/dot-agent-deck-dispatch-issue-709/…/scratchpad/`), 16 cores at load average **62–96**. Once those exited (load average 1.17) the identical command on the identical tree gave 9381/9382 in **263s** — 2.6x faster with 43 fewer failures.

The failure mode under load was uniform and recognisable: `did not see "<string>" within 10s. Final grid:` with an *empty* grid, i.e. the spawned TUI produced no output at all inside a 10-second budget. Sampled at the time, `modal_001_help_close_closes_overlay` took 19.8s and failed cold under load, then passed in 4.6s alone.

Worth knowing for anyone else running this tier on the shared box: the e2e budgets are wall-clock and do not survive a neighbour saturating the CPU, so a red run there says nothing until it is repeated quiet.

</details>

## Rule 12

Not applicable: no daemon, TUI↔daemon protocol, orchestration or hook change. `PROTOCOL_VERSION` is untouched and no `.breaking.md` fragment is warranted. No cross-version run was performed, deliberately.

## Known limitations, stated rather than hidden

- **Intel Macs lose `devbox shell`.** The flake covers `x86_64-linux`, `aarch64-linux` and `aarch64-darwin`. nixpkgs 26.11 dropped `x86_64-darwin` and `throw`s on import for it with no config opt-out, which is the same reason the root `flake.nix` already excludes it. This extends that exclusion from `nix build` to `devbox shell`. Apple Silicon resolves to `emptyDirectory` and is unaffected. The remedy, if anyone needs it, is written in the flake: a second nixpkgs input pinned to `nixpkgs-26.05-darwin` for that system's empty output only. It is not done pre-emptively because flake inputs are fetched during evaluation on every platform, so every Linux contributor would pay a second nixpkgs fetch for a platform this repository already declares unsupported.
- **The `.deb` bundler set is not included.** Issue #771 notes `tauri build` additionally wants `patchelf`, `fakeroot`, `file` and `desktop-file-utils`. No gate in this repository runs `tauri build`, so I could not verify them the way everything above is verified; adding unverified packages is the failure mode this issue is about. They are documented as a manual step in `docs/develop/desktop-gui.md` instead.
- **`PKG_CONFIG_PATH` is set, not appended.** devbox's `env` expands only `$PATH` and `$PWD`, so appending is not expressible. A value from your login shell therefore does not reach the devbox shell — which is the wanted behaviour, since a silent fallback to `/usr/lib`'s `.pc` files reproduces exactly the build-succeeds/run-fails trap this fixes.


vfarcic:
Correcting the e2e caveat in the description: I have a quiet baseline now.

The original run reported 42 failed / 2 timed out in 684s while another agent session's eight `burner.sh` processes had this 16-core box at load average 62–96. Those exited; the identical command on the identical tree now gives:

```
     Summary [ 263.297s] 9382 tests run: 9381 passed (3 slow), 1 failed, 0 skipped
```

2.6x faster, 43 fewer failures. The one remaining failure is `shell_activity_005_real_claude_bash_child_trips_the_descendant_scan`, a real-agent canary that passes on an isolated re-run (`PASS [26.435s]`). `cargo test-fast` afterwards on the same quiet machine: 3472 passed, 0 failed.

No code changed — this is the same commit, re-measured. The description's verification section has been updated, with the loaded run kept in a collapsed block since the contrast is useful to anyone else running this tier on the shared box.

prageethw (dismissed):
Reviewed the diff (devbox.json/lock, tauri-deps/flake.nix, devbox-smoke.sh, desktop-gui.md, CONTRIBUTING.md, changelog fragment) plus the PR body's reproduction. The pkg-config-closure-via-flake approach is well-reasoned (transitivity + Linux-only platform gating without breaking check-pin-lockstep.sh's array-form parsing), the no-LD_LIBRARY_PATH rationale is verified against the built binary's RUNPATH, and devbox-smoke.sh's pkg-config assertion is the right regression guard since it's the only CI job that would otherwise stay green with the stack missing. All CI green, mergeable, no inline comments or unresolved threads.

prageethw (approved):
Re-approving after the merge-from-main commit (5e99502d) dismissed my prior approval — the diff is unchanged (same 8 files: devbox.json/lock, tauri-deps/flake.{nix,lock}, devbox-smoke.sh, desktop-gui.md, CONTRIBUTING.md, changelog fragment), no new commits touch this PR's own changes. Same unrelated `desktop-web` pnpm supply-chain failure as #778/#783, not caused by this diff.

PR #784, opened 2026-08-31
test: wait on the child, not the clock, for a spawned stand-in's first output (#709)

### LINKED ISSUE #709: Three fast-tier tests use fixed deadlines for a spawned child's first output and fail under heavy machine load
## Problem

Three tests wait a **fixed** wall-clock deadline for a freshly spawned child process's first output. Under heavy machine load the child has not been scheduled yet when the deadline expires, and the test fails with a starvation signature rather than a behavioural one.

Measured during the #702/#687 work, on a 16-core box at **load average 44** driven by concurrent agent sessions:

| Test | Failures in a 12-run full-tier streak |
|---|---|
| `delegate_012_slow_agent_toggle_proves_delivery_and_submission` | 3 |
| `delegate_022_delegate_during_an_in_flight_close_brings_the_role_back` | 2 |
| `idle_worker_008_closed_orchestrator_pane_id_reuse_receives_nothing` | 1 |

The signature is unambiguous: `delegate_012` panicked on a 2 s wait for a freshly spawned stub with `snapshot = ""` — the child produced *nothing at all*, rather than producing something wrong.

## Why this is not attributable to the change that found it

The #702/#687 branch's production diff at the point of measurement was **mechanically comment-only** — `git diff src/state.rs` was two hunks, a comment block and a new test inside `mod tests`, with zero added non-comment lines outside `mod tests`. A comment cannot flake a test in a different binary, and two of the three tests live in test binaries whose source was never edited.

Controls run alongside:

- a pristine-`HEAD` control run: **8/8 clean**
- a repeat 12-run streak at load ~4–9: **12/12 clean, 3194 tests each**

So the failures track machine load, not the branch.

## Why it is worth fixing rather than tolerating

These are fast-tier tests, which CLAUDE.md rule 5 makes the **per-task gate**. A per-task gate that fails when the developer's machine is busy is the reliability problem rules 5 and 6 exist to prevent: it trains people to re-run until green, which is exactly the habit that lets a real regression through. The risk is not theoretical here — during this same work a genuine `/013` discrimination defect was nearly misdiagnosed as "a low-frequency flake", and treating flakes as background noise is how that misdiagnosis becomes permanent.

This repo also runs several agent sessions concurrently on one box by design, so high load is a normal operating condition rather than an anomaly.

## Expected

Replace the fixed deadlines for a spawned child's *first output* with either:

- a wait that is bounded by an observable readiness condition rather than a duration (the pattern `orchestration/delegate/013`'s fixture now uses after #702 — it waits for the observer's own readiness marker, and separately proves the marker's termios rather than assuming it); or
- a load-aware deadline, if a genuine duration bound is unavoidable.

The goal is that "the child has not been scheduled yet" is never indistinguishable from "the child produced the wrong thing".

## Notes

- Not a regression, and not urgent — all three pass reliably on an unloaded box.
- Found while running the #702/#687 pre-PR gates; filed separately so it does not widen that PR's scope.

vfarcic:
Fixes #709.

Three fast-tier tests waited a **fixed wall-clock deadline** for a freshly spawned child to reach a state the test depends on. On a contended box the child has not been scheduled inside that window, so the assertion underneath reported an *empty* snapshot — a starvation signature indistinguishable from a real delivery defect. Measured in #709 on a 16-core box at load average 44: `delegate_012` failed 3 times, `delegate_022` 2, `idle_worker_008` 1, across a 12-run full-tier streak; 12/12 clean on the same commit at load 4–9.

## What changed, per test

**`orchestration/delegate/022` had TWO fixed deadlines, and both are now condition-driven.**

- A flat `sleep(400 ms)` before the close. What the scenario needs is not "400 ms passed" but "the stand-in has installed its `trap '' TERM`" — until it has, `StopAgent` kills it on the first signal and the close is over in milliseconds. The stand-in now prints a marker **after the trap and before the `exec`**, so its appearance proves the disposition is already `SIG_IGN`, and the test waits for that fact.
- A flat `sleep(200 ms)` standing in for "the close has entered its grace window". `stop_agent` runs through a spawned task and a socket round trip, so on a loaded box neither had necessarily reached `begin_pane_close` when the sleep expired — and the assertion then said the close was *not in flight*, which was true for the opposite of the reason it named. It now waits for `pane_close_in_flight`, the exact state the 200 ms was approximating.

While chasing the second one I also found that the neighbouring `assert_eq!` labelled "the worker stand-in must still be alive" only ever checked **registration**. That difference is the whole scenario: `close_agent` spends its grace only against a *running* child, so an already-dead one makes the window unobservable. It now checks liveness, which is what the sentence always meant.

**`orchestration/delegate/012` and `scheduler/idle-worker/008` were already polling on the condition; the fixed part was the *ceiling*** (2 s for a Python stand-in, 5 s for a `sh` one). A child's first byte has no observable precursor to wait on instead, so per #709's second option these take a **load-aware** ceiling — `common::child_boot_budget()`, a base scaled by `/proc/loadavg` ÷ CPUs and clamped to 6× — plus an **early return once the child is no longer live**. That second half is what keeps the wider ceiling honest: a stand-in that *died* rather than printed still fails about as fast as it did against 2 s, so the extra patience is spent only on the starvation case. `idle_worker_008` had a second such ceiling in the harness constructor every test in that file boots through (`ORCH-READY`, flat 5 s) — it is on this test's path and fires first, so it got the same treatment.

## Evidence

**A/B on a deterministic model of the failure.** Rather than wait for a lottery, each stand-in was given an injected boot delay larger than the ceiling it was waited for with (`time.sleep(3)` before the Python stub's first byte, `sleep 6` before the `sh` observer's marker, `sleep 1` before the trap). That is exactly "the child has not been scheduled yet", made repeatable. On the pre-fix code **all three fail**, and `delegate_012` reproduces the issue's exact signature:

```
delegate_012 … panicked: replacement slow-readiness stub never entered raw discard mode; snapshot = ""
idle_worker_008 … panicked: orchestrator raw-cat stub never became ready; snapshot = ""
delegate_022 … panicked: precondition: the close must still be in flight when the delegate lands …
```

With the fix and the **same** injected delay: 3 passed, 0 failed. The injection was reverted; it is not in the diff.

**Streaks under real load.** The box was carrying 40–200 load average from concurrent agent sessions plus CPU/fork-storm stressors throughout.

- `delegate_022` alone: **25/25 pass**, load 58–136 (and an earlier 12/12 at load 76–209).
- All three together: **25/25 pass**, load 60–121.
- Full fast tier (3437 tests), repeated: the three named tests did not fail in any run.

**What did fail, and why it is not this PR.** At load ≳60 the full tier does still produce failures — a different pair most runs — in tests #709 does **not** name: `delegate_013`, `delegate_021`, `dispatch_003`, `dispatch_018`, `codex_wrap_004`, `wrap_child_group_backstop_…`, `shell_activity_monitor_…`, `in_process_registry_spawn_arms_…`. Every one is the same defect class at a fixed deadline **outside this diff** — e.g. `delegate_021` dies on `wait_for_snapshot_needle(…, Duration::from_secs(2))` for a freshly spawned orchestrator with `snapshot = ""`, a line this PR does not touch. They are pre-existing and cannot be caused by a purely additive change to `tests/common/mod.rs` plus edits confined to the three named tests. I have **not** widened this PR to cover them: #709 names three tests, and widening silently would make this diff harder to review for no gain in what it claims. They want their own issue — say the word and I'll file it with this list.

## What was deliberately *not* widened

- **Negative windows keep their fixed lengths.** `delegate_012`'s zero-buffer control arm waits for a pointer that must never arrive; that window is always paid in full and its length is part of what it asserts, so it keeps its exact 1200 ms while only the buffered arm scales. The two arms now read from `POINTER_DELIVERY_SLACK` and spend it differently on purpose.
- **`delegate_012`'s `measured_readiness_window` band `[500 ms, 900 ms]` is untouched.** It is a fixture self-check on a 650 ms synthetic window and is load-sensitive in principle — but its two observation errors (noticing `RAW-READY` late shortens the measurement, noticing `CAT-READY` late lengthens it) come from the same 20 ms poll loop and cancel to first order. It was not the observed failure mode, and every way to make it load-proof either widens it (weakening) or rebuilds the fixture's clock. Flagging rather than touching it.
- No assertion was dropped, no matcher widened, no sleep lengthened. The two assertions that changed both got **stronger**: a new precondition that the trap is installed, and a liveness check where there was only a registration check.

## Rule 12

**Not applicable — test-only.** The diff touches `tests/**` and `changelog.d/` and nothing under `src/`, so no wire shape moved, no meaning behind a stable wire shifted, there is no `PROTOCOL_VERSION` bump and no cross-version run to make.

## Gates

- `cargo clippy --workspace --all-targets --features e2e -- -D warnings` — clean. Run with `--exclude dot-agent-deck-desktop`: that crate fails to load `libgdk-3.so.0` in this devbox, which is issue #771 and not this change.
- `cargo test-fast` (`--exclude dot-agent-deck-desktop`, same reason) — 3437 tests, clean on an unloaded box.
- `cargo test-e2e` — 9347 tests, 9343 passed / 4 failed. All four are **real-agent** tests in files this diff does not touch — `mode_live_002_real_haiku_user_journey`, `shell_activity_005`, `shell_activity_006`, `dispatch_015_three_real_claude_seeds_are_genuinely_confirmed` — and **all four pass when re-run in isolation** (4/4, 142 s). Three of them are already documented in `.config/nextest.toml` as dying under full-tier contention; `dispatch_015`'s failure was `did not see "No active sessions" within 10s` at `tests/common/mod.rs`'s pre-existing `WAIT_TIMEOUT`, i.e. the same class again, at a line this PR does not modify. This is the flaky-tolerant tier rule 5 describes.


prageethw (approved):
Reviewed the full diff — test-only (tests/common/mod.rs, delegate_prompt_injection.rs, delegate_respawn_recovery.rs, idle_worker_detector.rs) plus a changelog fragment, no production code touched. The design cleanly separates three wait kinds and treats each correctly: boot waits (load-scaled, condition-driven, early-exit on child death), behavior waits (load-scaled but still condition-driven), and negative windows (deliberately left unscaled — POINTER_DELIVERY_SLACK's zero-buffer control arm stays fixed). The trap '' TERM / SIG_IGN-survives-execve reasoning behind the new STUBBORN_WORKER_ARMED marker is correct. Empirical evidence (12-run streaks at specific load averages) backs the fix. No unresolved threads, no prior review, CI fully green.

Issue #656, opened 2026-08-23
The four real-agent Codex e2e tests fail when actually run — and currently hide behind a silent skip


vfarcic:
The four real-agent Codex e2e tests fail when actually run — and currently hide behind a silent skip

## Summary

`e2e_codex_hooks::codex_hooks_001` and its three siblings **do not pass**. On a ChatGPT-subscription host they never run at all: `CODEX_TEST_MODEL_DEFAULT` is `gpt-5.1-codex-mini`, which subscription auth cannot reach, so `check_codex_available` fails its probe and the tests SKIP. A skip prints `SKIP:` to stderr, which nextest hides on a green run — so this reads as a pass.

Set `DOT_AGENT_DECK_CODEX_TEST_MODEL=gpt-5.4-mini` (verified reachable) and they run, and fail.

**This is not root-caused.** Filing it with the evidence and — more usefully — the list of things that are *not* the cause, each disproved by measurement, because several are individually plausible and cost hours.

## Symptom

The deck receives only wrapper-sourced events. A representative trace:

```
Codex  SessionStart   metadata: session_start_origin = wrapper_fork   <- wrapper, not hook
Codex  Thinking       user_prompt: None                              <- stdout classifier
None   ShellBusy                                                     <- deck-side
```

No hook-sourced event ever arrives, so assertions wanting a `Thinking` that carries `user_prompt` time out. `ShellBusy` proves the turn genuinely ran and executed a shell command — the agent works; only the rich events are missing.

**The failure point moves between runs**, which matters for whoever picks this up:

- sometimes `tests/common/mod.rs` `wait_for` (the 120s event wait)
- sometimes the earlier grid assertion "the dashboard card never showed Thinking", where the card reads `● Working`

Same command, same machine, `-j 1`. So there is a timing component and it is **not** simply "hooks never fire".

## NOT the cause (each disproved by measurement)

- **Model choice.** `gpt-5.4-mini` and `gpt-5.4` fail identically. `user_prompt` comes from Codex's `UserPromptSubmit` payload, emitted before any model runs.
- **Parallel-load flakiness (#351 / #415).** Fails the same at `-j 1`, in isolation.
[...]

PR #792, opened 2026-08-31
fix(devbox): drop nodejs and pnpm so the Node plugin stops racing at startup (#791)

### LINKED ISSUE #791: devbox: cold .devbox materialization is not concurrency-safe, so a multi-role orchestration loses a role at startup
## Symptom

Starting a multi-role orchestration in a **fresh** worktree kills one role at launch, with one of two errors:

```
Error: error running script "agent-orchestrator" in Devbox: remove
  <worktree>/.devbox/virtenv/bin/setup-corepack.mjs: no such file or directory

Error: error running script "agent-orchestrator" in Devbox: symlink
  <worktree>/.devbox/virtenv/nodejs/bin/setup-corepack.mjs
  <worktree>/.devbox/virtenv/bin/setup-corepack.mjs: file exists
```

The two are complementary halves of the same non-atomic operation — one process removing the link while another creates it.

**It is silent unless you count roles.** The deck reports the orchestration as started; the unit simply runs with five roles instead of six, and an orchestration with no `agent-orchestrator` has workers and nobody to delegate to. Observed twice in one session, taking `agent-orchestrator` both times.

## Why now: `daf94f0`, 2026-08-29

`git show daf94f0^:devbox.json` contains **neither `nodejs` nor `pnpm`**; `daf94f0` ("Add visual agent control deck (#416)") adds both. `setup-corepack.mjs` is, per its own header, "the nodejs plugin's init_hook". Before that commit there was no Node plugin, no corepack hook, and nothing to race on — which is why this has no history before 29 August and why every orchestration prior to it was unaffected.

Same commit, same class of fallout as #771 (devbox.json lacking Tauri's GTK/WebKit packages).

## Root cause is broader than corepack

Disabling the nodejs plugin removes the corepack symlink entirely, but concurrent cold `devbox run` in the same directory then fails differently:

```
Error: error running script "true" in Devbox: compact json for hashing: unexpected end of JSON input
```

That is devbox reading a partially-written JSON state file. So the underlying issue is that **devbox's per-project `.devbox` materialization is not safe against concurrent invocations**, and the corepack symlink is one manifestation rather than the whole problem.

## What only bites once

The failure is confined to the **first** materialization of a cold `.devbox`. Once the directory is warm, every subsequent invocation succeeds — which is why the main checkout never fails, why a failed orchestration succeeds on restart (the failed attempt materialized the directory as a side effect), and why this only appeared once we began creating fresh worktrees per unit.

## Hypotheses already falsified — do not re-run these

1. **"Concurrency between roles."** Six simultaneous `devbox run -- true` in a cold worktree passed 6/6 with the plugin enabled. The race is real but timing-dependent, so a clean run proves nothing; a later 6-way run failed 1/6.
2. **"devbox.json content differs between the failing and working branches."** `agent/dispatch-e2e-ci-tiers` and `agent/dispatch-prd-745` carry identical `nodejs@24.12.0` + `pnpm_10@10.34.5`; one failed, one worked.

## Fixes

**Partial, free, and worth doing** — set `disable_plugin: true` on `nodejs` (`devbox add nodejs@24.12.0 --disable-plugin`). Verified in a throwaway worktree: the corepack symlink is gone, and `node --version` / `pnpm --version` still report `v24.12.0` / `11.22.0`.

This costs nothing here because **corepack is provably unused in this repo**: `DEVBOX_COREPACK_ENABLED` is set nowhere, `desktop/package.json` has no `packageManager` field, nothing references corepack outside `.devbox`, and CI invokes `pnpm install --frozen-lockfile` / `pnpm test` directly via `pnpm/action-setup`. The plugin's only effect is a symlink to a script that returns immediately.

**General** — warm the environment once, serially, before spawning roles: a single `devbox run -- true` in the new worktree, after which no role can hit the window. Doing this inside `dispatch` would fix it for every unit; until then it is a manual step for a fresh worktree.

Note the first fix does **not** subsume the second: the JSON-state race survives it.

## Not the fix

Removing `nodejs`/`pnpm` from `devbox.json` and relying on OS-installed versions was considered and rejected. It would not address the remaining race, and it would drop the version pins that keep a devbox shell aligned with CI's `node 24` / `pnpm 11` — the two-halves drift #648 exists to prevent, in a lane `scripts/check-pin-lockstep.sh` does not currently cover.

vfarcic:
Closes #791's corepack half. The underlying devbox limitation stays open there — see **What this does not fix** below.

## The failure

Starting a multi-role orchestration in a **fresh** worktree kills one role at launch:

```
Error: error running script "agent-orchestrator" in Devbox: remove
  <worktree>/.devbox/virtenv/bin/setup-corepack.mjs: no such file or directory

Error: error running script "agent-orchestrator" in Devbox: symlink
  <worktree>/.devbox/virtenv/nodejs/bin/setup-corepack.mjs
  <worktree>/.devbox/virtenv/bin/setup-corepack.mjs: file exists
```

Two halves of one non-atomic operation — several agents entering the same cold environment race on creating and removing that single path. **It is silent unless you count roles**: the deck reports the orchestration as started, and the unit simply runs with five roles instead of six. An orchestration missing `agent-orchestrator` has workers and nobody to delegate to. Observed twice in one session, taking that same role both times.

## Why now

`git show daf94f0^:devbox.json` contains neither `nodejs` nor `pnpm`; `daf94f0` ("Add visual agent control deck", 2026-08-29) adds both. `setup-corepack.mjs` is, per its own header, "the nodejs plugin's init_hook". Before that commit there was no Node plugin and nothing to race on — which is why this has no history before 29 August, and why it only appeared once we began creating a fresh worktree per unit.

## Why removal rather than `disable_plugin`

`devbox add nodejs@24.12.0 --disable-plugin` was tried first and rejected on evidence. It rewrites the entire packages block from array to object form, changing pins from `"cargo-nextest@0.9.143"` to `"cargo-nextest": "0.9.143"` — which **breaks `scripts/check-pin-lockstep.sh` outright**. It then reports DRIFT for every Rust pin and exits 1, failing `cargo test-fast` (via `xtask/linkage-check/src/pin_lockstep.rs`, which runs it against the real checkout) and CI's `devbox` job. To its credit the guard refuses to pass vacuously and says so. The minimal hand-edit that would avoid the reformat — a single object inside the array — is rejected by devbox itself (`Value looks like object, but can't find closing '}' symbol`).

Removal keeps the array form, so the guard is untouched. The `devbox.json` diff is three lines.

## Why nothing is lost

- The corepack hook was **already a no-op here**: `DEVBOX_COREPACK_ENABLED` is set nowhere, `desktop/package.json` pins no `packageManager`, and nothing outside `.devbox` referenced corepack. Its only effect was a symlink to a script that returns immediately.
- **No devbox script used `node` or `pnpm`.** Every agent CLI those scripts launch — `claude`, `opencode`, `pi`, `codex` — is already installed outside devbox. Node and pnpm now sit in that same category rather than in one of their own.
- **CI never used devbox's Node**, taking it from `actions/setup-node` and `pnpm/action-setup`.
- Devbox is optional, so Node and pnpm had to be OS-available for non-devbox contributors regardless. This removes a divergence rather than creating one.

## The one real cost, and how it is covered

`xtask/linkage-check/src/issue_labeler_memory.rs` shells out to `node` and prints `SKIP: issue-labeler memory validator tests need 'node' on PATH` when it is absent — a silent-skip hazard this repo keeps meeting (#656; the `SKIPPED`-counts-as-passing trap in CLAUDE.md rule 8). So Node and pnpm are now documented as prerequisites in `CONTRIBUTING.md` and `docs/develop/desktop-gui.md`, naming the versions CI pins. Note `devbox.json`'s `init_hook` already prepends `$HOME/.local/bin`, so a user-level install is visible **inside** `devbox shell` as well as outside — no separate setup for devbox users.

`docs/develop/desktop-gui.md` also had to change because it explicitly promised devbox supplies "Node.js 24.12.0, and pnpm 10.34.5". Unrelated staleness I did **not** fix while here: that same line says CI's job runs Node 20 (it pins 24), and describes pnpm 10.x as the line that reads `lockfileVersion: '9.0'` (devbox had moved to 11.22.0 before this PR).

## Verification

- Six concurrent cold `devbox run` in a fresh worktree: **0/6 failures** after this change. Before, the same shape produced the errors above.
- `scripts/check-pin-lockstep.sh`: **passes**.
- `node` and `pnpm` resolve **inside** `devbox shell` from `$HOME/.local/bin`.
- The corepack machinery is gone — `.devbox/virtenv` now holds only `runx` and `rustc`.
- `cargo fmt --check` clean; clippy clean; `cargo test-fast` **3437 passed, 0 skipped**, up from 3426, because the ten `issue_labeler_memory` tests now RUN instead of emitting their `SKIP:` line.

Both cargo gates ran with `--exclude dot-agent-deck-desktop`: `dbus-1`/GTK are absent on this machine, which is **issue #771** — a sibling of this one from the same `daf94f0` — and not caused by this change.

## What this does not fix

The root limitation: devbox's per-project materialisation is not safe against concurrent invocations. With the Node plugin already out of the picture, concurrent cold runs can still fail with `compact json for hashing: unexpected end of JSON input` — devbox reading a partially-written state file. That is rarer (1/6 in one test, versus corepack hitting repeatedly) and is not Node-related. **#791 stays open for it**, with `devbox run -- true` in a fresh worktree as the mitigation.

Rule 12 does not apply: no daemon, protocol, orchestration or hook change.


vfarcic:
Closing in favour of #780's approach.

#780 keeps the devbox environment self-sufficient — a `tauri-deps` flake supplying the GTK/WebKit closure — and its rationale is that `devbox shell` should need nothing installed by hand. This PR argued the opposite for Node and pnpm: drop them and make them an OS prerequisite. Both cannot be the project's stance, and #780's is the stronger one now that it is verified working: with it, `cargo clippy --workspace --all-targets --features e2e` passes with no `--exclude`, and `cargo test-fast` runs 3488 tests including the desktop crate's, which had never executed in a devbox shell.

The two also collided directly on `devbox.json`, `devbox.lock`, `CONTRIBUTING.md` and `docs/develop/desktop-gui.md`.

**What was learned here is worth keeping, and is recorded in #791** rather than lost with this branch:

- The corepack race is a real, dated regression: `daf94f0^` had neither `nodejs` nor `pnpm`, so before 2026-08-29 there was no Node plugin and nothing to race on.
- `devbox add --disable-plugin` is not a viable minimal fix: it rewrites the packages array into object form, which breaks `scripts/check-pin-lockstep.sh` outright (it reads the toolchain pins as `"rustc@1.97.1"` string entries) and thereby fails `cargo test-fast` and CI's `devbox` job. #780 hit and documented the same constraint independently.
- The underlying limitation is broader than corepack: with the Node plugin out of the picture, concurrent cold `devbox run` still fails with `compact json for hashing: unexpected end of JSON input`. So devbox's per-project materialisation is not concurrency-safe in general, and warming a fresh worktree once with `devbox run -- true` remains the mitigation.

#791 stays open for that residual.

Issue #818, opened 2026-09-02
e2e-deterministic has been red on every branch and on main since #789 merged


vfarcic:
## Symptom

`e2e-deterministic` (lane 1, added by #800) fails on **every** branch and on **`main` itself**, so every open PR carries a red CI job that has nothing to do with its own change.

## Bisect: it starts exactly at `d47bc75` (#789)

The job was green five consecutive times, then went red five seconds after #789 merged:

```
2026-09-01T01:26:27Z  edaa2b4  #800 — adds the e2e-deterministic job

2026-09-01T20:34:43Z  success  agent/e2e-live-credential-sinks
2026-09-01T21:20:27Z  success  agent/e2e-live-credential-sinks
2026-09-01T21:39:17Z  success  agent/e2e-live-credential-sinks
2026-09-01T22:13:47Z  success  agent/e2e-live-credential-sinks
2026-09-01T22:34:58Z  success  agent/e2e-live-credential-sinks

2026-09-01T23:27:42Z  d47bc75  #789 — orchestrator remit on compaction/clear  ← merged
2026-09-01T23:27:47Z  failure  main                                           ← 5s later
2026-09-01T23:40:39Z  failure  docs/verify-absolutes-before-asserting
2026-09-01T23:47:38Z  failure  fix/813-renovate-minimum-release-age
2026-09-01T23:59:17Z  failure  agent/dispatch-prd-745
2026-09-02T00:13:32Z  failure  agent/dispatch-issue-815
2026-09-02T00:13:54Z  failure  renovate/github-actions
2026-09-02T00:31:44Z  failure  main
```

Worth stating because CLAUDE.md rule 8 invites the opposite reading: this is **not** lane 1 being inherently unstable on a runner. By the time #789 landed, lane 1 had five honest green runs. The instability was introduced.

### 1. `orchestration_remit_001` / `_007` — a timeout too tight for a runner

```
panicked at tests/e2e_orchestration_remit.rs:244:5:
the daemon's own ListAgents/live-status join never reported Compacting for pane 1
(agent_id 1) within 10s — the hook socket write was accepted, but
AppState::apply_event may have rejected it or applied it to the wrong session.
```

The wait is a correct, Decision 21-compliant harness poll — the bound is simply too small:

```rust
// tests/e2e_orchestration_remit.rs:237 and :316
let applied = common::wait_until(Duration::from_secs(10), || { … });
```

[...]

PR #839, opened 2026-09-02
fix(xtask): neutralise git location discovery in the fixture git sandbox

### LINKED ISSUE #834: xtask/linkage-check's Sandbox::git() neutralizes git config discovery but not location discovery (GIT_DIR etc.) — read/write escape from the fixture
Found while auditing a fork of this project (fork issue: https://github.com/prageethw/dot-agent-deck/issues/579). Confirmed present unchanged on this repo's current `main`.

## What's wrong

`xtask/linkage-check/src/repo_state.rs`'s `Sandbox::git()` (the shared fixture-git-invocation test helper) neutralizes git's **config** discovery:

```rust
.env("GIT_CONFIG_GLOBAL", self.at("no-such-gitconfig"))
.env("GIT_CONFIG_SYSTEM", self.at("no-such-gitconfig"))
```

but not its **location** discovery: `GIT_DIR`, `GIT_WORK_TREE`, `GIT_COMMON_DIR`, `GIT_INDEX_FILE`, `GIT_OBJECT_DIRECTORY`, `GIT_ALTERNATE_OBJECT_DIRECTORIES`, `GIT_CEILING_DIRECTORIES`, `GIT_NAMESPACE` are all left ambient.

The module's own doc comment (a few lines above `Sandbox::git()`) claims tests using this helper can't read or write the checkout they're running inside. That claim is inaccurate given the above.

## Evidence

- **Read escape**: with ambient `GIT_DIR` set (pointing at a real repo's `.git`), `git show origin/main:tests/CATALOG.md` inside the sandboxed helper returns the *real* repo's catalog, not the fixture's — `Sandbox::git()`'s own `cwd` argument is ignored by git once `GIT_DIR` is set.
- **Write escape**: against a disposable stand-in repo with ambient `GIT_DIR` pointed at it, the fixture's `init`/`add -A`/`commit` sequence moved that repo's HEAD to a new fixture commit, replacing its tree — i.e. the sandbox helper wrote into a repository it did not create and was not meant to touch.
- Reachable in practice from a pre-commit hook, `git rebase --exec`, or `git bisect run` — any context that legitimately sets these vars in the parent process before spawning the test binary that then calls into this helper.

## Fix

Add the missing env vars to `Sandbox::git()`'s `Command` builder alongside the existing `GIT_CONFIG_GLOBAL`/`GIT_CONFIG_SYSTEM` treatment — `env_remove` (if git treats unset the same as "not location-overridden") or an explicit sandbox-local/non-existent value, whichever git honors correctly for each var:

```rust
.env_remove("GIT_DIR")
.env_remove("GIT_WORK_TREE")
.env_remove("GIT_COMMON_DIR")
.env_remove("GIT_INDEX_FILE")
.env_remove("GIT_OBJECT_DIRECTORY")
.env_remove("GIT_ALTERNATE_OBJECT_DIRECTORIES")
.env_remove("GIT_CEILING_DIRECTORIES")
.env_remove("GIT_NAMESPACE")
```
(worth verifying each var's actual git-honored behavior under `env_remove` vs. an explicit override before landing — some of these may need a concrete non-existent path rather than plain removal to fully pin the invoked `cwd`.)

## Severity note

Same-uid, requires a specific ambient-env precondition set in the parent process before the test binary spawns (not the default state of an ordinary local or CI test run) — not an urgent/exploitable-by-default issue, but the doc comment's safety claim is currently false and the demonstrated write escape is a real (if narrow) data-loss risk for anyone running this test suite with one of those vars set for an unrelated reason (e.g. mid-rebase, mid-bisect).

vfarcic:
Fixes #834.

## What was ambient

`xtask/linkage-check`'s `repo_state::Sandbox::git()` switched off git's **config** discovery but left its **location** discovery inherited from the parent process. Location discovery outranks the `current_dir` each fixture command passes, so both escapes reproduce:

- **Read** — with `GIT_DIR` aimed at another repository, `git -C <fixture> log` reports *that* repository's history.
- **Write** — the fixture's own `init`/`commit` sequence moves that repository's HEAD to a fixture commit.

That falsified a guarantee written down in two places: this module's doc comment, and CLAUDE.md rule 5, which justifies these tests living in the fast tier on the grounds that "none of them can read or write the checkout it is running in". Both now say what the code does.

## What is now neutralised

- **Eight discovery variables cleared** — `GIT_DIR`, `GIT_WORK_TREE`, `GIT_COMMON_DIR`, `GIT_INDEX_FILE`, `GIT_OBJECT_DIRECTORY`, `GIT_ALTERNATE_OBJECT_DIRECTORIES`, `GIT_NAMESPACE`, and `GIT_DISCOVERY_ACROSS_FILESYSTEM` (not in the issue's list; it lets the upward walk cross a mount point, which is one of only two things bounding it). Cleared rather than overridden because unset *is* git's default for each — verified.
- **`GIT_CEILING_DIRECTORIES` is set, not merely cleared.** Clearing restores git's default *unbounded* walk, which would leave the claim true only contingently — it would hold because no current fixture command does upward discovery from a non-repository directory (`git init` and `git clone` do none), and would break silently the day one does or the day `TMPDIR` points inside a checkout. Bounding it makes the claim structural. It is set to the sandbox root's **parent**, not the root: the variable names the directories git must not `chdir` up *into* and never excludes the invocation's own cwd, so a root-listed ceiling escapes exactly as an absent one does (measured, and pinned by injection — see below).
- **`GIT_DEFAULT_HASH`/`GIT_DEFAULT_REF_FORMAT` cleared** — one extra line for the *same* doc comment's other claim, that fixtures are byte-identical on a developer's machine and on a runner. Ambient `GIT_DEFAULT_HASH=sha256` builds SHA-256 fixtures against SHA-1 on a runner (measured; likewise `GIT_DEFAULT_REF_FORMAT=reftable`). Flagging it explicitly as slightly beyond the issue — it was the same defect, in the same helper, falsifying the same doc comment.

## How the regression tests prove it

The failure mode is silent by construction — nothing errors when a fixture escapes — so both tests were held to this module's stated bar and **observed failing** against four one-at-a-time injections:

| injection | goes red |
|---|---|
| drop the `AMBIENT_LOCATION_VARS` loop | read half: `show-toplevel` reports the ambient `GIT_DIR`'s repository |
| that, plus reduce the child to building its fixture | write half: the victim's HEAD moves to a fixture commit |
| clear `GIT_CEILING_DIRECTORIES` instead of setting it | discovery escapes to the outer repository |
| set the ceiling to the root instead of the root's parent | escapes identically — this is what pins the parent |

`sandbox_git_ignores_ambient_location_env` **re-execs the test binary** with all nine variables aimed at a victim repository. The re-exec is the point: the variables must be in the environment *before* the process starts to reproduce the real condition (mid-`rebase --exec`, a pre-commit hook, `bisect run`), and setting them in-process would be `unsafe` and would race every other test in the process. The child asserts it operated on its own fixture; the parent asserts the victim's HEAD and commit count held. Two fail-green guards: the child asserts the variables really are set (otherwise every assertion passes proving nothing), and the parent asserts `1 passed` because a zero-match libtest filter also exits 0.

`sandbox_git_cannot_discover_a_repository_above_its_root` places a sandbox inside another repository's working tree — what you get when `TMPDIR` points under a checkout — and carries a **control** proving the escape is actually stageable, so it cannot pass by the scenario being impossible on an ordinary machine. It also asserts discovery *inside* the sandbox still resolves normally.

## Scope and gates

Test-harness only: every changed line of Rust is inside `#[cfg(all(test, unix))] mod real_git`, so no shipped behaviour can change. Nothing under the root package's `src/` was touched.

- `cargo test-fast` — 3974/3976. Two failures, both in the root package's PTY/timing tests, both passing when rerun alone, and different tests on different runs (`spawn_time_role_prompt_submits_after_input_readiness_buffer` on one run; `delegate_012`/`delegate_022` on the next). This branch changes only the xtask crate's test module, CLAUDE.md and a changelog fragment, so those binaries are identical to `main`'s — the documented parallel-contention flake class (rule 5, #415/#351/#701), not this change. Note `e2e-deterministic` is red on `main` for unrelated reasons (#818/#832, fix in PR #837).
- Full `xtask-linkage-check` suite: 216/216. The two new tests run in 0.030s and 0.038s; all 16 `real_git` tests together stay under 0.2s, so rule 5's fast-tier budget is unchanged.
- **Rule 12 does not apply** — no daemon, protocol, orchestration or hook change, so no `PROTOCOL_VERSION` bump and no cross-version manual test.
- CLAUDE.md rule 5's paragraph is updated: the count 14 → 16, and its "ambient git configuration switched off" claim now names the location half that makes its "none of them can read or write the checkout it is running in" conclusion true.


prageethw (approved):
Verified: the fixture sandbox now clears all 8 git location-discovery variables (GIT_DIR and friends) plus GIT_DEFAULT_HASH/GIT_DEFAULT_REF_FORMAT, and sets GIT_CEILING_DIRECTORIES to the sandbox root's *parent* (correct — ceiling never excludes the invocation's own cwd, so listing the root itself would not bound the walk). The two new regression tests are held to the module's own mutation-style bar (four one-at-a-time injections, each observed failing) and change is confined to #[cfg(test)] xtask code plus CLAUDE.md/changelog — no shipped behavior affected. The red e2e-deterministic check is documented as a pre-existing main flake (#818/#832) fixed separately in PR #837, not caused by this change. LGTM.

Issue #405, opened 2026-08-07
Local daemon attach ignores PROTOCOL_VERSION: a newer daemon silently breaks an older TUI's event stream


prageethw:
`PROTOCOL_VERSION` is not enforced on the **local (same-machine)** daemon↔TUI attach path. It is wired only into the SSH / `remote add` flow. So an older TUI can attach to a newer daemon, silently fail to decode events it does not understand, and sit in an **invisible reconnect loop while the dashboard looks perfectly normal**.

Found by running the CLAUDE.md rule 12 cross-version manual test for a `PROTOCOL_VERSION` bump.

## The gap

Two unconnected mechanisms guard the two attach paths:

| Path | Guard | Checks `PROTOCOL_VERSION`? |
|---|---|---|
| SSH / `remote add` | `connect::probe_remote_protocol` | **Yes** — refuses cleanly with `ProtocolMismatch` |
| Local same-machine daemon | `build_version_handshake::ensure_compatible_daemon_or_die` | **No** — compares `DAD_BUILD_ID` only |

Neither `daemon_attach.rs` nor `build_version_handshake.rs` references `PROTOCOL_VERSION` or `probe_remote_protocol` anywhere. `grep -n "server_version" src/build_version_handshake.rs` returns **zero hits** — the field is never read on any branch (`Match`, `Recovered`, or `ProceedOnExisting`).

When the build-version handshake detects a mismatch it offers to restart the daemon. **Declining** — which is the correct choice when you do not want to kill live agents — returns `ProceedOnExisting` and attaches anyway, with no further version check of any kind.

The doc comment on `PROTOCOL_VERSION` states the bump "forces `probe_remote_protocol` to refuse the old-reader/new-daemon pairing with a clean `ProtocolMismatch`." True as written, but it does not describe what happens on an ordinary **local** upgrade — which is what rule 12's procedure actually exercises.

## Reproduction (verified empirically)

Built two versions with different protocol numbers. `daemon hello` confirms the split:

```
{"server_version":6,"build_version":"0.35.5-g525cc4c", ...}   # old
{"server_version":7,"build_version":"0.35.5-g4adc7ea", ...}   # new
```

New daemon (protocol 7) + old TUI (protocol 6):

1. Launch the old TUI against the new daemon → build-version mismatch prompt appears.
2. Decline the restart (any key but `s`) → **attaches anyway**, dashboard renders normally.
3. Trigger activity that produces an event variant the old client does not know.
[...]

Issue #815, opened 2026-09-01
#780's PKG_CONFIG_PATH is empty under Available scripts:
* agent
* agent-auditor
* agent-auditor-oc
* agent-big
* agent-coder
* agent-coder-oc
* agent-orchestrator
* agent-orchestrator-oc
* agent-plan
* agent-release
* agent-reviewer
* agent-reviewer-oc
* agent-small
* agent-tester
* agent-tester-oc
* codex-big
* oc-big
* oc-kimi
* oc-release
* pi-big
* pi-sol, so pkg-config resolves GTK against the system and cargo test-fast still dies on libgdk-3.so.0


vfarcic:
## Symptom

After #780 merged, `cargo test-fast` inside devbox still fails before running a single test:

```
error: creating test list failed
  target/debug/deps/dot_agent_deck_desktop-<hash>: error while loading shared libraries:
  libgdk-3.so.0: cannot open shared object file: No such file or directory
```

That is the runtime half of #771, which #780 was supposed to close. Measured on 2026-09-01 on `main` at `edaa2b4`, with `tauri-deps/flake.nix` present and both `pkg-config@0.29.2` and `path:tauri-deps#tauri-deps` in `devbox.json`.

Cleaning the crate does not help — `cargo clean -p dot-agent-deck-desktop` removed 1357 files / 2.9 GiB, and the rebuilt binary fails identically.

## Mechanism, traced

`devbox.json` carries the env block #780 describes:

```json
"env": { "PKG_CONFIG_PATH": "$PWD/.devbox/nix/profile/default/lib/pkgconfig" }
```

But inside `devbox run --`, **that variable is empty**:

```
$ devbox run -- sh -c 'echo "PKG_CONFIG_PATH=${PKG_CONFIG_PATH}"'
PKG_CONFIG_PATH=
```

So `pkg-config` falls back to its built-in search path and **silently resolves against the system**:

```
$ devbox run -- pkg-config --variable=libdir gtk+-3.0
/usr/lib/x86_64-linux-gnu
```

`pkg-config --exists gtk+-3.0` therefore returns **success**, the compile succeeds, and the whole thing looks fixed — right up until a test binary tries to load. The linked artifact has **no gtk3 path in its RUNPATH**:

```
$ readelf -d target/debug/deps/dot_agent_deck_desktop-<hash> | grep -oE '/nix/store/[a-z0-9]+-gtk[^:]*'
(no output)
```

[...]

PR #840, opened 2026-09-02
fix(connect): drop the vacuous laptop<->remote PROTOCOL_VERSION check (#491)

### LINKED ISSUE #491: connect: drop the vacuous laptop<->remote PROTOCOL_VERSION check — the laptop is only ssh + a terminal
`connect` hard-fails on a laptop↔remote `PROTOCOL_VERSION` difference, but those two constants never meet on a wire. The laptop is only `ssh` + a terminal — the TUI↔daemon attach conversation happens entirely on the remote box, between two invocations of the *same* remote binary.

The result is a spurious fatal error that blocks connecting to a remote which is internally consistent and fully working, and whose recommended recovery (`remote upgrade`) can cost the user their live remote agents.

Companion to #405, which covers the opposite defect on the local path. Together they would leave `PROTOCOL_VERSION` enforced exactly where skew is possible, and nowhere else.

## Why the comparison is vacuous

`build_connect_command` (`src/connect.rs:614`) runs:

```
ssh -t -- <user@host> env DOT_AGENT_DECK_VIA_DAEMON=1 ~/.local/bin/dot-agent-deck
```

That execs the **remote** binary's TUI on the remote's pty. Every attach-socket client in the tree resolves its path through `attach_socket_path()`, which is local-only — there is no laptop-side client of a remote daemon socket. The module doc says so directly (`src/connect.rs:19`):

> No socket relay or laptop-side daemon ever runs in this flow — that's M2.7's deliberate simplification.

So on the remote machine, TUI and daemon are the same install and agree on `PROTOCOL_VERSION` by construction. The laptop's compiled-in constant is never a party to that conversation.

## This is a leftover from the deleted socket bridge

Before the 2026-05-09 pivot the laptop *did* attach to the remote daemon's socket over a bridge, and the comparison was correct. M2.7 deleted the bridge; the check survived it.

PRD #161 M1.2 already applied this exact reasoning to the sibling check in the same function and removed it (`src/connect.rs:474`):

> That laptop↔remote build-id comparison guarded nothing (**the laptop is only ssh + a terminal**), so a matched-protocol remote is accepted regardless of build id.

It then kept the `PROTOCOL_VERSION` arm as "the structural floor" — which the same sentence disproves. The build-id half and the protocol half are vacuous for identical reasons; only one was removed.

## User-visible cost

Remote on protocol v5, laptop on v6, `connect` dies at `src/connect.rs:554-565`:

```
Remote 'box' speaks attach protocol v5; laptop speaks v6. Run `dot-agent-deck remote upgrade box`
```

The remote is not broken. Its TUI and daemon are both v5 and interoperate perfectly. Nothing about the ssh session requires the laptop to understand the attach wire.

The recovery is worse than the disease: `remote upgrade` swaps the remote binary, after which the remote's *own* build-id handshake (`ensure_compatible_daemon_or_die`, running on the remote) sees a stale daemon and prompts to restart it — putting the user's live remote agents at risk to satisfy a check that was measuring nothing. The reverse direction (`local < remote` → "Upgrade your laptop binary") is spurious in the same way.

The soft version nudge that *should* own this axis already exists three stages later: `maybe_nudge_upgrade` (`src/connect.rs:825`) is newer-only, TTY-only, defaults to N, and names the running-agent count as the restart cost. That's the right UX for "your remote is older" — and it is currently unreachable whenever the protocol digits happen to differ, because the fatal fires first.

## Suggested fix

Delete only the version-comparison arms of `probe_remote_protocol` (`src/connect.rs:488`). **Keep the rest of the probe** — the non-version arms are a genuine connect floor and are not affected by any of the above:

| Arm | Keep? | Why |
|---|---|---|
| ssh transport failure | keep | reachability floor, feeds the reconnect/backoff path |
| non-zero exit (pre-M2.21 binary has no `daemon hello`; broken install) | keep | "remote binary is too old or broken", real signal |
| stdout doesn't parse as `AttachResponse` | keep | stub script / corrupted install |
| `ok: false` → `RemoteHandshakeError` | keep | remote daemon signalling a runtime error |
| `Some(v) if v != PROTOCOL_VERSION` → `ProtocolMismatch` | **drop** | compares two constants that never share a wire |
| `None` (reply lacks `server_version`) → `ProtocolMismatch` | **drop as a protocol verdict** | a pre-M2.21 reply is an *age* signal, not an incompatibility; fold into the missing/too-old arm above if it should stay fatal |

After that, the function keeps returning `running_agents` for the nudge, and `RemoteConnectError::ProtocolMismatch` (`src/connect.rs:162`) either loses its version fields or is replaced by the too-old/broken variant, since no live caller would still be comparing versions.

Worth doing in the same pass: the doc comments that justify the check (`src/connect.rs:152-159`, `src/connect.rs:463-480`) and the `PROTOCOL_VERSION` doc block in `src/daemon_protocol.rs:144-195` all describe `probe_remote_protocol` as the enforcement point. `src/daemon_protocol.rs:170-174` in particular presents it as the guard that turns an old-reader/new-daemon pairing into a clean connect-time refusal — which, given #405, means the constant's own documentation currently points at the one path where it cannot help. Those blocks should end up pointing at whatever closes #405 instead.

## Test impact

The existing fake-executor tests in `src/connect.rs` (~lines 1432, 1487, 1636) all return a *matching* protocol from `daemon hello`, so they exercise the success arm and should be unaffected. Any test that asserts `ProtocolMismatch` on a numeric difference is asserting the behaviour being removed and should be retired with it. Add coverage for the case this unblocks: a remote whose `daemon hello` reports a different `server_version` connects normally, and — when the laptop is the newer of the two by `--version` — reaches `maybe_nudge_upgrade` rather than dying.

## Ordering note

Independent of #405 — this is `connect.rs`, that is `build_version_handshake.rs` / `daemon_attach.rs`, no shared code path. Either can land first.

In particular there is **no** "don't leave `PROTOCOL_VERSION` enforced nowhere" argument for sequencing them, which an earlier draft of this issue claimed. That reasoning contradicts the premise above: the check being removed here guards nothing today, so removing it subtracts no coverage. `PROTOCOL_VERSION` is already unenforced against every skew that can actually occur, and remains so until #405 lands either way.

The two fix different things on different timescales: #405 is what makes the *next* protocol bump safe; this issue is what stops a working remote from being refused right now.

vfarcic:
Closes #491.

## The evidence that the check was vacuous

`connect` hard-failed when the laptop binary's compiled-in `PROTOCOL_VERSION` differed from the remote binary's. **Those two constants never share a wire**, so the comparison could not fail for a real reason. This is a deletion, so I verified the premise rather than trusting the issue or the module doc:

1. **`connect` runs the *remote* binary's TUI, not a laptop client.** `build_connect_command` (`src/connect.rs:757`) emits exactly `ssh -t -o ConnectTimeout=N -o ServerAlive* -p PORT [-i KEY] -- user@host env DOT_AGENT_DECK_VIA_DAEMON=1 <install_path>`. **No forwarding of any kind** — `grep` over `src/` for `-L` / `-R` / `StreamLocalBind` / `LocalForward` returns two hits, both irrelevant (`src/event.rs:138` is an unrelated argv list; `src/remote.rs:444` is a doc comment about the *user's* ssh-config forwards, i.e. reverse SOCKS for the remote's agents).

2. **No laptop-side attach client can even address a remote socket.** Every attach client resolves `attach_socket_path()` (`src/platform/paths.rs:1269`), which **takes no arguments** and resolves only `DOT_AGENT_DECK_ATTACH_SOCKET`, `$XDG_RUNTIME_DIR/…`, or `/tmp/…-{uid}.sock` — all local. I enumerated every non-test `DaemonClient::new(` call site (`main.rs:1679/1946/2105/2124/2238`, `build_version_handshake.rs:193/566`, `embedded_pane.rs:502`); all local, and no host-parameterised variant exists anywhere.

3. **The module doc agrees** (`src/connect.rs:13-21`): step 3 "runs the deck TUI on the remote with stdio inherited", "the laptop process blocks until ssh exits", "No socket relay or laptop-side daemon ever runs in this flow — that's M2.7's deliberate simplification."

So both ends of the attach conversation are the **same install on the same machine** and agree on `PROTOCOL_VERSION` by construction. The laptop's constant was never a party to it.

PRD #161 M1.2 had already deleted the *sibling* build-id comparison on this exact reasoning — "That laptop↔remote build-id comparison guarded nothing (**the laptop is only ssh + a terminal**)" — while keeping the protocol half as "the structural floor". This is that same sentence applied to the half it spared.

## The three questions a reviewer will ask

**Was the check load-bearing for anything?** No. It compared two constants across two machines, but the protocol it purported to guard is a conversation that happens entirely *on the remote*. The pairing that genuinely *can* skew — the remote's TUI meeting the remote's still-running daemon after an in-place upgrade — is guarded **on the remote** by `build_version_handshake`'s `DAD_BUILD_ID` comparison, which is strictly stronger than protocol equality when it matches (same build ⇒ same protocol). Its residual hole is #405's "decline the restart prompt → attach anyway", which this PR does not touch.

**Does the `--version` probe (step 2) rely on it?** No. `probe_remote_version` parses `dot-agent-deck <semver>` and returns a `String`; its error arms are `RemoteBinaryMissing` / `HostUnreachable` / `ProbeOutputTruncated`, and it never reads `PROTOCOL_VERSION`. It runs *before* the protocol probe. Removal actually **unblocks its consumer**: `maybe_nudge_upgrade` was unreachable whenever the digits differed.

**Does removal weaken a message that tells the user something true?** No. The old message — `Remote 'box' speaks attach protocol v5; laptop speaks v6. Run 'dot-agent-deck remote upgrade box'` — stated two true constants but a **false implication** (that they block attaching) and a **harmful remedy**: `remote upgrade` swaps the remote binary under the user's live agents, after which the remote's *own* build-id handshake prompts to restart the daemon serving them. What replaces it on that axis is strictly better: `maybe_nudge_upgrade` is newer-only, TTY-only, defaults to No, and names the running-agent count as the restart cost.

## What is deliberately KEPT

The probe still refuses a remote that cannot answer `daemon hello` **at all** — non-zero exit (pre-M2.21 binary, broken install), unparseable stdout (stub script), or a reply carrying no `server_version`. That is an *install floor*, not a version verdict: "something ran there, but nothing we can identify as a deck", and `connect` is about to hand over the user's terminal. `ProtocolMismatch` becomes `RemoteHandshakeUnsupported`, carrying **no version numbers**, because quoting them described a comparison that measured nothing.

Kept the `None`-`server_version` arm fatal on purpose (the issue left this open): `AttachResponse::hello` always sets the field, so its absence is a stub or a hand-rolled reply, not an older deck disagreeing about a number.

## `remote doctor`

Its `ProtocolCompatible` check already carried a comment anticipating this issue, and its FAIL verdict was justified *only because* `connect` refused. Left unchanged it would have reported FAIL for something that no longer blocks anything — the inverse of the misdiagnosis that command exists to prevent. It now reports what `connect` actually enforces (did the remote answer), and **FAIL stays FAIL** because an unanswerable handshake is genuinely unattachable. `CheckId::ProtocolCompatible` **keeps its name**: it is part of the report's scriptable output contract (pinned by `tests/e2e_remote_doctor.rs`), it still reads true, and renaming it would break anyone grepping the report for zero diagnostic gain. Only prose moved (`docs/remote-recipes.md`'s sample output updated to match).

**Not a TUI↔daemon contract change**, and here is the reasoning rather than an assertion:

- `PROTOCOL_VERSION` stays **8**. No request or response variant, field, `KIND_*` code, or handler semantic moves.
- The change to `src/daemon_protocol.rs` is **doc-comments only** — verified mechanically: `git diff -U0` filtered to non-comment lines returns **zero** changed lines, and the `pub const PROTOCOL_VERSION: u32 = 8;` line appears in the diff only as unchanged context.
- The deleted code is a **laptop-side ssh pre-flight that never spoke the wire**. The daemon's own handshake is untouched, on both the local and remote sides.

So: no `PROTOCOL_VERSION` bump, no `.breaking.md` fragment, and the previous-release-daemon manual test would exercise nothing this PR changed (it drives the *local* attach path; every byte of that path is untouched).

**Tests asserting the removed behaviour:** one, and it was the doctor's *classifier* unit test driven by an input field rather than by a real comparison — updated to match the new meaning. No e2e test asserted a numeric mismatch: `tests/e2e_remote_doctor.rs` always set `SSH_STUB_PROTOCOL` to the matching `PROTOCOL_VERSION`.

## Doc corrections (rule 17)

The `PROTOCOL_VERSION` doc block named `probe_remote_protocol` as its enforcement point in three separate bump rationales, and `docs/develop/versioning.md` called it "the structural **fatal floor**" whose "attach handshake refuses across a mismatch … mandatory and non-negotiable". Those became false, so they are narrowed to the verified claim: **no call site refuses on a `PROTOCOL_VERSION` difference today.** Enumerated to check that absolute — every remaining reference in `src/` is a doc comment, a *sender* (`client_version: PROTOCOL_VERSION`, `AttachResponse::hello(...)`), or a test fixture; the daemon's handler destructures `client_version: _`; and `server_version` appears **zero** times in `build_version_handshake.rs`, re-confirming #405's claim on this tree. The bump discipline stays mandatory — the bump is what makes a skew *nameable*, and #405 is what will make it *refused*.

`CHANGELOG.md` and `prds/` still mention `ProtocolMismatch` and are left alone on purpose: both are point-in-time historical records (as `prds/103` already is, still describing the build-id comparison #161 deleted).

## Changelog fragment

`changelog.d/491.bugfix.md` — **not `.misc`**. The check could and did fire: `PROTOCOL_VERSION` has gone 3→4→5→6→7→8, so any laptop/remote pair straddling one of those bumps hit a fatal error that blocked connecting — an ordinary situation, since you upgrade your laptop and the remote box stays on the older release. It is user-visible behaviour with a user-visible message, so it is a bugfix (patch bump; not `breaking`, per rule 12 above).

## Gates

| gate | result |
|---|---|
| `cargo test-fast` | **3977 passed, 0 failed** (no `libgdk` failure in this worktree, so #815 did not bite) |
| `cargo xtask linkage-check` | ok (618 catalog ids, 546 annotations, 11 rules) |

**Scoped tests, named** (rule 5 — "ran the related tests" is not a claim anyone can check):

- `connect::tests::*` — 27 passed, including the three new ones: `differing_remote_protocol_connects_and_reaches_the_nudge`, `newer_remote_protocol_connects_without_touching_the_laptop`, `unanswerable_handshake_stays_fatal_without_naming_versions`.
- `remote_doctor::tests::*` — 35 passed, including the renamed `classify_unanswered_handshake_is_fatal_not_advisory`.
- `cargo test-e2e remote_doctor_0` — **12 passed** (`remote/doctor/001`–`012`), lane 1, no credentials.

New L2 coverage: **`remote/doctor/012`** drives the real binary against a stub remote reporting a protocol version one *below* and one *above* this build's own — both skew directions, since the removed check had a distinct arm and a distinct (equally spurious) remedy for each — and asserts `ProtocolCompatible` PASS, exit 0, and that no laptop-side version is quoted at the user. It is a genuine regression guard: under the old code `Some(v) if v == PROTOCOL_VERSION` fails the guard, so the probe returned `Err(ProtocolMismatch)`, the doctor reported FAIL, and the run exited 1.


prageethw (approved):
Verified the vacuousness claim directly: `attach_socket_path()` (platform/paths.rs) takes no host argument and every `DaemonClient::new(` call site is local-only, and `build_connect_command` requests no port/socket forwarding — so the laptop's `PROTOCOL_VERSION` genuinely never joins the attach conversation the check was comparing. The real floor ("remote can't answer `daemon hello` at all") is kept as `RemoteHandshakeUnsupported`, and the replacement upgrade path (`maybe_nudge_upgrade`) is pre-existing, not new-and-unproven. Tests cover both skew directions plus the floor (connect::tests + remote/doctor/012 e2e, both directions). Rule-12 no-bump reasoning checks out: `git diff -U0` on daemon_protocol.rs is comment-only. The one CI red (e2e-deterministic: `orchestration_remit_001_start_role_compaction_reasserts_remit`) is an unrelated pre-existing flake — the exact race PR #837 fixes — not caused by this diff.

Issue #747, opened 2026-08-29
PTY resize: the client parser is not bounded by PTY_RESIZE_DIM_MAX, so above ~4096 columns the pane renders at a different width than the child uses


vfarcic:
Found by the **PRD #313 audit** (zoom the focused agent pane). It is a **pre-existing** bug in the shared resize path, not something zoom introduced — zoom only lowers the terminal width at which it becomes reachable. Deliberately **not** fixed in PRD #313, because changing the clamp changes it for every layout and that is the scope creep that PRD names as its top risk.

## The mismatch

`PTY_RESIZE_DIM_MAX` (`src/agent_pty.rs:163`, `= 4096`) bounds the **daemon-side child PTY** and nothing else. `AgentPtyRegistry::resize` (`src/agent_pty.rs:6384-6385`) silently clamps with `rows.min(PTY_RESIZE_DIM_MAX)` / `cols.min(PTY_RESIZE_DIM_MAX)` and reports success. The **client** applies no such bound: `EmbeddedPaneController::resize_pane_pty` (`src/embedded_pane.rs:~837-850`) pushes the requested dimensions onto the per-pane coalescing channel *and* applies them locally with `vt100::Screen::set_size(rows, cols)`.

So above 4096 in either axis the local vt100 parser is parsing and rendering the agent's output at one width while the child process has been told a different, smaller one. The child wraps its output at 4096 columns; the parser lays it out for 4097+. The result is misaligned/rewrapped content in the pane, and it is silent — nothing logs, nothing errors, and `resize` returns `Ok`.

## The two thresholds

Both are a function of terminal width, because the inner width of a pane is its rect minus the two border columns.

- **Zoomed** (PRD #313): the pane column is 100% of the frame, so inner width `= terminal_width - 2`. The mismatch begins at a terminal **4099 columns** wide (inner 4097).
- **Unzoomed**, default 34/66 orchestration split: inner width `≈ floor(terminal_width × 0.66) - 2`, so the mismatch begins at roughly **6.2k columns** (the audit measured ~6209).

Neither is a width you hit on a laptop. Both are reachable on a very wide multi-monitor terminal or a tmux window spanning several displays, and the zoomed threshold is only ~2/3 of the unzoomed one — which is the only thing PRD #313 changed here.

## The four places that must agree

A resize is correct only when all four carry the same number:

[...]

Issue #832, opened 2026-09-02
e2e_orchestration_remit 001/003/007 fail intermittently on main, reddening e2e-deterministic


vfarcic:
The `e2e-deterministic` job (CI lane 1, PRD #502) is failing intermittently on `main` in `tests/e2e_orchestration_remit.rs`. Discovered on PR #830, where it failed on a branch whose diff cannot be the cause.

## Measured

On PR #830 head `d230964` (run `33597048735`), exit 100:

```
Summary [261.696s] 9399 tests run: 9397 passed (1 leaky), 2 failed, 0 skipped
  FAIL [14.504s] dot-agent-deck::e2e_orchestration_remit orchestration_remit_003_reassertion_waits_for_confirmed_delivery
  FAIL [15.286s] dot-agent-deck::e2e_orchestration_remit orchestration_remit_007_compaction_reassertion_preserves_a_dispatched_task
```

Both are 10-second wait assertions (`tests/e2e_orchestration_remit.rs:244` and `:755`).

**That branch cannot be the cause** — its diff touches no file under `tests/` or `src/` at all, only `desktop/`, `desktop/src-tauri/`, `xtask/linkage-check/`, `docs/`, `prds/`, `changelog.d/` and `Cargo.lock`. Confirmed by checking `main` directly, where the job fails on **all three** of the most recent post-merge runs with a varying subset of the same family:

| Run on `main` | `e2e-deterministic` | Failing tests | `build-windows` |
| --- | --- | --- | --- |
| 33582926884 | failure | remit 001, 007 | success |
| 33575697073 | failure | remit 001 | success |
| 33571127245 | failure | remit 007 | success |

It then passed cleanly on PR #830 head `25b39b5` (run `33598267667`), which is consistent with flake rather than with anything having been fixed.

## Reading

**The varying subset across runs is the signature of contention flake**, which is what CLAUDE.md rule 5 records about this tier under parallelism — #415 measured 6–7 of 40 files failing in parallel against 40/40 at `-j 1`, with the same contention in #351, #322 and #701. These are wall-clock wait assertions, which is exactly the shape that degrades first when a runner is saturated.

Worth checking against `d47bc75` (*"feat(orchestration): re-assert the orchestrator remit on compaction and /clear (#787)"*), since these tests and the code under them are recent — a genuinely racy assertion and a contention-sensitive one look identical from the CI log, and only reading the tests separates them.

## Why this matters more than an ordinary flake

[...]

PR #841, opened 2026-09-02
fix: bound the exported *_to_buffer render seams' allocation (#748)

### LINKED ISSUE #748: Exported *_to_buffer render seams allocate a caller-sized TestBackend with no upper bound (u16::MAX x u16::MAX is an OOM/abort)
## What

`src/ui.rs` exports a family of `*_to_buffer` L1 render seams. Every one of them builds a `ratatui::backend::TestBackend` from dimensions the **caller** supplies, with no upper bound:

```rust
let backend = TestBackend::new(width, height);
```

`TestBackend::new` allocates exactly `width * height` cells up front, so a call at `u16::MAX` x `u16::MAX` asks for **4,294,836,225** cells. That is an OOM/abort, not an error a caller can handle. `src/lib.rs:66` is `pub mod ui;`, so all of these are reachable as `dot_agent_deck::ui::render_*_to_buffer` by anything that depends on the crate.

This is the repo-wide sibling of the bound that PRD #313 added to its own new seam. It predates PRD #313 by a long way, which is why it is filed here rather than fixed quietly inside that PR.

## Affected functions

Three allocation shapes, all in `src/ui.rs`. Line numbers are as of `d3f25fc`.

### A. Via the private helper `draw_to_buffer` (`src/ui.rs:19192`) — both dims caller-controlled

| Function | Line | Visibility |
| --- | --- | --- |
| `render_stats_bar_to_buffer` | 19213 | `#[doc(hidden)] pub` |
| `render_experimental_footer_to_buffer` | 19249 | `#[doc(hidden)] pub` |
| `render_quit_confirm_to_buffer` | 19268 | `#[doc(hidden)] pub` |
| `render_close_confirm_to_buffer` | 19288 | **plain `pub`** |
| `render_stop_confirm_to_buffer` | 19301 | `#[doc(hidden)] pub` |
| `render_star_prompt_to_buffer` | 19315 | `#[doc(hidden)] pub` |
| `render_config_gen_prompt_to_buffer` | 19324 | `#[doc(hidden)] pub` |

### B. Via the private helper `render_overlay_to_buffer` (`src/ui.rs:21229`) — both dims caller-controlled

| Function | Line | Visibility |
| --- | --- | --- |
| `render_help_overlay_to_buffer` | 21251 | **plain `pub`** |
| `render_dir_picker_to_buffer` | 21263 | **plain `pub`** |
| `render_new_pane_form_to_buffer` | 21279 | **plain `pub`** |
| `render_new_pane_orchestration_guard_to_buffer` | 21320 | **plain `pub`** |
| `render_new_pane_orchestration_name_collision_to_buffer` | 21368 | **plain `pub`** |
| `render_new_pane_form_schedule_to_buffer` | 21416 | **plain `pub`** |

Every wrapper in this group is plain `pub` with no `#[doc(hidden)]` at all — they are the most exposed part of the family.

### C. Construct their own `TestBackend::new(width, height)` — both dims caller-controlled

| Function | Line | Visibility |
| --- | --- | --- |
| `render_help_overlay_with_bindings_to_buffer` | 17348 | **plain `pub`** |
| `render_hints_bar_for_mode_to_buffer` | 17386 | **plain `pub`** |
| `render_card_with_declared_agent_to_buffer` | 19438 | `#[doc(hidden)] pub` |
| `render_dashboard_cards_to_buffer` | 19493 | `#[doc(hidden)] pub` |
| `render_card_grid_to_buffer` | 19577 | `#[doc(hidden)] pub` |
| `render_button_bar_with_bindings_to_buffer` | 19876 | **plain `pub`** |
| `render_button_bar_for_mode_to_buffer` | 19919 | **plain `pub`** |
| `render_command_banner_pane_to_buffer` | 20037 | `#[doc(hidden)] pub` |

Three further exported wrappers reach group C by delegation and carry no allocation of their own — fixing the callee fixes them: `render_hints_bar_to_buffer` (17374, **plain `pub`**) → `render_hints_bar_for_mode_to_buffer`; `render_card_to_buffer` (19352, `#[doc(hidden)] pub`) → `render_card_for_mode_to_buffer` (19389, `#[doc(hidden)] pub`) → `render_card_with_declared_agent_to_buffer`.

**Total: 24 exported entry points, 10 places to change** (2 shared helpers + 8 direct constructors).

### Not affected, for the record

- **Width-only seams** — these pass `TestBackend::new(width, 1)`, so the worst case is 65,535 cells, which is not an allocation concern. Listed only so a fixer does not have to re-derive it: `render_button_bar_to_buffer` (19847), `render_filter_bar_to_buffer` (21131), `render_rename_bar_to_buffer` (21158), `render_tab_bar_to_buffer` (21189) — all plain `pub`.
- **`render_orchestration_frame_to_buffer`** (19686, `#[doc(hidden)] pub`) — **already bounded** by PRD #313; see below.
- **`render_scroll_seam_pane_to_buffer`** (20246) and **`render_sized_scroll_seam_pane_to_buffer`** (20269) — module-private, not exported, and their dimensions are derived from constants rather than from a caller.
- `render_focused_pane_cursor_for_mode_to_position` and the in-module `#[cfg(test)]` call sites use fixed constants.

## Severity — read this before prioritising

These are **test seams**. Nothing in production calls them, no path reaches them from untrusted input, and there is no exploit here. The honest argument for fixing is narrower than "it is a vulnerability":

- they are **exported library API** — `pub mod ui` puts them in the public surface of the crate;
- **`#[doc(hidden)]` is not a compile barrier** — it hides an item from rustdoc and nothing else, so it does not make an item unreachable; and
- **half of them are not even hidden** — 12 of the 24 exported entry points, including all six of group B, are plain `pub` with no `#[doc(hidden)]` at all.

A `pub fn` that aborts the process on an in-range argument is a bad thing to leave in a library, and the fix is two lines per site. That is the whole case. Not a release blocker.

**Totality is a separate question and appears to be fine.** During the PRD #313 audit, `render_card_grid_to_buffer` was probed at `(0,0)`, `(0,10)`, `(10,0)` and `(1,1)` with an empty card slice, and `render_help_overlay_to_buffer` at the same degenerate dimensions — no panics in either. So this is an **allocation bound only**, not the totality flaw that `render_orchestration_frame_to_buffer` also had. The other seams in the family have not been probed; whoever picks this up should do so, but the two samples suggest the panic-on-degenerate-input problem was specific to the orchestration frame (which indexes `role_names[focused_role_index]` and draws a tab strip).

## Suggested fix

Clamp the dimensions **before** anything is allocated, in as few places as possible:

1. **One clamp inside `draw_to_buffer`** (`src/ui.rs:19192`) — covers all 7 of group A.
2. **One clamp inside `render_overlay_to_buffer`** (`src/ui.rs:21229`) — covers all 6 of group B.
3. **The same two lines in each of the 8 group-C functions**, which construct their own backend.

The clamp itself is what PRD #313 already landed:

```rust
let width = width.min(RENDER_SEAM_DIM_MAX);
let height = height.min(RENDER_SEAM_DIM_MAX);
```

`RENDER_SEAM_DIM_MAX` is already `#[doc(hidden)] pub const RENDER_SEAM_DIM_MAX: u16 = 1024;` at `src/ui.rs:19641` — 1024 x 1024 is ~1M cells, comfortably above any real terminal and comfortably below 4.3 billion. Reuse it rather than inventing a second cap; if it ends up shared by the whole family, consider renaming its doc comment away from the orchestration-frame-specific wording.

Clamping is preferable to returning an error: every real caller is asking for a terminal-sized frame, no real terminal is near the cap, and these seams are called from tests where a `Result` would be pure friction. Changing the return type of 24 exported functions to `Result` would be a much larger and less useful change.

Worth a small test per group asserting that a `u16::MAX` call returns a buffer of the clamped size rather than hanging or aborting — the group-A and group-B clamps can each be covered by one test through any one wrapper.

## Provenance

Surfaced by the audit on the **PRD #313** ("Zoom the focused agent pane") branch, alongside sibling issue #747.

**PRD #313 deliberately fixed only its own new seam.** Commit `d3f25fc` made `render_orchestration_frame_to_buffer` total and bounded — clamped to `RENDER_SEAM_DIM_MAX` (1024) on both dimensions and `RENDER_SEAM_ROLES_MAX` (64) on the role count, with a defined blank-buffer answer for degenerate inputs instead of a panic. It left every function listed above untouched, on the grounds that a repo-wide sweep of a pre-existing condition does not belong in a zoom PRD. This issue is that sweep.

vfarcic:
Closes #748.

## The bound

Every exported `*_to_buffer` render seam sized its `ratatui` `TestBackend` from caller-supplied dimensions with no upper limit. `TestBackend::new` allocates exactly `width * height` cells, so `u16::MAX` on both axes asked for 4,294,836,225 of them — a process abort, not an error a caller could handle. `pub mod ui` makes all 24 entry points reachable as `dot_agent_deck::ui::render_*_to_buffer`, and 12 of the 24 are plain `pub` with no `#[doc(hidden)]` (verified by enumerating them, not from the issue).

Both axes are now clamped to the existing `RENDER_SEAM_DIM_MAX` (1024) before anything is allocated, via one new private helper `clamp_render_seam_dims`, at all ten places the family allocates: the shared `draw_to_buffer` (covers 7 wrappers) and `render_overlay_to_buffer` (covers 6), plus the eight seams that build their own backend.

## Why clamp, not error or log

- **Not a `Result`:** every real caller asks for a terminal-sized frame and no real terminal is near the cap, so changing the return type of two dozen exported functions would be friction at every call site in exchange for bounding a case none of them reach.
- **Not silent, so no log needed:** the returned `Buffer`'s own `area` *is* the clamped size — the bound is observable in the value the seam already hands back, which is what the tests assert.
- **One rule for the family**, because #747 is about the harm done when one side of a resize clamps and another does not. Same cap as the daemon-side sibling `PTY_RESIZE_DIM_MAX` clamps with, and the same constant `render_orchestration_frame_to_buffer` already applies inline (PRD #313) — that function is deliberately left untouched, since it pairs the cap with a role-count bound and a degenerate-input guard.

Two group-A wrappers (`render_stats_bar_to_buffer`, `render_experimental_footer_to_buffer`) rebuilt a `Rect` from their raw arguments, which after the clamp would describe an area larger than the buffer; they now take `frame.area()`.

## How the test proves it

`tests/render_seam_bounds.rs` — **L1** (rule 4). The change is an allocation bound with no visual difference, so in-process `TestBackend` through the real exported seams is the right tier; no PTY, no subprocess, nothing to snapshot. Twelve seams covering nine of the ten change sites, plus a sibling test for `render_dashboard_cards_to_buffer`, whose height is derived from the card count rather than passed.

The failure mode today is a process abort, so the tests are ordered to **demonstrate the bound instead of reproducing the crash**: each seam is oversized on **one axis at a time**, which unbounded allocates only ~65k cells, and only after every one of those has passed does the test make the issue's own `u16::MAX` x `u16::MAX` call. A missing clamp is therefore diagnosed by a clean assertion failure, never by the allocator. Verified by neutering the helper to `(width, height)`:

```
assertion `left == right` failed: render_stats_bar_to_buffer: an absurd width must be bounded to 1024 and the height left alone
  left: (65535, 8)
 right: (1024, 8)
```

In-range requests are asserted to be honoured exactly, which is also what catches a `max`-for-`min` slip. New catalog entry `render/seam-bound/001`; both tests carry `/// Scenario:` comments (rule 7).

**Rule 12 does not apply** — no daemon, protocol, orchestration or hook change, and no wire shape or semantics touched.

## Gates

- `cargo clippy --workspace --all-targets --features e2e,e2e-live -- -D warnings` — exit 0.
- `cargo test-fast` — **3976 passed, 0 failed**. The new tests take 1.8s and 4.3s under full parallel load. (Note: `libgdk-3.so.0` / #815 did **not** reproduce here — the desktop crate's tests ran. An earlier run of this same tier had two `delegate_respawn_recovery` failures — `delegate_022` and `dispatch_003` — which pass in isolation and passed on the final full run; contention flakes of the shape rule 5 documents, untouched by this change.)
- `cargo xtask linkage-check` — ok (618 catalog ids, 547 annotations, 11 rules).
- Tests run, named per rule 5: the whole fast tier, plus the new `render_seam_bounds` file and the two rerun-in-isolation delegate tests above. No lane-2/credentialed test is in scope — this touches no agent path.

The diff is deliberately narrow in `src/ui.rs`: ten one-line clamps, two `Rect` → `frame.area()` swaps, one generalized doc comment on `RENDER_SEAM_DIM_MAX`, and the helper next to it. No formatting sweep, no refactoring of neighbouring code, no renames.


vfarcic:
## Addressing Greptile's P2 — valid facts, and the fix is the prose

> `render_button_bar_to_buffer` passes its public width directly to `TestBackend::new`, as do the filter, rename, and tab-bar seams. Widths above `RENDER_SEAM_DIM_MAX` therefore produce buffers wider than 1024, leaving these exported functions outside the newly documented family-wide bound and outside its regression coverage.

**Confirmed, and thank you — but the defect is in what I wrote, not in what the code does.** I enumerated all 29 exported `*_to_buffer` entry points rather than take either side on trust:

| shape | count | bounded? |
|---|---|---|
| two caller-controlled axes, via `clamp_render_seam_dims` | 24 | yes (this PR) |
| two axes, clamped inline (`render_orchestration_frame_to_buffer`) | 1 | yes (pre-existing, PRD #313) |
| one row — `TestBackend::new(width, 1)` | 4 | **no, deliberately** |

**Not clamping the four is the right call, and it is not mine to reverse here.** `u16::MAX * 1` is 65,535 cells (~2 MB) — not an allocation concern, which is the entire subject of #748. Capping the width would silently truncate a legitimately wide bar, and each of those seams documents its contract as "the returned buffer is `width × 1`". Issue #748 measured this and scoped them out explicitly under *"Not affected, for the record"*, so widening the code to rescue a sentence I wrote would be the tail wagging the dog — especially in `src/ui.rs`, where several PRs are open and diff width is a real cost. Nor is there a #747-style hazard: that harm comes from a renderer and a PTY disagreeing about **the same pane**; two unrelated seams with independent buffers have no invariant between them.

**What was genuinely wrong was four over-claiming sentences**, all mine, all from this PR — a textbook instance of the class CLAUDE.md rule 17 exists for (*"a narrower fact written one quantifier too wide"*), whose prescribed remedy is to write the narrower claim:

- `RENDER_SEAM_DIM_MAX`: "Largest dimension **any** exported `*_to_buffer` render seam will render" → scoped to two-axis seams, with the four exclusions named, the 65,535-cell measurement recorded, and an explicit *"do not write 'every exported seam is bounded'"* for the next reader.
- `clamp_render_seam_dims`: "**the whole family** goes through this function" → "of the 25 exported two-axis seams, 24 route through here", naming the 25th.
- Catalog headline "**Every** exported `*_to_buffer` render seam…", and its Asserts bullet "**every** seam that builds its own `TestBackend`" (it is seven of eight). The test's Scenario line and module doc over-claimed identically.
- Changelog: scoped, with the exclusion and its reason.

**And because prose decays, the exclusion is now executable.** Your "outside its regression coverage" half I have fixed as stated: `seam_bound_001_one_row_seams_are_deliberately_unbounded` calls all four one-row seams at `u16::MAX` columns and asserts they return that width verbatim, one row tall. I verified it bites by clamping `render_filter_bar_to_buffer`:

```
assertion `left == right` failed: render_filter_bar_to_buffer: a one-row seam honours its
width verbatim and is NOT capped at RENDER_SEAM_DIM_MAX (1024) — if this now fails because
the seam was capped, that is a deliberate scope change and RENDER_SEAM_DIM_MAX's doc comment
has to move with it
  left: (1024, 1)
 right: (65535, 1)
```

So the boundary is now asserted in **both** directions, and whoever later decides those four *should* be capped gets told what else has to move. 0.08s.

`587c4dd` is doc-comments-only in `src/ui.rs` plus that one test. Gates: `cargo fmt --check` clean, `cargo clippy --workspace --all-targets --features e2e,e2e-live -- -D warnings` exit 0, `cargo test-fast` **3977 passed / 0 failed**, `cargo xtask linkage-check` ok.

Per CLAUDE.md rule 8, `greptile.json` sets `triggerOnUpdates: false`, so this push will not draw a re-review — noting it so nobody waits for one.

vfarcic:
## CI on `587c4dd` — all four required checks green; the one failure is #832

**All four required checks pass:** `build` (6m6s), `build-macos` (6m10s), `build-windows` (12m30s), `security` (19s). Also green: `nix`, `devbox`, `windows-cross-check`, `desktop-web`, `changes`, `label`.

**`e2e-deterministic` failed — 9346 of 9347 passed, and the one failure is not from this PR.** It is `e2e_orchestration_remit::orchestration_remit_001_start_role_compaction_reasserts_remit`. Four independent lines of evidence, since "known flake" is worth proving rather than asserting:

1. **It is named verbatim in the tracking issue.** #832 — *"e2e_orchestration_remit **001**/003/007 fail **intermittently** on main, reddening e2e-deterministic"*. #818 (open) — *"e2e-deterministic has been red on every branch and on main since #789 merged"*. The fix is unmerged PR #837.
2. **`main` itself alternates on this job**, which is what "intermittently" means in practice: `e43fd09a` failure, `5993cbaf` failure, `f06ab8f4` success. Nothing about those commits involves render seams.
3. **This branch cannot reach it.** The diff is four files: `src/ui.rs` (`*_to_buffer` seams and their doc comments), the new `tests/render_seam_bounds.rs`, a catalog entry and a changelog fragment. `tests/e2e_orchestration_remit.rs` contains no reference to any `*_to_buffer` seam or to `RENDER_SEAM_DIM_MAX` — grep is empty. There is no call path from anything changed here to that test.
4. **It passes locally on this branch**: `cargo test-e2e orchestration_remit_001_start_role_compaction_reasserts_remit` → `PASS [2.286s]`.

Per CLAUDE.md rule 8, `e2e-deterministic` is deliberately advisory rather than required, so this does not gate the merge — but it should not be read as green either, and the reason it is red is #832, waiting on #837.

**No second Greptile review is coming, and nothing is waiting on one.** `greptile.json` has `triggerOnUpdates: false`; the summary comment still reads `Reviews (1)`, last reviewed commit `d6ff70d`, and `587c4dd` has no Greptile check-run. Its one P2 is answered in the comment above — the facts were right, the over-claim was in my prose, and the exclusion is now asserted in both directions.

Ready for review. I have not armed auto-merge; say the word if you would rather it land itself once approved.

prageethw (approved):
Verified: all ten allocation sites for the two-axis `*_to_buffer` seams (draw_to_buffer, render_overlay_to_buffer, and the seams building their own TestBackend) route through `clamp_render_seam_dims`, including the derived-height case in `render_dashboard_cards_to_buffer`. The two `Rect`→`frame.area()` swaps in `render_stats_bar_to_buffer`/`render_experimental_footer_to_buffer` are needed and correct — a raw `Rect` from the pre-clamp arguments would otherwise describe an area larger than the buffer. The one-row bar exclusion (max 65,535 cells) is reasoned and pinned by its own test. Test file exercises both axes independently before the combined u16::MAX case, which is the right order for diagnosing a missing clamp via assertion rather than allocator abort. The lone CI failure (`orchestration_remit_001_start_role_compaction_reasserts_remit`) is the pre-existing race PR #837 is fixing, unrelated to this change.

PR #842, opened 2026-09-03
ci: pin every hand-written workflow's actions to immutable commit SHAs (#769)

### LINKED ISSUE #769: Pin GitHub Actions to immutable SHAs on the release path
Raised by Greptile as a P2 on #768 and deferred there deliberately, because the fix is repo-wide rather than specific to that PR.

## The problem

`uses: actions/checkout@v7` resolves a **git tag** at workflow runtime, and tags are mutable — the owner can repoint `v7` at any commit, and publishers routinely do to ship patches under a major. So whatever code sits at that tag when a release runs is what executes on the runner.

In CI that produces a wrong or failed check: visible, recoverable. In `release.yml` it executes inside the steps that **compile and upload binaries users download and run**, and inside the job holding the Homebrew and Scoop publishing tokens. The artifact is the payload.

This is not hypothetical. The `tj-actions/changed-files` compromise (March 2025) worked exactly this way: version tags were retroactively repointed at a malicious commit, and every workflow referencing the tag picked it up silently.

## Current state

`release.yml` is uniformly tag-pinned:

| Job | Actions |
| --- | --- |
| `build` (produces the four CLI binaries) | `actions/checkout@v7`, `actions/upload-artifact@v7`, `dtolnay/rust-toolchain@v1`, `Swatinem/rust-cache@v2` |
| `finalize` (creates the release, holds `HOMEBREW_TAP_TOKEN` / `SCOOP_BUCKET_TOKEN`) | `actions/checkout@v7`, `actions/download-artifact@v8`, `arduino/setup-task@v3.0.0`, `softprops/action-gh-release@v3` |
| `desktop-bundle` / `desktop-publish` (added by #768) | the four above plus `pnpm/action-setup@v4`, `actions/setup-node@v4` |

**`finalize` is the highest-value target in the file** — it is the only job that both creates the public release and holds credentials for two external package registries.

Repo-wide there are already 113 SHA-pinned `uses:` lines against 168 tag-pinned ones, but every one of the SHA pins is in an **auto-generated agentic workflow lock file** (`issue-labeler.lock.yml` and friends). Nothing hand-written is pinned.

## Why it was not fixed in #768

Pinning only the two actions that PR introduced would leave `release.yml` carrying two conventions while concentrating the protection on the least-installed artifact — an alpha GUI nobody has downloaded yet — and leaving unpinned the CLI binaries that saw 78 downloads of `linux-amd64` in a single release. That is worse than either consistent state.

The counter-argument is worth recording, because it is the reason gaps like this persist: "the existing code is already unpinned" is precisely the reasoning by which nothing ever gets pinned. #768 added two entries to the gap, which makes it a contributor rather than a bystander.

## Suggested shape

- Pin every `uses:` in `release.yml` to a commit SHA with a trailing `# vX.Y.Z` comment. Renovate reads that form and keeps the pins current, so this does not become manual maintenance.
- Consider `ci.yml` in the same pass, though the argument there is weaker — a compromised CI action produces a visible failure rather than a shipped artifact.
- Leave the auto-generated `*.lock.yml` files alone; they are already pinned by their generator.

Do `finalize` first if this is split: it is the job where a compromise reaches both users and two external registries.

Refs #768

vfarcic:
Closes #769.

`uses: actions/checkout@v7` resolves a mutable git **tag** at workflow runtime, so whatever code sits at that tag when the job runs is what executes. In CI that produces a visible failure. On the release path the runner's output is the artifact rather than a verdict: `release.yml`'s `build` job compiles the four CLI binaries users download and run, and its `finalize` job both creates the public release and holds `HOMEBREW_TAP_TOKEN` and `SCOOP_BUCKET_TOKEN` — the one job where a compromise reaches users and two external package registries at once. The `tj-actions/changed-files` compromise of March 2025 worked exactly this way.

## What is pinned

All 66 `uses:` references across the eight hand-written workflows are now accounted for: **63 carry a full commit SHA with the tag in a trailing `# vX.Y.Z` comment**, and the remaining three are local `./.github/workflows/…` references, which GitHub resolves within the calling commit and does not accept a SHA for. 61 lines changed; the two DeterminateSystems pins added under #450 were already in this form and were re-verified rather than touched. The auto-generated `*.lock.yml` files are untouched — their generator already pins them, and a hand edit would be overwritten.

| file | `uses:` | SHA-pinned | local `./` |
| --- | --- | --- | --- |
| `release.yml` | 17 | 16 | 1 |
| `ci.yml` | 34 | 34 | 0 |
| `docs-publish.yml` | 4 | 4 | 0 |
| `docs.yml` | 4 | 4 | 0 |
| `aarch64-crossbuild-check.yml` | 3 | 3 | 0 |
| `labeler.yml` | 1 | 1 | 0 |
| `stale.yml` | 1 | 1 | 0 |
| `issue-labeler-batch.yml` | 2 | 0 | 2 |

The 18 distinct pins:

| action | tag | commit | tag object |
| --- | --- | --- | --- |
| `actions/checkout` | v7 | `3d3c42e5aac5ba805825da76410c181273ba90b1` | lightweight |
| `actions/upload-artifact` | v7 | `043fb46d1a93c77aae656e7c1c64a875d1fc6a0a` | lightweight |
| `actions/download-artifact` | v8 | `3e5f45b2cfb9172054b4087a40e8e0b5a5461e7c` | lightweight |
| `actions/setup-node` | v7 | `820762786026740c76f36085b0efc47a31fe5020` | lightweight |
| `actions/labeler` | v7 | `bf12e9b00b37c5c0ca2b87b79b2daf7891dbda13` | lightweight |
| `actions/stale` | v11 | `4391f3da665fdf50b6810c1a66712fb9ba21aa93` | lightweight |
| `dtolnay/rust-toolchain` | v1 | `6c977a6ca4077a0ceb28ffbe03f59d46e9ac8772` | lightweight |
| `Swatinem/rust-cache` | v2 | `6323deb102c322ba6fcbdcafc7e3dddab59af2b6` | **annotated** |
| `arduino/setup-task` | v3.0.0 | `c0bc642852239c2689f73f4ea6459c29405f3c52` | **annotated** |
| `softprops/action-gh-release` | v3 | `efb35369e0ad2afab669f228072c1b0d510eae64` | **annotated** |
| `pnpm/action-setup` | v6 | `0977fd99725f1db4007ccb2928dbb4e90d06cc86` | **annotated** |
| `taiki-e/install-action` | v2 | `e67fa11c4b9316fa714ddf0abed07a0c3143b95b` | lightweight |
| `jetify-com/devbox-install-action` | v0.15.0 | `8c6a66ed6273138b1915457069de78cb52fe3bd7` | lightweight |
| `docker/setup-buildx-action` | v4 | `37fe631027851001ddb9b187196cc803df7f5f0e` | lightweight |
| `docker/login-action` | v4 | `dbcb813823bdd20940b903addbd779551569679f` | lightweight |
| `docker/build-push-action` | v7 | `53b7df96c91f9c12dcc8a07bcb9ccacbed38856a` | lightweight |
| `DeterminateSystems/nix-installer-action` | v22 | `ef8a148080ab6020fd15196c2084a2eea5ff2d25` | *pre-existing, re-verified* |
| `DeterminateSystems/magic-nix-cache-action` | v14 | `908b263ff629f4cc17666315b7fd3ec127c6244d` | *pre-existing, re-verified* |

## How each SHA was resolved and checked

This file is not exercised by pull-request CI, so a wrong SHA would surface as a broken release. Nothing was written from memory. Every tag went through **three** checks that had to agree:

1. **`gh api repos/<owner>/<repo>/git/ref/tags/<tag>`**, with `.object.type` inspected and any `tag` object dereferenced through `git/tags/<sha>` to the commit it points at. This is load-bearing: four of the sixteen tags are annotated tag objects, and pinning an annotated tag's *own* SHA produces a ref that looks pinned and is not the commit Actions checks out.
2. **`gh api repos/<owner>/<repo>/commits/<tag>`** — an independent server-side resolution of the same tag. Route 1 and route 2 had to return the same SHA for every one of the sixteen.
3. **Read-back**: after writing, the pins were parsed back *out of the files as committed*, and each `owner/repo@sha # version` triple was re-resolved through both routes above, plus `gh api repos/<owner>/<repo>/commits/<sha>` to confirm the SHA in the file is a real commit in that repository. All 18 distinct pins verified, including the two pre-existing ones.

The version in every trailing comment is the tag whose SHA was actually pinned — that is what step 3 checks, since it re-resolves *the comment* and compares against *the SHA on the same line*.

## Renovate

The trailing `# vX.Y.Z` comment is the exact form Renovate's `github-actions` manager reads, so this does not become manual maintenance. The comment carries the ref that was there before (`# v7`, `# v1`, `# v3.0.0`), matching the convention the two pre-existing DeterminateSystems pins already use, so each action keeps the same version semantics it had — Renovate proposes a `digest` update when the tag moves and a version update when a new tag appears. Both `digest` and `pin` sit inside `renovate.json`'s automerging `GitHub Actions` group (`matchDepTypes: ["action"]`), which is unchanged.

Nothing here touches the `uses-with` lane: no `toolchain:` or `tool: cargo-nextest@` input changed, so the toolchain pins stay held for a human and `scripts/check-pin-lockstep.sh` still reports both sides in lockstep.

## Why `ci.yml` is included

The issue treats `ci.yml` as optional because a compromised CI action yields a visible failure rather than a shipped artifact. It is included anyway, for the reason the issue itself gives for not deferring: leaving some workflows tag-pinned and others SHA-pinned is the "two conventions" state that is worse than either consistent one, and `ci.yml` was *already* mixed (two SHA pins from #450 among 32 tag pins).

Two files beyond the issue's list are included for reasons worth stating rather than assuming:

- **`docs-publish.yml` is on the release path.** `release.yml`'s `docs` job calls it via `uses:` with `secrets: inherit`, which forwards *every* repository secret — `RELEASE_TOKEN`, `HOMEBREW_TAP_TOKEN`, `SCOOP_BUCKET_TOKEN` — so its `docker/*` actions run with the release credentials in scope even though nothing in that file names them. That places it above `ci.yml` in value, not below it.
- **`docs.yml`, `aarch64-crossbuild-check.yml`, `labeler.yml`, `stale.yml`** cost two additional SHA resolutions between them (`actions/labeler`, `actions/stale`; everything else was already resolved), and including them is what makes "every hand-written `uses:` in this repository is immutable" a checkable property instead of a list of exceptions.

## Behaviour is unchanged, and that is verified rather than asserted

No workflow does anything different. Same actions, same versions, same inputs — the ref after `@` is the only thing that moved, and that was established mechanically: all eight files were parsed with a YAML loader before and after, and compared structurally (job topology, step ordering, every key outside `jobs`, and every job with `uses` values normalised to the action name). The comparison is clean. `git diff --numstat` is 61 insertions / 61 deletions, and every changed line matches `^[+-]\s*(-\s+)?uses:`.

## Gates

- `cargo clippy --workspace --all-targets --features e2e,e2e-live -- -D warnings` — clean (all four flags, per CLAUDE.md rule 2).
- `cargo test-fast` — **3974 passed, 0 failed**, including the two `xtask/linkage-check` tests that read `.github/workflows/`: `pin_lockstep::repository_pins_are_in_lockstep` and `pin_lockstep::yaml_extension_and_nested_dirs_match_renovates_file_set`. `scripts/check-pin-lockstep.sh` was also run directly: *"ok: Rust toolchain pinned at 1.97.1 on both sides (9 workflow site(s)) / ok: cargo-nextest pinned at 0.9.143 on both sides (4 workflow site(s))"*.
- One intermediate `cargo test-fast` run showed two failures in `delegate_respawn_recovery` (`delegate_022_…`, `dispatch_003_…`). Both pass alone, pass as a module (111/111), and the full tier passes on re-run — the known parallel-contention flake class in CLAUDE.md rule 5 (#415/#351/#701). Neither test reads a workflow file, and the tier had already passed clean on the same tree before the Markdown-only fragment was added.

No e2e tests were run: this change compiles nothing and alters no runtime surface, so no `tests/CATALOG.md` entry covers it. CI's `e2e-deterministic` job will exercise lane 1 on this PR as usual.

## Changelog fragment: included

`changelog.d/769.misc.md`. The instinct was to skip it as CI-only with no user-facing effect, but the repo's own convention decides otherwise: towncrier's `misc` type is defined in `pyproject.toml` as *"Internal improvements, CI/CD, tooling, refactoring (improves the project but users don't directly see it)"*, and the closest precedent — `ebca5a7`, "pin the two dependencies that resolved from mutable refs" (#450, #323) — shipped two `.misc.md` fragments for exactly this class of change.

## What this deliberately does not fix

A commit pin fixes the fetcher, not the fetched. Several of these actions download an executable payload at run time: `arduino/setup-task` resolves `version: 3.x` to the newest 3.x release when `finalize` runs, `actions/setup-node` and `pnpm/action-setup` resolve `24` and `11` the same way, and `taiki-e/install-action` fetches a prebuilt cargo-nextest. Narrowing those is a behaviour change with its own upstream verification to do, so it is left out of a pinning-only PR — the same boundary #450 drew for the DeterminateSystems payload channel.

The **nested** layer is closed, and was checked rather than assumed: of the sixteen actions, thirteen are JavaScript actions with no internal `uses:` at all; of the three composite ones, `dtolnay/rust-toolchain` and `taiki-e/install-action` reference no other action, and the four internal references in `jetify-com/devbox-install-action` are themselves already SHA-pinned upstream (`actions/cache/restore` and `actions/cache/save` at `cdf6c1fa…`).

No bug was found in any workflow while doing this, so nothing was filed.


vfarcic:
## CI outcome

**All four required checks pass** — `build`, `build-macos`, `build-windows`, `security` — as do `desktop-web`, `devbox`, `nix`, `aarch64-crossbuild`, `windows-cross-check`, `changes` and `label`.

**Greptile** reviewed at `4881dd0` and reported **5/5 confidence with no findings**. Checked on all four surfaces rule 8 names, not just the green check: the check-run, the summary comment, `gh pr view --json reviews` (empty), and `gh api repos/{owner}/{repo}/pulls/842/comments` (empty — no inline P1/P2).

### `e2e-deterministic` failed, and it is a known pre-existing flake

One test out of 9344: `orchestration_remit_007_compaction_reassertion_preserves_a_dispatched_task`, panicking at `tests/e2e_orchestration_remit.rs:244:5` with

> the daemon's own ListAgents/live-status join never reported Compacting for pane 1 (agent_id 1) within 10s

That is a 10-second wait on a live-status transition — the contention-sensitive class CLAUDE.md rule 5 records (#415/#351/#322/#701), and rule 8 is why this job is advisory rather than required.

It is not caused by this PR, and the evidence is not "it looks unrelated". The same job fails intermittently on **`main` itself**, on `push` runs with no PR and no workflow change involved, and **two of those failures panicked at the identical line**:

| run | branch / event | failing location |
| --- | --- | --- |
| this PR | `agent/dispatch-issue-769` | `tests/e2e_orchestration_remit.rs:244:5` |
| `5993cbaf` | `main`, push | `tests/e2e_orchestration_remit.rs:244:5` |
| `46ac28c5` | `main`, push | `tests/e2e_orchestration_remit.rs:244:5` |
| `e43fd09a` | `main`, push | `tests/delegate_respawn_recovery.rs:368:5` |

Across the last twelve `ci.yml` runs the job failed on four unrelated branches (`agent/dispatch-issue-748`, `agent/dispatch-issue-491`, `agent/dispatch-issue-834`, `agent/dispatch-prd-803`) as well as on `main`.

Independently of the history: this PR compiles nothing and changes no runtime surface, so it has no mechanism by which to affect a daemon status transition. The structural before/after comparison in the description establishes that the ref after `@` is the only thing that moved.

Every pinned action step in that very job **succeeded** — the job reached and failed inside `cargo test-e2e`:

```
✓ Run actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1
✓ Run dtolnay/rust-toolchain@6c977a6ca4077a0ceb28ffbe03f59d46e9ac8772
✓ Run Swatinem/rust-cache@6323deb102c322ba6fcbdcafc7e3dddab59af2b6
✓ Run taiki-e/install-action@e67fa11c4b9316fa714ddf0abed07a0c3143b95b
✗ Run cargo test-e2e          ← 9343/9344 passed
```

## Including `ci.yml` bought an execution test, not only consistency

`release.yml` is not exercised by pull-request CI — the reason the description spells out how each SHA was verified. But `ci.yml` **is**, so this run actually executed 10 of the 18 pins, and six of them are pins `release.yml` shares:

- **Shared with `release.yml`, proven to run here:** `actions/checkout`, `actions/setup-node`, `actions/upload-artifact`, `dtolnay/rust-toolchain`, `pnpm/action-setup`, `Swatinem/rust-cache`.
- **Also exercised:** `taiki-e/install-action`, `jetify-com/devbox-install-action`, and the two pre-existing DeterminateSystems pins (via `nix`).
- **Not reachable from PR CI:** `actions/download-artifact`, `arduino/setup-task`, `softprops/action-gh-release` (release-only), plus `actions/labeler` and `actions/stale` — though `actions/labeler` did run and pass, in the `label` check on this very PR, and the three `docker/*` pins are only reached by the docs publish path.

So six of the eight distinct actions `release.yml` depends on have now been observed resolving and running at their new SHAs.

prageethw (approved):
Spot-checked all 16 distinct actions pinned in this diff by independently resolving each `@vX` tag via the GitHub API (dereferencing the 4 annotated-tag cases — Swatinem/rust-cache, pnpm/action-setup, arduino/setup-task, softprops/action-gh-release — to their commit objects) — every SHA matches the tag it claims. Grepped the PR-head content of all 7 changed workflow files for remaining `uses:` lines: every external action is SHA-pinned, and the one unpinned reference (`uses: ./.github/workflows/docs-publish.yml`) is a local reusable-workflow call that can't take a SHA. LGTM.

PR #846, opened 2026-09-03
test: run the harness unit tests once, not once per binary (#806)

### LINKED ISSUE #806: The harness unit tests are re-run once per test binary: 104 tests become 6656 executions in every lane-1 CI run
> **BLOCKED — do not start this until PR #805 is merged.**
>
> This issue moves ~104 harness unit tests out of `tests/common/mod.rs`. PR #805 changes **3,129 lines of that same file** (the diagnostic redaction seam, the wrapped-credential matcher, the credential importers). Starting before #805 merges means resolving a very large conflict by hand in security-critical redaction code, which is the worst possible place to do it — thirteen audit rounds went into getting that code right.
>
> Check first: `gh pr view 805 --json state,mergedAt`. If it is not `MERGED`, stop and pick something else.
>
> This is the **first** follow-up to take once #805 lands — it is the largest available reduction in local and CI test cost.

---

## The defect

`tests/common/mod.rs` carries a `harness_unit_tests` module of pure-function unit tests, and **every test binary that links the harness re-runs all of them**. There are 71 `tests/e2e_*.rs` files plus 17 non-e2e test files that link `tests/common/mod.rs`, so those unit tests are executed once per binary rather than once.

The cost is **multiplicative, not additive**: every harness unit test added from now on costs 64× in every CI run of lane 1, and 88× in a local `cargo test-e2e-live`, forever.

## Measured

On `agent/e2e-live-credential-sinks` at `0f64c79` + this PR's changes, 16-core box:

| selection | binaries linking the harness | `harness_unit_tests` executions | total selected | share |
| --- | --- | --- | --- | --- |
| `cargo nextest list --workspace --features e2e` (lane 1, what CI runs) | **64** = 47 lane-1 e2e + 17 non-e2e | **6656** | **8961** | **74%** |
| `cargo nextest list --workspace --features e2e,e2e-live` (lane 1 + lane 2) | **88** = 71 e2e + 17 non-e2e | **9152** | **11518** | **79%** |
| `cargo nextest list --workspace` (fast tier) | — | — | 3894 | — |

**104 distinct** harness unit tests produce those 6656 and 9152 executions — exactly 104 × 64 and 104 × 88. So of lane 1's ~8961 selected tests, roughly **2305 are distinct scenarios and 6656 are the same 104 unit tests re-run 64 times**.

The 17 non-e2e binaries in the count: `agent_detection`, `agent_event`, `agent_lifetime_bound`, `daemon_status`, `delegate_prompt_injection`, `delegate_respawn_recovery`, `foreign_pane_detection`, `grid_box_helpers`, `harness_isolation`, `idle_worker_detector`, `orchestration_delegate`, `real_agent_preflight`, `render_button_bar`, `render_layout`, `spawn_time_role_prompt_submit_after_session_start`, `work_done_reporting`, `wrap_io`.

The owner independently measured **106 × 64 = 6784** against a ~7478-test lane-1 total shortly before this PR. The 106 → 104 difference is this PR removing two containment guards; the total differs because that count and this one were taken with different exclusions. Either measurement gives the same shape.

## Why it went unnoticed

The symptom is visible but the cause is not. PR #805's coder observed "the +64 is the two new guard tests × 32 binaries" while explaining a test-count delta, and treated it as arithmetic rather than as a defect: adding two pure-function assertions to a shared module moved the suite's test count by 64, and nothing about that reads as wrong at the time.

## Why it matters

These are tests of **pure functions** — redaction range arithmetic, credential-value collection, env-var trimming, name parsing. They neither exercise nor depend on the binary they are linked into, so 63 of the 64 executions prove nothing the first did not. Meanwhile:

- CLAUDE.md rule 5's argument for CI running lane 1 rests on it being an affordable per-PR job; 74% of its work is duplication.
- Every future harness unit test silently costs 64×. That is the wrong incentive on a module this repo deliberately keeps well-tested.

## The fix (not in PR #805)

Move `harness_unit_tests` somewhere it runs **once**: either a single dedicated test binary that includes the harness (`tests/harness_unit.rs` with a `#[path]` include, the arrangement `src/test_temp.rs` already uses per issue #474), or into the fast tier alongside the code they cover. `#[cfg]`-gating them to one binary is the cheap version and would also work.

Care needed on two points: the harness module is included by `mod common;` in each test file rather than being a crate, so whatever gates them must not change what the other binaries compile; and `tests/CATALOG.md` / `linkage-check` rules should be checked for anything keyed to the current arrangement.

Deliberately **not** fixed in PR #805 — that PR is about taking the credentialed lane out of CI and rewriting the instructions around it, and this is a separate, mechanical change with its own risk of quietly dropping coverage.

Refs: #502, #785, PR #805.

vfarcic:
Closes #806.

The 108 pure-function unit tests covering `tests/common/mod.rs` executed once per test binary that linked the harness. They now execute once, from a single dedicated binary. No test's assertions changed and no test was dropped.

## The mechanism, and why the fix has the shape it does

`cfg(test)` is true in **every** integration-test target, not only in the one being tested. So `#[cfg(test)] mod harness_unit_tests { … }` inside `tests/common/mod.rs` was never "the tests for this file" — it was a module that every binary writing `mod common;` compiled, and that nextest then ran, once each.

The fix is the one the issue proposes: the module moves to `tests/common/harness_unit_tests.rs`, and `tests/harness_unit.rs` — a new two-declaration test binary — is the only place that names it. The module tree is the mechanism because nothing else can single out one test binary: `cfg(test)` is true in all of them, and feature, `RUSTFLAGS` and `build.rs` cfgs are per-package rather than per-target.

That makes the moved module a **sibling** of `common` rather than a child, so `use super::*` no longer reaches the harness's private helpers. The helpers under test therefore carry `pub(crate)` — 66 declarations of 64 distinct items (two are declared twice, under `#[cfg]` pairs) plus four struct fields. This is the part worth reviewing, and it is the price of the arrangement rather than a preference: a private item cannot be re-exported at wider visibility (E0364/E0365, verified in a scratch crate), so there is no allowlist-shaped alternative that leaves the declarations alone. Every widening is additive; nothing narrowed, and no item outside this set changed.

## Measured, before and after, on the same 16-core box

`cargo nextest list --workspace --exclude dot-agent-deck-desktop [features]`. The exclusion is issue #771 — the devbox has no Tauri GTK/WebKit deps for a `--workspace` *run*; CI's counts will be higher by the desktop crate's tests, and identically so on both sides. Before is `45528bc`, after is this branch.

| selection | command | before | after | change |
| --- | --- | --- | --- | --- |
| fast tier | `cargo test-fast` | 3939 | **2211** | −1728 (−43.9%) |
| lane 1 — what CI runs per PR | `cargo test-e2e` | 9309 | **2397** | −6912 (−74.2%) |
| lane 1 + lane 2 | `cargo test-e2e-live` | 11962 | **2458** | −9504 (−79.4%) |

Broken out, the same three selections:

| selection | harness-unit executions | binaries compiling them | distinct tests |
| --- | --- | --- | --- |
| fast tier | 1836 → **108** | 17 → **1** | 108 → **108** |
| lane 1 | 7020 → **108** | 65 → **1** | 108 → **108** |
| lane 1 + lane 2 | 9612 → **108** | 89 → **1** | 108 → **108** |

The issue measured 104 × 64 on the pre-merge #805 branch; on `main` today it is 108 × 65, so the numbers above are mine taken the same way on both sides rather than the issue's carried forward.

## Coverage is preserved, and that is checked rather than asserted

Comparing the `nextest list` JSON before and after, per selection:

- the **set of harness unit test names** is identical — `missing_after=[]`, `new_after=[]` for all three selections;
- every **non-harness `(binary, test)` pair** is identical too — 2103 / 2289 / 2350 pairs, none lost, none added.

So this removes duplicate executions and nothing else.

The moved code is also verified to be a pure move. Compared whitespace-stripped against `HEAD:tests/common/mod.rs`, the entire 4065-line body differs in exactly two places, both `cargo fmt` normalisations that the freed 4-column indent enabled: a redundant `{ … }` dropped from a closure body, and a `=> { i += 3 }` match arm rewritten to `=> i += 3,`. Nothing else — no assertion, message or literal was touched.

## Gates

- `cargo fmt --check` — clean. I checked that rustfmt actually *visits* the new file rather than silently skipping it (it is reached through the `#[path]` declaration in `tests/harness_unit.rs`): injecting a deliberately misformatted function made `cargo fmt --check` report it, and removing it went clean again. That mattered — an `include!`-based arrangement, which I tried first, would have taken 9000 lines of harness out of the fmt gate.
- `cargo clippy --workspace --all-targets --features e2e,e2e-live -- -D warnings` — exit 0, all four flags as rule 2 spells them, desktop crate included.
- `cargo xtask` (linkage-check, 11 rules) — ok.

## Tests

Rule 5's shared-harness exception applies — `tests/common/mod.rs` is linked by every `tests/e2e_*.rs` file — so this is a broad pass rather than a scoped one.

- **`cargo test-fast --exclude dot-agent-deck-desktop`** — `2211 tests run: 2211 passed, 0 skipped` (34.3s). All 108 harness unit tests are in that number.
- **`cargo xtask`** (linkage-check, 11 rules) — `ok (617 catalog ids, 545 annotations, 100 allowlisted, 11 rules)`.
- **`cargo test-e2e`** (lane 1 in full) — `2432 tests run: 2424 passed, 8 failed` (127.9s).

**The 8 lane-1 failures are pre-existing on `main` and not caused by this change — verified by A/B on this box, not assumed.** They are `orchestration_remit_001`–`007`, `focus_007` and `new_pane_015`, all failing on a 10s wait for a spawned deck or daemon. Three pieces of evidence:

1. **The identical scoped command fails the identical set on `main`.** `cargo test-e2e -E 'binary(e2e_orchestration_remit) + test(new_pane_015…) + test(focus_007…)'` on `45528bc`: `117 tests run: 108 passed, 9 failed` — the same nine test names, same order of magnitude of runtime. On this branch, same command: `9 tests run: 0 passed, 9 failed`. (The 108 extra tests in `main`'s run of that command are the harness unit tests riding along in those three binaries — the defect this PR fixes, visible in the control.)
2. **A full `cargo test-e2e` on `main` on this same box also fails**: `9344 tests run: 9339 passed, 5 failed`, including `orchestration_remit_007` — which failed in every run of this tier I did today, on both revisions.
3. **They pass individually on this branch**: `orchestration_remit_001` in 2.6s and `orchestration_remit_007` in 3.4s, versus 13–16s (i.e. into the 10s timeout) when run alongside their siblings.

The box was shared with sibling agent worktrees throughout — up to 28 concurrent `rustc` processes and a load average of 73 on 16 cores at the worst point, which is the contention CLAUDE.md rule 5 records from #415/#351/#322/#701. My first lane-1 run, taken at load 73, failed **108** of 2432; the identical command at load ~9 failed **8**. Same code, same command.

Lane 2 was not run: nothing here touches a real-agent path, and those 24 files are type-checked by the clippy gate above.

### Incidental: the local tier got faster

Full `cargo test-e2e` on this box, both runs on the same machine within about twenty minutes of each other: `main` 325.2s for 9344 tests, this branch 127.9s for 2432. Ambient load differed between the two, so treat 2.5x as an indication rather than a benchmark — the reliable part is the selection count, which is exact.

## Notes for the reviewer

- **Three tests re-run their own binary** (`the_first_allocation_in_a_fresh_process_is_contained`, `harness_temp_root_is_removed_when_the_process_exits_normally`, `the_panic_seam_redacts_a_key_this_process_read_from_the_environment`). Each derives its child filter from `module_path!()` rather than naming a binary, so they followed the move without edits, and they pass.
- **Nothing in `src/` changed**, so rule 12's cross-version contract check does not apply — no daemon, protocol, orchestration or hook surface is touched.
- **No `.config/nextest.toml` override is affected**: every filter there targets an e2e test function by name, none matches this module path.
- **Left alone deliberately:** CLAUDE.md says `tests/common/mod.rs` "is linked by all 71 `tests/e2e_*.rs` files" in rules 2 and 5. There are 72 today, and after this PR the harness is linked by 90 files rather than 89 — the new binary is not an e2e file. That drift predates this branch, and rewriting rule 5's counts is a separate edit I did not want to bury in a test-infrastructure PR.


vfarcic:
## CI result, and what the advisory lane says

**All four required checks pass**: `build`, `build-macos`, `build-windows`, `security`. Greptile reviewed at 5/5 with no inline findings.

**`build-windows` failed on the first attempt and passed on re-run, and the failure was not this PR's.** Both failures were `spawn::tests::dispatch_018_user_input_disarms_detached_submit_probe` and `dispatch_019_writer_release_does_not_expose_unstamped_user_input` — unit tests inside `src/spawn.rs`, a file this PR does not touch and which is byte-identical between my base (`45528bc`) and current `main` (`0153758`). The panic was `precondition: attempt 1 must physically reach the detached pane`, a timing precondition. Re-run: green.

**The CI numbers confirm the local measurement.** `e2e-deterministic` on this PR selects **2474** tests; the same job on `main` (run `33706828334`) selects **9386**. That is the same shape as the local figures in the description — they differ only because CI includes the desktop crate, which the local runs exclude for issue #771.

`build-windows`'s fast tier tells the same story from the other side: **2467** tests on `main`, **1797** here. The 670 that went are 67 Windows-eligible harness unit tests (108 minus the 41 gated `#[cfg(unix)]`) that were running in 11 Windows binaries and now run in one.

**`e2e-deterministic` is red on this PR and also red on `main`, on the same class of test.**

| | tests run | failed | which |
| --- | --- | --- | --- |
| this PR | 2474 | 2 | `work_done_004`, `orchestration_remit_007` |
| `main` @ `0153758` | 9386 | 2 | `pane_input_007`, `orchestration_remit_001` |

`orchestration_remit_*` is in both. It is also what failed locally on **both** revisions: running `cargo test-e2e -E 'binary(e2e_orchestration_remit) + …'` on `main` fails the identical nine tests it fails here. The job is advisory rather than required for exactly this reason — CLAUDE.md rule 8 records that it has no honest baseline yet — and nothing here changes its behaviour, only how many tests it selects.

I have not touched `e2e_orchestration_remit`; its flakiness looks worth its own issue, but it is not this PR's to fix.

vfarcic:
@prageethw — ready for review whenever you have a moment. Flagging the red check up front so it does not read as a reason to wait.

**All four required checks are green** (`build`, `build-macos`, `build-windows`, `security`), and there are no unresolved review threads.

`e2e-deterministic` is **red, and it is not this PR.** It failed on `orchestration_remit_007_compaction_reassertion_preserves_a_dispatched_task` and `work_done_004_unsolicited_completion_is_visibly_labelled_in_the_attached_tui` — the pre-existing breakage tracked as #818, which has been red on `main` and on every branch since #789 merged. The same lane is red on #739 and #847 for the same remit tests. Per CLAUDE.md rule 8 that job is deliberately advisory rather than required. **PR #837 is the fix**, so this clears once #837 lands.

Nothing in this PR touches the remit or work-done paths — it moves the `harness_unit_tests` module so those pure-function tests execute once instead of once per linking test binary.

prageethw (approved):
Clean test-infra-only move: the harness unit tests go from linked-and-re-executed by every e2e binary to a single dedicated `tests/harness_unit.rs` binary, mechanically justified (cfg(test) is target-wide, not file-scoped) and measured (43.9%/74.2%/79.4% reduction across the three selections). Coverage preservation is checked via nextest-list diffing, not asserted. Only test files changed, nothing in src/. The e2e-deterministic CI failure (work_done_004, orchestration_remit_007) is unrelated to this diff — main's own most recent independent CI run fails a different pair of e2e tests (pane_input_007, orchestration_remit_001), consistent with the pre-existing runner-contention flakiness the PR body already documents with A/B evidence.

PR #854, opened 2026-09-03
fix(tests): wait for user input to reach the PTY instead of sleeping 75ms (#850)

### LINKED ISSUE #850: dispatch_020 hard-reds the required build job: type_user_bytes uses a fixed 75ms sleep as a synchronisation primitive
## Observed

PR #843's **required** `build` job failed on:

```
test spawn::tests::dispatch_020_payload_guards_are_scoped_to_one_delivery ... FAILED
test result: FAILED. 0 passed; 1 failed; 0 ignored; 0 measured; 1442 filtered out; finished in 1.37s
thread 'spawn::tests::dispatch_020_payload_guards_are_scoped_to_one_delivery' (28820)
  panicked at src/spawn.rs:4186:9
```

That PR was a Renovate no-op whose entire diff was byte-identical to `main`'s `Cargo.lock` (closed for that reason), so a dependency change cannot be the cause. **Not reproduced locally** — this is a read of the code and the CI log, not a bisect.

## The failing assertion is a precondition, not the property under test

`src/spawn.rs:4186` is the `assert!` guarding the setup, whose message is:

> `precondition: the production-shaped bracketed multiline paste must physically reach the PTY`

The sequence is: write a bracketed-paste draft to a real PTY, then `snapshot()` it and assert the bytes are physically present before the test proceeds to exercise payload-guard scoping.

## Mechanism: a fixed sleep used as a synchronisation primitive

`type_user_bytes` (`src/spawn.rs:3117-3136`), reached via `type_user_draft`, is the write half:

```rust
writer.write_all(bytes).expect("write detached user input bytes");
writer.flush().expect("flush detached user input bytes");
drop(writer);
registry.note_user_input(pane_id);
tokio::time::sleep(Duration::from_millis(75)).await;
```

It writes, flushes, then **sleeps a fixed 75 ms** and returns. Nothing waits for the bytes to be observable. The caller snapshots immediately afterwards and asserts they arrived. On a contended runner a PTY round trip exceeding 75 ms produces an empty or partial snapshot and the precondition fires — which is exactly the failure text seen.

This makes the test load-sensitive by construction, with no relationship to what it is trying to verify.

## Blast radius

- `type_user_bytes` / `type_user_draft` are called **11 times**, from **two** tests: `dispatch_020_payload_guards_are_scoped_to_one_delivery` and `dispatch_018_user_input_disarms_detached_submit_probe`. Both share the flake.
- `src/spawn.rs`'s test module contains **29** `tokio::time::sleep` / `thread::sleep` occurrences. This issue is about the one on the observed path, but the pattern is broad enough that a survey is worthwhile before assuming 4186 is the only reachable instance.

## Why this reaches a required check, which is what makes it worth fixing

- It runs in `cargo nextest run --workspace` — the fast tier — which the Linux `build` job runs, and `build` is one of the four **required** status checks (CLAUDE.md rule 8).
- `.config/nextest.toml` sets `retries = 0` under `[profile.default]` (line 69). The `retries = 1` at line 284 is a name-scoped override for an unrelated e2e test; **neither `dispatch_020` nor `dispatch_018` is covered by any override**, so a single flake hard-reds a blocking check with no retry.
- Direct precedent: issue #402, where `delegate_011` was load-flaky on the macOS fast tier and "the fast tier has no `--retries`, so it hard-reds a blocking job".

## Structural cause: `src/` is exempt from the no-sleep rule, by design

CLAUDE.md and `xtask/linkage-check` rule 5 (Decision 21) forbid raw `std::thread::sleep` / `tokio::time::sleep` as a wait mechanism — but only in `tests/e2e_*.rs`. `xtask/linkage-check/src/main.rs:689-692` says so explicitly:

> Scan tests/ AND src/ for `#[spec(...)]` annotations. PRD #83 added per-tab-selection `#[spec]` unit tests in `src/tab.rs`; the e2e-only checks below key off the `e2e_` filename prefix, so **library sources never trip the sleep/polling rules**.

That exemption exists so `#[spec]` unit tests under `src/` are not subjected to e2e-only rules. The sleep ban is collateral: it means a `#[cfg(test)]` module inside `src/` can use fixed sleeps as synchronisation with nothing objecting, which is how 29 of them accumulated in one file while the identical pattern is a build failure two directories away.

## Suggested fix

1. **Replace the fixed sleep with a wait-until-observed poll** in `type_user_bytes`: wait, bounded, until the written bytes appear in `registry.snapshot(agent_id)`, and fail with the observed buffer if the bound expires. That removes the race rather than widening it, and turns the precondition assert into something that cannot fire spuriously. Raising 75 ms to a larger constant is **not** a fix — it moves the threshold and keeps the shape, which is the trap CLAUDE.md rule 5 records for #818 ("the bound is NECESSARY BUT NOT SUFFICIENT").
2. **Then survey the other 28 sleeps** in that module and convert the ones used as synchronisation. Sleeps that deliberately let a timeout elapse are legitimate and should stay — the distinction is whether the sleep is waiting for an event that has an observable completion.
3. **Consider whether the linkage-check exemption should be narrowed** so a `#[cfg(test)]` module under `src/` is covered by the sleep rule while non-test library code stays exempt. That is a separate decision with its own cost (29 existing violations would need addressing or annotating) and should not block fix 1.

## Not claimed

The failure has not been reproduced, so the 75 ms sleep is the *most likely* cause rather than a proven one. Anyone picking this up should reproduce first — CLAUDE.md's `reproduce-first` skill applies — by running `dispatch_020` under parallel load, or by temporarily lowering the sleep to force it. If it turns out to be something else at `4186`, the mechanism above is still a real latent flake and worth fixing on its own.

vfarcic:
Closes #850.

## The defect, reproduced before it was fixed

`dispatch_020_payload_guards_are_scoped_to_one_delivery` hard-red the **required** `build` job on PR #843 — a Renovate no-op whose diff was byte-identical to `main`'s `Cargo.lock`, so no dependency change could explain it. The panic was at a *precondition*, not the property under test:

```
panicked at src/spawn.rs:4186:9
precondition: the production-shaped bracketed multiline paste must physically reach the PTY
```

`type_user_bytes` wrote a user draft to a real PTY, flushed, and then slept a fixed 75 ms. Nothing waited for the bytes to become observable, so every caller's "must physically reach the PTY" assertion was a bet on a PTY round trip beating that constant. `.config/nextest.toml` sets `retries = 0` for `[profile.default]` and no scoped override covers `dispatch_018` or `dispatch_020`, so losing that bet once reddens a blocking check.

The issue was filed from a code read and one CI log, so the first deliverable was proof the defect exists:

| what was run | load | result |
| --- | --- | --- |
| unmodified `dispatch_020` | 3x CPU oversubscription, 20 runs | 20 pass |
| unmodified `dispatch_018` + `dispatch_020` | 4x, 25 runs | 25 pass |
| **sleep shortened to 0 ms** | **4x, 15 runs** | **15 FAIL** |

The race is real but rare, which is why luck was not the route. With the sleep at 0 ms the dominant panic is `src/spawn.rs:4186` with the exact CI message — `output="automatic payload before bracketed paste"`, i.e. the draft had not arrived at all. Two runs also failed at `src/spawn.rs:3173` (`retry_after_user_frame`'s draft precondition) and two at the final payload-safety assertion. That pins the causal chain rather than inferring it.

## The fix — three parts, none redundant

**1. Wait until observed, not until a clock.** The sleep is replaced by a content-keyed bounded poll: each maximal printable run of the payload's **text** must appear in `registry.snapshot(agent_id)`. `echo_text` strips terminal escape sequences first, so a landmark is text rather than protocol (see the Windows section below for why that matters). `type_user_bytes` now *asserts* rather than silently doing nothing when handed a payload with no text landmark, and `type_user_frame` covers the bare newline control frames — `Ctrl+J` encodes to `\n`, `Alt+Enter` to `ESC \r`, plain `Enter` to `\r` — which carry no text at all.

**2. `wait_for_quiescent_lines` waits out the byte target's copy — by observing, not predicting.** `completed_lines` counts two terminators per completed line, so a pane with nothing in flight always holds an **even** count, and a caller that drained the pane first need only wait for the count to become even again. That absorbs the platform difference instead of encoding it: Unix honours the `\n` inside a bracketed paste and produces two more terminators, Windows' ConPTY consumes the paste markers and produces none, and both are simply "even". Sound only because of part 3, and only for a write carrying at most one terminator — two would take the count from even straight to even — which both callers assert.

**3. Drain the pane before writing.** This surfaced only because part 1 turned a silent 75 ms bet into a loud 8 s diagnostic, and it is the part easiest to mistake for tidiness. `/bin/cat` and the echo are two independent writers into one PTY output queue, and the kernel commits echo in batches, so a pending `cat` copy can land **in the middle** of the write's echo:

```text
"automatic payload before plain Enter\r\n"   <- echo of the earlier line
"user"                                       <- this write's echo, cut off
"automatic payload before plain Enter\r\n"   <- cat's copy, interleaved
" turn completed with plain Enter"           <- the rest of this write
```

The scrollback is append-only, so that split is **permanent**: no amount of waiting reassembles the draft, and every `windows(draft.len())` search in the file — the tests' own preconditions included — fails for good. Each `type_user_draft` call site now states how many input lines precede it, and `wait_for_drained_lines` drains them first, leaving `cat` with nothing to copy until the write terminates a line. That helper is `wait_for_detached_delivery_attempt`'s existing poll renamed to the vocabulary of a caller counting *writes* rather than delivery attempts; the old entry point delegates to it, so its behaviour is unchanged.

**Raising the constant was not the route.** CLAUDE.md rule 5 records for #818 why that moves the threshold and keeps the shape — "the bound is NECESSARY BUT NOT SUFFICIENT". No nextest retry or `slow-timeout` override was added either: this is a fixable race, not inherent cost.

## Evidence the race is gone

All at 4x CPU oversubscription on 16 cores — the harness that failed 15/15 with the sleep removed:

| configuration | runs | failures |
| --- | --- | --- |
| all three parts | 40 | **0** |
| drain (part 3) alone disabled | 30 | 2 — at the final exact-equality assertion, not the precondition |
| quiescence wait (part 2) alone disabled | 30 | 8 — same assertion |
| landmark poll disabled (i.e. the pre-fix sleep at 0 ms) | 15 | 15 |

No part is redundant. The middle rows are worth reading: with only the landmark poll, the precondition failures are gone and what remains is a `cat` copy landing inside an `after == before` comparison window — a different assertion, which is why parts 2 and 3 are not cosmetic.

`dispatch_020` also got slightly faster (2.83 s vs 3.42 s before), since the waits return as soon as the bytes are observed instead of always paying 75 ms × 11.

## The first version failed `build-windows`, and why that is recorded here

**The first push passed `build` and failed `build-windows` on the same commit** — the same class of error as the bug being fixed: a platform assumption written as if it were universal. That version cut its landmarks from the *raw* payload, so the bracketed paste's landmark was `[200~first line of an unsent draft`. Unix's line discipline knows nothing of the marker and echoes its printable tail literally; ConPTY **parses** it and echoes nothing, so the landmark could never match on Windows. The callers' own preconditions had only ever looked for the draft text. The observed Windows snapshot:

```
"...paste\x1b[39X\r\n...paste\r\nfirst line of an unsent draftsecond line of an unsent draft"
```

Windows also honoured no line terminator inside the paste, so the "twice per completed line" arithmetic that version predicted was Unix-specific too — hence the observed even-count rule in part 2, which needs no platform knowledge.

**One assumption is left explicitly unverified rather than guessed a second time.** Whether ConPTY echoes a *bare* control frame is not something this change has evidence for — the Windows run that caught the bug panicked at the paste, before reaching any frame — so `type_user_frame`'s echo proof is `#[cfg(unix)]` and Windows keeps the 75 ms sleep it already had, with the quiescence wait additive on top. `scripts/windows-cross-check.sh` type-checks the workspace for `x86_64-pc-windows-msvc` from Linux with **no errors and no new warnings**.

## Survey of the other sleeps, per the issue's step 2

The issue says `src/spawn.rs`'s test module holds 29 `sleep` occurrences. **29 is the file total: 25 were in the test module and 4 are production timers** outside it (`src/spawn.rs:1187/1262/1267/2668` — delivery buffer and debounce waits, not test synchronisation). Nothing here converts the remaining ones; the classification is the deliverable.

**Same defect — a fixed sleep guarding a POSITIVE observation that has an observable completion (6):**

| line | sleep | what it guards |
| --- | --- | --- |
| 3113 | 250 ms | `observe_dispatch_016_rearm_case`'s `after_replay` delta. A short read silently *shrinks* the delta rather than panicking, so this one fails as a false negative. `wait_for_detached_delivery_attempt` is already used twice in the same function. |
| 3904 | 750 ms | asserts `SPAWNED_PROMPT` **did** arrive (#570's late-native-claim case). |
| 4087 | 75 ms | asserts `REPLACEMENT_PROMPT` "must physically reach the detached pane" — literally #850's shape. |
| 4167 | 800 ms | asserts attempt 2's `PROMPT` reached the target before the user types. |
| 4245 | 75 ms | `dispatch_019`: asserts the draft "must already be physically visible" — an inline copy of the write-then-sleep pattern this PR removed from `type_user_bytes`. |
| 4314 | 75 ms | asserts `after_delivery_b.len() > before_delivery_b.len()`. |

(Line numbers are pre-change, as the issue's are.)

**Mixed polarity — one sleep, two contracts (1):** line 3383, `retry_after_user_frame`'s post-retry window. `plain_enter_retry` asserts `after.len() > before.len()` (positive, convertible); `ctrl_j` and `alt_enter` assert `after == before` (negative, must stay). Splitting it means telling the helper which polarity its caller wants.

**Correct and must stay — deliberate negative-observation windows (12):** lines 3584 (50 ms), 3627/3701/3788 (550 ms), 3850 (750 ms), 4133 (800 ms), 4183 (1200 ms), 4262 (75 ms), 4373 (75 ms), 4421 (75 ms), 4544 (75 ms), 4693 (50 ms). Each asserts that **nothing** arrived, so the sleep *is* the observation window and there is no completion event to poll for. They are sized against production's 500 ms `unconfirmed_retry_delay(1)`.

**Not a wait at all (1):** line 4773 (200 ms) burns 200 ms of a 300 ms deadline so the confirmation task starts against an almost-expired one. The sleep is the fixture input.

**Legitimate but fragile, a different class (1):** line 3575 (100 ms) must land *inside* production's 500 ms retry window — "Rebind while it is waiting, before the 500ms retry is resolved". "The retry has not yet resolved" has no observable completion, so a poll cannot replace it; making it deterministic needs `tokio::time::pause`, as `dispatch_021` does. Under enough load 100 ms can overshoot 500 ms and flip which case is exercised. Worth its own issue.

**Poll intervals inside bounded content-keyed waits (3, now 5):** backoff inside the wait loops themselves, legitimate by construction.

That is 6 + 1 + 12 + 1 + 1 + 3 = 24 of the 25, plus the one converted here. The test module now holds 27: 25, minus the two removed on the observed path, plus two new poll intervals and the Windows-only frame sleep.

The issue's step 3 — narrowing `xtask/linkage-check`'s sleep-rule exemption for `src/` — is deliberately **not** done here. It would surface all of the above at once and is a separate decision; left to the issue.

## Rules and scope

- **Rule 12**: every hunk is inside `#[cfg(test)] mod tests` (which starts at `src/spawn.rs:2678`; the earliest hunk is at 2910), so no production code changed, the TUI↔daemon contract is untouched, `PROTOCOL_VERSION` needs no bump and no `.breaking.md` fragment applies. No cross-version manual test is required.
- **Rule 4**: no user-visible TUI behaviour changed, so no new TUI test. `tests/CATALOG.md` is unchanged — no test was added or renamed, and `scheduler/dispatch/018` and `/020` still assert exactly what they did; only the fixture's synchronisation changed.
- **Rule 9**: no new user-facing surface, so no experimental-flag question.

## Tests run

- `cargo test-fast` — **4015 passed, 0 failed**
- `cargo nextest run --workspace -E 'test(dispatch_020) or test(dispatch_018) or test(dispatch_016)'` — 3 passed (`dispatch_016` included because it consumes the renamed poll's return value)
- `cargo nextest run --workspace -E 'test(~spawn::tests::)'` — all **42** `spawn::tests` passed
- `cargo fmt --check` — clean
- `cargo xtask linkage-check` — ok (621 catalog ids, 551 annotations, 100 allowlisted, 11 rules)
- `scripts/windows-cross-check.sh` — `cargo check --workspace --tests` for `x86_64-pc-windows-msvc`, no errors
- plus the load harness above: 40 runs of `dispatch_018` + `dispatch_020` at 4x oversubscription, 0 failures

### On `e2e-deterministic`

The first CI run showed it failing on `orchestration_remit_001`, `orchestration_remit_007` and `idle_worker_008` — the #818 set. It is **not** a required check. One correction to the framing this task was given, though: the lane is **intermittent on `main`, not permanently red** — `main`'s two most recent runs are `de719d17` **success** and `01537583` failure.

Either way this change cannot cause those three. Verified rather than asserted: `grep` finds **zero** references from anywhere under `tests/` to any helper this PR touches, the two files holding the failing tests are not in this branch's diff, and the change lives in the library's `#[cfg(test)] mod tests`, which integration-test binaries never compile.


vfarcic:
## Final CI state

Everything green except the non-required `e2e-deterministic` lane:

| check | result |
| --- | --- |
| **build** (required) | pass — 6m12s |
| **build-macos** (required) | pass — 5m23s |
| **build-windows** (required) | pass — 5m49s |
| **security** (required) | pass — 25s |
| nix | pass — 7m17s |
| windows-cross-check | pass — 1m58s |
| devbox / desktop-web / changes / label | pass |
| Greptile Review | pass — 5/5 confidence, **zero inline findings** |
| e2e-deterministic (not required) | fail — see below |

`build-windows` passing is the point of the second commit: the first push passed `build` and failed `build-windows` on the same commit, because the landmarks were cut from the raw payload and so included the bracketed-paste marker's printable tail — literal on Unix, parsed and swallowed by ConPTY. The mechanism is now observed rather than predicted.

### `e2e-deterministic`

**1 failed of 9386** (`427.6s`): `orchestration_remit_001_start_role_compaction_reasserts_remit` — the test CLAUDE.md rule 5 and issue #818 name explicitly as the known-failing one. Round 1 of this PR failed on three (`orchestration_remit_001`, `orchestration_remit_007`, `idle_worker_008`); round 2 is down to the named one, on an unchanged diff apart from the Windows portability commit — which is itself the signature of the load-dependent contention #818 tracks.

Worth correcting one framing, since a reader may have the opposite impression: this lane is **intermittent on `main`, not permanently red**. `main`'s two most recent runs are `de719d17` **success** and `01537583` failure.

Either way this change cannot cause it, and that is verified rather than asserted:

- `grep` finds **zero** references from anywhere under `tests/` to any helper this PR touches (`type_user_bytes`, `type_user_draft`, `type_user_frame`, `echo_text`, `echo_landmarks`, `printable_runs`, `line_terminators`, `wait_for_drained_lines`, `wait_for_quiescent_lines`, `wait_for_echo_bytes`, `wait_for_growth`);
- `tests/e2e_orchestration_remit.rs` and `tests/e2e_idle_worker_detector.rs` are not in this branch's diff (`src/spawn.rs` and `changelog.d/850.misc.md` are the only two files);
- every hunk is inside the library's `#[cfg(test)] mod tests`, which integration-test binaries never compile.

prageethw (approved):
Test-only change (all inside `#[cfg(test)] mod tests`), well-evidenced with measured before/after failure rates (0/40 with both parts, 2/30 and 8/30 with each disabled). Content-keyed waits replace the fixed 75ms sleep soundly; the Windows/Unix asymmetry (echo_text stripping bracketed-paste markers, `#[cfg(unix)]` frame proof) is reasoned through rather than assumed. The only failing check, `e2e-deterministic`, fails on `orchestration_remit_001_start_role_compaction_reasserts_remit` — unrelated to this PR's `src/spawn.rs` test fixtures and already the subject of #837.

Issue #862, opened 2026-09-04
PRD #386 M5 follow-up: shell-activity poll overhead measurement and cadence (deferred 2026-08-06)


vfarcic:
PRD #386 deferred M5 to a follow-up issue on 2026-08-06 and that issue was never filed. This is it, and it arrives with the field measurement M5 asked for.

## M5 as written

> **M5 — Overhead measurement and cadence. DEFERRED** ... Measure the poll's cost with a realistic process table and pane count; confirm or revise the cadence; decide Route A vs Route B on the number, and record the number in this PRD.

and the rationale behind it:

> A `ps -A` per second on a busy machine is not free, and the deck's own workers routinely drive load averages above 20. M5 exists to put a number on the whole thing before this ships.

That prediction held exactly.

## The measurement

Taken on the `dot-agent` box during a genuine multi-agent build storm: load average 24-28 on 16 cores, PSI `io full avg300=65.95`, ~440 processes / ~1300 threads, 8 live panes.

| | value |
| --- | --- |
| `ps -A` sample, idle baseline (per the code comment in `src/platform/proc/unix.rs`) | ~49 ms |
| `ps -A` sample, under load | **19-20 s** |
| `SAMPLE_TIMEOUT` (`src/daemon.rs:1330`) | 2 s |
| `MAX_TABLE_AGE` (`src/daemon.rs:1381`) | 3 s |
| `POLL_INTERVAL` (`src/daemon.rs:1325`) | 500 ms |
| `shell-activity` warnings in one day of `deck.log` | **1716** |

Representative lines:

```
WARN shell-activity: process-table sample overran its deadline; leaving every pane's status
     alone and continuing to wait on the SAME sample ... timeout_ms=2000 panes=8
WARN shell-activity: discarding a process-table sample that answered too late to trust;
     leaving every pane's status alone ... age_ms=20340 max_age_ms=3000
```

The daemon's behaviour here is **correct** — it declines to classify against a stale table rather than misattributing a reused pid. The defect is that under sustained load the signal stops working entirely and the log fills with 1716 warnings saying so.

## A caveat on Route B

M5 frames the decision as Route A (`ps` per poll) vs Route B (native enumeration). I do not think Route B addresses this particular stall, and this issue should not be picked up assuming it will.

[...]

PR #872, opened 2026-09-04
test(e2e): re-read the grid when locating a target an input event painted (#807, #395)

### LINKED ISSUE #395: e2e: find_in_grid never retries, so find_in_grid(..).expect(..) races the render — scheduler form_002 now fails 3/3, and ~51 call sites share the footgun

### LINKED ISSUE #807: Flaky lane-1 e2e: single-shot find_in_grid after an input event races the repaint (form_002, plus 36 more in lane 1)
## Symptom

`e2e-deterministic` (the new lane-1 per-PR job from #502) failed **1 of 8017** on PR #779:

```
thread 'form_002_add_spawns_authoring_agent_in_picked_dir' panicked at tests/e2e_scheduler_manager.rs:931:10:
the mode-locked schedule form must render a [Submit] button
```

Triaged as a **pre-existing, load-sensitive flake** — unrelated to the PR it appeared on. **35/35 passes** locally across both that branch and `origin/main`, including 10 runs at 4x CPU oversubscription and two whole-module runs. Never reproduced solo, however loaded.

## Mechanism — already diagnosed and fixed once in this repo

`tests/e2e_scheduler_manager.rs:929-931` reads the grid **once**, immediately after an input event:

```rust
deck.send_keys(b" ");                 // Space -> mode-locked schedule form
deck.wait_for_string("New Schedule"); // snapshot A
let (scol, srow) = deck
    .find_in_grid("[Submit]")         // snapshot B - separate read
    .expect("the mode-locked schedule form must render a [Submit] button");
```

Ratatui flushes a frame as a byte stream; the harness reader thread consumes it in chunks. Under PTY contention one frame's bytes can split across reads, so snapshot B can land after the title cell is painted and before the button row — which this form paints in a **second pass**, `render_modal_button_row` drawing over the reserved blank line after the `Paragraph`. The preceding `wait_for_string` narrows the window but cannot close it: the wait and the lookup take two separate snapshots.

This is not inference. `TuiDeck::wait_for_in_grid`'s doc comment (`tests/common/mod.rs:1755`) says it:

> Prefer this over a bare `find_in_grid(..).expect(..)` whenever the lookup follows an input event (a click, a keystroke) rather than a wait that already proved the target is painted. A single-shot read landing mid-repaint sees a transiently cleared region and the `expect` fires — a load-sensitive flake, not a real failure.

That helper was added by **`6480f9d`** (2026-08-27) to fix `buttonbar_007`, whose fingerprint matches line for line:

| | `buttonbar_007` (fixed) | `form_002` (this issue) |
|---|---|---|
| failed inside a tier run at | 4.831s | 6.382s |
| passes solo at | 0.659s | ~1.3s |
| solo reruns survived | 22/22 | 35/35 |

Two further facts from the triage worth keeping:

- **Duration alone does not decide it.** One `origin/main` solo run took **8.097s and passed** — longer than CI's failing 6.382s. The race is narrow and orthogonal to how slow the test is.
- **Solo runs structurally cannot reproduce it.** CPU starvation is not the trigger; a solo run has one deck writing one PTY, so there is no competing writer to fragment the flush. It needs many PTY decks interleaving in one nextest pool — which is exactly what the new job does on `ubuntu-latest`'s 4 vCPUs.

`.config/nextest.toml:25` sets `retries = 0` globally and `form_002` has no override, so one bad frame is fatal.

## Fix

**1. The immediate one-liner** — use the helper that already exists for this:

```rust
let (scol, srow) = deck.wait_for_in_grid("[Submit]");
```

No production code, no new dependency, and it buys a grid dump on genuine failure — which CI lacked here, because `.expect()` dumps nothing while `wait_for_in_grid` does.

**2. Do NOT reach for `retries = 1`.** `.config/nextest.toml:208` permits retries only for "load that is not the tier's" and calls it "a bounded concession… not a rerun-until-green". This load **is** the tier's, and the harness can close the race outright.

## The sweep — why this will recur

**44 bare single-shot `find_in_grid(..).expect(..)` lookups remain across `tests/e2e_*.rs`, 36 of them in lane 1** — i.e. in the job that now runs on every PR:

```
12  tests/e2e_scheduler_manager.rs      4  tests/e2e_mouse_dashboard.rs
 6  tests/e2e_mouse_preserve.rs         3  tests/e2e_mouse_picker.rs
 4  tests/e2e_mouse_tabstrip.rs         + 13 more lane-1 files at 1-2 each
```

36 is the upper bound on exposure rather than a count of live defects — only lookups that follow an **input event**, rather than a wait that already proved the target painted, carry the race. But `e2e_scheduler_manager.rs` alone holds 12, and a job sampling them on every PR will keep finding them.

Expect more single-test reds of this exact shape until the pattern is swept.

## Context

Found on PR #779 (PRD #745). Lane 1 is advisory, not one of the four required checks, so it did not block that PR — which is the disposition rule 8 anticipated when it noted lane 1's runner stability is unmeasured and "its first ever run is the one on the PR that introduces it". This is the first instance of that anticipated noise.

vfarcic:
Closes #807. Closes #395.

Both issues describe the same mechanism from two angles — #807 has the diagnosed failure and the matched fingerprint against `buttonbar_007`, #395 has the call-site census — so they close together.

## The mechanism

`find_in_grid` reads the terminal grid **once**. Ratatui flushes a frame as a byte stream and the harness reader thread consumes it in chunks, so under PTY contention one frame's bytes split across reads and a single-shot lookup lands mid-repaint — seeing a transiently cleared region, firing the `expect`, and reddening a lane-1 job for no real defect. `wait_for_string` in front narrows the window but cannot close it: the wait and the lookup take two separate snapshots.

The fix already existed. `TuiDeck::wait_for_in_grid` was added by `6480f9d` (2026-08-27) for `buttonbar_007` and polls the lookup itself. Only **3** sites used it.

## What was classified, and how

A same-line grep finds one site; the statement almost always spans three lines. A multi-line scan (skipping comment lines) over `tests/**/*.rs` finds **52 panic-on-miss `find_in_grid` sites across 21 files** — `.expect(`, `.unwrap(`, or `unwrap_or_else(|| panic!(..))`. That is the same population the issues describe; #807's "44 bare, 36 in lane 1" counted only `.expect(`/`.unwrap(`, and #395's "~51" is within one of this.

Every one was read against the discriminator on `wait_for_in_grid`'s own doc comment:

| bucket | n | disposition |
| --- | --- | --- |
| the preceding wait proved a **different** string than the one located | **39** | **migrated** |
| the preceding wait proved the **same** target | 10 | left — converting adds noise, removes no race |
| in `e2e_shell_activity_real_agent.rs`, being edited for #862 | 3 | deferred, see below |

Two further `find_in_grid` calls (`e2e_session_restore.rs:403-404`) are `.map()`-ed into an `Option` comparison rather than unwrapped, so they are not this bug and are untouched. Nothing that asserts absence was converted: waiting for something that should be absent would hang to `WAIT_TIMEOUT`.

Migrated, by file: `e2e_scheduler_manager` 9/12, `e2e_mouse_preserve` 3/6, `e2e_mouse_tabstrip` 3/4, `e2e_mouse_dashboard` 3/4, `e2e_mouse_form` 2/2, `e2e_mouse_buttonbar` 2/2, `e2e_pane_close` 2/2, `e2e_login_path` 2/2, `e2e_dispatcher_mode` 2/2, `e2e_mouse_picker` 1/3, and one each in `e2e_mouse_inline`, `e2e_mouse_modal`, `e2e_render_contract`, `e2e_prompt_delivery_confirmation`, `e2e_pane_input_live`, `e2e_new_pane_seed`, `e2e_new_pane_schedule_option`, `e2e_mode_watch_stream`, `e2e_mode_tab_layout`, `e2e_card_layout`.

Four of the 39 are the `click_target` / `click_button` helpers in the mouse-parity files, so one edit each covers 17 call sites between them.

## The mechanism-level evidence

Not "I ran it and it was green" — #807 records that trap explicitly (one `origin/main` solo run took 8.097s and passed, longer than CI's failing 6.382s), and it never reproduced solo however loaded.

Instead the retry property is now pinned directly. `TuiDeck::find_in_grid`'s scan and `wait_for_in_grid`'s polling are split out as `grid_find(grid, needle)` and `poll_grid_for(read_grid, needle, timeout, interval)`, so a test can hand them a scripted frame sequence with no PTY. Five tests in `tests/common/harness_unit_tests.rs` drive a two-frame schedule-form repaint — `MID_REPAINT` has the `New Schedule` title and a blank reserved button line, `SETTLED` has both:

- `a_mid_repaint_frame_shows_the_title_without_the_button_row` — the wait's needle is present and `[Submit]` is `None` on the same frame. That is the bare `find_in_grid(..).expect(..)` aborting, reproduced deterministically.
- `polling_the_lookup_survives_a_target_that_appears_after_the_first_read` — the same frame sequence through `poll_grid_for` returns the settled frame's real coordinates, and asserts it read more than once.
- `the_polling_lookup_gives_up_with_the_grid_it_last_read` — a timeout returns the last grid for the panic message (a diagnostic `.expect()` never produced), having retried on the way.
- `an_already_painted_target_costs_exactly_one_read` — the common case costs nothing; the poll interval is deliberately longer than the deadline, so a second read would blow the test's own budget.
- `the_returned_column_counts_characters_not_bytes` — pins the char-vs-byte offset arithmetic that moved into `grid_find`.

Then the migrated tests were run to show they still assert what they did.

## What this does and does not do for #818

`e2e-deterministic`'s failing set rotates. This should remove the repaint-race family from it: **`form_002`** (#807's own), **`form_006`**, **`manager_016`**, and the **`inline_001` group** — all five `inline_001_*` tests drive their buttons through `click_button`, including `inline_001_rename_save_commits`, which #818 names.

It cannot touch:
- **`orchestration_remit_001` / `_007`** — PR #837 is open for those, and `tests/e2e_orchestration_remit.rs` is untouched here.
- **`work_done_004`** — `tests/e2e_work_done_reporting.rs` contains no `find_in_grid`.
- The contention and starvation class in **#351** and **#415**, which is a different mechanism.

#818 stays open. So do #351, #415, #247, #240 and #656.

## Collision handling

`tests/e2e_shell_activity_real_agent.rs` has 3 racy sites and is being modified by the `issue-862` agent. No PR exists for #862 yet, so **I left the file alone** rather than merge against work that has not landed. #873 tracks its three sites, with the exact lines and the one-line fix.

## Not a contract change

Tests and test harness only — no `src/` change, no protocol, daemon, orchestration or hook change, so rule 12's cross-version check does not apply.

## Test commands run

Local, all green:

- `cargo fmt --check`
- `cargo clippy --workspace --all-targets --features e2e,e2e-live -- -D warnings` — all four flags, so both lanes are type-checked.
- `cargo test-fast` — **2406 passed**, full workspace, desktop crate included. The 113 `harness_unit` tests run here, so the five new ones are in this number.
- `cargo nextest run --workspace --test harness_unit -E 'test(/grid|repaint|read|characters/)'` — 12 passed, covering all five new tests by name.
- `cargo nextest run --features e2e,e2e-live --test-threads 4 -E '<the 16 migrated lane-1 binaries> + binary(harness_unit)'` — **188 tests across 17 binaries, 188 passed**.
- `cargo nextest run --features e2e,e2e-live --test-threads 2 -E 'binary(e2e_scheduler_manager)'` — **20/20**, the heaviest migrated file, including `form_002`, `form_006` and `manager_016`.

**Lane 2, which runs nowhere in CI** (rule 5), so writing the migration is only half the job. Each of the 5 migrated credentialed sites has a green run against a real agent:

| file | migrated site | green test |
| --- | --- | --- |
| `e2e_card_layout.rs` | 1 | `card_stats_005_real_agent_card_narrows_without_restructuring` — 31.5s |
| `e2e_pane_input_live.rs` | 1 | `pane_input_022_ctrl_w_does_not_tear_down_interactive_claude` — 12.5s |
| `e2e_prompt_delivery_confirmation.rs` | 1 | `dispatch_014_concurrent_swallowed_seeds_retry_until_confirmed` — 14.1s |
| `e2e_dispatcher_mode.rs` | 2 | `orchestration_dispatch_004_list_targets_marks_the_declared_default` — 9.6s (the `open_cat_caller_pane` helper) and `new_pane_016_dispatcher_opens_dashboard_card_with_real_agent` — 26.2s |

CI on this commit: **all 12 checks green**, including the four required ones and `e2e-deterministic`, which is `cargo test-e2e` — lane 1 in full — at **2594 tests run, 2594 passed** in 314s. That job is the broad lane-1 pass rule 5's shared-harness exception asks for, and it is broader than a local run of the same command (CI's `--workspace` includes the desktop crate).

Greptile: pass, 5/5 confidence, **no inline comments** on `repos/.../pulls/872/comments`.

## One thing worth knowing about the local runs

Several local attempts failed before this settled, and the cause is not this branch. Every PTY-deck-launching e2e test on the dev box was failing at the *first* wait with `did not see "No active sessions" within 10s` and an **empty** grid — the lookup is never reached. Measured with a properly-sized PTY: the deck took **15.0s** to emit its first byte (a daemon-connect failure), against the harness-wide 10s `WAIT_TIMEOUT`. It is the lazy daemon spawn of a 196 MB debug binary while three agents thrash ~5 GB of swap.

Three things make it clear this is the box and not the diff: the identical symptom hits `tests/e2e_dashboard_selection.rs`, which **this branch never touches**; `git diff origin/main -- src/ desktop/` is **empty**, so the binary is byte-identical to `main`'s; and the same tests pass in CI on this commit. Reruns alternate pass/fail on an unchanged tree — `dispatch_014` went FAIL, PASS, FAIL — which is the coin-flip signature #818 records. This is the second marginal 10s bound that #818's fifth comment already names, and the contention class of #351/#415. Not addressed here, and not made worse by this change.


prageethw (dismissed):
Verified the core refactor in tests/common/mod.rs (grid_find / poll_grid_for split, with the discriminator documented on wait_for_in_grid's doc comment) and the new harness_unit_tests.rs coverage (mid-repaint vs settled frame, retry-until-timeout, already-painted fast path, UTF-8 column offset). Spot-checked several of the ~39 migrated call sites (buttonbar_003/004, etc.) — the .expect(...)/.unwrap() → wait_for_in_grid(...) conversions are mechanically correct and match the stated discriminator (converted only where the preceding wait proved a different string). e2e-deterministic is clean on this PR. LGTM.

vfarcic:
Heads-up on one test in this PR's victim list, found while clearing the review backlog on #895 — **this PR does not actually fix `manager_016`**, and I would rather flag it than let it look fixed.

`changelog.d/807.misc.md` lists the family it removes as "`form_002`, `form_006`, **`manager_016`**, the `inline_001` group". The first three and `inline_001` look right. But this PR's only change inside `manager_016_wheel_over_dialog_does_not_scroll_side_pane` is the `[Submit]` lookup:

```
@@ -731,9 +727,7 @@ fn manager_016_wheel_over_dialog_does_not_scroll_side_pane() {
-    let (submit_col, submit_row) = deck.find_in_grid("[Submit]")
+    let (submit_col, submit_row) = deck.wait_for_in_grid("[Submit]");
```

That lookup runs **before the Scheduled Tasks dialog is ever opened**. The read that actually fails is later and is not a `find_in_grid` at all:

```rust
deck.send_keys(b"s");
deck.wait_for_string("NEXT FIRE");
let before = visible_side_scroll_markers(&deck.snapshot_grid());   // single-shot
```

`NEXT FIRE` is the dialog **header**, so it lands before the overlay interior paints. The centered dialog paints *over* the mode side pane and occludes marker rows, so a mid-paint `before` holds a marker set the settled frame does not, and the later comparison fails with no defect present.

So this is the **same race as #807/#395 reaching a raw `snapshot_grid()` comparison** rather than a `find_in_grid` — outside this PR's sweep by construction. Your `--retries 1` would mask it rather than fix it, so `manager_016` would most likely have kept failing after this merged and looked like a regression in your fix.

**Measured rather than reasoned about:** inducing a genuine wheel leak (disabling both wheel swallows while leaving the selection moving) shifts every marker by 3 and keeps the occluded band **11 rows wide**. The observed CI failure kept its band *start* fixed and grew the band **9 → 11**. A pane scroll cannot change the band width — only the overlay's painted extent can — so the failing read caught the dialog two rows short of settled.

Tracked as **#936**, fixed in **#937**. Nothing is needed from this PR: I test-merged #937 against this branch and it is clean, and the composed tree compiles with **21/21** `e2e_scheduler_manager` tests passing — `manager_016` ends up carrying both fixes. `tests/common/mod.rs` is untouched by #937, deliberately, so it cannot collide with your harness rewrite.

Two suggestions, both optional and neither blocking:

1. Consider softening the `manager_016` mention in `changelog.d/807.misc.md`, since as written it claims a fix this PR does not deliver.
2. Worth a sweep for other raw `snapshot_grid()` comparisons taken **across an overlay paint** — the `find_in_grid` sites are the bulk of the problem, but they are not the whole of it. I noted that as a follow-up in #936 rather than widening #937.

Lane-wide status and the merge order across #872/#933/#937/#895 are in #818, so there is one place to read it: https://github.com/vfarcic/dot-agent-deck/issues/818#issuecomment-5574871550

This PR's own five required checks are green with zero unresolved threads, so it needs only review.

vfarcic:
Correction to my note above: **#937 is closed** and the `manager_016` fix now lives in **#895** (`954f941d`), because keeping it as its own PR forced an extra serialized approval round on #895 (`dismiss_stale_reviews_on_push` voids an approval when `main` is merged in).

Nothing changes for this PR — it still needs nothing from me, and #895 still merges cleanly with it (verified again after the fold: **21/21** `e2e_scheduler_manager` on the composed tree, `manager_016` carrying both fixes).

**The two optional suggestions from my earlier comment still stand, and one has grown slightly more relevant:**

1. `changelog.d/807.misc.md` still claims this PR removes `manager_016` from the failure family, which it does not. Worth softening — and it now matters a little more, because if #895 stalls (it is a breaking `PROTOCOL_VERSION` 8→9 change), the sensible move is to cherry-pick `954f941d` onto **this** PR, at which point the claim would become true.
2. The raw `snapshot_grid()`-across-an-overlay-paint sweep is still open in #936, and belongs with this PR's family rather than with a protocol PR.

Also worth recording here since it is your family: **`form_002_add_spawns_authoring_agent_in_picked_dir` failed on #937's CI run** at `tests/e2e_scheduler_manager.rs:979` — the single-shot `find_in_grid("[Submit]").expect(...)` that this PR replaces with `wait_for_in_grid`. That test did not appear in my earlier 28-run survey, so this PR's family bites more often than that survey suggested. `manager_016` passed on that same run at 1.121s, the first CI evidence the settle holds.

vfarcic:
Fixed in `2937b99`. The changelog fragment no longer claims this removes `manager_016`; it now says explicitly that `manager_016` is a distinct `snapshot_grid()`-across-overlay-paint race tracked in #936 and fixed separately. `cargo fmt --check` and clippy both clean.

vfarcic:
New live evidence for this change, recorded in the fragment as `29852f4a`.

#990 — a Renovate PR whose entire diff is one GitHub Actions version bump — failed `e2e-deterministic` at 2026-09-10T02:04Z on `preserve_modal_click_miss_is_consumed`, at `find_in_grid("[Cancel]")`, `tests/e2e_mouse_preserve.rs:191`. That is precisely the call site this PR rewrites to `wait_for_in_grid`. Since the branch touches no Rust at all, the test cannot have been broken by its own change — which makes this the cleanest evidence in the fragment for the family being real and this being the fix.

Run: https://github.com/vfarcic/dot-agent-deck/actions/runs/34427420793

Issue #526, opened 2026-08-12
Confirm a submitted prompt by digest instead of a 200-byte prefix


vfarcic:
Split out of issue #424's review (auditor MEDIUM, second half of "A spoofed, stale, or merely colliding hook event can falsely confirm an undelivered prompt"). Ruled out of scope for the #424 fix batch because adding a field to `AgentEvent` is a wire change.

`crate::hook` truncates a reported `user_prompt` to `USER_PROMPT_MAX_LEN` (200 bytes) before it ever reaches `AgentEvent`, so `prompt_delivery::prompt_submission_matches` must accept the truncated form. Two DIFFERENT prompts sharing their first 200 bytes therefore produce the same hook report: an expected `prefix(200) + task-A` is confirmed by a submitted `prefix(200) + task-B`. Dispatch and orchestrator prompts routinely share long boilerplate preambles, so the shared-prefix case is not exotic.

The #424 batch narrowed this rather than closing it. A false confirmation now additionally requires the same pane, an exact non-optional registry agent id, and an event that postdates the delivery's causal watermark — i.e. the *same agent* submitting a different prompt sharing the first 200 bytes inside the delivery's own window. The residual is documented at the comparison site.

Remedy: carry a collision-resistant digest of the FULL submitted prompt (computed by the hook before display truncation) on the event, and compare that instead of the prefix. Additive optional field, but it is still a wire change and wants a `PROTOCOL_VERSION`/`AGENT_EVENT_SCHEMA_VERSION` decision plus a fallback for producers that do not send it.

PR #874, opened 2026-09-04
fix(prompt-delivery): log a duplicated turn distinctly from a clean confirmation (#685)

### LINKED ISSUE #685: A duplicated prompt turn is reported as a clean confirmation in 3 of 4 repetition shapes

vfarcic:
Closes #685.

## What this is

Log-only. **No behaviour change.** Every shape that confirmed a delivery before still confirms it, every shape still terminates the retry loop, and no delivery decision reads anything added here.

`prompt_submission_accumulated` fired its `warn` in **one** of the four repetition shapes the issue measured, and in **none** of them for a prompt over `USER_PROMPT_MAX_LEN` — which a `dispatch` or issue-dispatch prompt routinely is. So a duplicated turn reached the operator as an ordinary `info` success.

## The classification

`classify_prompt_submission` is `prompt_submission_matches`'s body with the `true`s labelled; `prompt_submission_matches` is now `.is_some()` over it. Same accept set, same evaluation order, same short-circuit — the matcher was **not** widened and **not** narrowed.

| shape | when | level | line |
| --- | --- | --- | --- |
| `SingleCopy` | report is the prompt verbatim | `info` | unchanged byte for byte |
| `RepeatedCopies` | repetition is VISIBLE in what arrived | **`warn`** | "...arrived as repeated newline-separated copies; the turn the agent is acting on states the task more than once" |
| `TruncatedPrefix` | report is a 200-byte prefix | `info` | "...reported only as a truncated prefix that one copy and any number of copies render identically; whether the turn was duplicated is undecidable from it" |

Each line carries a greppable `confirmation="single-copy" | "repeated-copies" | "truncated-report"`, and all three keep the `prompt delivery confirmed by the agent's submitted prompt` prefix that existing delivery-log readers key on.

### Knowable vs unknowable

The two are deliberately not one line. `RepeatedCopies` states a fact; `TruncatedPrefix` states that the question is unanswerable. The weaker one is not dressed up as the stronger — a line claiming a duplicate that may not have happened is worse than the current silence — and it is **not** a `warn`, because it is emitted for every long-prompt confirmation and warning on it would be the #818 shape in a new place. It stays at `info` rather than dropping to `debug` so the line an operator gets today does not vanish below the default level.

`RepeatedCopies.copies` is `Option<u32>`: `Some(n)` exactly when the report carried the whole repeated text. It is deliberately `None` for the truncated match — the loop counter there is the first candidate that outgrew the report and is **not** a floor on what was submitted (measured: a 67-byte prompt submitted 3 times matches at counter 4). A wrong number in a warning is worse than no number.

### Determinism

`ui::prompt_submission_evidence` scans a whole journal across a `HashMap` of sessions, so it no longer returns on the first confirmation — it takes the most notable shape (`ConfirmedSubmission::more_notable`). The evidence returned is unchanged: any confirmation still beats any accumulation. Without this, which line got logged depended on hash order.

## The two pinned tests are unmodified

Neither `repeated_copies_of_the_seed_count_as_delivered` nor `bare_concatenation_is_accumulation_and_not_confirmation` appears in the diff except as a prose reference in a doc comment. Both still call the `bool` API and both pass — that is the evidence the accept set did not move.

**No contract change.** `PromptWatch` is `pub(crate)` and never serialized; `SubmissionEvidence` is private to `ui.rs`; `src/daemon_protocol.rs` is untouched and `PROTOCOL_VERSION` is unchanged; no wire field changed shape or meaning. The only new artifact is text in the per-process log file, which crosses no process boundary. No `.breaking.md`; `bugfix` fragment, patch bump.

## Tests

- `cargo test-fast` — full workspace.
- `prompt_delivery::tests::*` (23, all) — including the two pinned ones unmodified.
- New: `confirmation_classification_covers_every_repetition_shape` — the issue's own fixture set (both repetition shapes × 2 and 3 copies × short and >200-byte, each through `crate::hook`'s truncation function), asserted against the classification and cross-checked against the `bool` matcher per row.
- New: `confirmation_log_lines_distinguish_the_three_shapes` — captures real `tracing` output through a subscriber and asserts level and message per shape, that the undecidable line is not a `WARN` and does not claim a duplicate, and that all three lines differ.
- New: `a_visible_duplicate_outranks_a_clean_confirmation_in_the_log`.
- `ui::tests::pane_input_023/024/025/026/027/028` — the delivery-confirmation consumers, covering the `prompt_submission_evidence` restructure.

## Not this issue

`#526` is the real closure: a digest of the full prompt computed before display truncation removes the `TruncatedPrefix` cell entirely. This makes it visible, not answerable.


prageethw (approved):
Reviewed the diff, changelog fragment, and new tests. The accept set is genuinely unchanged — `prompt_submission_matches` is now `.is_some()` over `classify_prompt_submission`, and both pinned tests (`repeated_copies_of_the_seed_count_as_delivered`, `bare_concatenation_is_accumulation_and_not_confirmation`) are untouched and still call the bool API, which is the right way to prove a pure observability change didn't widen or narrow anything. The knowable/unknowable split (RepeatedCopies vs TruncatedPrefix) is well-reasoned — not claiming a duplicate the evidence can't support is the right call, and keeping the undecidable case at info rather than warn to avoid an #818-shaped flood makes sense. `more_notable`'s tie-break avoids hash-order dependence in the journal scan. Rule 12 checked: no wire/protocol change, log-only. New tests (`confirmation_classification_covers_every_repetition_shape`, `confirmation_log_lines_distinguish_the_three_shapes`, `a_visible_duplicate_outranks_a_clean_confirmation_in_the_log`) cover the actual claim. The e2e-deterministic failure on this PR is the pre-existing `orchestration_remit_001` race (fixed by #837 in this same batch), unrelated to this diff.

prageethw (approved):
Verified this is a mechanical relabelling, not a widening/narrowing: prompt_submission_matches is now .is_some() over classify_prompt_submission, same accept set and evaluation order. Confirmed the two pinned tests (repeated_copies_of_the_seed_count_as_delivered, bare_concatenation_is_accumulation_and_not_confirmation) appear in the diff only as a doc-comment reference — they're unmodified. The three new tests (classification-covers-every-shape, log-lines-distinguish-the-three-shapes, visible-duplicate-outranks-clean-confirmation) match what's claimed. No wire/protocol touch (PROTOCOL_VERSION untouched, no daemon_protocol.rs changes), consistent with the log-only claim. LGTM.

Issue #892, opened 2026-09-05
src/spawn.rs test module: five more positive-arrival assertions are still bounded by a fixed sleep


vfarcic:
`src/spawn.rs`'s test module still asserts a **positive** physical-arrival condition after a fixed `tokio::time::sleep` in five places. Each is the defect #850 and #851 fixed, on a path neither issue reached. All are inside `#[cfg(test)] mod tests`, so none is a shipped-behaviour bug — they are blocking-check flakes waiting for a loaded runner.

This is the remainder of the survey PR #854 published, minus the two PR #893 converts. It is filed rather than folded into #893 because widening that patch silently was explicitly out of scope.

## The sites

Line numbers are against `main` after #893 lands; each is a `sleep` immediately followed by a snapshot and an assertion that something **did** arrive.

| line | sleep | what it guards | why it is the same defect |
| --- | --- | --- | --- |
| 3148 | 250 ms | `observe_dispatch_016_rearm_case`'s `after_replay` delta | a short read silently **shrinks** the delta instead of panicking, so this one fails as a false negative rather than a loud one. `wait_for_detached_delivery_attempt` is already used twice in the same function. |
| 3527 | 75 ms | `retry_after_user_frame`'s post-retry snapshot | **mixed polarity, one sleep, two contracts.** `plain_enter_retry` asserts `after.len() > before.len()` (positive, convertible); `ctrl_j` and `alt_enter` assert `after == before` (negative, must stay). Splitting it means telling the helper which polarity its caller wants. |
| 4048 | 750 ms | `spawned_output.windows(SPAWNED_PROMPT.len()).any(..)` — #570's late-native-claim case | asserts the payload **did** reach the target. |
| 4324 | 800 ms | `dispatch_018`'s "precondition: attempt 2 must have reached the byte target before the user types" | **inside the very test #851 de-flaked**, and the most likely of these to bite next: same test, same shape, 800 ms of budget instead of 75 ms. |
| 4489 | 75 ms | `dispatch_020`'s `after_delivery_b.len() > before_delivery_b.len()` | asserts delivery B's first attempt physically wrote. |

[...]

PR #893, opened 2026-09-05
fix(tests): wait for the PTY in dispatch_018/019 instead of sleeping 75ms (#851)

### LINKED ISSUE #851: build-windows: dispatch_018/019 assert physical PTY arrival after a fixed sleep(75ms), so a required check flakes under load
`build-windows` is one of the four **required** checks (verified on ruleset `20587589`: `build`, `build-macos`, `build-windows`, `security`), so unlike the advisory `e2e-deterministic` lane a flake here actually blocks a merge — and per CLAUDE.md rule 8 the `write`-role maintainer cannot bypass it.

## Measured

PR #830, head `8034713`, run `33779968467`. Attempt 1, `build-windows` — `cargo build` and `cargo clippy` both passed, `cargo nextest run --workspace` failed:

```
FAIL [0.416s] (819/2516) dot-agent-deck spawn::tests::dispatch_018_user_input_disarms_detached_submit_probe
  panicked at src\spawn.rs:3892:9:
  precondition: attempt 1 must physically reach the detached pane; output="\u{1b}[?9001h\u{1b}[?1004h"

FAIL [0.427s] (820/2516) dot-agent-deck spawn::tests::dispatch_019_writer_release_does_not_expose_unstamped_user_input
  panicked at src\spawn.rs:4045:9:
  precondition: the user's unsent bytes must already be physically visible before the clock stamp; output="\u{1b}[?9001h\u{1b}[?1004h"
```

Attempt 2 — the **same job re-run on the same commit with no code change** — passed. That is the flake demonstrated rather than asserted.

Both panics are on a *precondition*, not on the behaviour under test, and in both the pane snapshot holds only the mode-set escapes the pane emits at startup (`ESC[?9001h ESC[?1004h`) — i.e. the write had not propagated yet, not that it went wrong.

## Not introduced by the branch it surfaced on

`src/spawn.rs` is **byte-identical** across the PR head, `main` and the merge — `sha256` `c0660c806963` on `15f77d4`, `0153758` and `8034713` alike. Corroborating:

- `build-windows` passed on **both** merge parents independently (`main` @ `0153758`, PR @ `15f77d4`).
- On the failing commit, Linux `build` and `build-macos` both passed.
- Locally on Linux both tests pass — `cargo test-fast` on `8034713` was 4073/4073, with `dispatch_018` at 3.294s and `dispatch_019` at 0.362s.

## Mechanism: a fixed sleep where the module already has a poll

Both sites do `sleep(75ms)` and then assert the bytes have physically landed, with no poll at all:

```rust
// src/spawn.rs:3888
tokio::time::sleep(Duration::from_millis(75)).await;
let after_initial_delivery = replacement_registry.snapshot(&replacement_agent)…;
assert!(after_initial_delivery.windows(REPLACEMENT_PROMPT.len()).any(…),
        "precondition: attempt 1 must physically reach the detached pane; output={:?}", …);

// src/spawn.rs:4043 — same shape after write_all + flush
tokio::time::sleep(Duration::from_millis(75)).await;
let before_writer_release = registry.snapshot(&agent_id)…;
assert!(before_writer_release.windows(USER_DRAFT.len()).any(…), …);
```

75 ms is a guess about PTY propagation on an unloaded machine. That Windows job runs 2516 tests in parallel, and `.config/nextest.toml` sets `retries = 0` in the default profile (line 69) — the only `retries = 1` override targets `dispatch_015`, so nothing absorbs a miss.

This is notably *not* the shape #818 describes for the `orchestration_remit` failures, which it calls "a correct, Decision 21-compliant harness poll — the bound is simply too small". Here there is no poll to widen.

## Suggested fix

The same `mod tests` already has the right helper, `wait_for_detached_payload_echo` (`src/spawn.rs:2935`): it polls the pane snapshot every 20 ms against an 8 s deadline for exactly this payload-reached-the-pane condition, and its timeout message prints the snapshot. Replacing the two `sleep(75ms)` + `assert!` blocks with it removes the race and raises the budget ~106x while keeping the diagnostic.

Worth a wider look while in there: `src/spawn.rs` contains **29** `tokio::time::sleep` calls, and any other one sitting in front of a physical-arrival assertion is the same latent flake on a loaded runner. I have not classified the other 27 — that is the check this issue is asking for, not a claim about them.

## Related but distinct

- #818 — `e2e-deterministic` red since #789. Same *class* (wall-clock assertions on a saturated runner) but a different lane, different tests, and that lane is advisory whereas this check is required.
- #832 — closed `NOT_PLANNED`; `orchestration_remit` specifically.
- #564, #642, #402 — macOS **fast-tier** timing flakes, all closed. #247 — `spawn_005` in the fast tier.
- #511 — a previous `build-windows` redness, but a Unix-only-path cause, closed.

Searched all 400 issues open and closed by title, plus a body search for `dispatch_018`/`dispatch_019`, and found nothing covering these two tests or a Windows timing flake in the fast tier.

Closing this means: neither test can fail on a loaded runner without a real regression behind it, and the required check stops costing a re-run.

vfarcic:
Closes #851.

## The defect, reproduced before it was fixed

`build-windows` is one of the four **required** status checks, so unlike the advisory `e2e-deterministic` lane a flake there actually blocks a merge. On PR #830, head `8034713`, run `33779968467`, attempt 1 failed on two fast-tier unit tests at once:

```
FAIL dot-agent-deck spawn::tests::dispatch_018_user_input_disarms_detached_submit_probe
  panicked at src\spawn.rs:3892:9:
  precondition: attempt 1 must physically reach the detached pane; output="\u{1b}[?9001h\u{1b}[?1004h"

FAIL dot-agent-deck spawn::tests::dispatch_019_writer_release_does_not_expose_unstamped_user_input
  panicked at src\spawn.rs:4045:9:
  precondition: the user's unsent bytes must already be physically visible before the clock stamp; output="\u{1b}[?9001h\u{1b}[?1004h"
```

Attempt 2 — the same job re-run on the same commit with no code change — passed. Both panics were on a *precondition*, not the property under test, and the snapshot in each held nothing but the mode-set escapes the pane emits at startup: the write had not propagated yet, rather than gone wrong.

The mechanism was a fixed sleep where the module already had a poll. Each test wrote to a real PTY and then, after `tokio::time::sleep(Duration::from_millis(75))`, asserted the bytes had physically arrived — a bet on a PTY round trip beating that constant. `.config/nextest.toml` sets `retries = 0` and no scoped override covers either test, so one lost bet reddens a blocking check. This is the defect #850 fixed for `dispatch_020`, on the two sites that fix did not reach.

**The issue was filed from CI logs, so the first deliverable was proof the defect exists.** Everything below is 16 cores, CPU oversubscribed with spinners:

| what was run | change | load | result |
| --- | --- | --- | --- |
| `dispatch_018` + `dispatch_019` | none (baseline) | idle | pass, 3.29 s / 0.38 s |
| `dispatch_019` | **sleep → 0 ms** | 4x | **14 FAIL in 15** |
| `dispatch_018` | sleep → 0 ms | 4x | 0 in 15 |
| `dispatch_018` | sleep → 0 ms | 20x + fork storm | 0 in 10 |
| `dispatch_018` | sleep → 0 ms **and `SUBMIT_DELAY` → 0** | 4x | **2 FAIL in 12** |

`dispatch_019` reproduces with no production change at all, panicking at the precondition with the exact CI message (`output="automatic payload already delivered once"`). Two of its 14 failures landed one assertion further on, at the final exact-equality check — which is what surfaced the second half of the fix below.

**`dispatch_018` needed one more step, and it is worth knowing why.** Production's `SUBMIT_DELAY` is 150 ms (`src/pane_input.rs:82`), and `write_and_submit_guarded` spends it between the payload write and the submit CR — so by the time that call returns, the payload has had 150 ms of head start that has nothing to do with the test. That incidental budget is why 0 ms alone survives even a fork storm on Linux. Setting `SUBMIT_DELAY` to 0 removes the accident and leaves the sleep as the whole budget, at which point the test fails with the exact CI message, `precondition: attempt 1 must physically reach the detached pane; output=""`. Neither the sleep nor `SUBMIT_DELAY` is modified by this PR; both were local edits for the measurement, reverted before the fix went in.

## The fix

`dispatch_018` waits for attempt 1's own echo through `wait_for_detached_payload_echo`, which polls `registry.snapshot` every 20 ms against an 8 s deadline and prints the snapshot on timeout. Nothing precedes attempt 1 on that pane, so there is no earlier byte-target copy that could cut into its echo and no drain to do first; the `type_user_draft` that follows already owns that half for the write after it.

`dispatch_019` gets **two** halves, and neither is redundant:

- `wait_for_echo_bytes` replaces the sleep before the draft's precondition. The draft carries no line terminator, so it completes no input line and the byte target has nothing to copy back — no quiescence wait is needed on top.
- `wait_for_drained_lines` drains the initial delivery **before** the draft is written. This one looks like tidiness and is not. The byte target and the line discipline's echo are two independent writers into one PTY output queue, so a copy still in flight lands inside the exact-equality window at the end of that test, where it reads as bytes the refused retry sent.

That second half is measured, not assumed. With the drain alone disabled, `dispatch_019` fails **4 runs in 30** at 4x — every one at the final `assert_eq!`, never at the precondition, with a delta of exactly `"automatic payload already delivered once\r\n"`. That is the byte target's copy of the *initial* delivery arriving late, and it cannot be the retry: the preceding `assert_eq!(retry_outcome, GuardedSend::Stale)` passed on each of those runs, and `Stale` returns before any byte is written. So the drain removes a measurement artifact and hides no regression.

**After the fix, at the same load:**

| what was run | change | load | result |
| --- | --- | --- | --- |
| `dispatch_019` | fix | 4x | **0 FAIL in 30** |
| `dispatch_019` | fix, drain disabled | 4x | 4 FAIL in 30 (final assertion) |
| `dispatch_018` + `dispatch_019` | fix | 4x | 0 FAIL in 15 |
| `dispatch_018` | fix, `SUBMIT_DELAY` still 0 | 4x | **0 FAIL in 12** |

## Windows verification is CI's

These are `#[cfg(test)]` unit tests in the fast tier, so `build-windows` runs them on every PR and no local gate replaces it. I cannot run that job and have not claimed to; **read the run on this PR rather than treating the Linux numbers above as Windows evidence.**

What can be said locally is that the change introduces no new platform assumption — the class of error that made #850's first attempt pass `build` and fail `build-windows` on the same commit. Each of the three helpers called here is already exercised on Windows today by a test that passes that job, in the same shape:

- `wait_for_detached_payload_echo` — by `scheduler/dispatch/016` (platform coverage mac+linux+windows).
- `wait_for_drained_lines` — by `scheduler/dispatch/020`, both through `type_user_draft(.., 1)` and as a direct `wait_for_drained_lines(.., 2)` call, since #850.
- `wait_for_echo_bytes` — by `type_user_bytes`'s landmark waits in the same test. `USER_DRAFT` is plain printable text with no escape sequences, so it is echoed literally on both platforms; the assertion that follows already demanded exactly that and passes on Windows today.

The byte target is `/bin/cat` on Unix and `more.com` on Windows. `scripts/windows-cross-check.sh` type-checks the workspace for `x86_64-pc-windows-msvc` from Linux, exit 0, with no warning touching `src/spawn.rs`.

## Scope — five more sites, deliberately not fixed here

`git grep from_millis(75) -- src/spawn.rs` reported nine sites before this change and seven after. Two of the seven are the same defect; three more sleeps at other intervals are too. All five are enumerated in **#892** rather than folded in here, because widening this patch silently was out of scope:

| line (post-merge) | sleep | what it guards |
| --- | --- | --- |
| 3148 | 250 ms | `observe_dispatch_016_rearm_case`'s `after_replay` delta — fails as a *false negative*, since a short read shrinks the delta rather than panicking |
| 3527 | 75 ms | `retry_after_user_frame` — **mixed polarity**: `plain_enter` asserts `after.len() > before.len()`, `ctrl_j`/`alt_enter` assert `after == before` |
| 4048 | 750 ms | `SPAWNED_PROMPT` arrival, #570's late-native-claim case |
| **4324** | **800 ms** | **`dispatch_018`'s "attempt 2 must have reached the byte target before the user types"** |
| 4489 | 75 ms | `dispatch_020`'s `after_delivery_b.len() > before_delivery_b.len()` |

The 4324 one is worth flagging explicitly: it is inside the very test this PR de-flakes, same shape, with 800 ms of budget instead of 75 ms. It is the next most likely of the five to bite. Say the word and I will fold it into this PR instead — I left it out only because the task scoped this change to the two sites CI actually failed on.

The remaining sleeps are correct and stay: the 12 deliberate negative-observation windows (where the sleep *is* the observation and there is no completion to poll for, including `dispatch_019`'s own 75 ms window before `after_clock_stamp`), the 200 ms fixture input in `dispatch_021`, the poll intervals inside the bounded waits, and the `#[cfg(windows)]` frame sleep at 3483 that #850 left explicitly unverified rather than guessed a second time.

## Rules

- **Rule 12** — answered explicitly rather than left implicit. Every hunk is inside `#[cfg(test)] mod tests`, which starts at `src/spawn.rs:2712` and is the only `cfg(test)` in the file; the earliest hunk is at 4228. So no production code changed and the TUI↔daemon contract is untouched: **no wire shape moved, so `PROTOCOL_VERSION` needs no bump, and no field's meaning shifted, so no `.breaking.md` fragment applies.** No cross-version manual test is required — there is no shipped code path for an older daemon to disagree with.
- **Rule 4** — no user-visible TUI behaviour changed, so no new TUI test. `tests/CATALOG.md` is unchanged: `scheduler/dispatch/018` and `/019` still assert exactly what they did, and only the fixture's synchronisation moved.
- **Rule 6** — each test was rerun alone by filter first, then the module, then the tier.
- **Rule 9** — no new user-facing surface, so no experimental-flag question.
- **Rule 10** — the changelog fragment is unwrapped prose.

## Tests run

- `cargo test-fast` — **2401 passed, 0 failed** (the count is unchanged by this PR; the same 2401 were selected before the edit)
- `cargo test-fast dispatch_018` and `cargo test-fast dispatch_019` — each alone by filter, pass
- `cargo nextest run --workspace -E 'test(~spawn::tests::)'` — all **43** `spawn::tests` pass, `dispatch_016` and `dispatch_020` included (they share the helpers this change calls)
- `cargo fmt --check` — clean
- `cargo xtask linkage-check` — ok (626 catalog ids, 556 annotations, 100 allowlisted, 11 rules)
- `scripts/windows-cross-check.sh` — exit 0
- plus the load harness in the tables above

`cargo test-e2e` was **not** run unfiltered: per CLAUDE.md rule 5 the full-tier obligation before a PR is gone, and lane 1 runs in CI on every PR. This change lives in the library's `#[cfg(test)] mod tests`, which integration-test binaries never compile, so it cannot reach lane 1 — read `e2e-deterministic` on this PR rather than treating that as a claim.


prageethw (approved):
Reviewed the diff and PR description. Both hunks are confined to `#[cfg(test)] mod tests` in `src/spawn.rs` (lines ~4228 and ~4374/4386) — no production code changed. The fix replaces two fixed 75ms sleeps with the module's existing content-keyed polling helpers (`wait_for_detached_payload_echo`, `wait_for_drained_lines`, `wait_for_echo_bytes`), consistent with the #850 pattern already used by `dispatch_016`/`dispatch_020`. The reasoning for the drain-before-write ordering in dispatch_019 (avoiding a byte-target copy landing mid-echo) is sound and the PR's own load-testing data backs it. No inline comments or open threads to address. LGTM.

prageethw (approved):
Well-reproduced flake (CI logs cited, load-tested repro showing 14/15 fail at 0ms sleep) and a real fix — waits for the actual PTY-write signal instead of a fixed sleep, and correctly accounts for the incidental 150ms SUBMIT_DELAY head-start in dispatch_018 rather than just padding the wait. LGTM.

prageethw (approved):
Correct fix for the CI flake: replaces a fixed 75ms sleep with content-keyed bounded waits (wait_for_detached_payload_echo / wait_for_echo_bytes / wait_for_drained_lines) that the module already had. The reproduction methodology is rigorous — reproduced the failure under CPU oversubscription before fixing, confirmed the drain-before-write half is load-bearing by disabling it and watching a different assertion fail (4/30) rather than assuming it, and re-confirmed 0 failures post-fix at the same load. Test-only change (inside #[cfg(test)] mod tests), no protocol/wire impact. Scope is deliberately limited with the remaining sleep-based assertions tracked in #892 rather than folded in here.

prageethw (approved):
Verified the diff: both fixed 75ms sleeps in dispatch_018/019 are replaced with content-keyed polling (wait_for_detached_payload_echo, wait_for_echo_bytes) plus the additional wait_for_drained_lines(1) drain before writing the draft in dispatch_019, matching the measured race described (a still-in-flight byte-target copy landing inside the exact-equality window). Confirmed the change is entirely inside #[cfg(test)] mod tests (no production code touched, rule 12 correctly N/A). The reproduction methodology (load-testing at 4x/20x with sleep zeroed) is rigorous and the scope decision to defer the other 5 flaky-sleep sites to #892 rather than widen this patch is reasonable. LGTM — build-windows is the real gate for this one.

PR #899, opened 2026-09-05
ci: stop persisting the job token in every ci.yml checkout (#811)

### LINKED ISSUE #811: Harden e2e-deterministic's checkout with persist-credentials: false, plus PR #805 sweep residuals
> **BLOCKED — do not start this until PR #805 is merged.**
>
> This issue touches `.github/workflows/ci.yml` (which PR #805 changes by 52 added / 17 removed lines) and several prose files #805 rewrote. It also contains a workflow **behaviour** change — `persist-credentials: false` — which #805 deliberately deferred so that its final rounds could stay prose-only.
>
> Check first: `gh pr view 805 --json state,mergedAt`. If it is not `MERGED`, stop and pick something else.
>
> Take this **third**, after #806 and #810.

---

Two residuals from PR #805's absolutes sweep (#502, #785). Neither is a defect on `main`; both were judged out of scope for a prose-only round.

## 1. Set `persist-credentials: false` on `e2e-deterministic`'s checkout

`.github/workflows/ci.yml`'s `e2e-deterministic` job checks out with an unqualified `actions/checkout@v7`. That receives the automatic `GITHUB_TOKEN` as its default `token:` input, and `persist-credentials` defaults to `true`, so the token is written into `.git/config` as an auth header for the life of the job — reachable by contributor-authored test code that this lane, uniquely among the Rust jobs, actually executes at PTY scale.

What bounds it today is scope, not absence: `ci.yml`'s top-level `permissions:` block grants `contents: read` and `pull-requests: read` and nothing else, and a fork `pull_request` run gets a read-only token regardless. So this is hardening, not an open hole — which is why PR #805 documented it in the job's comment rather than changing the workflow inside a prose-only commit.

The change is one line under that `- uses: actions/checkout@v7`:

```yaml
      - uses: actions/checkout@v7
        with:
          persist-credentials: false
```

Before merging it, confirm no step in the job needs the persisted credential — the tier builds and runs from the checked-out tree and does not fetch or push, but that should be verified by a run rather than asserted. The generated `issue-labeler*.lock.yml` files already set it, so the pattern is established here. Worth considering for the other `ci.yml` jobs in the same pass; `e2e-deterministic` is the one that matters most because it runs the most contributor code.

## 2. Prose nits deliberately left alone by the sweep

Recorded so a later reader knows they were seen and judged, not missed:

- **"and never will"**, said of the in-process worker paths having no `TuiDeck` and so no `recording_redactions` — `tests/common/mod.rs` (`seed_claude_worker_home`), `tests/e2e_codex_worker.rs`, and `docs/develop/e2e-lanes.md`. It is a prediction about future code rather than a checkable property, but it reads as a design statement about a deliberately deck-free path (the same category as "no test credential will be registered"), and all three say it consistently. Left as written.
- **`ci.yml`'s Magic Nix Cache comment**: "so there is no third-party service and no secret". In context this means the cache backend needs no repository secret — true — but it is the same phrase shape as the P1 finding, in a job that also holds a `GITHUB_TOKEN`. Rewording it would be clearer; it is not wrong.
- **`tests/CATALOG.md`** carries ~200 prose absolutes on lines this PR did not touch (per-test "Asserts"/"Does not assert"/"Agent: none" descriptions). A 26-line sample found none of the flagged class — they are bounded claims about individual tests, not about CI or credential policy — but the file was not read exhaustively.

vfarcic:
Closes #811.

## What changed

All **nine** `actions/checkout` steps in `.github/workflows/ci.yml` now set `persist-credentials: false`.

The action defaults that input to `true`, which writes the job's automatic `GITHUB_TOKEN` into the checkout's **local** git config as an auth header for the life of the job, readable by anything the job runs. `e2e-deterministic` is the sharp case — it executes the PTY e2e tier, which is contributor-authored code — and its own comment block already named this and explicitly deferred the change to "its own change rather than folded into a prose pass". This is that change.

### Why all nine, not just one

Decided per step, not applied blindly:

| job | what it runs | needs the persisted credential? |
| --- | --- | --- |
| `desktop-web` | `pnpm install --frozen-lockfile`, `pnpm test`, `pnpm build` | no — `desktop/pnpm-lock.yaml` has no `git+` entry, so every dependency comes from the npm registry |
| `build` | `cargo fmt/clippy/build/nextest`, `cargo xtask linkage-check` | no — `Cargo.lock` has no `source = "git…"`; linkage-check's git is `worktree list` / `rev-parse --show-toplevel`, both local |
| `e2e-deterministic` | `cargo test-e2e` | no — the harness's git is `git init` inside its own tempdirs; `lifecycle/version/001`'s nested `cargo build` injects env vars and shells out to no git |
| `windows-cross-check` | `scripts/windows-cross-check.sh` | no — `cargo check` for the MSVC target |
| `build-windows` | symlink check, `cargo build/clippy/nextest` | no |
| `build-macos` | `cargo build/clippy/nextest` | no |
| `security` | `cargo install cargo-audit`, `cargo audit` | no — the advisory-DB clone lands in `~/.cargo/advisory-db`, a **separate** repository that a `--local` config in this checkout never reached |
| `nix` | `nix flake check -L --all-systems` | no — flake inputs are fetched by Nix's own fetchers, not through git's repo-local config |
| `devbox` | `scripts/check-pin-lockstep.sh`, devbox install, `scripts/devbox-smoke.sh` | no — none of those scripts contains a `git clone/fetch/pull/push/ls-remote/submodule` |

Two properties hold across all nine and are the real argument:

- **Nothing here pushes.** `ci.yml` contains no `git push`, and its single top-level `permissions:` block grants `contents: read` + `pull-requests: read` with no job overriding it — so no job had write scope to push *with* even before this change.
- **Nothing here reaches the remote from inside the checkout**, as far as the usual routes go: no `.gitmodules`, no git-sourced cargo or pnpm dependency, and the git these jobs do run is local (`build.rs`'s `describe`/`rev-parse`/`status` version stamping, linkage-check's reads, the harness's tempdir `git init`).

### Deliberately NOT changed

`release.yml` and `docs-publish.yml` push to `main` under the `RELEASE_TOKEN` admin PAT and **must** keep a usable credential in the checkout (CLAUDE.md rule 8). `docs.yml` and `aarch64-crossbuild-check.yml` are `contents: read` and look like the same safe case, but they are outside this issue's scope and — importantly for the verification bar below — `aarch64-crossbuild-check.yml` is path-filtered and would not run on this PR, so a change to it could not be verified here.

### The comment block is part of the change

The paragraph in `e2e-deterministic` that argued "what bounds it is SCOPE, not absence" was correct for the old state and is now describing history. It has been rewritten to say what is true after the change: the token still exists and `actions/checkout` still authenticates the fetch with it; what moved is where it comes to **rest**. Scope is still the outer bound and still does the heavier lifting — the flag is a layer under it, not a replacement. The surrounding nuance about the Codex issue-labeler (a separately credentialed workflow that genuinely does reach a real agent on a runner) is untouched, per CLAUDE.md rules 5 and 17.

A file-level note above `jobs:` carries the survey behind the flag, so the per-step `with:` blocks stay bare.

## PR #805 sweep residuals

- **`ci.yml`'s Magic Nix Cache comment** (issue body): "no third-party service and no secret" → "no third-party service and no **repository** secret to register", with both halves explicitly scoped to the *cache backend* rather than to the job, which still holds the automatic `GITHUB_TOKEN` like every other job here.
- **`tests/harness_isolation.rs`** (issue comment 1): the launch-path claim said "the 57 of the 71 … (the other 14 …)". Measured now: 73 e2e files, 58 launch a `TuiDeck`, 15 do not. Rather than swap in numbers that will rot again, the sentence now carries **no counts at all**.
- **`xtask/linkage-check/src/junit_strip.rs`** (issue comment 2): the comment enumerated where the stripper's usage is documented but omitted `.config/nextest.toml`'s `store-failure-output` comment — the one place a person meets the instruction *in the course of a nextest run*. Added, and named by setting rather than by line number so it does not rot. (The issue cited `.config/nextest.toml:55`; the instruction is at line 99 today.)
- **Lane-2 alias wording** (issue comment 3): CLAUDE.md rule 5, `CONTRIBUTING.md` and `docs/develop/e2e-lanes.md` now say the alias is the convenient spelling, not the definition — a lane is selected by the cargo **feature set** the `#![cfg(…)]` attributes read, so a raw `cargo nextest run --workspace --features e2e,e2e-live <filter>` selects the same files and discharges the same obligation.

The other two issue-body residuals ("and never will" about the in-process worker paths; `tests/CATALOG.md`'s ~200 per-test prose absolutes) were recorded there as *seen and judged, left as written* — no change, deliberately.

## Verification

- **A workflow change cannot be proven locally**, so the honest evidence is this PR's own CI run plus a `workflow_dispatch` run on the branch. Run URLs are in a comment below. Every job that depends on its checkout for later work — all nine — has to be green for the change to be sound, and `security`'s advisory-DB clone and `nix`'s flake-input fetches are the two that would fail first if the reasoning above were wrong.
- `cargo test-fast` — **2401 passed, 0 skipped**. This covers the Rust doc-comment edits and, via `xtask/linkage-check/src/pin_lockstep.rs`, re-reads `.github/workflows/` after the YAML edit. It covers **none** of the workflow behaviour itself.
- Scoped tests for what was touched, named per rule 5: `junit_strip::junit_strip_self_test_passes`, `junit_strip::a_real_failure_report_loses_every_output_surface`, `junit_strip::a_missing_report_is_a_no_op`, `harness_isolation::tui_deck_launch_discards_the_previous_recording_before_it_spawns`, and the fifteen `pin_lockstep::*` including `repository_pins_are_in_lockstep` — 23 run, 23 passed.
- `cargo xtask linkage-check` — ok (626 catalog ids, 556 annotations, 100 allowlisted, 11 rules).
- `cargo test-e2e` was **not** run unfiltered, per the task and CLAUDE.md rule 5; CI's `e2e-deterministic` lane is the run to read.

**Rule 12**: this touches a workflow file, prose and doc comments. It moves no TUI↔daemon protocol, no handler contract, no orchestration or hook path, so `PROTOCOL_VERSION` is unchanged and no `.breaking.md` fragment applies. The changelog fragment is `changelog.d/811.misc.md`.

## Discovered, not fixed here

The e2e tier is now **73** files (49 lane 1 + 24 lane 2). CLAUDE.md, `CONTRIBUTING.md`, `docs/develop/e2e-lanes.md`, `tests/CATALOG.md` and `ci.yml`'s own comments still say 71 (47 + 24) — two lane-1 files landed since (`e2e_orchestration_remit.rs`, `e2e_orchestration_orphan.rs`). That is a repo-wide count refresh touching ~20 sites, several of which are historical records that must **not** move (`CHANGELOG.md`, `prds/`), so it is deliberately left out of this PR rather than half-done. Filed as #900.


vfarcic:
## CI evidence — and the one red job is #818, not this change

**Run URLs.** Two runs on this branch, deliberately: the PR run exercises the `pull_request` token path this change is about, and a `workflow_dispatch` run exercises the `push`/dispatch path where the `changes` job takes its no-PR-payload branch. Different concurrency groups, so neither cancelled the other.

- `pull_request` — https://github.com/vfarcic/dot-agent-deck/actions/runs/33943030304
- `workflow_dispatch` — https://github.com/vfarcic/dot-agent-deck/actions/runs/33943033727

**Attempt 1 board (PR run):** `changes`, `desktop-web`, `build`, `build-windows`, `build-macos`, `security`, `windows-cross-check`, `nix`, `devbox` — **all green**. `e2e-deterministic` — **red**.

That is every job in the file, each having done its work against a checkout carrying no persisted credential. The two that would fail first if the survey in the PR body were wrong are green:

- **`security` (18s, pass)** — `cargo audit`, whose advisory-DB clone is the only network git operation any job in `ci.yml` performs. It resolved fine, which is the evidence for the claim that the clone lands in `~/.cargo/advisory-db` and never consulted this checkout's `--local` config.
- **`nix` (pass)** — `nix flake check -L --all-systems`, i.e. the flake inputs fetched from `github:` URLs by Nix's own fetchers rather than through git's repo-local config.

`build` also passed, which covers `cargo xtask linkage-check` and `build.rs`'s `git describe` / `rev-parse` / `status` version stamping — the local git this change had to leave working.

### The `e2e-deterministic` failure is pre-existing and tracked

Four failures, all in one file:

```
FAIL orchestration_remit_001_start_role_compaction_reasserts_remit
FAIL orchestration_remit_003_reassertion_waits_for_confirmed_delivery
FAIL orchestration_remit_004_start_role_clear_reasserts_remit
FAIL orchestration_remit_005_non_start_role_clear_reasserts_nothing
```

2585 of 2589 passed. Not asserting this is unrelated — showing it:

| run | branch | event | `e2e_orchestration_remit` failures |
| --- | --- | --- | --- |
| [33878486262](https://github.com/vfarcic/dot-agent-deck/actions/runs/33878486262) | **main** | push | yes |
| [33880370511](https://github.com/vfarcic/dot-agent-deck/actions/runs/33880370511) | **main** | push | `remit_001` |
| [33912904800](https://github.com/vfarcic/dot-agent-deck/actions/runs/33912904800) | **main** | push | `remit_007` |
| 33943030304 | this PR | pull_request | `remit_001`, `003`, `004`, `005` |

`e2e-deterministic` is red on `main` itself in each of its last three push runs, and a **different subset of the same file** fails each time. That is the signature of [#818](https://github.com/vfarcic/dot-agent-deck/issues/818) — still open, titled "e2e-deterministic has been red on every branch and on main since #789 merged" — whose `.config/nextest.toml` comment records the cause as an isolation bug where concurrent decks interfere on identifiers every test in that file reuses.

Those tests touch no git path at all. This change removes an auth header from `.git/config`; it moves nothing the orchestration remit path reads.

Attempt 2 of the job is running. Per CLAUDE.md rule 8 `e2e-deterministic` is deliberately **not** one of the four required checks, so this does not gate the merge either way — but the four that are required (`build`, `build-macos`, `build-windows`, `security`) are all green on attempt 1.

### Fresh data point for #818

Both halves of #818's containment are on `main` at `d1baecd` and were in effect for these runs — the 30s `STATE_APPLIED_TIMEOUT` / `DELIVERY_TIMEOUT` bounds in `tests/e2e_orchestration_remit.rs`, and the `orchestration-remit = { max-threads = 1 }` group whose `test(/^orchestration_remit_/)` filter does match all seven. So the containment measured green at `-j 1` on a dev box is **not sufficient on a 4-vCPU GitHub runner**. Noted on #818 rather than acted on here.

vfarcic:
## Attempt 2 — same file, and now the same single test as `main`

Re-ran `e2e-deterministic` once. It is red again, and the result is a cleaner control than attempt 1:

```
Summary [507.072s] 2589 tests run: 2588 passed (1 slow, 1 leaky), 1 failed, 0 skipped
  FAIL [47.661s] dot-agent-deck::e2e_orchestration_remit
       orchestration_remit_001_start_role_compaction_reasserts_remit
```

`orchestration_remit_001` is **the same single test** that failed on `main`'s push run [33880370511](https://github.com/vfarcic/dot-agent-deck/actions/runs/33880370511) — same test, same file, unrelated commit. Across the two attempts on this branch the failing set went `{001, 003, 004, 005}` → `{001}` with no code change between them, which is the non-determinism [#818](https://github.com/vfarcic/dot-agent-deck/issues/818) describes.

I am **not** re-running a third time: two attempts, both confined to `tests/e2e_orchestration_remit.rs`, one of them matching a `main` failure exactly, is enough evidence. Further attempts would only overwrite the job conclusion again for no new information.

**Board as it stands:** the four required checks — `build`, `build-macos`, `build-windows`, `security` — are green, as are `desktop-web`, `windows-cross-check`, `nix`, `devbox`, `changes` and `docs-build`. `e2e-deterministic` is red for #818, which is not a required check and is red on `main` too.

**Auto-merge is deliberately not armed.** The board has a red check on it, and whether to accept it is a judgment call that belongs to the reviewer rather than to a queue that would land this the instant an approval arrives. Happy to arm it if you would prefer that.

prageethw (approved):
Reviewed the checkout hardening and the docs/prose sweep together.

- Per-job survey in the `ci.yml` comment holds up: none of the nine jobs pushes (the workflow's `permissions:` block is `contents: read` + `pull-requests: read`, no override), and none reaches a remote from inside the checkout — no `.gitmodules`, no git-sourced `Cargo.lock`/`pnpm-lock.yaml` entries, and the git these jobs run is local (build.rs version stamping, linkage-check's `worktree list`/`rev-parse`, the e2e harness's per-test `git init`). `security`'s advisory-DB clone lands in `~/.cargo/advisory-db`, outside this checkout's `--local` config, so that's correctly unaffected.
- Correctly leaves `release.yml`/`docs-publish.yml` alone — those push to `main` under `RELEASE_TOKEN` and need a usable credential (rule 8).
- The rewritten `e2e-deterministic` comment block is accurate: the token still authenticates the fetch, only where it *rests* changed. No overclaim.
- `e2e-deterministic`'s single failure (`orchestration_remit_001_start_role_compaction_reasserts_remit`, tests/e2e_orchestration_remit.rs:283) is the same pre-existing race PR #837 fixes elsewhere in this queue — unrelated to this CI-config/prose diff.
- No inline comments or open review threads on this PR.

Approving.

prageethw (approved):
Careful per-job justification for persist-credentials:false across all nine checkout steps (each job's actual git usage traced, none needs the persisted token), and the two structural properties (no git push anywhere, no job overriding the read-only top-level permissions) hold up. Sensible, low-risk hardening. LGTM.

prageethw (approved):
Sound, well-scoped hardening: persist-credentials: false on all nine actions/checkout steps in ci.yml, with a per-job survey justifying each (no pushes anywhere given the read-only top-level permissions block, no remote git op needed from inside any checkout). Correctly distinguishes this from release.yml/docs-publish.yml, which push under RELEASE_TOKEN and must keep persisted credentials — explicitly calls that out so nobody copies the flag there. The e2e-deterministic failure on this PR is the known cross-PR orchestration_remit flake (fixed by #837), unrelated to this workflow-only change.

prageethw (approved):
Verified: persist-credentials: false is applied to all 9 actions/checkout steps in ci.yml, and this PR own CI run confirms the security (cargo audit advisory-DB clone) and nix (flake input fetches) jobs — the two that would fail first if the per-job survey were wrong — are both green. release.yml/docs-publish.yml correctly left untouched since they need the RELEASE_TOKEN credential to push. The PR #805 residual sweep items are small, honest doc corrections. No review threads open. LGTM.

Issue #905, opened 2026-09-05
Restructure CLAUDE.md and the project-local skills: global authority, task-scoped skills, evidence in docs/develop


vfarcic:
## Problem

`CLAUDE.md` is read in full by every agent at the start of every session — a one-line typo fix pays the same cost as a PRD. It has grown 5x in a month:

| date | bytes |
| --- | --- |
| 2026-08-04 | 15,317 |
| 2026-08-11 | 29,810 |
| 2026-08-18 | 38,097 |
| 2026-08-30 | 50,739 |
| 2026-09-04 | **76,905** |

Half of that arrived in the last five days. The numbered rule count only went 14 → 17 over the same period, so this is **not more rules — it is the same rules getting longer**. Bytes per rule went from ~1.1KB to ~4.5KB.

Measured per-rule sizes today:

| rule | bytes |
| --- | --- |
| 8 — Greptile / PR / branch protection | 20,310 |
| 5 — test tiers and lanes | 19,512 |
| 2 — fmt + clippy gate | 8,444 |
| 12 — cross-version contract | 5,894 |
| 14 — tmpfs / worktrees / temp reaper | 5,151 |
| 17 — verify absolutes | 3,935 |
| 4 — TUI tests | 3,595 |
| 1, 3, 6, 7, 10, 16 | 215–854 each |

Rules 8 and 5 alone are **52%** of the file; adding 2 and 12 reaches **70%**. Rules 1, 3, 6, 7, 10 and 16 are the model to aim at: short, imperative, unambiguous.

## The defect is restatement and misplacement, not length

The goal is **clarity**. Shorter should be a *symptom* of clearer, never the target. No fact may be deleted to save bytes, and nothing normative may be weakened.

Two concrete defects:

**Duplication.** The clippy gate `cargo clippy --workspace --all-targets --features e2e,e2e-live -- -D warnings` is stated in 5 files (`CLAUDE.md`, `issue-queue/SKILL.md`, `run-dot-agent-deck/SKILL.md`, `prd-queue/SKILL.md`, `verify-pr/checks.sh`). `cargo test-fast` appears 12 times across 4 files; `tests/CATALOG.md` is referenced in 14 files. These have **not** drifted — every normative copy today reads the full four-flag command. That is the argument *for* this work: the command has changed four times (bare → `--all-targets --features e2e` → `+--workspace` → `+e2e-live`) and is currently correct in five places only by diligence.

[...]

PR #909, opened 2026-09-05
docs: the e2e lane must be green before a PR is done (#908)

### LINKED ISSUE #908: Failing e2e tests accumulate because nobody must fix one they did not cause: red should never go unexplained
## Problem

Failing e2e tests accumulate, because nobody is obligated to fix one they did not cause.

Measured 2026-09-05 across the 21 open PRs: **16 were red on `e2e-deterministic`**, and almost all of it was one family — `orchestration_remit_00{1,2,3,5,7}`, plus `pane_input_007`, with one-off failures in `form_002`, `manager_016` and `manager_010`. `orchestration_remit_001` alone was failing on **11** PRs.

The cost is not the red itself, it is the loss of signal. Determining whether any given PR was actually broken required reading CI logs on every one of the 16 to see whether its failure was its own or the shared one. A lane that is red by default trains everyone to skip it, which is the state it was in.

## This is a policy change, not a gap

The current behaviour is deliberate and should be changed knowingly:

- `.config/nextest.toml` sets `retries = 0` globally. Flakes are **not** masked — the repo already refuses to hide them.
- `orchestration_remit_*` is **already quarantined to serial execution** (`test-groups.orchestration-remit = { max-threads = 1 }`). Someone already identified it as contention-sensitive, **and it still fails across 11 PRs**. The existing mitigation is demonstrably insufficient.
- `.claude/skills/verify-pr/SKILL.md` already carries a flake protocol: rerun in isolation per CLAUDE.md rule 6, then "passes in isolation → report it as a suspected flake with the log, **not as a blocking failure**".
- CLAUDE.md describes the tier as "flaky-tolerant by design", which is *why* lane 1 is advisory and lane 2 gates nothing.

So today's rule is tolerate, report, do not block. Nobody must fix anything. That is exactly how `orchestration_remit_001` survived across 11 PRs while an open fix (#837) sat unmerged.

## Why "always fix it, whoever caused it" is the wrong shape

Three concrete failure modes:

1. **Unbounded cost.** A compaction/reassertion race can take a day to diagnose. Blocking a one-line docs PR on someone else's deep race creates an incentive to avoid touching anything.
2. **Duplicated work under parallelism.** Several dispatched agents running at once all hit the same pre-existing flake, and all stop to fix the same race — producing conflicting fixes.
3. **Contention ghosts.** #415 measured **6-7 of 40 files failing in parallel against 40/40 at `-j 1`**. Those are artifacts of load, not defects. A rule to "fix it" sends agents chasing failures they manufactured themselves — and any deliberate parallel-load measurement produces these by design.

## Proposed rule

> **Red never goes unexplained, and nothing stays red without an owner.**

Graduated, composing with the existing rule 6 isolation protocol rather than replacing it:

- **You caused it** → fix it.
- **You did not cause it** → you may not ignore it. Rerun in isolation, classify flake vs defect, then ensure a tracked issue exists: file one if it does not, link it if it does, and record the determination in the PR.
- **A cap on tracked-failing tests.** This is the part that actually bounds growth. Tracking alone documents accumulation without limiting it; a ceiling on the count (or an age limit per tracked issue) forces prioritisation when it creeps up.

Optionally, quarantine known flakes into a nextest group that does not redden the lane — but **only with a named owner and an expiry**. Without the expiry, quarantine becomes the graveyard the cap exists to prevent.

## Where it belongs

Applying issue #905's own criterion (does an automated gate backstop this?): nothing does — it is an instruction, not a gate — so it cannot live only in a skill. But it is task-scoped, and `verify-pr` and `pr-review-queue` already carry flake protocol.

So: extend those two skills, plus a **short** imperative in CLAUDE.md pointing at them. Deliberately **not** a new multi-KB rule — #905 exists because CLAUDE.md grew 5x in a month, and this change should not add to that.

vfarcic:
Closes #908.

## What this changes

A red e2e test now has exactly **two** outs: **fix it in the PR**, or **quarantine it with a named owner and an expiry issue**. Leaving it red-but-tracked is no longer sufficient.

This overturns `verify-pr`'s older instruction that a test passing in isolation is *"a suspected flake with the log, not as a blocking failure"* — that tolerance is what produced the backlog. Measured 2026-09-05: **16 of 21 open PRs red** on `e2e-deterministic`, almost all one family, `orchestration_remit_001` alone failing on **11** — already pinned to `max-threads = 1`, with an open fix (#837) nobody was obliged to land.

## Why "fix", not "track"

Tracking documents accumulation without bounding it. And most flakiness here is a **test defect, not a fact of nature**: pinning the remit group to serial execution treated the symptom, while #837's title names the cure — *make the tests wait for the confirmation they race*, rather than sleep. #837 is now green on `e2e-deterministic`, which is the evidence that these are fixable.

So "passes in isolation" is reclassified: not an absolution, but a defect in the *test* — one that only passes on an idle machine is making a timing assumption.

**Quarantine is the pressure valve.** Without an escape hatch a strict rule gets quietly ignored, which is worse than a rule with a bounded exception. The expiry is what stops quarantine becoming the graveyard.

**From a green baseline this is cheap**, which is the point: a lane green before your change and red after it is almost certainly yours. The cross-PR case mostly disappears once the lane is genuinely green.

## The required check, and why merging this does not enable it

Also adds `e2e-deterministic` to `REQUIRED_CHECKS` in `scripts/apply-branch-protection.sh`. Only a gate produces "always" — #908 notes instructions "work reasonably well on agents; on humans they are a hope."

**Merging this PR does not require the check.** The ruleset moves only when someone runs `apply`, which sends a full `PUT` (warning 3 in that script). This records the intent; an operator picks the moment. That is deliberate — **15 PRs are red right now** (904, 902, 899, 895, 894, 890, 889, 886, 881, 880, 879, 874, 871, 870, 869), almost all because they branched before #837. Running `apply` today makes every one of them unmergeable at once. Wait until they have rebased past the flake fixes, then run `apply` and `status` and read the contexts back — a ruleset missing its checks looks identical from the PR page.

### Stated caveat

That script warns: *"a context that never reports is worse than one that fails"*, citing #416, a fork whose workflows had never run and which became permanently unmergeable. **A fifth required check widens that trap.** A skipped check counts as success, so `ci.yml`'s `changes` gate is fine — but a fork whose CI never produces `e2e-deterministic` at all would be stuck. Worth weighing before running `apply`.

## Placement

Per #905's criterion: nothing automated backstops the instruction half, so it cannot live only in a skill — but it is task-scoped, and `verify-pr` already carried the flake protocol. CLAUDE.md gets the short imperative and a pointer; the evidence and quarantine mechanics live in the skill. **Rule 6: 455 → ~940 bytes**, deliberately not another multi-KB rule of the kind #905 exists to unwind.

## Testing

**None run, deliberately.** Markdown and one shell default — no `.rs` files, so `cargo fmt --check` and clippy are inapplicable. A measurement run for #906 was in flight on this machine and any local build would have contaminated it; CI runs on GitHub runners, so it validates this at zero cost to that measurement. `xtask/linkage-check` has documentation rules — if this trips one, CI is where it surfaces.


prageethw (approved):
Docs-only change (SKILL.md, CLAUDE.md, checklist.md, apply-branch-protection.sh comment) — no .rs touched, so the e2e-deterministic failure here is unrelated to this diff: it's `orchestration_remit_007_compaction_reassertion_preserves_a_dispatched_task`, another member of the same pre-existing `e2e_orchestration_remit` flaky family this PR itself documents (and that #837 in this batch fixes). The policy change is sound: replacing 'passes in isolation = not blocking' with 'fix or quarantine with a named owner + expiry issue' closes exactly the tolerance gap that let 16/21 PRs go red. The apply-branch-protection.sh change is inert until someone runs `apply` (correctly called out), and the fifth-required-check trap for forks with no e2e-deterministic CI is explicitly flagged as a caveat to weigh later. No inline comments or open threads.

prageethw (approved):
Well-measured policy tightening (16/21 PRs red, 11 from one already-fixable family per #837) with a real escape hatch (named-owner + expiry quarantine) so the stricter rule doesn't just get ignored. Correctly notes merging this doesn't itself flip branch protection — records intent only. LGTM.

prageethw (approved):
Sound process tightening: red e2e now requires fix-or-quarantine-with-expiry rather than track-indefinitely, backed by a real measurement (16/21 open PRs red, mostly one fixable family per #837) that corroborates what this review pass independently found across this same PR batch. Correctly separates recording the new REQUIRED_CHECKS default from actually applying it — apply/status is a deliberate operator action taken later, once PRs have rebased past the flake fixes, rather than something this merge triggers automatically. No source code change.

prageethw (approved):
Doc/skill/script-only change. Reviewed the full diff: the policy shift (fix-or-quarantine-with-expiry, no more indefinite tracking) is well-motivated by the measured 16/21 red-PR backlog, and the REQUIRED_CHECKS default change is correctly deferred from actually taking effect (apply is a separate, explicit operator action) with the risk clearly flagged (a never-reporting fork check would become permanently unmergeable). No source code touched. LGTM.

Issue #917, opened 2026-09-06
Retire the now-callerless unguarded pane-write primitives (write_to_pane_and_submit, write_to_pane_notice)


vfarcic:
Housekeeping fallout from #617's fix (commit `f9907cd8`), raised independently by both the reviewer and the implementer.

After the guarded-write migration, **both** unguarded primitives have zero production callers under `src/`:

- `write_to_pane_notice` — its last production caller was `src/state.rs:5041` (the respawn-failure notice), now guarded.
- `write_to_pane_and_submit` — likewise.

Evidence: `git grep -nE '\.write_to_pane_and_submit\(|\.write_to_pane_notice\(' f9907cd8` returns 7 hits repo-wide — 5 in `src/agent_pty.rs` (9640, 9694, 9706, 9790, 9794), all inside `#[cfg(all(test, unix))] mod spawn_tests` (opened at `src/agent_pty.rs:8366`), plus one in `prds/done/249-*.md` and one in `tests/e2e_delegate_work_done_chain.rs:320`.

Both remain `pub` and both are still unit-tested, so nothing warns and nothing breaks — which is exactly why this needs an issue rather than being noticed later. **The type change removed the permissive *argument*, not the permissive *API*.** A future caller can still reach an unguarded pane write without going through the identity gate, and the compiler will not object.

## Why it was not done in #617's PR

Removing a public API is out of that issue's scope, and the change is mechanical enough to be reviewed on its own.

## Shape of a fix

Decide between:

1. **Delete both**, migrating the five `mod spawn_tests` call sites and `tests/e2e_delegate_work_done_chain.rs:320` to the guarded equivalents; or
2. **Keep them but make the unsafety explicit** — rename to something self-documenting and non-inviting, and add a doc comment stating that no production path may use them, so a future call site is a visible decision rather than an accident.

Option 1 is cleaner if the test call sites migrate without contortion. Option 2 is the fallback if some test genuinely needs an unauthenticated write to a non-reusable target.

Issue #916, opened 2026-09-06
get-seed's native pull resolves a pane id with no identity check and no exited filter


vfarcic:
Found during the security audit of #617's fix (commit `f9907cd8`). It is **not** covered by that change and was deliberately left out of its PR.

## The hole

`take_pending_seed_native(&req.pane_id)` (reached from `src/daemon.rs:2049`, implemented at `src/agent_pty.rs:7463`) takes a **caller-supplied pane id off the hook socket**, does a linear `find` on `pane_id_env`, and applies **no identity check and no `exited` filter**.

That filtering is the norm everywhere else in the same registry: `writer_target_for_pane` (`src/agent_pty.rs:5568`), `pane_current_agent_id` (`:5539`), `agent_records` (filters at `:5211`) and the spawn-side duplicate check (`:5222`) all skip exited entries. This path does not.

`pending_seed` lives on the `RunningAgent` record, and an agent that **exits on its own** is not removed from `inner.agents` until something reaps it — while the duplicate check lets a **successor** claim the same `pane_id_env` in the meantime. In that window two entries share the pane id, `find` order over the `HashMap` is arbitrary, and a successor calling `dot-agent-deck get-seed` can be handed **the previous agent's seed**.

Note the window is specifically *self-exit then successor*: `close_agent` (`src/agent_pty.rs:6712`) **removes** the entry, so a deliberately-closed agent's seed goes with it.

## Why this is filed separately rather than fixed in #617

#617's fix closes the **injection** route — the daemon writing bytes into a pane. This is the **pull** route: an agent process asking the daemon for its seed over the hook socket. Same data, different direction, different mechanism.

It matters that they are named together, because #617's own new documentation reasons about exactly this route. `src/agent_pty.rs:133-137`, added by that commit, justifies *dropping* the seed on a refused injection because "putting it back would leave it stashed under a pane id whose new occupant could pull it via `get-seed`". That reasoning is correct, but a reader can come away believing the native route is therefore safe. It is not safe independently — #617 does not claim to cover it, and this issue records that gap rather than leaving it implied.

## Shape of a fix

Two things, both matching what the rest of the registry already does:

- **Filter exited entries** in the `find`, as `writer_target_for_pane` and `pane_current_agent_id` do.
[...]

Issue #915, opened 2026-09-06
Close the two residual permissive holes in the #608 write-authorization fix (deferred from #617)


vfarcic:
Issue #617 closed the unguarded and permissive pane-keyed *write sites* (findings 1, 2, 3, 6, 7) by making the guarded API take `&str` rather than `Option<&str>`. Two residual holes in the #608 fix itself were deliberately left open and remain live on `main`: the `(expected None, current None)` arm of `compute_write_and_submit_outcome` cannot tell a sessionless agent from a conversation that has just ended (`SessionEnd` removes `pane_hook_session` at `src/state.rs:7148` while agent and pane stay alive), and the paned re-validation closure samples `pane_has_live_attach` outside the `AppState` guard, so it can authorize on a stale `false`.

Both are documented in comments in `src/daemon_protocol.rs` rather than fixed, and both fail **permissive**.

## Why these were split out of #617 rather than folded in

Not because they are hard. An audit of `origin/main` found #617's stated rationale — that both "need new daemon state or a lock-ordering change" — **overstated for both**:

- Finding 4 needs new state, but the two hard parts are already solved. `AgentEvent` already carries `agent_id` (`src/event.rs:756`), populated from the `DOT_AGENT_DECK_AGENT_ID` env var the daemon injects at spawn, and registry agent ids are monotonic and **never recycled within a daemon process** (`src/agent_pty.rs:5042-5043`). So the stable agent-scoped key the tombstone needs is already on the wire.
- Finding 5's *recommended* remedy needs **neither** new state nor a lock-ordering change. It is a two-line deletion — see below.

The reason to split is **scope discipline**. Findings 1/2/3/6/7 are a mechanical type-driven refactor whose correctness argument is "it compiles". Findings 4 and 5 are policy changes whose correctness argument is "here is who we now refuse, and why that is right". Those want separate review. Together they are perhaps 60–100 lines including tests.

### 4 — `(expected None, current None)` conflates a sessionless agent with a just-ended conversation

[...]

PR #919, opened 2026-09-06
fix(delivery): close the permissive-on-absent-identity pane-write family (#617)

### LINKED ISSUE #617: Pane-keyed writes are still permissive on absent identity at three call sites outside the #608 handler

vfarcic:
Closes #617.

Several daemon-side paths resolved a **pane** and wrote into it without checking that the same agent identity still owned that pane. If the pane changed hands between the decision and the write — an agent exits, a new one claims the recycled pane id — the payload landed in an unrelated process, and a tool-capable agent could act on a delegation, scheduled prompt, seed or dispatch result intended for someone else.

## Approach: make the invariant a type

#617 notes that **two independent reviews each found sites the other missed**, and argues the compiler should enumerate them instead. That is what this does: `write_and_submit_guarded`, `write_and_submit_guarded_detailed`, `write_notice_guarded` and the shared `write_guarded` body now take `&str` rather than `Option<&str>`, so an absent identity is no longer expressible and the pre-lock identity gate is unconditional on the paned arm.

A cold `cargo check --workspace --all-targets --features e2e,e2e-live` then reported **31 errors** (1 lib, 30 lib-test) — the authoritative call-site list.

## Sites closed — 14 production, not the 5 the issue named

| # | site | what changed |
|---|---|---|
| 1 | `write_guarded` gate | unconditional on the paned arm; paneless resolves *by* the concrete id, so the skipped comparison is tautological |
| 2 | `deliver_on_idle` (scheduled reuse) | identity resolved at decision time, intersected with `ReuseEntry.agent_ids`, carried through the debounce, delivered guarded |
| 3 | dispatch-result delivery | extracted to `daemon::deliver_dispatch_result`, `(agent_id, cwd)` captured from **one** `AgentRecord` so two lookups cannot straddle a hand-over |
| 6 | `arm_seed_fallback` | takes the expected id; a refusal **drops** the seed rather than restoring it |
| 7 | `handle_work_done` | bound to the commissioning orchestrator; refuses when the two routes disagree |
| +8 | mechanical `Some(&x)` → `&x` unwraps, plus two sites **beyond the issue's list** (`src/state.rs` dead-replacement and respawn-failure notices) | the first was a *guarded* call passing a possibly-absent identity, which put it back on the unguarded path in exactly the hazardous case |

Every refusal fails **closed** — no bytes — and no site retries against whichever process now occupies the pane. No sentinel, `.unwrap_or("")` or manufactured identity was introduced; both reviewers swept for that independently.

## Tests

Four `#[spec]` regression tests, all **fast tier** — deliberately not lane 2, which runs on no CI runner:

| id | site | file |
|---|---|---|
| `scheduler/reuse/004` | 2 | `src/spawn.rs` (`deliver_on_idle` is private) |
| `orchestration/dispatch/005` | 3 | `tests/delegate_respawn_recovery.rs` |
| `prompt/pane-input/033` | 6 | `src/agent_pty.rs` |
| `orchestration/work-done/006` | 7 | `tests/work_done_reporting.rs` |

Each asserts **both halves**: the refusal outcome, *and* the payload's absence from the successor's scrollback — barriered by a readiness write asserted visible first, so "payload absent" cannot pass because a stub never started.

**RED-verified 4/4.** With the identity gate temporarily disabled, all four failed. They catch the bug rather than merely passing alongside the fix.

One existing test was **deleted**: `guarded_send_with_no_expected_identity_writes_to_the_live_pane` asserted the removed permissive contract (and could not compile under `&str`). Its replacement, `guarded_send_refuses_a_pane_that_changed_hands_before_the_write`, pins the new one.

## Gates

- `cargo fmt --check` and `cargo clippy --workspace --all-targets --features e2e,e2e-live -- -D warnings` — clean immediately before each commit, no flag dropped
- `cargo test-fast` — **2406/2406**, five green runs
- `cargo test-e2e -E binary(e2e_scheduler_reuse)` — **3/3**, including `reuse_003_deliver_on_idle_debounce`
- scoped `cargo test-e2e` over 11 delivery/dispatch/pane binaries — **27/27**
- lane 2: `cargo test-e2e-live -j 1 -E binary(e2e_dispatcher_mode)` — **7/7** (runs nowhere in CI, so it is on trust unless run; it was run)

The full tier was **not** run locally — lane 1 runs in CI on every PR.

## Rule 12 — daemon-internal, no bump, no breaking fragment

`PROTOCOL_VERSION` stays **8** (`src/daemon_protocol.rs:259`). The wire shape did not move: no request/response variant or field added, and the three Rust types whose shape changed (`spawn::ExistingTab`, `spawn::ReuseDecision`, `agent_pty::DelegationRetirement`) carry no serde derives. `WriteAndSubmit`'s handler already refused an absent identity per #608, so no value an older peer can send is newly refused.

**The cross-version manual test was run rather than the classification merely asserted**, in both directions:

| | old daemon (v0.39.3) + branch TUI | branch daemon + v0.39.3 TUI |
|---|---|---|
| `Attach protocol listening` lines | **1** | **1** |
| daemon pid / start time | unchanged across the run | unchanged |
| handshake | `client 0.39.3-gbe690582 / daemon 0.39.3-ga0233770` | `client 0.39.3-ga0233770 / daemon 0.39.3-gbe690582` |
| mismatch prompt | shown, naming 3 live agents, **declined** | shown, declined |
| delegate (`clear=false` / `clear=true`) | worked | worked |
| work-done feedback | worked | worked |
| status hook (run last) | worked | worked |

Both sides reported `server_version: 8`, so this exercised **semantics behind a stable wire** — what rule 12 exists to catch. Neither documented false-green trap fired (`grep -niE 'idle window|shutting down|restart|SIGTERM|orphan'` matched nothing). The reverse direction was added because #617's changes are daemon-side, so the prescribed direction has the *old* daemon executing the delivery code.

**What that run does not establish:** no identity mismatch was manufactured, so the new refusal arms were not taken — those are covered by the unit tests above. All roles were `cat` stand-ins.

## Rule 4

Not user-visible in the sense rule 4 enumerates — no new pane, field, command, tab, footer or keybinding, and a refusal produces only a `warn!` with no card notice. What changes is *which* pane receives bytes, which every test above asserts directly via PTY snapshot. The one visible consequence — scheduled reuse now opens a fresh tab when the delivery pane's live agent is not one the task spawned — reuses the existing spawn-fresh rendering path and is covered by `scheduler/reuse/004` and the lane-1 `scheduler/reuse/*` suite.

## Review findings addressed

A review and a security audit ran against `f9907cd8`; both concluded **no blockers** and that the commit closes the property. `be690582` addresses six agreed findings:

- **`deliver_on_idle`'s unconditional pre-write `note_payload_settled`** made the path exempt from #424's repeat-payload guard in *every* outcome, including `Ambiguous`, where a partial write leaves half a prompt in the box. `CONFIRMATION_TASKS` now carries the prompt and the release fires only when the cancelled watch was delivering those exact bytes. (Deleting the release outright was considered and rejected — `AbortHandle::abort` only *schedules* cancellation, so nothing orders the watch's `PayloadRecordRelease` against the write two lines below; that is the race `scheduler/reuse/003` caught.)
- **`Ambiguous` folded into the refusal arm at the two notice sites** — now broken out, since `Ambiguous` means partial bytes *did* reach the target and the log line said otherwise.
- **`handle_work_done` degraded to the weaker binding** when the commissioned and routed orchestrator panes disagreed, failing *open* on a report still labelled solicited. Now refuses with both pane ids logged.
- **`expect` on the `Option → &str` conversion** replaced with a `SendResult::NoLiveTarget` refusal — it could not fire today, but the failure mode was a panic rather than a refusal.
- **Three rule-17 over-claims in the changelog fragment** narrowed, plus one omission added (an orchestrator replaced *in place* keeps its pane but gets a new registry id, so outstanding completions become **lost reports** — the analogue of the reuse path's extra tab).

## Deliberately out of scope — filed separately

- **#915** — the two residual holes in #608's fix (the `(expected None, current None)` conflation; `pane_has_live_attach` sampled outside the `AppState` guard). Both are *same-wire/different-meaning* breaks needing a `.breaking.md` and their own cross-version run, so folding either in would import that obligation into a PR that otherwise avoids it. #915 also records the trap that would bite a naive fix: `pane_generation_closures` is **never cleaned on pane close**, so a pane-keyed witness would permanently refuse every future sessionless agent on that pane id.
- **#916** — `get-seed`'s native pull resolves a caller-supplied pane id with no identity check and no `exited` filter. Same data, opposite direction. Named because this PR's new comment reasons about that route and could be read as covering it.
- **#917** — `write_to_pane_and_submit` and `write_to_pane_notice` now have zero production callers but remain `pub`; the type change removed the permissive *argument*, not the permissive *API*.

## Note on `eee99099`

That commit is unrelated to #617 — it moves the `auditor` orchestration role off a GPT model onto the all-Claude default, after GPT credits were exhausted mid-run (mirroring what the `tester` role already carries with a dated comment). It is isolated to `.dot-agent-deck.toml` and kept as its own commit so it can be dropped or cherry-picked independently if you would rather it not ride along here.


vfarcic:
Thanks — the P1 on `src/agent_pty.rs:147` is correct and is being fixed.

To confirm the sequence in our own words, since it is the same defect class this PR is about:

1. Agent A spawns on pane P, seed stashed for A; `arm_seed_fallback` armed with A's id.
2. A exits, successor B claims pane P and stashes **its own** seed under P.
3. The grace expires and the timer's take is **pane-keyed**, so it removes **B's** seed.
4. The guarded write then correctly refuses with `WrongSession` — no bytes reach B.
5. But B's seed is already gone, so B's native `get-seed` pull and its own fallback both find nothing and **B silently loses its opening task.**

So this PR made the *write* identity-guarded while leaving the *take* pane-keyed and destructive — exactly the pane-keyed-should-be-identity-keyed shape #617 exists to close, one layer earlier than we looked. The original justification for dropping the seed on refusal (not leaving it stashed under a pane id whose new occupant could pull it) is right about not *restoring* it, but it treats the symptom; the fix is to never take a seed that was never ours.

Being fixed by making the take identity-scoped **and atomic** — the identity check happens under the same registry lock acquisition that resolves the pane, since a take-then-compare leaves the same window. Drop-on-refusal is kept for a seed that genuinely *was* ours. A regression test is being added beside `prompt/pane-input/033` (which pins the refusal) asserting the other half: after a hand-over during the grace, the **successor's own seed survives and is still retrievable**.

Two notes for the record:

- The related **native `get-seed` pull** hole — a caller-supplied pane id resolved with no identity check and no `exited` filter — was found independently in our security audit and is tracked separately as #916. It is deliberately not fixed here.
- The advisory `e2e-deterministic` failure on `orchestration_remit_001` is the known #818 flake, not this change: our diff touches no remit/compaction/`apply_event` code, and that same test file has gone red on a Markdown-only commit before (recorded on #906).

vfarcic:
**Ready for review** @prageethw — CI has settled and Greptile's P1 is addressed. Flagging the timing per rule 8: this note is deliberately last, after CI and the fixes, since `dismiss_stale_reviews_on_push: true` would void an approval given before them.

## Required checks — all green (head `0de97cfb`)

| check | result | duration |
|---|---|---|
| build | pass | 5m40s |
| build-macos | pass | 8m00s |
| build-windows | pass | 5m55s |
| security | pass | 22s |

## Greptile's P1 — fixed, and it uncovered a second bug

`acad396b` makes the seed-fallback take **identity-scoped and atomic**: identity resolution and removal happen under one acquisition of the registry `inner` lock, keyed by agent id rather than by scanning `pane_id_env`, so a seed belonging to another agent is never removed — not even transiently. It returns a three-way `Taken`/`Absent`/`StrangersSeed` instead of `Option<String>`, because folding the third case into `Absent` made the fallback log a hand-over as *"seed already delivered natively"* — describing the failure as a success.

Fixing it surfaced a **second, independent bug**: `set_pending_seed` filed the seed by *unfiltered linear scan* over a `HashMap`. Since an agent that exits on its own keeps its record (flagged `exited`) and the duplicate-pane-id guard admits a successor once the incumbent is exited, two records can carry one `pane_id_env` — so the successor's seed could be filed on the **dead predecessor's** record and then correctly taken and dropped. Greptile's exact bug through a different door, reachable on a coin flip. `set_pending_seed` now files on the pane's live occupant, pinned by its own test.

The old pane-keyed `take_pending_seed_fallback` is now `#[cfg(test)]`, so no production path can remove a seed it does not own — by construction rather than by review.

## One test was asserting the bug

Worth calling out explicitly. `prompt/pane-input/033` asserted:

```rust
assert!(registry.take_pending_seed_fallback(PANE).is_none(),
    "the refused seed must be DROPPED, not restored");
```

The seed store has one slot per pane, and in that fixture the seed in the slot was the **successor's own** — so the test was pinning precisely the data loss Greptile reported. It passed because production and test agreed on the wrong answer.

`0de97cfb` re-pins it by changing the *fixture* rather than weakening the assertions: the predecessor is now `sh -c 'read _line'` retired by one notice write, so it exits **on its own** and its record is retained as `exited`. The seed in play is then genuinely its own, and 033 keeps its original thesis — armed injection → identity gate → `WrongSession` → seed dropped — which the new `prompt/pane-input/034` does not cover.

The replacement is also **stronger**: the old single pane-keyed probe was nondeterministic (two records, one `pane_id_env`, `HashMap` iteration order), and it is now two identity-scoped reads asserting the seed was neither restored to its own record nor re-filed under the successor.

`cargo test-fast`: **2408/2408**. `cargo xtask linkage-check`: ok (631 catalog ids, 561 annotations, 11 rules).

## `e2e-deterministic` is red — the known #818 flake, not this PR

It failed on **both** runs of this PR, each time on `orchestration_remit_001`, at *different* assertion lines within the same test (`:283` then `:436`) — the same 30s-bound shape #818 documents. This PR touches no remit, compaction or `apply_event` code, and the second run's two commits are confined to the seed-fallback path. The job is advisory and not one of the four required checks. Data point for #818: that file has now gone red on 2 of 2 runs here, and previously on a Markdown-only commit (recorded on #906).

vfarcic:
**Scope change — #915, #916 and #917 are being folded into this PR** rather than left as follow-ups, so one review round covers the whole family. @prageethw has not reviewed yet, so this costs nothing in re-review.

There is also a substantive argument for it: **#915 and #916 are each a compatibility event.** Shipped separately, operators absorb two of them. Together, this becomes a single release that closes the entire identity-authorization family.

### What this PR turns into

| | before | after |
|---|---|---|
| changelog | `617.bugfix.md` | **`617.breaking.md`** |
| `PROTOCOL_VERSION` | 8 | **likely 9** — `GetSeedRequest` is a wire type and #916 needs an identity on it |
| release bump | patch | **minor** (while `0.x`, breaking → minor) |

Whether #916 truly requires the wire change is being **established rather than assumed** — the daemon receives only `pane_id` and cannot know who is asking unless the client says so, but if an existing route attributes a hook-socket connection to an agent, that avoids the bump. Reporting either way.

### Rule 12 re-runs

The cross-version manual test passed earlier for the *previous* change set and will be re-run. With a `PROTOCOL_VERSION` bump it changes character: an old daemon should now **reject** the new client at handshake rather than silently degrade — which is the point of bumping, and is what the re-run will verify.

### The one way this goes wrong, called out up front

#915's finding 4 wants an ended-generation witness. Keyed by **pane** it is a disaster: `AppState::pane_generation_closures` is never cleaned on pane close (all six mutation sites enumerated; the only `.remove()` calls are a snapshot-and-restore inside `apply_daemon_report_event`), so a pane-keyed witness would permanently refuse every future sessionless agent on that pane id — the exact behaviour the carve-out exists to protect. It must be keyed by **agent id**, which is already available: `AgentEvent.agent_id` (`src/event.rs:756`) is on the wire today, and registry ids are monotonic and never recycled within a daemon (`src/agent_pty.rs:5042-5043`).

For #915's finding 5 the remedy is a **deletion** — dropping the `None if has_live_attach` arm retires the only production consumer of `pane_has_live_attach` rather than synchronising the race. The alternative (making `subscribe` take the writer) was assessed and rejected: lock-order inversion against `write_guarded`'s writer→inner ordering, plus a reintroduced TOCTOU from splitting phases.

Unrelated, for the record: `e2e-deterministic` remains red, and that is **#818** — red on every branch *and on main* since #789. Verified against main directly: it failed on `d1baecd4` (this branch's base), `79198191` and `d6396bad`. Not this PR's.

vfarcic:
**Correction to my previous comment: `PROTOCOL_VERSION` stays 8.** I wrote "likely 9" and flagged it as an inference to be verified. It was verified, and it was wrong.

`GetSeedRequest` travels the **hook socket**, not the attach socket, and `PROTOCOL_VERSION` is scoped to the attach socket. Its own doc exempts this exact field class — *"Additive `#[serde(default, skip_serializing_if = \"Option::is_none\")]` fields do NOT require a bump — they're forward-compatible by design"* (`src/daemon_protocol.rs:143-155`). The sibling message says the same in-tree: `ListTargetsRequest`'s doc reads *"Additive on the hook socket, so it does NOT move the attach `PROTOCOL_VERSION`"*. No `deny_unknown_fields` applies, and the additivity is pinned by a test in both directions rather than argued — a new client with no id omits the key entirely (so a pre-#916 daemon receives byte-identical JSON), and a legacy payload deserializes to `agent_id: None`.

It also avoided a collision I had not considered: **`agent/dispatch-issue-882` is unmerged and plans its own 8 → 9**, so taking 9 here would have clashed.

So the final classification is: **`617.breaking.md`, minor bump, `PROTOCOL_VERSION` unchanged at 8** — a semantic break behind a stable wire, the same call #608 itself made (`src/daemon_protocol.rs:2169`).

One consequence for the rule 12 re-run, which I had also anticipated wrongly: with no bump, an old daemon does **not** reject the new client at handshake. The cross-version test keeps its original character rather than becoming a rejection test.

### Where the three items landed

- **#915 finding 5** — deleted the `None if has_live_attach` arm, retiring the only production consumer of `pane_has_live_attach`.
- **#915 finding 4** — `agent_generation_closures`, keyed by **agent id**, read under the same `AppState` guard as the generation. The pane-keyed trap was re-enumerated independently before implementing: one declaration, one accessor, one insert, two call sites, and exactly one `.remove()` — inside `apply_daemon_report_event`'s snapshot-and-restore, not a cleanup path. Absent agent id fails **open** to today's behaviour, stated in the changelog as a known limit.
- **#916** — both halves. The `exited` filter is the load-bearing one and was **reproduced**: reverted, the successor is handed the predecessor's seed verbatim. Identity scoping is defence in depth on top. Absent id **falls back rather than refuses**, deliberately — with the filter, the id-less arm resolves to *the* unique live occupant, so it is a complete resolution rather than a skipped check, and refusing would cost a lost opening task for any producer the daemon injected no id into.
- **#917** — `write_to_pane_and_submit` **deleted**; all 7 call sites migrated, including the lane-2 one, which was **run against a real agent** rather than assumed. `write_to_pane_notice` is retained `#[cfg(test)]` — stronger than the issue's rename option, since a future production call site becomes a *compile error* rather than a visible decision.

`cargo test-fast`: **2413/2413** (+5). clippy (4-flag), fmt, `linkage-check` green. Scoped e2e 18/18 and 10/10.

vfarcic:
## Round-2 review + audit: no blockers found, but I am treating one audit finding as one

Both reviewers agree the change set **closes the security property it claims**, and independently verified the load-bearing calls — the `PROTOCOL_VERSION` reasoning (all five legs), the agent-scoped keying of the new witness, and #916's uniqueness guarantee (`spawn_agent`'s duplicate check rejects a second `!exited` agent on a `pane_id_env` *inside the same lock acquisition as the insert*, and the id-less scan uses the same predicate under that lock — so the fallback is a complete resolution, not a skipped check).

**But the audit found that #915's new witness is forgeable, cross-pane, and permanent, and I am upgrading it from the auditor's own "suggestion" to a must-fix.**

### The chain, traced branch by branch

1. Post a `SessionStart` for an **invented** pane `P'`. `generation_ownership` → `Unclaimed`, `registration_admits(P')` → false, so it takes the `SessionStart` branch and **inserts `P'` into `managed_pane_ids`** (`src/state.rs:7053`).
2. Post a `SessionEnd` for `P'` naming a **victim agent on a different pane**. `generation_ownership` returns `Unclaimed` on the pane mismatch — but `registration_admits(P')` is **now true**, so it is admitted, and `note_agent_generation_closed(victim)` runs.
3. Nothing ever removes the entry.

Every unnamed automatic delivery bound to that agent is then refused, permanently. Registry ids are sequential decimal strings from `"1"`, so `1..N` is a single pass — and future agents can be poisoned pre-emptively.

**Why it is new rather than pre-existing:** the pane-keyed sibling `pane_generation_closures` is poisonable by the same route, but its poison is pane-local and is consumed as a *comparison against a snapshot*. This is the first daemon state where a forged event on one pane produces a **permanent adverse decision about an agent on a different pane** — and before #915 the same forgery made the guard *more permissive*, not less. That inversion is the change.

### Why "suggestion" was the wrong severity

The auditor marked it below blocker because it requires hook-socket access, which is a documented pre-existing unauthenticated same-uid boundary — while explicitly inviting disagreement. I disagree: **this PR's threat model is a prompt-injected, tool-capable agent**, which is exactly a process that can open that socket. Shipping a permanent, forgeable, cross-pane latch in order to close a sub-second race is a bad trade, and it lands hardest on precisely the class the carve-out exists to protect — an agent that never announces a generation would never receive its opening prompt again.

The fix is cheap and is the repo's own stated doctrine (`src/event.rs:558-562`: *"Provenance is established by the DAEMON, at the site that acts on it"*): record the witness only when `oracle_ownership(pane, agent)` is `Owned`. Free on the honest path, and it degrades to the existing fail-open where no oracle is installed.

### Also being fixed

- **The witness is permanent but code and changelog both describe it as transient** (*"has not announced itself yet"*, *"in the gap"*). Being evicted when the agent's registry record is removed — safe because ids are never recycled, and it cannot re-open the hole since a removed agent resolves to `NoLiveTarget` before the closure runs.
- **`apply_daemon_report_event` restores two generation maps and not the third.** Not reachable today (its only production caller synthesizes `EventType::Error`, verified), but the asymmetry matters: the siblings are *counters* compared against a snapshot, this one is a *latch*.
- Four nits: a changelog sentence that asserts something the check *can* actually do; three broken intra-doc links; an empty `DOT_AGENT_DECK_AGENT_ID` silently costing the seed; and a missing availability-cost paragraph plus a paned-only scope clause.
- The reviewer's suggestion that the `/clear`-gap case (same agent id, `pane_hook_session` momentarily `None`) is now refused too and is not in the changelog.

Pushing once these land, so this PR spends one more CI cycle rather than several.

vfarcic:
## Final round green — ready for review @prageethw

Head `f87ea5e7`. **All four required checks pass.**

| check | result | duration |
|---|---|---|
| build | pass | 5m52s |
| build-macos | pass | 5m38s |
| build-windows | pass | 5m14s |
| security | pass | 23s |

Local on the committed tree: `cargo test-fast` **2414/2414**, four-flag `clippy` clean, `fmt` clean, `cargo xtask linkage-check` ok.

### What the last round fixed

The round-2 audit found that #915's own fix introduced a worse problem than the one it closed: the new ended-generation witness was **forgeable, cross-pane and permanent**. Post a `SessionStart` for an invented pane (which inserts it into `managed_pane_ids`), then a `SessionEnd` naming a victim agent on a *different* pane — `generation_ownership` returns `Unclaimed`, but `registration_admits` is now true, so it is admitted and the witness is recorded against the victim. Registry ids are sequential from `"1"`, so `1..N` is one pass.

Before #915 that same forgery made the guard *more permissive*. After it, permanently restrictive. `8be02cf0` closes it by establishing provenance before witnessing — the repo's own doctrine (`src/event.rs:558-562`) — and adds a regression test that the coder verified fails with the guard reverted.

Also in: the third generation map is now restored in `apply_daemon_report_event` (latent, but a *latch* rather than a *counter*, so the asymmetry mattered), four prose/rule-17 corrections, three broken intra-doc links, and an empty `DOT_AGENT_DECK_AGENT_ID` no longer silently costing the seed.

Two places the implementer **overruled its reviewers, correctly**, both recorded here because the reasoning is load-bearing:

- The auditor's suggested predicate would have broken two of the three #915 tests — `oracle_ownership` returns `None` with no oracle installed (bare `AppState` in tests), so the literal `matches!(…, Some(Ownership::Owned))` refuses there. The auditor's *prose* said "degrades to the existing fail-open where no oracle is installed"; the shipped code follows the prose.
- I suggested evicting the witness on registry-record removal to bound the map. That was wrong: there are **two** `agents.remove` sites and *neither fires for an agent that simply exits*, so eviction would bound nothing while changing no decision (a removed agent is already refused earlier by `write_guarded`). Documented permanence instead; the unbounded-growth half is closed by the provenance fix, since keys can now only be daemon-verified registry ids.

### Rule 12

Cross-version manual test **re-run in both directions** after the wire gained a field, with both false-green traps avoided (one `Attach protocol listening` line, unchanging daemon pid and `/proc/stat` start time, real build-id mismatch declined). Wire bytes captured directly rather than inferred:

```
branch client : {"message_type":"get_seed","pane_id":"1","agent_id":"1"}
v0.39.3 client: {"message_type":"get_seed","pane_id":"1"}
```

Both directions serve the seed. And the fallback is not vacuous — against the branch daemon on one seeded pane: `agent_id=7` (a live agent, wrong pane) → refused; `agent_id=999` → refused; **no id** → seed returned, and still there after the refusals, so the refusal is non-destructive.

Verdict unchanged: **`617.breaking.md`, minor bump, `PROTOCOL_VERSION` stays 8.**

### Still red, still not ours

`e2e-deterministic` failed on `orchestration_remit_001` — **#818**, verified red on `main` itself (`d1baecd4` — this branch's base — plus `79198191` and `d6396bad`). Three of three runs on this PR, always the same remit test, and no commit here touches remit/compaction/`apply_event`. The job is advisory, not one of the four required.

Also filed while working, both pre-existing and neither blocking: **#921** (`delegate_015`'s readiness needle is ASCII `"Ask anything..."` but OpenCode now paints U+2026 — invisible to CI because it is lane 2) and **#922** (`OPENCODE_TEST_MODEL_DEFAULT` points at an exhausted model).

prageethw (approved):
Read the full diff (19 files, ~450 net lines in `src/`) and the PR body's own accounting of the 14 sites. The core fix is exactly right: `write_and_submit_guarded`/`write_and_submit_guarded_detailed`/`write_notice_guarded` moving `expected_agent_id` from `Option<&str>` to `&str` removes the permissive case from the type rather than a comment, and the identity gate at `write_to_pane_internal` (formerly behind `let Some(expected) = expected_agent_id`) is now unconditional on the paned arm — `if !is_paneless && expected_agent_id != target.agent_id`.

The seed-fallback take (`take_pending_seed_fallback_for`) correctly distinguishes `Taken`/`StrangersSeed`/`Absent` under one lock acquisition, keyed by agent id rather than a linear pane-id scan — this closes the real subtlety Greptile flagged (the old scan could consume a *successor's* seed and drop it via `WrongSession`, costing that successor its opening task even though no bytes leaked). `deliver_on_idle`'s reuse-decision fix (resolving the pane's live agent and intersecting with the stored `ReuseEntry.agent_ids`) and `handle_work_done`'s refuse-on-disagreement change both match the described threat model, and the `DelegationRetirement::Superseded` arm now carries the orchestrator pane+agent id together rather than letting a caller bind mismatched ones.

Verified independently rather than taking the PR body's word for it:
- `write_to_pane_and_submit` is gone from the diff (not just deprecated), and `write_to_pane_notice`/`take_pending_seed_fallback` are `#[cfg(test)]`-gated in the new diff — so a future permissive call site is a compile error, not a review miss, which is the actual point of #617's "two reviews each missed different sites" framing.
- This PR's own `e2e-deterministic` CI failure (`orchestration_remit_001_start_role_compaction_reasserts_remit`) is the same pre-existing race that #837 fixes elsewhere in this review batch — confirmed by reading this PR's own job log, not assumed from the PR body's claim. Unrelated to this diff.
- Rule 12 accounting holds: `PROTOCOL_VERSION` unchanged at 8, and the three retyped structs carry no serde derives, so no wire value is newly refused — a same-wire/different-meaning break, correctly paired with `changelog.d/617.breaking.md`.

One documentation nit, non-blocking: the PR body's "Rule 12 — daemon-internal, no bump, no breaking fragment" section header is stale — the diff *does* add `changelog.d/617.breaking.md` (correctly, since this is exactly the same-wire/different-meaning case that fragment type exists for). Worth fixing the header text before merge so a future reader doesn't take it as "no breaking-behavior fragment needed" when one was in fact added.

The unrelated `eee99099` commit (auditor role model swap in `.dot-agent-deck.toml`, GPT credits exhausted) is isolated to that one file/line as claimed and does not touch anything security-relevant — fine to ride along or drop, author's call.

LGTM.

prageethw (dismissed):
Rigorous approach — closing the identity gap by type (Option<&str> -> &str) rather than by review, using a cold compile's 31 errors as the authoritative call-site list, is exactly the right response to '#617: two independent reviews each found sites the other missed'. 14 production sites closed vs. 5 named in the issue. Correctly flagged as a behavior-breaking (not wire-breaking) change with PROTOCOL_VERSION unchanged and a thorough changelog.d/617.breaking.md documenting every new refusal cost. No unresolved review threads. LGTM.

Issue #241, opened 2026-07-28
PRD: Safe pane closing — scope Ctrl+W to command mode, confirm destructive close, unwedge NotFound


vfarcic:
## PRD: Safe pane closing

**Problem**: `close_pane` (default `Ctrl+W`) resolves from the global keybinding layer with no mode guard, so pressing it out of muscle memory while typing in a shell or agent instantly tears down the pane — `Ctrl+W` is delete-previous-word everywhere else — and the close is immediate, unconfirmed, and irreversible.

**Solution**: Scope `close_pane` to command mode so `Ctrl+W` forwards to the PTY in `PaneInput` (mirroring the existing `Ctrl+C` carve-out), add a confirmation dialog for the deliberate close, and fix the close path so a `NotFound` stop-agent response is treated as already-stopped instead of wedging the pane forever.

**Consolidates**: #88, #192, #218

**Detailed PRD**: See [prds/done/241-safe-pane-closing.md](https://github.com/vfarcic/dot-agent-deck/blob/main/prds/done/241-safe-pane-closing.md)

**Priority**: High

vfarcic:
Shipped in [PR #258](https://github.com/vfarcic/dot-agent-deck/pull/258), squash-merged as `a005907`.

What changed:
- `close_pane` (`Ctrl+W`) now resolves only in command mode — inside a pane it reaches the shell/agent as ordinary word-delete instead of tearing the pane down.
- Every close now goes through a confirmation dialog (keybinding, `[Close]` button, tab-strip `×`, and the modal's own `[Close]`), defaulting to Cancel.
- The dialog's wording matches the real blast radius: a plain pane reads "Close selected pane?"; a multi-pane orchestration tab (including one reached via a dashboard card) reads "Close this tab and all its panes?".
- A stop-agent `NotFound` is treated as already-stopped, so a stale/ghost card closes cleanly instead of wedging on a retry that could never succeed — including the slot-owner-chase case where a second respawn lands mid-close.
- Tab teardown now closes panes first (concurrently) and only removes the tab on a clean outcome; a partial failure keeps the tab with just the panes that couldn't be stopped.
- Command mode is now discoverable via mode-aware hints/help, and `Ctrl+D` is a genuine two-way dashboard/pane toggle.

Consolidated and closed #88, #192, #218 along with this one — thanks to everyone who reported it.

[...]

Issue #835, opened 2026-09-02
A scheduled task cannot choose its spawn shape, so a schedule in a repo defining [[orchestrations]] silently ignores its own command and fires a full team


vfarcic:
## Problem

A scheduled task cannot choose its spawn shape. The shape is derived entirely from the target directory's config, so **any schedule whose `working_dir` defines `[[orchestrations]]` fires a full orchestration tab and silently ignores its own `command`** — there is no way to say "fire one agent here".

`src/daemon.rs:900-903` hardcodes the absence of a selector:

```rust
// `None`: a scheduled task's shape still comes from its working dir's
// config. The PRD #220 selector is a `dispatch`-only surface.
resolved_target: None,
```

With `resolved_target: None`, `spawn` falls through to `decide_target` (`src/spawn.rs:374`), which returns an `Orchestration` target whenever `default_orchestration` finds a role-bearing block. `SpawnRequest.command` is then ignored (the role commands win), exactly as documented in `docs/scheduled-tasks.md`'s field reference.

## This is already biting a live schedule

On this repo, `.dot-agent-deck.toml` defines `mixed` with `default = true` and six roles. So a schedule pointed at the repo root fires **six panes** (orchestrator, coder, reviewer, auditor, tester, release) where one was intended.

My own `stale-prs` schedule is in exactly that state today:

```toml
[[scheduled_tasks]]
name = "stale-prs"
cron = "0 9 * * SAT"
working_dir = "/home/vfarcic/code/dot-agent-deck"
command = "devbox run agent"   # never used
```

It reads as a single `devbox run agent` card and has never once run as one. Nothing in the config, the CLI output, or `schedule list` says so — the only way to find out is to read `decide_target`.

## Why the workarounds are all bad

Three options exist today and each gives something up:

1. **Accept the orchestration.** Five idle worker processes per fire, and the orchestrator's role template opens with *"You NEVER do implementation, review, or audit work yourself — only delegate"*, which is not what a scheduled report or triage sweep wants. It also walks straight into #703 (*a dispatched orchestration inherits the interactive role template's human gates, so a unit can stop forever waiting for a user who is not there*) and #708 (an unattended unit cannot act on worker-exited notices). A scheduled orchestration is unattended by construction, so it is in that same class.
[...]

PR #924, opened 2026-09-06
fix(ui): arm close confirmation on the agent generation, not the session id (#317)

### LINKED ISSUE #317: Close confirmation arms on session id alone, so it can act on a replacement generation
## Problem

`CloseTarget::Session` captures the **session id** at arm time, and `resolve_close_plan` resolves it later by a direct key lookup. Pi's producer key, `{pane_id}-session` (`src/main.rs:727-748`), is **stable across respawns** — it does not change when the pane's agent generation changes. So an armed close target stays resolvable after a generation change, and confirming the close can act on whichever generation currently occupies the pane rather than the one the user actually armed against.

This is exactly the retargeting hazard `PRD #241` and `prompt/close-confirm/005` exist to guard against.

## Pre-existing, not introduced by #284

This predates #284 (https://github.com/vfarcic/dot-agent-deck/issues/284, shipped in PR #314) and is neither introduced nor worsened by it:

- **Before #284**: the armed key resolved to a stale corpse entry.
- **After #284**: it resolves to the live replacement.

In both cases, the resolved card is the pane's *current* card — the confirm action lands on the generation occupying the pane at confirm time, not necessarily the one armed at arm time.

## Why this is fixable now

#284's same-producer identity refresh means `agent_id` is now refreshed **in place** on the stable session-id entry, so a generation change is finally *detectable*. Before #284, the stale `agent_id` would have matched whatever identity was armed and concealed the fact that a generation change had even happened. #284 is a prerequisite for fixing this properly — it is not the fix itself, and deliberately does not attempt to be (see "Why #284 deliberately did NOT fix it" in #284 for the related discussion of narrowing the retire seam itself).

## Proposed fix

Arm on the **generation** — session id *plus* agent id — instead of the session id alone, so that resolving against a refreshed generation reads as "vanished" rather than "still there." This belongs at the arm/resolve seam (`CloseTarget` / `resolve_close_plan`), **not** at the #284 retire predicate in `AppState::apply_event`.

## Testing note

This needs close-flow coverage in the `prompt/close-confirm/*` tier (L2), not a state-level test:

- `status/supersede/002` is only a state-seam guard — it does not execute `CloseTarget` / `resolve_close_plan` at all.
- `prompt/close-confirm/005` remains the close-flow proof and should stay green.

During #284's review, a test named `status/supersede/006` was written to try to cover this at the state seam, then **removed** as logically unsatisfiable there — it demanded the same `HashMap` key be simultaneously present (for the stale generation) and absent (for the "vanished" armed target), which directly contradicts `status/supersede/005`. Its gap is documented in the "Does not assert" line of the `status/supersede/005` catalog entry in `tests/CATALOG.md` — that is the deliberate, documented gap at `/006`. Any fix here should add the missing close-flow-level test rather than resurrect `/006`.

## References

- Issue #284: https://github.com/vfarcic/dot-agent-deck/issues/284
- PR #314: https://github.com/vfarcic/dot-agent-deck/pull/314
- `src/main.rs:727-748` (Pi's stable producer key)
- `tests/CATALOG.md` — `status/supersede/005` entry

vfarcic:
Closes #317.

## The defect

`CloseTarget::Session` captured the **session id** at arm time and `resolve_close_plan` resolved it later by a direct key lookup. That is sound only while a session id goes stale once the pane changes hands — and Pi's does not. Its `agent-event` reports arrive under the pane-derived `{pane_id}-session` producer key, which is *deliberately* stable across respawns, so an armed target kept resolving after a generation change and the confirmed close landed on whichever generation occupied the pane rather than the one the user pointed at.

This is the retargeting hazard PRD #241 and `prompt/close-confirm/005` exist to prevent, reached through a key that never goes stale rather than through a moving selection. It predates #284 and is neither introduced nor worsened by it — before, the armed key resolved to a stale corpse; after, to the live replacement; in both cases to the pane's *current* card.

## The fix

`CloseTarget::Session` now carries the `agent_id` alongside the session id, and `resolve_close_plan` reads a refreshed identity as **vanished** — the same `Nothing closed — what this confirmation was armed against is gone` path a session id that genuinely disappeared already took.

#284 is what makes this detectable at all: it refreshes `agent_id` **in place** on the surviving stable-key entry. Before it, the stale identity would have matched whatever was armed and concealed that a generation change had happened.

**The comparison is deliberately the same test `AppState::apply_event` applies: a generation change is a differing `Some` → `Some`, and nothing else.** The two seams have to agree on what a generation *is*. A stricter rule here would make a card that merely **learned** its agent id between arm and confirm read as vanished, so an ordinary close would silently do nothing on a race the user can neither see nor control — and the state seam already ruled that transition out of scope in the same words. `same_generation` carries the reasoning and enumerates all four cases, including the `Some` → `None` arm it does *not* claim to have designed for.

Fixed at the arm/resolve seam, **not** at #284's retire predicate, as both the issue and the `apply_event` comment asked: deleting the card so the armed id reads as vanished would leave zero cards on a live pane, which `status/supersede/003` forbids. The `apply_event` comment that pointed forward to this fix now points *at* it and records that the two must move together.

## Rule 12 — answered, not skipped

**No `PROTOCOL_VERSION` bump and no `.breaking.md` fragment.** Evidence rather than assertion:

- `CloseTarget` and `ClosePlan` are referenced in **no** daemon/protocol file — `grep -rn "CloseTarget\|ClosePlan" src/daemon_protocol.rs src/daemon.rs src/daemon_client.rs` is empty.
- `CloseTarget` derives `Debug, Clone, PartialEq, Eq` — no `Serialize`/`Deserialize`, so it cannot reach the wire.
- `PROTOCOL_VERSION` stays **8**; `src/daemon_protocol.rs` is untouched.
- The Stop requests a confirmed close sends are unchanged in shape *and* meaning. What changed is only whether the TUI decides to send them, which is a TUI-local decision. A new TUI against an old daemon simply declines to send a Stop in one more case; an old TUI against a new daemon is unaffected.

`src/state.rs` is **comment-only** — verified by diffing with comment lines excluded.

## Tests

Rule 4: the change is user-visible (the close-confirmation outcome), so it carries L2 close-flow coverage as #317 asked for, rather than a state-level test.

**New — `prompt/close-confirm/009`** (L2, lane 1, credential-free): drives the real binary through **two `SessionStart` frames under one producer key differing only in `agent_id`**. Both generations are identical in every field the dashboard renders, which is the point — the test cannot pass by noticing an agent-type or session-id change, because there is not one. Its sync point is gen-1's prompt history disappearing, which is #284's identity refresh landing.

**New unit test** on `resolve_close_plan` pins all four controls: same generation still closes, refreshed generation reads as vanished, a `None`-armed target does not invent a change, and a gone session id still vanishes.

Catalog updated: `/009` added, and `status/supersede/005`'s "Does not assert" line — which documented precisely this gap — narrowed to point at `/009` instead of describing an open hole.

### Gates run

| gate | result |
|---|---|
| `cargo clippy --workspace --all-targets --features e2e,e2e-live -- -D warnings` | pass, 0 warnings |
| `cargo test-fast` | **2402 tests run, 2402 passed, 0 skipped** (139 binaries) |
| `cargo xtask linkage-check` | ok — 627 catalog ids, 557 annotations, 11 rules |

### Scoped tests (rule 5), located via `tests/CATALOG.md` and the `#[spec]` annotations

Lane-1 e2e, run locally — **named because the part of the tier CI does not run is on trust**:

- `tests/e2e_pane_close.rs` — `dashboard/pane/002`, `dashboard/pane/003`, `prompt/close-confirm/002`, `/003`, `/004`, `/005`, **`/009` (new)**
- `tests/e2e_tab_close_regressions.rs` — `prompt/close-confirm/006` (the `CloseTarget::Session`-that-is-really-a-tab path)
- `tests/e2e_mouse_tabstrip.rs` — `mouse/tabstrip/003` (the `CloseTarget::Tab` arm, unchanged by this)

In the fast tier: `tests/render_close_confirmation.rs` (`prompt/close-confirm/001`, `/007`, `/008`), `tests/card_supersession.rs` (the state seam whose comment changed), and `src/ui.rs`'s `dashboard/selection/012`, `/016` plus the new `resolve_close_plan` test.

**No lane-2 test was run and none was needed** — nothing here touches an agent, daemon, protocol or credential path.

### On the scoped e2e run, and why it went red once

Worth stating plainly rather than quietly re-running until green. The 13-test scoped set **failed 13/13 on its first attempt and passed 13/13 on the second, with no code change between them.** The timings are the whole story:

| | first attempt | second attempt |
| --- | --- | --- |
| every test | 14.5 – 15.3s (all FAIL) | **0.82 – 4.68s (all PASS)** |
| `close_confirm_009` | 15.135s FAIL | **1.264s PASS** |
| whole run | 15.314s | **4.679s** |

Every failure was a timeout on the **first** wait — `[Command Mode Ctrl+D]`, `No active sessions`, `alpha-agent`, `demo-agent` — i.e. nothing rendered at all, before any close logic ran. None was an assertion about close behaviour. Sampled during that window the box was at `io full avg10` **26–42** with `cpu some avg10` ~0.2 and no compiler running: I/O starvation, the #818/#415 class, not a defect in this change. This run shared the machine with a second agent (measurement context below), which is why it is worth recording rather than hiding.

**The new test was independently confirmed to have teeth**, which is the part that matters for trusting it. I disabled *only* the `same_generation` guard and re-ran it: it failed at `did not see "Nothing closed" within 10s` — precisely the assertion that pins #317 — then passed again with the guard restored (the file verified byte-identical to its pre-experiment state). So `/009` fails for the right reason on unfixed code, rather than merely passing on fixed code.

## Measurement

This PR was also a cost sample for #906 at the **N=2 concurrent** level; the numbers are posted there, not here.


vfarcic:
## CI analysed: the two failures are not from this change

`e2e-deterministic` on `d122745a` reported **2591 tests run, 2589 passed, 2 failed**. Both failures are `orchestration_remit_001` (42.494s) and `orchestration_remit_002` (23.094s), which this PR does not touch — it changes the close-confirmation arm/resolve seam and adds close-flow tests.

**The tests this change actually covers all passed on the runner:**

```
PASS [ 0.435s] (1738/2591) e2e_pane_close close_confirm_005_vanished_armed_session_closes_nothing
PASS [ 0.356s] (1739/2591) e2e_pane_close close_confirm_009_stable_key_respawn_closes_nothing
PASS [ 0.315s] (1740/2591) e2e_pane_close pane_002_confirmed_ctrl_w_removes_card
```

All four **required** checks are green: `build`, `build-macos`, `build-windows`, `security`. `e2e-deterministic` is advisory (CLAUDE.md rule 8).

**This is the pre-#907 baseline, not a regression.** PR #907 — which fixes exactly this starvation — is still **open and unmerged**, so its containment is not on `main`. My base `d1baecd4` carries only #818's older `orchestration-remit = { max-threads = 1 }`, not the `threads-required` / `priority` override #907 adds. So this is an independent reproduction of a live flake on a branch that does not contain its fix, which is the same thing #835's write-up recorded and was careful not to overstate.

The durations are the usual tell: 42.494s and 23.094s failing, against the 1.5–2.6s these tests take when scheduled with the pool reserved. They are starved, not slow. Per CLAUDE.md rule 5 I analysed the CI run rather than reproducing lane 1 locally.

## One further commit since Greptile reviewed

Greptile reviewed `d122745a` (5/5, no inline findings) and does not re-review on push (`triggerOnUpdates: false`). `dffc714d` is **test-only** and is called out here so the delta is not invisible: it replaces `close-confirm/009`'s synchronisation barrier — which waited for generation one's history to *vanish* — with a positive wait on a marker only the generation-two frame can produce. An absence wait is also satisfied by a card that is covered or not yet drawn, so it was a race dressed as an observation. It was sound as written (verified), but it would have stopped being sound the moment anyone widened the modal, with no test failing to say so.

vfarcic:
## Run 2 (`dffc714d`) — ready for review

All four **required** checks green again: `build` 322s, `build-macos` 361s, `build-windows` 344s, `security` 20s.

`e2e-deterministic` reported **2591 tests run, 2590 passed, 1 failed** — down from 2 failures on run 1, and the one remaining is again `orchestration_remit_001`, which this PR does not touch. The new test passed on the runner both times: `close_confirm_009` at 0.356s (run 1) and 0.383s (run 2).

Run 2 makes the starvation reading harder to argue with, because the spread is now visible **inside a single job** rather than across runs:

| test, run 2 | duration | result |
|---|---|---|
| `remit_001` | **46.896s** | **FAIL** |
| `remit_002` | 2.868s | PASS |
| `remit_003`–`007` | 2.381 – 4.056s | PASS |

Same runner, same job, same minute — one test at 46.9s while its six siblings finish in 2.4–4.1s. And `remit_002` itself went 23.094s-failing (run 1) → 2.868s-passing (run 2), an 8x spread on near-identical code. That is #907's territory, on a base that does not carry #907.

No review threads are open — Greptile reviewed `d122745a` at 5/5 with zero inline findings, and correctly did not re-review on push. Nothing here is waiting on me.

@prageethw this is ready whenever you have time. The two things most worth your eye:

1. **`same_generation`'s `_ => true` arm.** It deliberately mirrors `AppState::apply_event`'s rule that a generation change is a differing `Some` → `Some` and nothing else. The stricter alternative — treating a `None` → `Some` identity-learning transition as "vanished" — would make an ordinary close silently do nothing on a race the user cannot see. I argued it as a coupling between the two seams and said so in both the code and the `apply_event` comment, but it is a judgment call and it is the one I would most want a second opinion on.
2. **The rule-12 answer.** I concluded no `PROTOCOL_VERSION` bump and no `.breaking.md`, on the grounds that `CloseTarget`/`ClosePlan` reach no daemon file and derive no serde impls. Evidence is in the PR description if you want to check the reasoning rather than the conclusion.

prageethw (approved):
Traced the identity through `resolve_close_target` → `CloseTarget::Session{session_id, agent_id}` → `resolve_close_plan` → `same_generation`. The fix is correctly scoped: it captures `agent_id` from the same snapshot read as `pane_id` (avoiding the prior two-lookup TOCTOU), and `same_generation`'s Some/Some-differ-only rule matches `AppState::apply_event`'s own supersession test, which is exactly the invariant that has to hold for the two seams to agree on what a generation is. The new unit test (`resolve_close_plan_treats_a_stable_key_generation_change_as_vanished`) and L2 test (`prompt/close-confirm/009`) both pin the four cases described in the doc comment, including the deliberately-unasserted `Some`→`None` arm. Rule 12 evidence (no `CloseTarget`/`ClosePlan` reachable from daemon/protocol files, no Serialize/Deserialize) is convincing for skipping the protocol-version bump. The lone e2e-deterministic failure is `orchestration_remit_001`, the pre-existing race already tracked and fixed elsewhere in this review batch (#837) — unrelated to this diff. LGTM.

prageethw (approved):
Correctly identifies that a stable producer key (Pi's pane-derived session key) defeats session-id-based close-confirmation targeting, and the fix (carry agent_id alongside, treat a generation change as vanished) reuses the exact same_generation comparison AppState::apply_event already applies rather than inventing a second definition of 'changed'. No unresolved threads. LGTM.

prageethw (dismissed):
Correct diagnosis and fix: arming a close on session id alone worked only if the session id went stale on a pane handover, but Pi's producer key ({pane_id}-session) is deliberately stable across respawns, so the confirmed close could land on a different agent generation than the one the user pointed at. Carrying agent_id alongside the session id and treating a refreshed identity as vanished reuses the existing 'what this was armed against is gone' path rather than inventing a new state. The same_generation comparison is deliberately kept identical to apply_event's own generation-change test (Some→Some differing) rather than over-tightening to also flag a card that merely learned its id, which is correctly identified as a different, out-of-scope transition. Fixed at the arm/resolve seam as the linked issue and existing code comment both pointed to, not by touching the retire predicate, which would have violated status/supersede/003's zero-cards invariant. No findings raised.

PR #933, opened 2026-09-07
test(e2e): drop the orchestration_remit containment group (#818)

vfarcic:
Deletes issue #818's `orchestration-remit` containment group from `.config/nextest.toml`, which its own comment said to do "when that is found, not before". It has been found — **by PR #837, which merged this morning at 05:02Z**, hours before this branch started. That is the substantive fix and it is not mine; what this PR adds is the thing #837 structurally could not do, plus a correction to the record.

## The hypothesis in the containment comment is refuted

The group's comment named a suspected mechanism, and every open discussion of #818 has carried it forward as the standing lead:

> the daemon's `ListAgents`/live-status join never reporting the injected `Compacting` for `pane 1 (agent_id 1)`, identifiers every test in the file reuses

i.e. concurrent decks leaking into one another through reused identifiers. **It is wrong.** Measured on the pre-#837 file, that exact panic reproduces with **one deck in existence** — the single test alone, `-j 1`, under a 64-way CPU load on 16 cores:

```
[pre837-solo-001] r2 rc=100  the daemon's own ListAgents/live-status join never reported
                             Compacting for pane 1 (agent_id 1) within 30s.
DONE: 5/8 red
```

There is no second deck in that run to interfere with. It could not have been a leak in any case: each deck's hook socket is `work.join("hook.sock")` where `work` is a fresh 0700 `tempfile` per-test tempdir, and the daemon is addressed only through that same directory (`tests/common/mod.rs:589,775`) — so `pane 1` is scoped to one deck and an injection into it cannot reach another.

The real cause was event **ordering inside a single test**, which is exactly why serialising helped without fixing anything: the fixture appends a delivered pointer to its log *before* forking the `python3` that confirms the delivery, `apply_event`'s `Thinking` arm sets `session.status` unconditionally, and load decided whether that in-flight confirmation landed after the test's injected `Compacting` and destroyed it. A cap that removed load looked like an isolation fix. #837's barrier — wait for the confirmation's own *applied* footprint before injecting — is the fix.

## The verification #837 could not run

Every run in #837 had this group in force, so `max-threads = 1` serialised the module regardless of the `-j 4` on its command line. Full parallelism was therefore never tested. It is now, and the pre-#837 file is the control throughout:

| variant | concurrency | load | result |
| --- | --- | --- | --- |
| pre-#837 | full parallelism | 64-way | **2 of 6 rounds RED** |
| pre-#837 | solo, `-j 1` | 64-way | **5 of 8 rounds RED** |
| post-#837 | solo, `-j 1` | 64-way | 8 of 8 green |
| post-#837 | full parallelism | none | 2 of 2 green |
| post-#837 | full parallelism | 64-way | 12 of 12 green |
| post-#837 | full parallelism | 96-way | 8 of 8 green |

22 green module rounds at full parallelism, the loaded 20 with the load average climbing to **229** on 16 cores. The failing rounds carried both of #818's hazard classes and the documented varying subset (`_002`, `_004`, `_005`, `_007` across two rounds). Unserialised the seven now run in **6.5s** against the 18.7s the group's comment recorded for them serialised.

## Removing the cap changes tier interleaving — so the tier was run five times, with a control

Three full `cargo test-e2e` runs on this branch, then two more with the group restored as a control:

| run | group | failures |
| --- | --- | --- |
| 1 | removed | `dispatch_005` |
| 2 | removed | `dispatch_005`, `manager_016` |
| 3 | removed | `idle_worker_008` |
| control 1 | **present** | `dispatch_005` |
| control 2 | **present** | none — 2653/2653 |

**No remit test failed in any of the five.** `dispatch_005_respects_max_per_run` failed three times across *both* configurations, including with the group present, so it is not something the removal introduced — it is the co-tenant load contention tracked in #701 (which names `dispatch_005` and `manager_016` explicitly) and #818's own second problem class. Each passes when rerun alone per rule 6: `dispatch_005` in 0.31s against the 0.72s it took while failing in-tier, the duration-inflation signature #701 measures.

So this PR does not make `e2e-deterministic` green. It removes one of the two problems from the job and leaves #818 open for the other, which is what #837's own closing note asked for.

## Scope — what is deliberately untouched

- **The widened bounds stay.** They are positive waits on real PTY work, and #837 already resolved which of them to carry forward and why.
- **`pane_input_007` is not this cause**, despite #818 grouping it here. Its fixture emits no confirmation event at all (`tests/e2e_pane_send_result.rs:222` logs each line and nothing else), so nothing overwrites anything; its 10s wait covers a genuinely long chain — a `python3` fork plus the deck's retry cadence. That is a budget question, not a race, and widening it is explicitly not what this PR is for.
- **The 10s deck-launch wait** flagged in #818 as a second marginal bound never tripped in any run here, including at load average 138. Left alone rather than pre-emptively widened.
- **No production code changed.** The diff is three files: `.config/nextest.toml`, the existing `changelog.d/818.misc.md`, and `tests/e2e_work_done_reporting.rs` — the last added by `15a1b0fb` after this description was first written, which claimed "no production code and no test code changed" and was wrong about the second half from that commit onward. Nothing under `src/`. No TUI↔daemon contract question arises (CLAUDE.md rule 12), no `PROTOCOL_VERSION` bump, no `.breaking.md`.

## The second commit, and why it belongs here

`15a1b0fb` ("test(e2e): poll for each needle of the daemon's wrapped label") fixes `work_done_004`, which failed **this branch's own `e2e-deterministic` run**. It is in scope because it is one of the rotating families #818 tracks *separately* from the remit group — the same second problem class the section above leaves #818 open for — so it is the same job, the same issue, and the same kind of defect this PR is about.

The defect is a **single-shot grid read, not a timing budget**. The daemon writes the unsolicited-completion label as one long line, the orchestration surface wraps it over several rows, and its parts therefore become visible in paint order. The test waited for the *first* of them (`wait_for_pane_string(UNSOLICITED_NEEDLE, 20s)`) and then read the grid **once** for three later ones — asserting against a frame it had itself just proved was mid-paint. Each positive check now polls, bounded at the 20s the first one already used.

The negative check behind them stays a single read **deliberately**: polling for an absence passes on the first frame that has not painted it yet, most easily on the starved runner this test keeps failing on, and one read is sound there precisely because the three waits ahead of it have just proved the write finished painting.

**The evidence is a measured paint window, not a local red-to-green, and the PR should be read that way.** Polling from the instant the first wait returned, under 96-way CPU load: the last-painted needle was still absent for a further **27.6ms** and **37.4ms** in 2 of 5 runs while the earlier needles were already present. A single-shot read evaluates its predicate about one grid snapshot after that instant, so in those runs it would have read the sentinel as absent. CI's captured grid shows exactly that — the framing painted and cut off immediately before the sentinel. The old code was **not** reproduced as a local red (0 of 8 solo runs under the same load), which is consistent with the window only firing when the check lands inside it; the code comment says so rather than implying a reproduction that did not happen.

Scoped to that one file and existing helpers on purpose: PR #872 is mid-flight on `tests/common/mod.rs` for the #807 sweep, and touching shared harness code would also pull in rule 5's whole-tier obligation. #872 and #901 were both checked and neither touches this file.

## Gates

- `cargo test-fast` — 2462 passed, 0 failed (no `--exclude`; the desktop crate is in, per #815).
- `cargo test-e2e orchestration_remit` — 7 passed at full parallelism, 6.5s.
- `cargo test-e2e` in full — five runs, as tabulated above.
- `cargo xtask linkage-check` — ok (632 catalog ids, 562 annotations, 11 rules).
- `cargo nextest show-config test-groups` — empty, confirming no group survives and nothing silently references one.

Tests named per rule 5: `orchestration_remit_001`–`007` (`tests/e2e_orchestration_remit.rs`, lane 1), run solo, as a module at full parallelism, under 64-way and 96-way load, and inside five full lane-1 passes; plus `work_done_004_unsolicited_completion_is_visibly_labelled_in_the_attached_tui` (`orchestration/work-done/004`, `tests/e2e_work_done_reporting.rs`, lane 1) for `15a1b0fb`, run solo and as a module, and instrumented under 96-way load to measure the paint window quoted above. Lane 2 was not run — `tests/e2e_work_done_reporting.rs` is `#![cfg(feature = "e2e")]`, so both files are credential-free lane 1, and no shared harness file is touched.

Refs #818


vfarcic:
## Added: the `work_done_004` fix that reddened this PR's own `e2e-deterministic` run

At the maintainer's instruction, and after confirming it collides with nothing in flight.

**First, what that run actually says about this PR.** All seven `orchestration_remit` tests **passed**, at full parallelism, on a 4-vCPU runner, with the group deleted — the one configuration #818 said was untested, and the configuration its last comment predicted would be *worse* on a small box. Their timings overlap inside a ~16s window (`_001` 3.2s, `_004` 15.2s, `_005` 13.9s), so they genuinely ran concurrently rather than serialised. That hypothesis is now refuted from the runner side as well as locally.

The single failure was `work_done_004`, one of the seven rotating families #837's closing note left open on #818. It cannot have been caused by removing the group: `max-threads = 1` never reduced the tier's total parallelism — nextest runs `num-test-threads` tests concurrently either way, and the group only forbade two *remit* tests from sharing those slots. Total load is identical with and without it; only the mix changes.

**Checked for collisions before touching it**, per the instruction: PR #872 (#807's sweep) spans 23 files and none is `tests/e2e_work_done_reporting.rs`; #901 is desktop-only. #807's body never mentions this test either, so the call site was unowned. #872 *does* touch `tests/common/mod.rs`, so this fix deliberately uses only helpers that already exist in the test file — which also keeps it clear of rule 5's whole-tier obligation for shared harness edits.

**The defect is a single-shot grid read, not a timing budget.** The daemon writes the label as one long line and the surface wraps it over several rows, so its parts appear in paint order. The test waited for the *first* needle and then read the grid *once* for three later ones — asserting against a frame it had just proved was mid-paint.

Measured by polling from the instant that wait returned, under a 96-way CPU load: the last-painted needle was still absent for a further **27.6ms and 37.4ms in 2 of 5 runs**, while the earlier needles were already present. A single-shot read evaluates its predicate about one grid snapshot after that instant, so in those runs it would have read the sentinel as absent. CI's grid shows precisely that — framing painted, cut off immediately before the sentinel.

Each positive check now polls, bounded at the 20s the first already used. The negative check behind them stays a single read **deliberately**: polling for an absence passes on the first frame that has not painted it yet, most easily on the starved runner this test keeps failing on, and one read is sound there because the three waits ahead of it have just proved the write finished painting.

**One honest limit, stated in the code comment too.** I did **not** reproduce this as a local red on the old code — 0 of 8 solo runs under the same load, consistent with the window only firing when the check lands inside it. The fix rests on that measured window plus the CI grid, not on a local red-to-green. Worth weighing as such.

### Gates, re-run on the pushed tree

- `cargo test-fast` — 2462 passed, 0 failed.
- `cargo test-e2e` in full — **2653/2653 passed, 0 failed** (115.6s), the first fully green tier run of this branch.
- `cargo xtask linkage-check` — ok (632 catalog ids, 562 annotations, 11 rules).

Tests named per rule 5: `work_done_004_unsolicited_completion_is_visibly_labelled_in_the_attached_tui` plus `orchestration_remit_001`–`007`. Lane 2 not run — no credentialed test and no shared harness file is touched.

One consequence for review: Greptile reviewed at open and `triggerOnUpdates` is `false`, so this second commit carries **no automated review**. Worth a closer human read than the first.

vfarcic:
Cross-reference for merge order, no action needed here.

`work_done_004` is **3 of the 9** `e2e-deterministic` failures I measured across the last 28 runs on all branches — and two of those three were on Renovate JS-dependency branches (`renovate/major-vitest-monorepo`, `renovate/lucide-monorepo`) carrying no Rust, so this PR fixes a genuinely cross-cutting blocker rather than a local one. `e2e-deterministic` became a **required** check today, which makes it a merge blocker for unrelated PRs.

This PR and **#937** (`manager_016`) are the two smallest fixes and between them cover **7 of the 9** observed failures, so they are the top of the order. Verified this branch and #937 test-merge cleanly against each other.

Full lane status and order across #872/#933/#937/#895: https://github.com/vfarcic/dot-agent-deck/issues/818#issuecomment-5574871550

PR #937, opened 2026-09-07
test(e2e): settle the grid before comparing side-pane scrollback (#936)

### LINKED ISSUE #936: Flaky lane-1 e2e: single-shot snapshot_grid() comparisons race the repaint too, not just find_in_grid (manager_016)
Sibling of #807/#395, and **not covered by PR #872**.

#807 scopes the single-shot repaint race to `find_in_grid` call sites, and #872 sweeps those to `wait_for_in_grid`. But the same race reaches raw **`snapshot_grid()` comparisons**, which #872 does not touch — and `manager_016_wheel_over_dialog_does_not_scroll_side_pane` has one.

## Why #872 does not fix manager_016

#872 changes exactly one line inside that test — the `[Submit]` lookup at line 731, which runs **before the dialog is ever opened**. The read that actually fails is later and untouched:

```rust
deck.send_keys(b"s");
deck.wait_for_string("NEXT FIRE");
let before = visible_side_scroll_markers(&deck.snapshot_grid());   // single-shot
```

`NEXT FIRE` is the dialog **header**, so it appears before the overlay interior is painted. The centered dialog paints **over** the mode side pane and occludes marker rows, so a `before` sampled mid-paint sees a marker set the settled frame does not have — then `after_down` is compared against it and the assertion fails with no defect present. #872 adds `--retries 1` in CI, which would **mask** this rather than fix it.

## Evidence that this is occlusion, not a real scroll

Comparing a deliberately induced real leak against the observed CI failure. A genuine wheel leak scrolls the pane, so every marker shifts and the **occluded band keeps its width**; a half-painted overlay changes the band width while the marker above it stays put.

| | occluded band width | marker values |
|---|---|---|
| real leak (induced by disabling both wheel swallows) | `072→083` = 11, then `075→086` = **11, constant** | all shift by 3 |
| observed CI failure | `074→083` = 9, then `074→085` = **9 → 11, grew** | band start unchanged |

11 is the settled band width, so the failing `before` read caught the dialog 2 rows short of fully painted. A pane scroll cannot change the band width — only the overlay's painted extent can.

## Blast radius

`manager_016` failed `e2e-deterministic` on **four unrelated branches** in a 28-run window, one of them `renovate/pnpm-12.x` — a pnpm bump containing no Rust at all. With `e2e-deterministic` now a required check, this intermittently blocks unrelated PRs (it is currently the sole failure holding #895).

## Fix

Sample the marker set only once two consecutive reads agree (`wait_until_grid` polls at 20ms, far longer than a ratatui frame flush), on **both** sides of the comparison, and wait for the initial selection marker as well as the header so the overlay interior is drawn. This does not weaken the test: a genuine leak still moves the settled after-state relative to a settled before-state — verified by mutation.

Worth a follow-up sweep for other raw `snapshot_grid()` comparisons taken across an overlay paint; this issue fixes the one that is actively reddening CI.

vfarcic:
Closes #936. Fixes the `manager_016` flake that is currently the **sole** failure holding #895, and that has reddened `e2e-deterministic` on four unrelated branches.

## Why this is separate from #872

#807 scopes the single-shot repaint race to `find_in_grid` call sites, and #872 sweeps those to `wait_for_in_grid`. **The same race reaches raw `snapshot_grid()` comparisons, which #872 does not touch.** Its only change inside this test is the `[Submit]` lookup at line 731 — which runs *before the dialog is ever opened*. The read that actually fails is later and untouched:

```rust
deck.send_keys(b"s");
deck.wait_for_string("NEXT FIRE");
let before = visible_side_scroll_markers(&deck.snapshot_grid());   // single-shot
```

`NEXT FIRE` is the dialog **header**, so it appears before the overlay interior paints. The centered dialog paints **over** the mode side pane and occludes marker rows, so a `before` sampled mid-paint sees a marker set the settled frame does not have. #872 adds `--retries 1` in CI, which would **mask** this rather than fix it — so `manager_016` would most likely still fail after #872 lands.

## The cause was diagnosed, not assumed

I induced a real leak (disabling both wheel swallows while leaving the selection moving) and compared its signature against the observed CI failure. A genuine scroll shifts every marker and keeps the occluded band a fixed width; a half-painted overlay changes the band width while the marker above it stays put.

| | occluded band | markers |
|---|---|---|
| **real leak** (induced) | `072→083` = 11, then `075→086` = **11, constant** | all shift by 3 |
| **observed CI failure** | `074→083` = 9, then `074→085` = **9 → 11, grew** | band start unchanged |

11 is the settled band width, so the failing read caught the dialog **two rows short of fully painted**. A pane scroll cannot change the band width — only the overlay’s painted extent can.

## The fix

- Sample the marker set only once **two consecutive reads agree**, on **both** sides of the comparison — an unsettled sample on either side produces the same false failure. Sound because `wait_until_grid` sleeps 20ms between polls, orders of magnitude longer than a ratatui frame flush.
- Wait for the initial selection marker (`▶ alpha`) as well as the header, so the overlay *interior* is drawn before the baseline is taken.
- The settle deliberately accepts **any** value including an empty set: a genuine leak could scroll every marker out of view, and it must return that so the caller’s assertion reports the leak rather than spinning to a wait timeout that explains nothing.
- Failure messages now carry the settled **baseline** grid too, so a future occurrence is decidable from the log without re-deriving any of this.

Uses only existing harness primitives — **`tests/common/mod.rs` is untouched**, deliberately, so this cannot conflict with #872’s harness rewrite.

## The assertion is still live

This is the part that matters for a "make the test stop failing" change. With the settle in place, the induced leak **still fails** the test and reports both marker lists in full. The fix removes only the mid-repaint false positive.

| check | result |
|---|---|
| `manager_016`, unmutated | **6/6 pass**, runtime unchanged (~1.2s) |
| `manager_016`, real leak induced | **fails**, as it must |
| `scheduler_manager` + `mouse` families | **52/52** |
| `cargo test-fast` | **2462/2462** |

## Suggested merge order

This is small and independent. Landing it (with #933) unblocks #895 without needing #872 first, though #872 remains the right fix for the ~50 `find_in_grid` sites. A follow-up sweep for other raw `snapshot_grid()` comparisons taken across an overlay paint is noted in #936.


vfarcic:
Verified the merge-order claim rather than asserting it from line numbers.

**Test-merged this branch against both in-flight flake PRs:**

| against | text merge | composed tree builds + tests |
|---|---|---|
| #872 (`fix/find-in-grid-repaint-race`) | **clean** | **21/21** `e2e_scheduler_manager`, incl. `manager_016` |
| #933 (`agent/dispatch-flake-818`) | **clean** | — |

The #872 case is the one worth checking, since both PRs edit `tests/e2e_scheduler_manager.rs` and `manager_016` specifically. They compose: after merging, that test carries **both** fixes — #872's `wait_for_in_grid` on the `[Submit]` lookup and this PR's settle on the scrollback baseline — and it passes. `wait_until_grid`, which this PR's helper depends on, is untouched by #872, so a clean text merge is also a clean compile here (I checked, because those are not the same thing).

So this can land before, after, or alongside #872 with no rebase needed either way.

vfarcic:
Lane-wide context and where this sits in the order: https://github.com/vfarcic/dot-agent-deck/issues/818#issuecomment-5574871550

Short version — this PR and #933 are the two smallest fixes and between them cover **7 of the 9** `e2e-deterministic` failures measured across the last 28 runs on all branches. Neither depends on #872, and both test-merge cleanly against it.

vfarcic:
CI result, reported straight: **`e2e-deterministic` failed, but not on `manager_016` — and `manager_016` passed.**

```
PASS [ 1.121s] (1833/2653) dot-agent-deck::e2e_scheduler_manager manager_016_wheel_over_dialog_does_not_scroll_side_pane
```

2652/2653 passed. The single failure is **`form_002_add_spawns_authoring_agent_in_picked_dir`**, and it is at exactly the line #872 replaces:

```
tests/e2e_scheduler_manager.rs:979:10
    let (scol, srow) = deck
        .find_in_grid("[Submit]")
        .expect("the mode-locked schedule form must render a [Submit] button");
```

That is the #807 flagship case — a single-shot `find_in_grid` after an input event — and #872's diff replaces those three lines with `wait_for_in_grid("[Submit]")`. So it is #872's family, in #872's territory, and untouched by this PR: my diff only changes `manager_016`'s own reads plus a helper beside `visible_side_scroll_markers`.

**What this is and is not evidence for.** It is the first CI datapoint that the settle works — `manager_016` passed on a runner, at the same ~1.1s it takes locally. It is **one** sample, and this test only ever failed intermittently in CI, so it is not proof on its own. The actual argument for the fix remains the band-width diagnosis plus the mutation test (a real leak still fails the test), not this green.

**It also corrects the arithmetic in my #818 summary.** I wrote that #937 + #933 cover "7 of the 9" observed failures and should turn the lane green for most PRs. `form_002` did not appear in that 28-run window but has now, which makes #872's family bite more often than that survey suggested. The order in #818 is unchanged — #937 and #933 first, #872 still needed and still wanted — but #872 is doing more of the work than "2 of 9" implied. Noted on #818.

Re-running this job for a clean sample; the underlying fix for it is #872, not anything here.

vfarcic:
Closing in favour of folding this into **#895** (commit `954f941d`), which is the PR the flake was actually blocking.

**Why, since I argued the other way earlier.** Keeping it separate cost an extra serialized approval round, for a structural reason: `dismiss_stale_reviews_on_push` is `true`, so merging `main` into #895 after this landed would have voided #895's approval. That forces *approve #937 → merge #937 → merge `main` into #895 → only then request review on #895 → approve → merge* — two human approvals, the second unable to start until the first had merged. Fixing it inside #895 makes it one.

**And my stated reason for separation no longer applied.** I had written that patching the assertion inside #895 "would hide whatever raised this family's hit rate". That was a good reason while the cause was unknown — but I then measured it: a half-painted dialog overlay, not a wheel leak, and not #895's diff (the occluded band grew 9→11 with its start fixed, which a pane scroll cannot do). With the cause identified there was no signal left to preserve, so the argument for a separate PR had quietly expired and I did not re-examine it. Adding a sixth PR to a queue I had myself diagnosed as the bottleneck was working against my own conclusion.

**Nothing is lost from the change or the reasoning.** The commit is cherry-picked unmodified, `changelog.d/936.misc.md` travels with it, #895's body now says `Closes #936`, and the full diagnosis lives in #936. Re-verified after folding: `manager_016` **5/5** at ~1.2s, `cargo test-fast` **2472/2472**, fmt and clippy clean, **58/58** across `scheduler_manager` + `mouse` + `pty_size_policy` + `tab_close_regressions` + `resize_layout_002`, and #895 still merges cleanly with #872 with **21/21** `e2e_scheduler_manager` passing on the composed tree.

**The one thing this trades away, stated plainly:** the fix now reaches `main` only when #895 does, and #895 is a breaking protocol change (`PROTOCOL_VERSION` 8→9) that may attract more scrutiny than a three-line test fix would. Until it lands, `manager_016` can still redden other PRs. If #895 stalls, the right move is to cherry-pick `954f941d` onto #872 — its family, its file, and its changelog already claims to fix this test.

vfarcic:
Superseded by #895 — see the comment above for the reasoning and what was verified after folding.

Issue #927, opened 2026-09-06
Every dispatched worktree pays a full cold build: measure whether a shared target dir or pre-warmed worktrees is the right lever


vfarcic:
## Problem

Every dispatched worktree starts with an **empty `target/`**, so each one pays a full cold build before it can do anything. Measured on this repo:

| | |
| --- | --- |
| dependencies in `Cargo.lock` | **613** (was 378 before `desktop/src-tauri` landed 2026-08-29) |
| `target/debug` per worktree | **12–21 GB** observed during the #906 measurement runs |
| shared `CARGO_TARGET_DIR` | **none** — `.cargo/config.toml` sets no `target-dir`, so every worktree builds its own |

With one unit dispatched per issue and several dispatched per day, this is now the **largest remaining build cost in the workflow**. Issue #906 measured the per-task gates in detail and found them comparatively cheap once warm — `clippy` 8.8s incremental, the desktop crate 2.7s, and PR #926 took linking from 34.5s to 16.0s. None of that helps a worktree that has to build everything from scratch first.

It is also the reason the #906 concurrency measurements looked the way they did: at N=2 the only reliable overlap between two agents was their **cold-start builds**, because that is the one phase both are guaranteed to be in at the same time. Measured overlap for the rest of the run was 20.8%.

## Two candidate levers

**1. A shared `CARGO_TARGET_DIR` across worktrees.** Obvious saving, real hazard: cargo takes a lock on the target directory, so concurrent invocations from different worktrees **serialise** rather than run in parallel. That interacts directly with the machine-wide link-slot pool from #863 and with #906's N=2 contention data — it could easily trade a one-off cold build for permanent serialisation, which is worse.

It also raises correctness questions worth answering before adopting: different worktrees can be on different commits with different feature sets, and `cargo` keys artifacts by fingerprint rather than by worktree, so cross-worktree reuse is exactly what makes it fast *and* exactly what makes a stale or mismatched artifact possible.

**2. Pre-warmed worktrees**, so a dispatch starts from a populated cache — e.g. seeding a new worktree's `target/` from a maintained warm one, or keeping a pool of pre-built worktrees to hand out.

Cheaper on the concurrency side because each worktree keeps its own lock, more expensive on disk (12–21 GB per worktree already, and #906 recorded 608 GB free at the time).

## What this issue asks for

Measure before adopting either. Specifically:

[...]

PR #956, opened 2026-09-09
perf(test): schedule version_001 first instead of 150s into the tier (#928)

### LINKED ISSUE #928: version_001 is 60% of the e2e tier's test time: a cold 613-crate build nested inside one test
## Problem

**One e2e test is 60% of the deterministic tier's test time.** Measured from CI run `34003192570` (`e2e-deterministic`, successful, on `main`):

| | |
| --- | --- |
| e2e tests in the run | **193** |
| total e2e test time | **589s** |
| `version_001_injected_build_env_reaches_the_binary` | **351.1s** |
| next slowest test | 12.8s |
| the other 192 tests, combined | **238s** |
| median e2e test | **0.27s** — 155 of 193 finish under 1s |

Because it is a single test, no amount of parallelism helps: it sets a **~6 minute floor** on the job regardless of runner width.

**The rest of the tier is healthy.** There is no broad inefficiency here to hunt — 80% of e2e tests finish in under a second. This is one outlier plus a well-behaved tail, which is why this issue is scoped to the one test.

## Where the time goes

`tests/e2e_version_injection.rs` runs **three real `cargo build`s** into a scratch `CARGO_TARGET_DIR`. Build 1 compiles the full dependency graph — **613 crates** — from scratch; builds 2 and 3 only re-run `build.rs` and relink ("tens of seconds", per the file's own header).

So most of the 351s is a **cold compile of the whole project, nested inside a test**. That is the same cost #927 is about, surfacing here rather than in a worktree.

## Two levers, both visible in the code

**1. The scratch target directory defeats CI's cache.** `ci.yml` runs `Swatinem/rust-cache`, which warms `target/` — but this test deliberately uses its **own** `CARGO_TARGET_DIR` in a tempdir, so it is cold on every run, forever.

The isolation itself is correct and must be preserved. The header states why: the build sets `DAD_VERSION` / `DAD_BUILD_ID`, and `cargo:rerun-if-env-changed` would otherwise invalidate the *main* build script and force a full rebuild of the real `target/` on the very next plain `cargo build`.

But **isolated and cold are not the same requirement.** A stable, cached scratch path outside `target/` would keep the isolation and lose the cold start. Whether `rust-cache` can be pointed at a second directory, or whether a separate `actions/cache` entry is needed, is the open question.

**2. Nested build parallelism is capped at half the machine's cores.** From the header: *"because it runs while nextest is executing the rest of the e2e suite and a full-parallel nested compile can starve the timed tests around it into flaking."*

That reasoning was sound when written, and the starvation it describes is real — it is the same mechanism behind the `orchestration_remit` flakes (#818, #832) and #415's contention measurements. But it is worth re-checking, because **this is the slowest test in the suite**: at 351s against a 12.8s runner-up, it is almost certainly still running long after the tests it was capped to protect have finished. On a 4-vCPU runner the cap means ~2 jobs for most of six minutes, protecting nothing.

If the cap could be lifted *after* the rest of the suite drains — or scaled by observed load rather than fixed at half — the compile would finish substantially sooner without reintroducing the starvation.

## Constraints for whoever takes this

- **Do not weaken what the test proves.** Its header calls it the only tier that observes the whole chain: `build.rs` reading the injected env, the `cargo:rerun-if-env-changed` directives, and the produced binary reporting the injected values with those vars *absent* at runtime. The unit tests in `tests/build_version.rs` cover precedence and validation but not that chain.
- **The three builds are load-bearing.** Builds 2 and 3 change `DAD_VERSION` and `DAD_BUILD_ID` **one at a time**, precisely so a single missing `rerun-if-env-changed` line fails the test — changing both at once would prove neither. Collapsing them defeats the test.
- **Preserve the target-dir isolation**, for the invalidation reason above.
- `.config/nextest.toml` carries a widened `slow-timeout` for this test (`period = "120s"`, `terminate-after = 10`) which exists for the cold case. If the cold case goes away, that override should be revisited rather than left as a 20-minute leash.

## Related

- #927 — the same cold-build cost, per dispatched worktree.
- #906 — the measurement series this came out of; PR #926 (mold) already speeds the relink half of builds 2 and 3.

vfarcic:
Closes #928.

## What was actually slow

`lifecycle/version/001` is the one e2e test that has to `cargo build` the binary itself. #928 measured it at 351s of the deterministic tier against a 12.8s runner-up, and since it is a single test no runner width splits it — a ~6 minute floor on `e2e-deterministic`.

**The dominant cost was not the build. It was when nextest got round to starting it.** I read 13 `e2e-deterministic` runs from their own logs (2026-09-06 → 2026-09-09, `ubuntu-latest`):

| | |
| --- | --- |
| last of ~2,680 tests to **complete** | 13 of 13 runs |
| when nextest **started** it | t = 148.6–156.9s, every run (mean 152.0s) |
| when every *other* test had finished | t = 198–292s |
| the tier's own wall clock | 220.6, 257.9, 270.8, 285.8, 318.8, 336.7, 439.6, 440.9, 441.2, 444.5, 448.3, 449.9, 454.9s |

The last row is exactly `that offset + this test's duration` in all 13. nextest schedules in binary-name order and `dot-agent-deck::e2e_version_injection` sits ~70% down that queue, so the tier's longest test was also its last-started.

## The change

`priority = 25` on the existing override in `.config/nextest.toml`. The tier's wall clock becomes `max(this test, everything else)` instead of `queue depth ahead of it + this test` — about **150s off a ~445s job** on the seven runs above 439s.

**Verified, not assumed.** With `--test-threads 1` completion order *is* schedule order, so `cargo test-e2e --test-threads 1 -E 'binary(e2e_version_injection) or binary(e2e_handshake)'` answers it directly (`e2e_handshake` sorts earlier by binary name):

* with the key: `version_001 … (1/8)`, then the seven handshake tests
* with the key commented out: handshake tests at 1/8…7/8, `version_001` last

25 rather than 100 because `dispatch_015` holds `priority = 50` **plus** `threads-required = "num-test-threads"`, and its comment explains it must be scheduled first so its exclusive reservation does not drain the tier — putting this test above it would make that drain wait out a multi-minute compile. In lane 1 `dispatch_015` is `e2e-live`-gated and not compiled, so second is first.

**The cost, plainly:** the nested compile now overlaps the whole tier instead of only its tail (on the run measured in detail: 113s of its 288s overlapped, then 175s alone on an idle runner — which is why #928 called the `--jobs` cap "protecting nothing"). That cap is unchanged, and its premise has gone from mostly false to true. On the fastest sampled run the gain is only 220.6s → 198.1s, and a tier that slowed more than 11% under the added contention would come out behind; the seven 440s runs are the ones worth fixing.

## Both of the issue's named levers turned out not to be the answer

**Lever 1 — "the scratch target directory defeats CI's cache" — is false.** `Swatinem/rust-cache` caches the whole `target/` tree and its pre-save cleanup *recurses* into a nested target dir: it calls a directory a profile only if that directory holds `build`, `.fingerprint` or `deps`, and `target/version-injection-e2e` holds none of the three at its root, so it is walked rather than pruned, and the keep-list spares every dependency entry. I transcribed `cleanTargetDir` / `cleanProfileTarget` / `rmExcept` from the action at the SHA `ci.yml` pins and ran it against a real populated scratch dir: **995 entries kept, 24 removed**, the 24 being this package's own artifacts and fingerprints plus `incremental/`, `examples/` and the two cargo lock files. A cache-hit run therefore starts build 1 with the dependency graph already compiled — which is also why the sampled runs show no clean split between "No cache found" (351, 370, 305, 185s) and full hits (72–297s). **Nothing in `ci.yml` changed, so there is no conflict with #872.**

**Lever 2 — the `--jobs` cap — is kept, and the issue's objection is what made keeping it correct.** The cap protected almost nothing while the test started at t≈152s and ran alone for its last 175s. Scheduled first, it overlaps the tier for its whole duration, which is precisely the situation the cap was written for.

**What actually dominates this test's cost is neither lever: machine load.** Same work, same flags, one 16-core box — the three nested builds took **65.0s cold quiet (load 3–8) against 400.9s with three other agents building (load 13–18)**. Warm: 24.0s against 142.3s. With `CARGO_INCREMENTAL=0` as CI sets it, 98.5s against 455.6s. That 6x is larger than every configuration lever put together, and the full table now lives in the test file so nobody tunes this test against a single measurement again.

## An experiment that is deliberately not shipped

A `debug = 0` cargo profile for these builds looks obviously right — the binary is executed twice per build, for `--version` and `daemon hello`. On the busy box it looked like a halving (400.9s → 199.3s). **At matched low load it is 10–15% where it shows at all and slightly negative in the configuration CI uses** (101.4s against 98.5s) — inside the noise. The ~50% disk saving is real (1.3 GB → 650 MB in CI's configuration) but taking it needs the profile declared in `Cargo.toml`, because rust-cache keys on that manifest and does not re-save on an exact key match; and then the first cache written carries the old `debug` artifacts forward for good, so CI would store *more*. The numbers are recorded in the test file so nobody redoes the work, and the disk half is pointed at #927, which owns per-worktree build cost. **#927 is not absorbed here.**

`slow-timeout` was revisited as the issue asked and **kept** at 120s × 10: the cold case is not gone (4 of the 16 runs whose cache status I checked restored no cache, and a fresh worktree pays it), the CI spread is 71.9–305.4s plus two no-cache runs at 351.1s and 369.9s, and scheduling the test first means it now runs contended throughout.

Also corrected: the issue's "613 crates" is the whole workspace's `cargo metadata --all-features` count (Tauri desktop crate, every dev-dependency). The nested build's graph is **224 packages** / 280 compilation units.

## What the test proves is untouched

The only change to `tests/` is documentation. Three real builds into an isolated target dir, `DAD_VERSION` and `DAD_BUILD_ID` still changed one at a time so a single missing `cargo:rerun-if-env-changed` line fails it, each produced binary still run with both variables absent. Two ways under the remaining floor were considered and rejected in the header: collapsing the builds (stops pinning either rerun directive), and seeding the scratch dir from the main `target/debug` (a test whose job is to prove a *build* behaves correctly should not start from another build's output).

## Gates

All on a 16-core devbox; load noted with every timing above because it turned out to be the largest variable.

* `cargo fmt --check` — clean
* `cargo clippy --workspace --all-targets --features e2e,e2e-live -- -D warnings` — clean
* `cargo test-fast` — **2483 passed**, 0 skipped
* `cargo xtask linkage-check` — ok (636 catalog ids, 566 annotations, 11 rules)
* `cargo test-e2e version_001` — **PASS**: 163.5s cold (395.3s on the pre-change baseline, busy box), **38.9s warm**
* `cargo test-e2e --test-threads 1 -E 'binary(e2e_version_injection) or binary(e2e_handshake)'` — 8 passed, and it is the control/treatment pair proving the reordering above


vfarcic:
### This PR's own `e2e-deterministic` run confirms the mechanism

Run [34312859383](https://github.com/vfarcic/dot-agent-deck/actions/runs/34312859383), 2674 tests, all passed:

| | before (13 runs) | this run |
| --- | --- | --- |
| `version_001` start offset | 148.6–156.9s, **every** run | **t = 0.0s** |
| its completion position | **last** in 13 of 13 | 1582 / 2674 |
| `version_001` duration | 71.9–305.4s | 52.9s |
| tier wall clock | 220.6–454.9s (median 439.6s) | **261.8s** |

**Read the wall clock carefully rather than as a 178s saving.** `version_001` was one of the fast ones on this runner (52.9s), so at the old offset it would have finished around t≈205s and the tier would have been bound by its other tests anyway. What is confirmed here is the *structural* change: this test can no longer add its duration on top of a ~152s queue wait, and it is no longer the last test to finish. The payoff lands on the runs where it is slow — the seven prior runs at 439–455s had `version_001` at 286–305s, which under this scheduling would put the tier at roughly `max(300, ~262) ≈ 300s`.

**And the flagged cost is visible, though not separable from runner variance on n=1.** The rest of the tier finished at 261.8s here, against 198.1–291.5s (median 223.5s) before — above the old median, which is the direction the added contention would push it. Worth watching over the next few runs on `main`; a tier that settles above ~292s would mean the overlap is costing more than the paragraph in `.config/nextest.toml` budgets for.

Unchanged and pre-existing: the one `SLOW` is `seam_bound_001_dashboard_cards_clamps_derived_height` (75.1s in the pre-change run too) and the one `LEAK` is `orphan_exit_001`, which the baseline runs also reported.

Issue #532, opened 2026-08-12
pane_hook_session treats a second producer's ordinary frames as a generation change


vfarcic:
Split out of #424 (round 3), where it is a documented residual rather than a fix.

## What

`AppState::pane_hook_session` advances to the session id of **any** event carrying a pane id, provided the timestamp is not older than the recorded one (`src/state.rs`, the `advance` block in `apply_event`). It therefore treats a non-announcing frame — `Thinking`, `ToolStart`, `Idle` — from a *second producer on the same pane* as a generation change.

That shape is normal, not exotic: a wrapped agent has two producers under ONE registry agent id. `dot-agent-deck wrap` emits under `{pane_id}-session` (`wrap::session_id_for`), while the wrapped agent's native hooks emit under their own session id. The pane's "current generation" then alternates between the two as events interleave.

## Why it matters

`pane_hook_session_id` is the daemon's send guard input: a guarded write carrying `expected_session_id` requires an EXACT match against it (`daemon_protocol.rs`). On a wrapped pane the guard therefore refuses deliveries intermittently, depending only on which producer emitted last — reported to the caller as `WrongSession`.

Issue #424 makes this more visible in two ways, both deliberate:

- the TUI's `delivery_target_changed` now compares the bound generation against the current one, so the alternation is read as a lost target and the delivery abandons **deterministically**, one frame earlier than the daemon would have refused it;
- `WrongSession` after a provisional write is now terminal on the seed path, so the refusal abandons instead of retrying to the deadline.

Neither introduces the defect; both make an existing intermittent failure legible.

## Suggested fix

Mirror the rule `state::latch_generation` already applies to the same question: **only a `SessionStart` announces a generation.** A non-start frame should refresh the high-water timestamp of the generation it names when that is the current one, and otherwise change nothing. `apply_event` already argues exactly this for card retirement ("`SessionStart` is the incoming generation ANNOUNCING itself: self-describing and authoritative"; anything else is inference).

## Care needed

This changes the meaning of a value a caller puts on the wire, so it is a cross-version question under CLAUDE.md rule 12: an old daemon computing `pane_hook_session` the old way, compared against a new TUI's `expected_session_id` computed the new way, can disagree where today they agree. Worth deciding deliberately — #424 left it alone specifically to avoid changing that answer mid-PR.

PR #1004, opened 2026-09-10
fix(prompt): a start that announces no conversation cannot lose one (#684, #1003)

### LINKED ISSUE #684: A TUI-owned prompt into a launcher-wrapped pane is dropped when the real agent announces itself

### LINKED ISSUE #1003: A TUI-owned prompt is dropped by the daemon's own card-surfacing SessionStart, not only by a wrapper's (second trigger for #684)
Split out of #684, which describes this defect class but attributes it to the wrapper-fork start alone. There is a **second establishing event**, it is the one that fires in ordinary single-agent and orchestration use with no wrapper anywhere, and #684's diagnosis would have led a fixer to miss it.

## Symptom

Creating an agent with an opening prompt reports `Seed prompt not delivered (the agent's conversation changed); abandoned`; the delivery log records `delivery-target-changed`. Both TUI-owned paths, exactly as #684 says: `ui::process_pending_seed_prompts` and `ui::deliver_orchestrator_prompt`.

## The establishing event is the daemon's own card-surfacing start

`spawn::surface_spawned_pane` broadcasts a synthetic `SessionStart` so a spawned pane gets a card immediately rather than at its first real hook. It has three properties that matter:

- its `session_id` **is the pane id**;
- it carries **no `agent_id`**;
- until this issue it carried **no origin marker**, so every `is_wrapper_session_start()` exclusion in the delivery path missed it and `apply_event` read it as a conversation announcing itself.

Critically it is **broadcast-only** — `spawn.rs` says so directly: it "never passes through `daemon::ingest_event`, so the daemon's own `AppState` never sees it." Only attached TUIs apply it. **That is why this trigger is invisible in `deck.log`:** the daemon logs the events it receives, and it never receives this one.

## Measured in the field

Five consecutive failures on 2026-09-10, three single-agent panes and one orchestration role, each abandoned 530–540 ms after the agent's genuine `SessionStart` — one render pass later:

```
12:55:41.372  SessionStart pane_id=8   1cbf38a0-…
12:55:41.902  WARN … path="seed" pane_id="8"  reason="delivery-target-changed"
12:57:06.179  SessionStart pane_id=10  e4d9355d-…
12:57:06.717  WARN … path="seed" pane_id="10" reason="delivery-target-changed"
12:57:57.725  SessionStart pane_id=11  b2244167-…
12:57:58.266  WARN … path="orchestrator" pane_id="11" reason="delivery-target-changed"
```

Pane 10 received **exactly one** event before the abandon — the genuine `SessionStart`. No wrapper event, no prior generation, nothing #684 describes. The launcher was `devbox run` → plain `claude`, i.e. no `dot-agent-deck wrap` in the chain at all.

## It is a race, which is why it looked like a regression

The prompt is bound to whatever generation exists when the write happens. If the agent announces itself *before* the readiness buffer elapses the delivery binds the genuine generation and succeeds; if it announces *after*, the delivery binds the surfaced card and is abandoned. The same log shows both outcomes 17 seconds apart on 2026-09-08 (`18:09:33` abandoned, `18:09:50` confirmed), and every TUI-owned delivery after `2026-09-09T00:56` failing — with **no change to the deck binary**, which was built 2026-09-05. Agent startup simply got slower relative to the buffer.

The daemon-owned paths (`dispatch`, scheduler, issue-dispatch) are unaffected, as #684 notes, because `latch_generation` refuses to bind a provenance-less start and permits one handoff.

## Fix

One mechanism covers both triggers: a `SessionStart` that announces no conversation may draw a card but may not **establish** the pane's generation. The card-surfacing start gets an explicit origin marker (`CARD_SURFACE_SESSION_START_ORIGIN`) so it is recognized by declared provenance rather than inferred from its shape — `session_id == pane_id` with no `agent_id` also describes a legacy untagged hook, and inferring would have taken the generation away from real conversations.

vfarcic:
## The report

> I'm getting "Seed prompt not delivered..." when I try to create dispatcher agent … I'm getting the same when I try to spin up an orchestration. That feels like a recent change that might have messed it up.

Creating an agent with an opening prompt discarded that prompt and left the pane healthy, idle and with nothing to do. Both TUI-owned delivery paths: `ui::process_pending_seed_prompts` (a `[[modes]]` `seed_prompt` and the new-pane form's prompt) and `ui::deliver_orchestrator_prompt`.

It was **not** a recent change — the deck binary was built five days before the first failure. It is a **race** that agent startup getting slower flipped from usually-winning to always-losing.

## Cause

Two events exist only to draw a pane's **card** before any conversation exists: the daemon's card-surfacing `SessionStart` (`spawn::surface_spawned_pane`) and the wrapper's boot-provenance start (PRD #225 M3). Both **established** the pane's hook generation.

A TUI-owned delivery binds `pane_hook_session_id` as its target *before* it writes, so the prompt claimed a conversation nothing had announced. The agent's genuine `SessionStart` moments later was a different generation replacing that one — which is simultaneously a `note_generation_closed` and a changed target in `ui::delivery_target_changed`. Either one alone abandons the delivery.

The daemon-owned paths (`dispatch`, scheduler, issue-dispatch) were never affected, because `state::latch_generation` already refuses to bind either wrapper origin and permits exactly one handoff.

## Fix

A `SessionStart` that announces no conversation may draw its card but may not **establish** a generation. The pane then simply has no conversation until a producer announces one, and the prompt is delivered into that genuine conversation when it arrives — the path `prompt/pane-input/030` already pinned. This makes the generation map agree with the daemon-side latch instead of adding a second policy.

Non-start frames are untouched and still establish on any pane id, so a producer whose native hooks emit no `SessionStart` at all is unaffected.

## Why #684's diagnosis would have missed it

#684 attributes the class to the wrapper-fork start alone. Your failures had **no wrapper in the chain** (`devbox run` → plain `claude`). The card-surfacing start is the second establishing event, filed as #1003, and it carried **no origin marker at all**, so every `is_wrapper_session_start()` exclusion in the delivery path missed it. It is now marked (`CARD_SURFACE_SESSION_START_ORIGIN`) and recognized by declared provenance rather than inferred from its shape — `session_id == pane_id` with no `agent_id` also describes a legacy untagged hook, and inferring would have taken the generation away from real conversations.

It is also invisible in `deck.log`: that event is broadcast straight to attached clients and never passes through `daemon::ingest_event`, so the daemon never logs it. Pane 10 below received **exactly one** event before the abandon.

## Evidence

Five consecutive failures, each abandoned 530–540 ms after the genuine `SessionStart` — one render pass later:

```
12:55:41.372  SessionStart pane_id=8   1cbf38a0-…
12:55:41.902  WARN … path="seed" pane_id="8"  reason="delivery-target-changed"
12:57:06.179  SessionStart pane_id=10  e4d9355d-…
12:57:06.717  WARN … path="seed" pane_id="10" reason="delivery-target-changed"
12:57:57.725  SessionStart pane_id=11  b2244167-…
12:57:58.266  WARN … path="orchestrator" pane_id="11" reason="delivery-target-changed"
```

The same log shows the race both ways 17 seconds apart on 2026-09-08 (`18:09:33` abandoned, `18:09:50` confirmed), then every TUI-owned delivery after `2026-09-09T00:56` failing, with the binary unchanged since 2026-09-05. `path="spawn"` deliveries were confirmed throughout.

## Tests

- **`prompt/pane-input/033`** (new, L1) drives the production seed path through **both** provisional establishing events and pins that the delivery survives the agent announcing itself, that the retry is delivered naming the genuine generation, and that the pane's generation-closure count stays 0 — the closure counter abandons the delivery on its own, so asserting only the bound id would have half-tested it. One mechanism, two inputs, so they are cases of one test rather than two near-duplicate tests.
- **`surface_spawned_pane_emits_session_start_for_attached_tuis`** (extended) pins the marker on the **producer**, because the behavioural test replays that event rather than calling the function — nothing else would have noticed the marker going missing.
- Reproduction came first and failed with the reporter's verbatim message. Both halves of the fix were then verified **load-bearing** by reverting each in turn (the core `None => !provisional_start`, and widening the predicate past the wrapper-only case). A third candidate change — excluding daemon-authored starts in `ui::pane_announced_generation` — left the test green, so it was **reverted rather than kept as dead defence**; the map-side fix subsumes it, and the residual it would have guarded is #532's non-announcing-frame shape.

## Gates

- `cargo test-fast`: 2558/2558.
- `cargo clippy --workspace --all-targets --features e2e,e2e-live -- -D warnings`: clean. `cargo fmt --check`: clean.
- Covering lane-1 e2e: `e2e_scheduler_spawn`, `e2e_scheduler_live_surface`, `e2e_orchestrator_setup` — 17/17. Includes `live_004_real_hook_supersession_keeps_friendly_title`, which is exactly the card-surface→real-hook handover this changes, so the card title still survives it.
- Cross-version contract check (rule 12) and a real-agent lane-2 run are reported in a comment below.

## Contract

No `PROTOCOL_VERSION` bump and no `.breaking.md`. This adds a fourth value to the existing `session_start_origin` metadata key, documented as additive in both directions; an old consumer matches none of the wrapper values and behaves exactly as it does today. Old and new builds interoperate unchanged — the fix simply requires both sides updated to take effect, which is stated on the predicate.

Closes #684
Closes #1003

### DIFF
--- a/changelog.d/684.bugfix.md
+++ b/changelog.d/684.bugfix.md
@@ -1,0 +1,9 @@
+## A Spawn-Time Prompt Is No Longer Discarded When the Agent Announces Itself
+
+Creating an agent with an opening prompt — a `[[modes]]` `seed_prompt`, the new-pane form's prompt, or an orchestration role's opening prompt — could report `Seed prompt not delivered (the agent's conversation changed); abandoned` and drop the prompt, leaving the pane healthy, idle and empty. The agent worked; it just never received what it was created to do.
+
+Nothing about the pane's conversation had actually changed. Two events exist purely to draw a pane's CARD before any conversation exists: the daemon's card-surfacing `SessionStart`, which it broadcasts to attached decks so a spawned pane appears immediately instead of at its first real hook, and the wrapper's boot-provenance `SessionStart` for a launcher-wrapped agent. Neither is a producer speaking for the pane — the card-surfacing one uses the pane id as its session id and carries no agent identity at all — but both established the pane's hook generation, so a prompt written before the agent announced itself bound one of them as the conversation it was entering. The agent's genuine `SessionStart` moments later was then a different conversation replacing that one, which reads as both a rollover and a closed conversation, and the delivery was abandoned on the safe side of a race it should never have been in.
+
+Such a start now draws its card without establishing a generation, so the pane simply has no conversation until a producer announces one, and the prompt is delivered into that genuine conversation when it arrives. This matches what the daemon-owned delivery paths — `dispatch`, the scheduler and issue-dispatch — already did, which is why they were never affected.
+
+Whether the bug fired was a race between the agent's startup and the deck's readiness buffer, so it grew more likely as agent startup got slower: a prompt lost this way had simply been written before the agent got around to announcing itself.
--- a/src/event.rs
+++ b/src/event.rs
@@ -509,6 +509,32 @@ pub const SESSION_START_ORIGIN_METADATA_KEY: &str = "session_start_origin";
 /// later.
 pub const WRAPPER_FORK_SESSION_START_ORIGIN: &str = "wrapper_fork";
 
+/// The [`SESSION_START_ORIGIN_METADATA_KEY`] value meaning "the DAEMON authored
+/// this `SessionStart` to draw a spawned pane's card, and no producer has spoken
+/// for this pane yet" (issue #684).
+///
+/// [`crate::spawn`]'s `surface_spawned_pane` broadcasts such an event straight
+/// onto the client fan-out so a spawned pane gets a card immediately instead of
+/// at its first real hook. It never passes through `daemon::ingest_event`, so the
+/// daemon's own `AppState` never applies it — only attached TUIs do. That is why
+/// the defect below appears in no daemon log.
+///
+/// It is a statement ABOUT a pane, not a producer speaking FOR one: its
+/// `session_id` IS the pane id and it carries no `agent_id`. Without this marker
+/// [`crate::state::AppState::apply_event`] read it as a conversation announcing
+/// itself, so a TUI-owned prompt bound the pane id as its generation and the
+/// agent's genuine `SessionStart` then read as a rollover — abandoning the prompt
+/// with "the agent's conversation changed" (`prompt/pane-input/033`).
+///
+/// MARKED rather than inferred from its shape. `session_id == pane_id` with no
+/// `agent_id` also describes a legacy untagged hook, and the permissive
+/// both-absent branch of `apply_event`'s reuse guard exists precisely to keep
+/// serving those — so inferring provenance from the shape would have taken the
+/// generation away from real conversations too. Producer-writable like every
+/// value on this key, and safe in the only direction it can be abused: a forged
+/// marker makes an event LOSE generation standing, never gain any.
+pub const CARD_SURFACE_SESSION_START_ORIGIN: &str = "daemon_card_surface";
+
 /// The [`SESSION_START_ORIGIN_METADATA_KEY`] value meaning "`dot-agent-deck wrap`
 /// watched the wrapped child take the inner PTY OUT OF COOKED MODE" (issue #243).
 ///
@@ -800,6 +826,19 @@ impl AgentEvent {
             .is_some_and(|origin| origin == WRAPPER_FORK_SESSION_START_ORIGIN)
     }
 
+    /// Issue #684: was this `SessionStart` authored by the DAEMON to draw a
+    /// spawned pane's card (see [`CARD_SURFACE_SESSION_START_ORIGIN`])?
+    ///
+    /// `false` for every event without the marker, which includes every event an
+    /// OLDER daemon relays — an old daemon's unmarked surface event therefore
+    /// behaves exactly as it does today rather than being silently reclassified,
+    /// and the fix this predicate gates needs both sides updated to take effect.
+    pub fn is_card_surface_session_start(&self) -> bool {
+        self.metadata
+            .get(SESSION_START_ORIGIN_METADATA_KEY)
+            .is_some_and(|origin| origin == CARD_SURFACE_SESSION_START_ORIGIN)
+    }
+
     /// Issue #770: does this event carry the daemon's ORPHANED-ROLE marker (see
     /// [`ORCHESTRATION_ORPHANED_METADATA_KEY`])? `false` for every event
     /// without it, which is every event an older daemon relays and every event
@@ -880,12 +919,17 @@ impl AgentEvent {
     /// Issue #424 D4: was this event SYNTHESIZED BY THE DAEMON rather than
     /// produced by the pane's agent?
     ///
-    /// The daemon emits identified events of its own through the same pipeline
-    /// real hook events take — [`EventType::ShellBusy`]/[`EventType::ShellIdle`]
-    /// from the shell-activity monitor (PRD #370/#386), and the delivery-notice
-    /// [`EventType::Error`] (issue #424). They carry the pane's registry
-    /// `agent_id` because that is how they land on the right card, and that is
-    /// exactly what made them indistinguishable from producer evidence to
+    /// The daemon emits events of its own through the same pipeline real hook
+    /// events take: [`EventType::ShellBusy`]/[`EventType::ShellIdle`] from the
+    /// shell-activity monitor (PRD #370/#386), the delivery-notice
+    /// [`EventType::Error`] (issue #424), and the card-surfacing `SessionStart`
+    /// (issue #684, [`CARD_SURFACE_SESSION_START_ORIGIN`]).
+    ///
+    /// The first two carry the pane's registry `agent_id` because that is how
+    /// they land on the right card — the card-surfacing start is the exception and
+    /// carries none, since it RUNS BEFORE any producer has identified itself, so
+    /// it is recognized by its origin marker instead. Carrying that identity is
+    /// exactly what made the first two indistinguishable from producer evidence to
     /// `crate::ui::evidence_channel_is_unidentified`: one of them arriving was
     /// enough to conclude the pane has a tagged reporting channel, when it proves
     /// only that the DAEMON can tag its own events. A pane behind a legacy
@@ -901,6 +945,7 @@ impl AgentEvent {
     pub fn is_daemon_synthetic(&self) -> bool {
         matches!(self.event_type, EventType::ShellBusy | EventType::ShellIdle)
             || self.metadata.contains_key(DELIVERY_NOTICE_METADATA_KEY)
+            || self.is_card_surface_session_start()
     }
 }
 
--- a/src/spawn.rs
+++ b/src/spawn.rs
@@ -2371,6 +2371,18 @@ fn surface_spawned_pane(
 ) {
     let mut metadata = HashMap::new();
     metadata.insert(DISPLAY_NAME_METADATA_KEY.to_string(), task_name.to_string());
+    // Issue #684: declare that the DAEMON authored this start to draw a card,
+    // rather than a producer announcing a conversation. `session_id` below is the
+    // PANE ID and there is no `agent_id`, so without the marker an attached TUI's
+    // `AppState` established the pane id as the pane's hook GENERATION — a
+    // TUI-owned seed or role prompt bound it, and the agent's genuine
+    // `SessionStart` then read as a rollover and abandoned the prompt with "the
+    // agent's conversation changed". See `CARD_SURFACE_SESSION_START_ORIGIN` and
+    // `prompt/pane-input/033`.
+    metadata.insert(
+        crate::event::SESSION_START_ORIGIN_METADATA_KEY.to_string(),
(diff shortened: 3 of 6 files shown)


vfarcic:
## Cross-version contract check (CLAUDE.md rule 12)

**Result: no contract break.** No `PROTOCOL_VERSION` bump and no `.breaking.md` fragment is owed.

Run with a fully isolated sandbox — `DOT_AGENT_DECK_SOCKET`, `DOT_AGENT_DECK_ATTACH_SOCKET`, `HOME`, `XDG_RUNTIME_DIR`, `XDG_STATE_HOME`, `DOT_AGENT_DECK_LOCK_DIR`, `DOT_AGENT_DECK_STATE_DIR`, `DOT_AGENT_DECK_SESSION`, **`DOT_AGENT_DECK_LOG`** and **`DOT_AGENT_DECK_EXPERIMENTAL`** all redirected, and `DOT_AGENT_DECK_IDLE_SHUTDOWN_SECS=0` exported for every process. The developer's real daemon was verified alive and untouched afterwards.

Sequence per the rule, which is the only one that actually reaches the scenario:

1. previous-release daemon (`0.39.3-ga023377`) started, then the **previous-release TUI** brought an orchestration up under it — two `cat` roles (`orchestrator`, `coder`), so no LLM tokens;
2. that TUI detached (**Detach**, not Stop) — daemon and both agents verified still alive;
3. the **branch TUI** (`0.39.3-ge76fe61c`) launched against it, which showed the mismatch prompt naming both live agents;
4. **declined** it, so the branch TUI attached to the older daemon unchanged.

**Both traps the rule warns about were checked, not assumed:**

- exactly **one** `Attach protocol listening` line for the whole run, so no replacement daemon was lazy-spawned and this was not a disguised same-version test;
- the same daemon pid (`3013437`, running the old binary) served it end to end, and the handshake reported two different build ids.

Core flows against the older daemon:

| flow | result |
| --- | --- |
| orchestration tab rebuilt | both role panes, correct role names |
| orchestrator role prompt | `prompt written to pane … path="orchestrator" pane_id="1"` — delivered, not refused |
| **delegate routes** | `Received delegate signal pane_id=1 targets=["coder"]`, role resolved, worker respawn path entered |
| **work-done arrives** | `Received work-done signal pane_id=2 done=false`, credited/retired against the outstanding delegations |
| **status hook arrives** | `Received event … event_type=Thinking pane_id=Some("2") agent_type=Pi`; the branch TUI rendered `1 active │ 1 thinking` |

The status-hook check was done **last**, since it reclassifies the pane's agent type as `Pi` and would otherwise change how prompt delivery routes for that pane.

### Why the wire is unchanged

This adds a fourth value to the existing `session_start_origin` metadata key, which is documented as additive in both directions. An old consumer compares that key only against the wrapper values, matches none, and treats the event exactly as it does today. Both mixed-version directions were reasoned through against the four `(expected, current)` arms of `compute_write_and_submit_outcome` and confirmed to interoperate:

- **new TUI + old daemon:** the old daemon's surface event carries no marker, so the fix does not take effect and behaviour is today's. For a wrapper pane the TUI now sends no generation against a daemon that has one, which the guard refuses as `Stale` — and `Stale` is classified retryable by both TUI paths, so the retry names the genuine generation once it announces. Self-heals.
- **old daemon → old TUI ordering** is unaffected; the marker is ignored.

This is stated on `is_card_surface_session_start` as well: the fix needs both sides updated to take effect, which is a deferred benefit rather than a break.

## Real-agent verification (lane 2), and a finding

`prompt/new-pane/016` — the PTY-attached dispatcher-mode test with a real Claude agent, i.e. the reported scenario — was run with `DOT_AGENT_DECK_REQUIRE_REAL_E2E=1` so a silent `SKIP:` could not pass as green. It passes on this branch.

**It also passes with the fix reverted.** Verified twice. So this test never covered the defect, and running it would not have caught the regression. With the fix reverted and `DOT_AGENT_DECK_RECORD=1`, the recorded card shows a prompt history of two entries — the seed *and* the typed goal — i.e. the seed was genuinely delivered, so the harness does not reach the failing state at all. The abandon needs the agent's genuine `SessionStart` to arrive *after* the write, and the harness's directly-exec'd `claude` announces sooner relative to the 500 ms readiness buffer than the reported `devbox run` → hooks → `claude` chain does (measured there at ~4.2 s). The test is structurally biased to win the race it exists to exercise.

Filed as #1006 rather than fixed here, since making announcement latency controllable in the harness is its own piece of work. The regression itself is now pinned deterministically **and in CI** by `prompt/pane-input/033`, which is an L1 test in `src/ui.rs` and therefore runs under `cargo test-fast` in the `build` job.

Also filed: **#1005** — the same card-surfacing event satisfies the TUI's `agent_ready` gate (it resolves `ClaudeCode` from the command), so the first payload is written ~3.7 s before the agent can read it. Different mechanism, touches the readiness gate shared with orchestration and the declared-no-signal producers, so deliberately not bundled into this PR.

Issue #1013, opened 2026-09-11
Pin `prompt/new-pane/016` to a cheap model: it runs a real agent on whatever the host selects (measured: Opus 5, xhigh), against CLAUDE.md rule 4


vfarcic:
## What is wrong

`prompt/new-pane/016` (`new_pane_016_dispatcher_opens_dashboard_card_with_real_agent`, `tests/e2e_dispatcher_mode.rs`) spawns a bare `claude` with no `--model`, so the agent it drives runs on whatever the host's managed settings select. Measured on the dev box while working #1006: the pane came up on **Opus 5 with xhigh reasoning effort**, visible in the test's own grid dump.

CLAUDE.md rule 4 asks real-agent e2e tests to run on a **cheap model** — Haiku for Claude, a mini model for OpenCode — through a cheap, deterministic-enough operation. A `[reel]`-marked lane-2 test on Opus at xhigh is the opposite of that, and the cost is silent: nothing in the test or in CI reports which model it picked, so the bill depends on whoever runs it.

## The working precedent

`scheduler/dispatch/015` pins exactly this, for exactly this reason: `tests/e2e_prompt_delivery_confirmation.rs:19` holds

```rust
const REAL_AGENT_COMMAND: &str = "claude --model claude-haiku-4-5-20251001 --allowedTools Bash";
```

and drives three concurrent real interactive Claude dispatches on it. Several other lane-2 files pin the same model id (`e2e_card_layout.rs`, `e2e_pane_input_live.rs`, `e2e_chain_smoke_claude.rs`, `e2e_idle_worker_detector.rs`, `e2e_project_launch_real.rs`, `e2e_delegate_respawn_readiness.rs`, `e2e_delegate_work_done_chain.rs`, `e2e_shell_activity_real_agent.rs`), so 016 is the outlier rather than the pattern.

## Why it is newly easy

This is **pre-existing** — 016 has always spawned bare `claude`. What changed is that #1006's launcher refactor made the exec'd command a one-line decision instead of an accident: the test now builds its agent through

```rust
let launcher = common::write_late_announcing_real_agent(
    staging.path(),
    LAUNCHER_LOG,
    LAUNCHER_DELAY_SECS,
    &format!("'{}'", real_claude_path().display()),
);
```

so the model pin is an edit to that last argument (and to `real_claude_path`'s contract, which today resolves the bare binary off the host PATH deliberately, so `AgentType::from_command` still resolves the launcher as a `claude` command).

[...]

Issue #1012, opened 2026-09-11
Remove the dead `process_pending_dispatches` queue: nothing in production writes to it, and its call-site comment claims a remit that belongs to `deliver_orchestrator_prompt`


vfarcic:
## What is dead

`UiState::pending_dispatches` (`src/ui.rs`, declared as `pending_dispatches: Vec<PendingDispatch>`), the `process_pending_dispatches` function that drains it, and the call to that function in the render loop. Nothing in production code ever pushes to the queue, so the `retain` loop iterates an always-empty `Vec` on every frame.

## Evidence

Grepped across the whole repository for writers (`grep -rn "pending_dispatches.push" src/`, and `grep -rn "pending_dispatches" --include=*.rs .` for every other mention). There are exactly three push sites and all three are tests inside `src/ui.rs`'s own test module:

- `pending_dispatch_timeout` (around `src/ui.rs:33455`)
- `pending_dispatch_waits_for_agent_ready` (around `src/ui.rs:33475`)
- `pane_input_037_the_other_two_spawn_time_gates_wait_for_a_conversation` (around `src/ui.rs:37369`) — added by #1005 to pin the gate across all three spawn-time paths

Line numbers are as of the #1005 branch and will drift; the symbol names are the durable anchors. Everything else that mentions the name is a doc comment or the declaration/initialisation/drain itself.

This was found while working #1005 and independently confirmed by that PR's reviewer.

## Why it is vestigial: PRD #93

PRD #93 round-5 moved delegation dispatch into the daemon. The comment block immediately above `process_pending_dispatches` records it: the TUI used to drain `AppState.delegate_events` / `AppState.work_done_events`, build the per-role prompt and route it through the pane controller, and "the daemon's hook loop now writes that prompt directly into the worker pane's PTY (see `AppState::handle_delegate` / `AppState::handle_work_done`), so the TUI is out of the dispatch business." The queue is what is left of the seam after that move — the drain survived, its producer did not.

## The call-site comment misdescribes it

At the call site (around `src/ui.rs:13819`) the comment ends:

> `process_pending_dispatches` still ferries the orchestrator's *initial* prompt across the agent-ready gate.

That remit belongs to `deliver_orchestrator_prompt`, which is called from the same render loop about twenty lines earlier. So the comment names a job this function does not have, which is how the queue reads as live to anyone skimming the loop.

## What #1005 deliberately did NOT do

[...]

PR #1014, opened 2026-09-11
fix(prompt): require an actual agent announcement before spawn-time delivery

### LINKED ISSUE #1005: A card-surfacing SessionStart satisfies the TUI's agent-ready gate, so the first spawn-time payload is written before the agent can read it
Found while fixing #684/#1003, and deliberately left out of that PR: same root cause, different mechanism, different blast radius.

## The defect

Both TUI-owned delivery paths gate their first write on `agent_ready`:

```rust
let agent_ready = snapshot.sessions.values().any(|s| {
    s.pane_id.as_deref() == Some(sp.pane_id.as_str()) && s.agent_type != AgentType::None
});
```

`src/ui.rs`, in `process_pending_seed_prompts` and again in `deliver_orchestrator_prompt`.

"A session on this pane has a resolved agent type" is not "the agent can accept input". The daemon's card-surfacing `SessionStart` (`spawn::surface_spawned_pane`) resolves one from the command — `agent_type.or_else(|| AgentType::from_command(command))` — so for a `claude` pane it arrives carrying `ClaudeCode` at spawn time. `agent_ready` is therefore true immediately, `ready_since` is stamped, and the write fires one `SPAWN_TIME_READINESS_BUFFER` (500 ms) later.

## Measured

Pane spawned `12:57:02`; the agent's genuine `SessionStart` arrived `12:57:06.179`. The seed was written at roughly `12:57:02.5` — about **3.7 seconds before the agent announced itself**, into a PTY whose program had not started reading. This is the same field incident as #1003; the abandoned delivery there is the consequence that was visible, and this is the premature write underneath it.

The spawn-time placeholder may reach the same state without any card-surfacing event, since `insert_placeholder_session` is called with a known `agent_type` on the interactive paths — worth checking as part of this rather than assumed.

## Why it is not simply the #424 problem

#424 made these writes provisional precisely so a swallowed one is retried, and after #684/#1003 the delivery survives and does recover on the retry. So this is not a lost prompt any more. What it still costs:

- one wasted payload write per spawn, against `MAX_PAYLOAD_SUBMISSIONS = 2` — it consumes the single bounded replacement payload that exists for the launcher case (`scheduler/dispatch/015`), so a genuinely swallowed second write has nothing left but blind submit probes;
- bytes sitting in a PTY before the agent's TUI is in submit-CR-aware mode, which is the accumulation shape (`seedseed`) the payload cap exists to bound.

## Suggested fix

Require that a producer has actually announced a conversation, not merely that some session on the pane has a type — i.e. add `snapshot.pane_hook_session_id(pane_id).is_some()` to the predicate. After #684/#1003 that map holds an entry only when a producer announced one or an ordinary frame arrived, which is exactly the fact this gate wants.

Deliberately NOT bundled into #684's fix: it touches the readiness gate shared with the orchestration path and interacts with the declared-no-signal producers (`spawn/005`, the OpenCode cold-boot path in #146) and with PRD #225/#243's readiness work, so it wants its own test pass rather than riding along on a delivery-target fix.

## Related

#684, #1003, #424 (B3), #225, #243, #146.

### LINKED ISSUE #1006: prompt/new-pane/016 passes with the seed-delivery bug present: the harness's fast-booting agent wins the race the test exists to exercise
Measured while fixing #684/#1003. `prompt/new-pane/016` is the PTY-attached, real-Claude test for the exact scenario a user reported broken — "create a dispatcher agent and its opening prompt is discarded" — and it **passes with that bug present**. Verified by reverting the one-line fix and running it twice.

## What it asserts vs. what broke

The catalog says it asserts "that the seed actually reached the pane (a distinctive `DISPATCHER_SEED_PROMPT` phrase, so it cannot pass on an unseeded agent)". That assertion is sound as written. The problem is that the bug it needs to catch does not occur in the harness at all, so nothing was asserted against.

With the fix reverted and `DOT_AGENT_DECK_RECORD=1`, the recorded card shows a prompt history of **two** entries — the seed *and* the typed goal — i.e. the seed was genuinely submitted and confirmed. The harness never reaches the failing state.

## Why the harness cannot reach it

The abandon requires the agent's genuine `SessionStart` to arrive **after** the deck has written the prompt. The write fires one `SPAWN_TIME_READINESS_BUFFER` (500 ms) after the readiness gate opens. So the defect needs an agent that announces itself later than that, which is the ordinary case in the field:

- reported environment: `devbox run agent` → devbox init hooks → `claude`, announcing ~4.2 s after spawn (measured: pane spawned `12:57:02`, `SessionStart` `12:57:06.179`);
- harness: `claude` exec'd directly with a pre-seeded HOME, announcing fast enough that the delivery binds the genuine generation and the sequence never arises.

So the test is **structurally biased to win the race** it exists to exercise. This is the same class as the note already on `prompt/pane-input/030` — the adjacent no-generation branch is pinned, the provisional-generation branch was not.

## Why this matters beyond one test

Two independent reasons the gap was invisible:

1. This test is lane 2, so per CLAUDE.md rule 5 it runs on a developer's machine and **nowhere in CI**. Even a version that did catch the defect would only catch it when somebody happened to run it.
2. Because it passes either way, running it would not have helped. It is not the accepted "lane 2 only runs when someone runs it" trade — it is a test that would have reported green on a broken product.

The regression itself is now pinned deterministically and in CI by `prompt/pane-input/033` (an L1 test in `src/ui.rs`, so it runs under `cargo test-fast` and therefore in the `build` job). What is still missing is an end-to-end test where a spawn-time prompt is genuinely delivered LATE, as a user experiences it.

## Suggested work

- Make the agent's announcement latency controllable in the harness so the provisional-generation ordering is reachable on purpose, rather than depending on how fast the fixture's agent boots. A launcher stand-in that delays its `exec` would do it without an LLM.
- Assert the seed was **acted upon**, not only that its text is on screen. The bytes are written before the delivery is abandoned, so screen text survives the failure — an adjacent artefact. The card's prompt history, or a sentinel the agent must produce, is the user-altitude fact.
- Consider whether the same bias affects the other spawn-time-prompt L2 tests (`scheduler/dispatch/014`, `/015`), which pin the daemon-owned path.

## Related

#684, #1003, #1005, #424, and CLAUDE.md rule 5 (lane 2 runs in no CI) and rule 4 (the "as a user actually uses and sees it" bar).

vfarcic:
## Summary

Three spawn-time delivery gates (mode seed, orchestration start-role remit, dispatched task) fired on a resolved agent type alone — which a *provisional* `SessionStart` satisfies before any agent has started reading the PTY. Measured in the field at ~3.7s early: a pane spawned via a launcher had its seed typed in well before the agent announced itself, landing bytes in a PTY whose program had not started reading.

Fix: new `fn spawn_time_agent_ready` in `src/ui.rs` adds `snapshot.pane_hook_session_id(pane_id).is_some()` to the predicate; the three gates call it. The 10s `timeout_ready` slow path is deliberately untouched, so producers that legitimately announce nothing are not stranded.

Adds the PTY-attached real-agent regression test `tests/e2e_seed_late_announcement.rs` (`prompt/new-pane/016`) for exactly this race — the prior version of that test passed with the bug present because the harness's directly-exec'd agent booted fast enough to win the race it exists to exercise.

Closes #1005
Closes #1006

## Conclusions for reviewers (not re-derivable from the diff)

- **CLAUDE.md rule 12 — answered explicitly: neither.** No `PROTOCOL_VERSION` bump (stays 9, `src/daemon_protocol.rs:346`) and no `changelog.d/*.breaking.md`. `src/ui.rs` is the only changed source file, so no wire shape moved and no field changed meaning; the gate reads state the TUI already derives from events the daemon already broadcasts. An older TUI against a newer daemon is unaffected; a newer TUI against an older daemon falls back to the 10s path at worst. The narrowing strictly *removes* a write moment, so a double-write is structurally impossible.
- **The cross-version manual test was deliberately not run**, and the reviewer judged it genuinely not owed: the semantic-break class rule 12 exists to catch requires both sides to read the same bytes differently *and* one side to be the daemon — but the daemon is byte-identical here, and the TUI-side interpretation that did change is already exercised against the real binary by `prompt/new-pane/017` and `prompt/pane-input/037`.
- **One pre-existing caveat the narrowing marginally widens** (reviewer's finding, not a defect introduced here): both `timeout_ready` and the 60s deadline for the orchestrator path use `.is_some_and(...)` over `orchestration_prompt_anchor_at`, so a *missing* anchor would neither deliver nor abandon — and the fast path can no longer rescue that case. All four production stamp sites were verified present and unconditional (`src/ui.rs:10512`, `:12584`, `:13633`, `:13765`), so this is theoretical.
- **Lane-2 verification**, since no runner anywhere runs it: `cargo test-e2e-live new_pane_016` was run locally with real credentials and passes, and was verified to genuinely FAIL when the fix is removed — `premature|<the entire dispatcher seed>`, sitting in the PTY three seconds before Claude started reading it. That demonstration is the point of #1006; without it the modified test would carry the same defect as the original.
- **Follow-ups filed, deliberately out of scope:** #1012 (remove the dead `process_pending_dispatches` queue — nothing in production writes to it) and #1013 (pin `prompt/new-pane/016` to a cheap model per rule 4; it currently runs on whatever the host selects, measured as Opus 5 at xhigh).

## Gates (re-run before commit, per CLAUDE.md rule 2)

| gate | result |
|---|---|
| `cargo xtask linkage-check` | `ok (661 catalog ids, 592 annotations, 100 allowlisted, 12 rules)` |
| `cargo test-fast` | 2738 / 2738 |
| `cargo test-e2e new_pane_017 mode_005 spawn_005` | 4 / 4 |
| `cargo test-e2e-live new_pane_016` | PASS, incl. under `DOT_AGENT_DECK_REQUIRE_REAL_E2E=1` |

Full e2e tier not run locally per issue #502 — lane 1 runs in CI as `e2e-deterministic`.

Demo reel for issues #1005 / #1006: https://youtu.be/uA08JZNr4k4. Watch before merging.


vfarcic:
Demo reel for issues #1005 / #1006: https://youtu.be/uA08JZNr4k4. Watch before merging.

prageethw (approved):
First review on this one. Confirmed spawn_time_agent_ready is called at all three named gate sites (process_pending_dispatches, process_pending_seed_prompts, deliver_orchestrator_prompt), each site's 10s timeout_ready slow path is untouched, and the diff leaves every production orchestration_prompt_anchor_at stamp site alone (only pre-existing test helpers were adjusted to keep exercising the fast vs. slow path now that a resolved agent type alone no longer satisfies agent_ready). The new e2e_seed_late_announcement.rs test drives a real claude-named launcher that withholds the actual agent for 3s behind a wrapper_fork SessionStart, which is a genuine regression harness for this race, not a trivially-passing one. No issues found. CI green.

PR #1099, opened 2026-09-15
test(e2e): pin prompt/new-pane/016's real agent to Haiku (#1013)

### LINKED ISSUE #1013: Pin `prompt/new-pane/016` to a cheap model: it runs a real agent on whatever the host selects (measured: Opus 5, xhigh), against CLAUDE.md rule 4

### LINKED ISSUE #1102: Flaky on build-windows: `a_watcher_whose_task_has_ended_does_not_hold_the_deck_hostage` bounds a cross-runtime wait by iteration count (~4 ms), not by time
## What is wrong

`a_watcher_whose_task_has_ended_does_not_hold_the_deck_hostage` (`desktop/src-tauri/src/lib.rs`, PRD #742 M14) fails intermittently on `build-windows` and passes on rerun. Met on PR #1099, whose diff is a one-line model pin in a Unix-only e2e test and cannot touch the desktop crate.

The cause is in the test, not the product. It releases a spawned task and then waits for the claim to become reclaimable with:

```rust
for _ in 0..1_000 {
    if let Some(token) = state.start_watcher_once_for(&deck) { … break; }
    tokio::task::yield_now().await;
}
```

That loop cannot make the thing it waits for happen. `tauri::async_runtime::spawn` does not use the runtime `#[tokio::test]` built — with no `async_runtime::set` anywhere in the binary it lands on tauri's global `default_runtime()`, which is `TokioRuntime::new()`, a **multi-threaded** runtime with its own OS threads. So the release must cross a runtime boundary: a worker thread there has to be scheduled and poll the released task to completion before `is_finished` flips for the handle `WatcherClaim::watching` reads. `yield_now` reschedules only the test's own task on its own current-thread runtime and gives that worker nothing.

## The iteration count is a wall-clock budget in disguise, and a tiny one

Measured by instrumenting the loop:

| condition | iterations used | elapsed |
| --- | --- | --- |
| idle, 16 cores | 1–11 | 17–41 µs |
| 4x CPU oversubscription | 46–174 | 0.25–16.8 ms |

So the 1000-iteration bound is worth roughly **4 ms** of wall clock. At 4x oversubscription the loop **exhausted all 1000 iterations once in ten runs** — i.e. the Windows failure reproduced on Linux.

Windows loses this race far more readily than Linux: its default scheduling quantum alone is ~15.6 ms, about four times the budget the loop actually had.

## Fix

Bound by time, and poll with a sleep that genuinely parks the thread so the other runtime's worker can be scheduled — `tokio::time::timeout(Duration::from_secs(5), …)` around a loop sleeping 5 ms, matching the `tokio::time::timeout(Duration::from_secs(5), …)` idiom already used elsewhere in this file's tests. At 20x oversubscription the reclaim then takes 10.6–110.7 ms against the 5 s budget (a 45–470x margin), and the fastest of those is already larger than the entire budget the old loop had.

No product code changes; the property under test is unchanged.

Fixed in #1099 under CLAUDE.md rule 6 (a red met while working on something else is in scope by default).

vfarcic:
Closes #1013. Also closes #1102 (see the second commit).

## Commit 1 — the issue: pin `prompt/new-pane/016` to Haiku

`prompt/new-pane/016` (`new_pane_016_dispatcher_opens_dashboard_card_with_real_agent`) built its agent through `common::write_late_announcing_real_agent` with a bare `claude` and no `--model`, so the dispatcher pane — and the unit it dispatches, which inherits the same `default_command` — ran on whatever the host's managed settings selected. Measured on the dev box while working #1006: **Opus 5 at xhigh reasoning effort**, on a `[reel]`-marked lane-2 test whose agent only has to decompose one goal and call `dot-agent-deck dispatch`. The cost was silent: nothing in the test or in CI reports which model it picked, so the bill depended on whoever ran it.

CLAUDE.md rule 4 asks real-agent e2e tests to run on a cheap model, and this same file already pins Haiku in three other places (`:1542`, `:1823`, and the `:437` comment describing exactly this intent). 016 was the outlier rather than the pattern.

One decision point moved: the `--model claude-haiku-4-5-20251001` pin now rides after the resolved binary path on the launcher's `exec` line, with a function-local `HAIKU_MODEL` const matching the spelling the sibling test at `:1535` already uses.

### Two things deliberately not done

**Pinned at the call site, not inside `real_claude_path`.** That helper must keep returning the bare binary's absolute path, because the launcher is itself named `claude` (so `AgentType::from_command` resolves it as the deck resolves a production `default_command`) and a command line there would be one step from re-entering the launcher.

**The launcher indirection is untouched.** `:537` explains why the agent is routed behind one at all (#1006: without it this test cannot reach the ordering it exists to exercise). The diff adds an argument and changes nothing else — same launcher, same three-second announcement delay, same tool set — so the ordering #1006 gave this test the ability to fail on is unchanged.

The other 23 files under `tests/` that pin this model id were left alone. Consolidating them into a shared constant is a wider refactor than this issue asks for.

### Verification

**Lane 2 runs in no CI anywhere** (rule 5), so nothing here — before or after this change — exercises 016 on a runner. Verified by hand against a real Claude credential:

- `cargo test-e2e-live new_pane_016` with `DOT_AGENT_DECK_REQUIRE_REAL_E2E=1`, **three times, all green**: 75.885s, 80.637s, 75.730s. Three runs rather than one because a weaker model's failure mode here is intermittent, not deterministic.
- **The flag actually took effect** — read off the recorded `full-stream.cast` (`DOT_AGENT_DECK_RECORD=1`) rather than assumed, since a silently-ignored flag would leave the bill unchanged and the issue falsely closed. Claude Code's own banner in the pane reads `Claude Code v2.1.272 … Haiku 4.5 · Claude Team`, and `Opus` / `Sonnet` appear **nowhere** in the 353 KB stream.

The dispatched unit's own pane is not attached in the dashboard view, so its banner does not reach the recorded PTY; it runs on Haiku by construction rather than by observation, inheriting the same `default_command` launcher.

## Commit 2 — a red met on the way: the `build-windows` watcher-reclaim flake (#1102)

The first push came back with `build-windows` red on `a_watcher_whose_task_has_ended_does_not_hold_the_deck_hostage` (PRD #742 M14) — a desktop-crate test this PR's diff cannot touch. Rerun in isolation first, per rule 6: it passed, so it is a flake rather than a standing break. Picked up here anyway, because rule 6 makes a red met while working on something else in scope by default and "that belongs in a separate PR" is how it stays unowned.

The bug is in the test. It releases a spawned task and then waits for the claim to become reclaimable with `for _ in 0..1_000 { tokio::task::yield_now().await }` — a loop that cannot make the thing it waits for happen. `tauri::async_runtime::spawn` does not use the runtime `#[tokio::test]` built: with no `async_runtime::set` in the binary it lands on tauri's global `default_runtime()`, a **multi-threaded** tokio runtime with its own OS threads. The release therefore crosses a runtime boundary — a worker there must be scheduled and poll the task to completion before `is_finished` flips for the handle `WatcherClaim::watching` reads — and `yield_now` reschedules only the test's own task, handing that worker nothing.

So the iteration count was a wall-clock budget in disguise. Instrumented:

| condition | iterations used | elapsed |
| --- | --- | --- |
| idle, 16 cores | 1–11 | 17–41 µs |
| 4x CPU oversubscription | 46–174 | 0.25–16.8 ms, and **1 of 10 runs exhausted all 1000** |

That last row is the Windows failure reproduced on Linux. The real budget was ~**4 ms**; Windows' scheduling quantum alone is ~15.6 ms, four times it.

It now waits on `tokio::time::timeout(Duration::from_secs(5), …)` over a loop polling every 5 ms with a sleep that genuinely parks the thread — the `timeout(Duration::from_secs(5), …)` idiom already used elsewhere in this file's tests. At **20x** oversubscription the reclaim takes 10.6–110.7 ms against that 5 s budget (45–470x margin), and the fastest of those already exceeds the entire budget the old loop had. No product code changed; the property under test is unchanged.

## Gates

`cargo fmt --check`; `cargo clippy --workspace --all-targets --features e2e,e2e-live -- -D warnings`; `cargo test-fast` (3094 passed, 0 failed); `cargo xtask linkage-check` (ok — 682 catalog ids, 12 rules); plus the three lane-2 `cargo test-e2e-live new_pane_016` runs above and 30 stress runs of the desktop test under 4x load.

## Rule 12

**Does not apply.** Both commits are test-only: no daemon, protocol, orchestration or hook surface is touched, so there is no TUI↔daemon contract to re-check and no `PROTOCOL_VERSION` bump or `.breaking.md` fragment is owed.


prageethw (approved):
Two independent, well-verified fixes: pins the real-agent e2e test to Haiku per rule 4 (verified via the recorded cast that the flag took effect), and replaces a flaky iteration-count wait with a proper tokio::time::timeout + real sleep, with the runtime-boundary root cause (tauri's own multi-threaded runtime vs the test's) correctly diagnosed and measured.

Issue #1028, opened 2026-09-11
Desktop: wait_for_coordinator_readiness wraps the cancel-unsafe next_event in a timeout, which can desync the event stream


vfarcic:
Found while reviewing PRD #741's M4 (branch `agent/dispatch-prd-741`). **Pre-existing — introduced by `daf94f09` (#416) and not touched by #741.** Filed separately because #741's own diagnosis implies this site.

## The mechanism

`EventSubscription::next_event` is **not cancel-safe**: `read_frame` (`src/daemon_protocol.rs:649`) accumulates a five-byte frame header across `await` points into a local buffer, so a future dropped mid-read loses those bytes and leaves the `EventSubscription` desynchronised — every subsequent frame is misparsed.

`wait_for_coordinator_readiness` (`desktop/src-tauri/src/lib.rs:425-436`) awaits `watch.next_event()` inside `tokio::time::timeout(...)`. `timeout` drops the inner future on expiry, so a timeout firing mid-header-read hits exactly that case.

## Failure scenario

A coordinator spawn is slow enough for the readiness timeout to expire while a frame header is partly read. The subscription is then desynchronised: subsequent events on that stream are misparsed rather than delivered. The function returns `Ok(None)`, so the caller reads it as "not ready yet" and nothing reports the desync.

## Blast radius — small, which is why this is filed rather than fixed inline

It is a one-shot coordinator-spawn path, not the long-lived snapshot watcher, and on timeout it returns `Ok(None)` rather than propagating a corrupted stream into the main event loop.

## The fix shape, already proven next door

PRD #741's M4(b) hit the same class deliberately and solved it: drain the subscription in a task of its own into a bounded `mpsc`, and `select!` only over cancel-safe operations (`mpsc::Receiver::recv` and `Interval::tick`). The same treatment applies here — or, more cheaply for a one-shot path, avoid dropping the `next_event` future at all (time out around the *whole* readiness operation and discard the subscription afterwards, rather than timing out a single read).

## Worth checking in the same pass

Whether any other site wraps `next_event` or `read_frame` in a `select!`, `timeout`, or anything else that can drop the future mid-read.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #1100, opened 2026-09-15
fix(desktop): bound the coordinator-readiness subscription (#1084)

### LINKED ISSUE #1084: subscribe_events on the coordinator-readiness path is still unbounded; a wedged daemon hangs it forever
## What is unbounded

PRD #742 M14 (`624b7bb6`) bounded three client RPCs that could previously never return, wrapping each in `bounded_reply(...)` at `DECK_REPLY_TIMEOUT` = 15s:

- `hello()` inside `establish()`
- `subscribe_events()` on the fleet-snapshot path (`lib.rs:868`)
- `list_agents()` in `snapshot_with`

The defect it fixed is worth restating because it applies unchanged here: a deck that is **down** fails immediately (`ECONNREFUSED`), but a deck that **accepts the connection and never answers** — a wedged daemon, a stalled ssh channel — was not a failure any caller could observe. `read_response` waits as long as the peer keeps the socket open, so it was a future that never completed.

**Two client RPCs on the coordinator/workflow path were deliberately left out of that change**, because it was scoped to the fleet snapshot:

- **`subscribe_events()` at `desktop/src-tauri/src/lib.rs:420`** (`begin_coordinator_readiness`) — unbounded.
- **`reconcile_workflow_agent` at `desktop/src-tauri/src/lib.rs:399`** — already carries its own explicit `timeout` parameter, so it is bounded, but separately and by a different mechanism.

## Why it matters

A wedged daemon on the coordinator-readiness path produces the same symptom the fleet path used to have: a future that never completes, with nothing on screen saying why. M14's report noted the fleet-side version had a blast radius wider than the milestone — before the bound, a wedged daemon also froze `desktop_get_snapshot` and every `DesktopAction`'s `refresh_and_emit` **indefinitely**.

This is not urgent: the coordinator path runs against the **selected** deck during a workflow launch, so it is a narrower surface than the fleet watcher, and a user watching a launch has more context for a hang than someone glancing at an overview. But it is the same class, and it is now the odd one out.

## What a fix should decide

Not simply "add the same 15s". Worth settling:

- **Whether all client RPCs get a uniform bound**, or whether the coordinator path wants a different one. `DECK_REPLY_TIMEOUT` was chosen at 15s for reasons specific to the snapshot path: three to four orders of magnitude above what an in-memory registry reply needs, **below** `FORWARD_READY_TIMEOUT` (30s, `src/remote_tunnel.rs:949`) so the two bounds add rather than multiply, and **above** `RECONCILE_INTERVAL` (5s, `desktop/src-tauri/src/agent_view.rs:111`) so a merely quiet deck is never mistaken for a stalled one. A coordinator-readiness wait has different neighbours and may want a different number.
- **Whether `reconcile_workflow_agent`'s own `timeout` parameter should fold into the same mechanism**, so there is one way to bound a reply rather than two.
- **The boundary rule, which M14 got right and a uniform sweep could easily get wrong:** only the single RESP that *confirms* a subscription may be bounded. The long-lived event stream after it must not be — `subscribe_events` awaits one RESP and returns `EventSubscription`, whose frames are read later by `spawn_event_reader`. A bound applied to the stream would kill every healthy subscription.

## Ordering

**Independent — after PRD #742's PR ([#1076](https://github.com/vfarcic/dot-agent-deck/pull/1076)) merges**, since `bounded_reply` and `DECK_REPLY_TIMEOUT` arrive with it and are what a fix here would reuse.

## Evidence

Found by a scoped review of `624b7bb6` during PRD #742's pre-merge verification, and flagged deliberately as a conscious exclusion rather than an oversight.

vfarcic:
Closes #1084.

## What changed

`begin_coordinator_readiness` (`desktop/src-tauri/src/lib.rs`) awaited `subscribe_events()` with no deadline. A deck that is **down** fails at once with `ECONNREFUSED`; a deck that **takes the connection and never answers** — a wedged daemon, a stalled ssh channel — left `read_response` waiting as long as the peer kept the socket open. A workflow launch parked forever, with nothing on screen saying why.

That call is now wrapped in `bounded_reply("SubscribeEvents for coordinator readiness", …)` at `DECK_REPLY_TIMEOUT`, the same mechanism PRD #742 M14 (`624b7bb6`) applied to the handshake, the fleet path's own `SubscribeEvents` (`lib.rs:931`) and `ListAgents`. The coordinator path now matches the *same call* on the fleet path.

`bounded_reply` already applies `safe_message` to the error arm, so the transport-failure path is byte-for-byte what it was; the only new outcome is the elapsed one.

## Which future this timeout can drop — it is not `next_event`'s

This is the question #1028 makes load-bearing, so stating it explicitly.

`bounded_reply` wraps **`self.subscribe_events()`** — the single round trip that awaits one RESP and returns an `EventSubscription`. That completes **before** the draining task exists, so the only future this timeout can ever drop is `subscribe_events()`'s own. It is never `next_event`.

Nothing inside the drain loop is wrapped, and nothing below the change moved. `next_event` is still owned solely by the spawned task, still read under no deadline — `read_frame` accumulates a five-byte header across await points, and a future dropped mid-read desynchronises the subscription permanently, which is exactly the defect #1028 fixed in #1069. The long-lived event stream after the confirming RESP is deliberately unbounded, which is also the boundary rule the issue names.

## Explicitly not in scope

- **`reconcile_workflow_agent`** already carries its own explicit `timeout` parameter. It is bounded, separately and by a different mechanism, and is left alone — no attempt to unify the two.
- No sweep for other unbounded calls across the crate. This is one call site.

## Rule 12 does not apply — stated rather than assumed

This adds a **client-side** timeout. No wire shape moves, no field meaning shifts, no handler contract changes: `PROTOCOL_VERSION` is untouched and no `.breaking.md` fragment is warranted. The cross-version manual test is therefore not run.

## Tests

New `#[cfg(unix)] #[tokio::test]` in `desktop/src-tauri/src/lib.rs`:

**`tests::a_deck_that_never_answers_the_readiness_subscription_is_reported_rather_than_awaited`**

A scripted daemon on a real unix socket accepts the subscription, reads the `SubscribeEvents` request and answers nothing, holding the write half bound so the client sees a stall rather than EOF. `begin_coordinator_readiness()` must come back with an error that says the deck took the connection and did not answer, naming the bound it exceeded.

Shaped after `daemon_bridge.rs`'s M14 sibling for the handshake, which is the nearest existing coverage of this property:

- The clock is paused only **after** `accepted_rx` resolves — the peer confirming it took the connection and read the request. `start_paused` would auto-advance from the first moment the runtime had nothing to poll, producing the same error for a scenario nobody wrote. Pausing at that point leaves exactly one thing outstanding, so the runtime advances to the only deadline left: real socket I/O, real silence, none of the fifteen seconds spent (runs in 0.021s).
- The outer `timeout(600s)` is what turns a regression into a red test instead of a hung job. Under a paused clock tokio advances to the nearest deadline, so the bound under test fires first while it has one.

**Verified red, not just green.** With the `bounded_reply` wrapper reverted and the test unchanged, it fails in 0.00s on `begin_coordinator_readiness() must bound its own wait rather than await a reply that is not coming: Elapsed(())`. With the fix, it passes.

### What was run

- `cargo test-fast` — **3095 run, 3095 passed, 0 skipped.**
- Scoped to what was touched: the whole `dot-agent-deck-desktop` unit suite, **260 run, 260 passed**. That includes `tests::a_readiness_timeout_mid_frame_header_leaves_the_watch_usable` — #1028's own regression test, which drives `begin_coordinator_readiness` against a real socket and is the one that would go red if this change had disturbed the drain — plus `non_pi_launch_waits_for_readiness_and_retries_with_one_identity`, `pi_coordinator_is_rejected_before_subscription_or_spawn`, `hookless_coordinator_uses_timeout_fallback_without_session_guard` and the rest of the workflow-launch set.

### Lane 2 — structurally unreachable, not skipped

No `e2e-live` test covers this surface, and none can. `begin_coordinator_readiness` is a private trait method on a type inside `dot-agent-deck-desktop`; `tests/e2e_remote_deck_transport.rs`'s own header already records why no `tests/e2e_*.rs` file can name such an item, and a dev-dependency the other way is ruled out there (it would make Tauri and its GTK/WebKit closure a build input of that tier on three platforms) and would not reach a `pub(crate)` item anyway. `tests/CATALOG.md`'s `fleet/observe/001` says the same from the other side: everything the desktop crate owns is pinned in that crate's own `#[cfg(test)]` module under `cargo test-fast`, which is where this test went. Lane 1 runs in CI on this PR.


prageethw (approved):
Sound fix using the existing bounded_reply helper (same pattern as the handshake/fleet-view bounds), correctly scoped to only the subscription confirmation and not the long-lived event stream. The new test is a real socket with a paused clock stopped at the right point (after accept+read, not before), which correctly reproduces the wedged-daemon scenario without spending real wall-clock time.

Issue #954, opened 2026-09-09
delegate_022's grace-window precondition can be missed entirely: wait_for_close_in_flight samples every 5 ms and bails on is_finished()


vfarcic:
## Symptom

`delegate_022_delegate_during_an_in_flight_close_brings_the_role_back` failed one full `cargo test-fast` run at **0.973s** on its *precondition*, not its subject:

```
precondition: the close never entered its grace window, so the delegate below would be
an ordinary post-close delegate instead of #606's race; stop_agent finished = true,
stand-in still live = true
```

It passes in isolation (4.2s) and passed 10/10 module runs immediately afterwards, plus a subsequent full-tier run at 2484/2484. So the lane is green and this is intermittent, not a standing red.

## Cause

`wait_for_close_in_flight` (`tests/delegate_respawn_recovery.rs:108`) polls `registry.pane_close_in_flight(pane_id)` on a **5 ms** sleep, and returns `false` as soon as `request.is_finished()`:

```rust
loop {
    if registry.pane_close_in_flight(pane_id) { return true; }
    if request.is_finished() { return false; }
    if Instant::now() >= deadline { return false; }
    tokio::time::sleep(Duration::from_millis(5)).await;
}
```

The close being waited on is entirely in-memory — the worker stand-in is already dead by an earlier precondition, so nothing spends a real termination grace. The in-flight window can therefore open and shut inside a single 5 ms tick, and the poller then observes only the finished request. The assertion's own text names the case it cannot distinguish: "the close already finished" versus "the close had not started".

This is the residual of #709, which reworked this exact deadline from a flat `sleep(200ms)` to this poller. #709's comment records the same signature — "a full-tier run whose whole failing case took 0.98 s" — so the window is narrower than what it fixed but has not closed.

Sibling precedent for the genre: #364 (`delegate_011`, wall-clock-sensitive), #531, #807, #936.

## Why polling cannot close it

Shrinking the interval narrows the window without removing it, and the failure is a *sampling* miss rather than a starved budget — the budget was nowhere near exhausted at 0.973s. Two shapes that would actually be total:

- have the registry record that the window **was** entered (a latch or a close-event subscription) so the observation does not depend on when the poller happens to look; or
- treat "the close completed without the window ever being observed" as **inconclusive** — retry the whole close rather than failing the precondition, since an unobserved window is not evidence that there was none.

## Not doing

[...]

Issue #828, opened 2026-09-02
desktop.toml: two processes racing on the same field silently lose one edit (blocks nothing until #741)


vfarcic:
Two processes — two windows of the desktop app, or the app and a hand edit — can each read `desktop.toml`, each change a *different* field, and each write the whole merged document back. The second write wins on every field it touched, including the one it never meant to change, and neither process sees an error.

## What is already protected, and what is not

`save_to` re-reads the document at save time and deep-merges the struct into it (`merged_document`), so an **unknown section** cannot be lost: a build that has never heard of `[voice]` still writes it back byte-identically, which is the property PRD #803's container promise rests on and it is tested. Within-process ordering is also now handled — `useDesktopSettings` serialises its saves and drops a stale response, so two rapid clicks cannot publish in reverse order.

What is unprotected is **two writers racing on the same known field**. The merge makes the struct authoritative over the file for every field the struct owns, which is correct for one writer and wrong for two: process A's stale `[endpoints]` overwrites process B's fresh one, because A's in-memory struct is the authority and it was loaded before B's write.

## What would fix it

The audit that raised this recommended two things, both deliberately deferred by PRD #803 as disproportionate for a preferences file:

- an **interprocess file lock** held across the read–merge–write, so the two saves serialise instead of interleaving;
- a **compare-and-retry transaction**: record what was read, and if the document changed underneath, re-merge onto the new content rather than publishing over it.

## Why it is deferred, and when it stops being deferrable

Today the whole document is one appearance mode, so the worst case is that a preference set in one window loses to another window's stale copy — annoying, invisible, and self-correcting the next time anyone changes it.

It becomes load-bearing when **#741 lands daemon endpoints**. A lost endpoint edit is not a cosmetic preference: it is configuration the user believes they saved, and the failure surfaces later as "the app is pointing at the wrong daemon" with nothing in the document to explain why. #802's model and backend choices are the same shape. Revisit this before either ships a section with more than one field a user edits.

Refs #803, #741.

Issue #827, opened 2026-09-02
BEFORE #802 stores an API key: the desktop secret guard is a naming tripwire, not a boundary


vfarcic:
**Read this before storing an API key in the desktop app.** PRD #802 needs a credential (an optional API key for a remote speech-to-text or intent backend). PRD #803 set the rule it has to obey — *a secret never goes in `desktop.toml` and never in `localStorage`* — and pinned it with a guard test. That guard is a **naming tripwire, not a security boundary**, and this issue is the list of checks that actually have to exist before #802 accepts real credential material.

## Why the existing guard is not enough

`no_settings_key_name_trips_the_credential_tripwire` (`desktop/src-tauri/src/settings.rs`) walks the **key names** of the serialised default document and fails on one containing `key`, `token`, `secret`, `password`, `credential`, `authorization`, `bearer` or `passphrase`. Its name, its doc comment, its failure message and `docs/develop/desktop-gui.md` all now say what it is, because the real danger is a future author reading "no secrets guard" and concluding the problem is solved. Four things it cannot see, none hypothetical:

- **a field whose name says nothing** — `endpoint`, `authorization`, `value` — whose *value* is a token. This is how a credential actually arrives;
- **a field serde omits** from the default document, which is invisible because the scan is *of* that document;
- **anything the TypeScript DTO carries that the Rust schema does not** — a frontend-only `apiKey` on `DesktopSettingsDto` is outside the Rust test entirely;
- **any value, ever.** It never looks at one.

## The checklist

An end-to-end test that submits a uniquely-named sentinel credential through the real secret-store flow, and then proves the sentinel string appears in **none** of:

- [ ] the serialised TOML on disk (`desktop.toml`), including after a load–modify–save round trip;
- [ ] any `window.localStorage` value, under any key, in either runtime mode;
- [ ] any IPC payload — the `desktop_get_settings` snapshot, the `desktop_set_settings` echo, and any emitted snapshot event;
- [ ] any log line (the deck log and stderr; note `SettingsWriteError::detail()` is deliberately the half that carries a path, so it is the one to check);
- [ ] any error message, on either side of the `SettingsWriteError` public/detail split, and any `safe_message` output.

Plus three structural checks:

[...]

Issue #825, opened 2026-09-02
Desktop: desktop.toml loses comments and inline-array formatting on save


vfarcic:
`desktop/src-tauri/src/settings.rs` saves by merging the serialised settings struct into the document already on disk rather than replacing it, so a section or field this build does not know about survives the write (the fix in `c51242a`, PRD #803). The merge round-trips through `toml::Table`, which models **data** — so no unknown *data* is ever lost, and the round-trip test proves it — but the document is re-rendered canonically on the way out.

What that loses:

- comments, including a user's own annotations;
- inline-array and inline-table formatting (`x = [1, 2]` comes back expanded to a multi-line array);
- key order and blank-line grouping inside a table.

The practical consequence is small but real: PRD #803 makes "a file a user can read, edit and delete without the app running" a success criterion, which invites hand-annotation — and those annotations disappear the next time the app writes any setting.

Fixing it means `toml_edit`, a format-preserving TOML DOM, in the desktop crate. That is a new dependency and was deliberately not taken in the PR that found this: the data-preservation property was what the container's promise depended on, and adding a second TOML library is a decision worth making on its own rather than as a footnote. The limitation is already pinned by `an_unknown_section_keeps_its_data_but_not_its_formatting` in `settings.rs`, so it is recorded behaviour rather than an assumption.

Closing this means: the save path preserves comments and formatting, and that test's expectation inverts.

PR #1106, opened 2026-09-15
fix(desktop): never overwrite a desktop.toml this build cannot read (#1072)

### LINKED ISSUE #1072: A malformed desktop.toml silently becomes all-defaults, and the next save overwrites the original
## The behaviour

`load_from` (`desktop/src-tauri/src/settings.rs:1513`) handles **every** failure to read `desktop.toml` the same way: log a locator to stderr, then return `DesktopSettings::default()`.

```rust
Ok(Some(contents)) => match toml::from_str(&contents) {
    Ok(settings) => settings,
    Err(error) => {
        eprintln!("{}; using defaults", invalid_document_log(path, &contents, &error));
        DesktopSettings::default()
    }
}
```

That arm catches **two different things**: a TOML syntax error, and a document that is valid TOML which this build's *schema* rejects (every field newtype runs its validator in `Deserialize`).

The app then starts, looking completely normal, with every setting at its default. **The next settings save writes those defaults over the user's file** — `merged_document` overwrites every key the struct owns, so appearance, zoom, endpoints and the deck selection are all replaced. The only signal is an `eprintln!` nobody reads.

So a malformed file does not present as an error. It presents as "my settings reset themselves", one save later, with the original already gone.

## What should happen

Show the user an error at desktop startup naming the file and what is wrong with it, and **do not let a failed load become a silent overwrite** — preserve the unreadable file rather than replacing it.

The existing diagnostic is careful about what it may print: `invalid_document_log` deliberately emits a **locator** (path + line) and never the document's own bytes, because `toml::de::Error`'s `Display` echoes the offending value twice and the deck log is a file people paste into bug reports (issue #827). Any user-facing message should keep that property.

## The related question: should `desktop.toml` be hand-editable at all?

Today it explicitly is. [`docs/develop/desktop-gui.md`](docs/develop/desktop-gui.md) carries a **"Configuring a remote deck by hand"** section, and every field newtype runs its validator inside `Deserialize` specifically so that *"a hand-edited `desktop.toml` cannot smuggle past what the settings form applies."*

The argument for dropping that: someone who prefers editing config by hand is not reaching for a desktop GUI. Supporting both means every field needs a hand-edit story, and it is the hand-edit path that makes the failure above reachable in the first place.

Worth deciding deliberately rather than leaving as an inherited default. Note that dropping it does **not** remove the need for the fix above — an interrupted write, a dotfile-sync conflict marker, or disk damage all produce the same malformed file with no human having typed anything.

## Why this is worth more than its likelihood suggests

**Tightening any field validator converts previously-valid documents into unloadable ones**, and this codebase tightens validators routinely — `port` moved from a bare `u16` to a newtype over `NonZeroU16` after a hand-edited `port = 0` was found being handed to OpenSSH as `-p 0`.

PRD #742 added a fresh instance: `EndpointId::parse` now reserves `all` (it previously reserved only `local`), so a document carrying `id = "all"` loaded before that change and does not after. That specific value is close to unreachable in practice — the app mints sixteen hex characters — but it is the second known instance of the pattern, and the pattern is what makes this worth fixing rather than any one value.

## Ordering

**Independent of PRD #742's PR — can be done before or after, in either order.** It touches `load_from` and the settings surface rather than anything in the fleet view. Expect light conflict in `settings.rs` if both are in flight at once.

## Evidence

Found while building PRD #742: the milestone that reserved the `all` token accepted "a hand-written `id = "all"` now refuses to load" as a clean trade, and the trade turned out not to be clean — refusing to load means silently resetting every setting on the next save.

vfarcic:
Closes #1072.

## The bug

`load_from` handled **every** failure to read `desktop.toml` the same way: log a locator to stderr, return `DesktopSettings::default()`. The app then started looking completely normal with every setting at its default, and the next settings save merged those defaults over the user's file — `merged_document` makes the struct authoritative over every key it owns, so appearance, zoom, endpoints and the deck selection were all replaced.

So a malformed document did not present as an error. It presented as *"my settings reset themselves"*, one save later, with the original already gone.

The arm caught two different things and both are reachable: a TOML syntax error, and a document that is valid TOML which this build's **schema** rejects (every field newtype runs its validator inside `Deserialize`). The second is the dangerous one, because tightening a validator converts previously-valid documents into unloadable ones and this codebase does that routinely — `port` became a newtype over `NonZeroU16` after a hand-edited `port = 0` reached OpenSSH as `-p 0`, and PRD #742 reserved `all` as an endpoint id. Neither needs a human to have typed anything, either: an interrupted write, a dotfile-sync conflict marker or disk damage produce the same file.

## The decision, and why

The issue asks for a deliberate choice between **refuse to start**, **start read-only**, **start with defaults but refuse to save**, and **prompt the user**. This PR takes **start with defaults, say so, and refuse to save**.

- **Not refuse to start.** Disproportionate for a preferences file, contradicts the module's own stated position ("a settings file is not worth failing an app launch over"), and hands anyone who can write that file a denial of service on the whole app. It is also the one option a GUI-only user cannot recover from without a terminal.
- **Not prompt.** A startup dialog offering "replace it?" puts the destructive action one click away at the exact moment the user has the least information — which is the failure this issue is about, with a confirmation step bolted on.
- **"Start read-only" and "defaults but refuse to save" are the same thing here**, and the second is the honest description: the app is fully usable, the in-memory document is this build's defaults, and nothing is written back until the file is readable again.

This is consistent with how the rest of the settings surface already behaves. A save that fails for any other reason (read-only directory, full disk) already leaves the change applied for the session and says it will not survive a restart; an unreadable *document* is the same shape of answer with a different reason, which is why it renders through the same channel.

**Deliberately not included: an in-app "replace it anyway" button.** The destructive action is exactly the one being prevented, the message names the line to open, and a user who genuinely wants defaults can delete the file. Adding a one-click path back to the data loss would undo the point.

**`desktop.toml` stays hand-editable**, which the issue raised as a linked question. The argument for dropping it — someone who edits config by hand is not reaching for a desktop GUI — does not survive the paragraph above: dropping it removes neither the failure nor the need for this fix, while `DOT_AGENT_DECK_DESKTOP_CONFIG`, "Configuring a remote deck by hand" and PRD #803's "a file a user can read, edit and delete without the app running" all depend on it. The validating newtypes are what make hand-editing safe, not what makes it costly. Written down in `docs/develop/desktop-gui.md` so it is a recorded decision rather than an inherited default.

## What changed

**Rust (`desktop/src-tauri/src/settings.rs`)**

- `load_document` replaces `load_from` and returns `(DesktopSettings, Option<SettingsDocumentProblem>)`. Loading still never fails; the reason is returned instead of dropped. An **absent** or **empty** document is deliberately *not* a problem — an empty file is valid TOML and genuinely means "everything at its default", and reporting it would put a permanent error in front of every first-run user and refuse their first save.
- `SettingsDocumentProblem` splits `detail` / `public` exactly the way `SettingsWriteError` does: the path goes in the log half and never in the half that crosses the bridge.
- `load_snapshot` carries the public half on `DesktopSettingsSnapshot.problem`, omitted from the wire when there is nothing to say.
- **`save_to` refuses.** It re-reads the document at save time and parses it as `DesktopSettings` — *not* as `toml::Table`, because only that catches the schema-rejection half. Re-reading rather than trusting the load also covers a file that became unreadable while the app was running. The refusal happens before `create_temp`, so nothing is created beside the file either.

  **The claim is "cannot be read *at save time*", and the qualifier is meant** (rule 17 — narrowed in the second commit after I wrote it too wide in the first). `save_to` reads, vets and then renames, so a document that becomes unreadable *inside that window* is still replaced. That is a microsecond-wide race between two writers — #828's subject, and the same shape of residual the path vet already accepts a few paragraphs up in the same module. What #1072 was about is the ordinary single-writer case, where the app read the document at launch, could not use it, and overwrote it anyway; checking at save time rather than inheriting the launch read is what shrinks the window to that race.
- `merged_document`'s `existing.and_then(…).ok().unwrap_or_default()` becomes an error. Unreachable through `save_to` now, and kept because defaulting is precisely what made the loss silent: it turns "I could not read this file" into "this file was empty", and the merge then publishes a whole-file replacement wearing a merge's clothes. Driven directly by a test, so it is a guarded invariant rather than dead code.

**Frontend**

- `DesktopSettingsSnapshotDto.problem` + its normaliser (`bridge.ts`), pinned in `PINNED_TS_FIELDS` with its reason — it is the second legitimate string on that DTO and it is a *diagnostic*, not a value.
- `useDesktopSettings` holds it as its own state and folds it into `saveError`, so all three settings panels render it with no change to `App.tsx` or `SettingsSheet.tsx`. It is kept separate from the save-failure state on purpose: a save begins by clearing the last failure, and sharing one variable would blink the explanation out on the click and bring it back when the refusal lands a round trip later — the one moment the user is reading it.
- The "This change is applied, but saving it failed…" lead-in moved from the three panels into the hook. The panels now render `saveError` verbatim, because a panel that composed that preamble itself would be wrong half the time: a document problem is already a whole sentence and has not earned a claim about a save the user never made.

## Tests

The regression test is `a_document_this_build_cannot_read_is_never_overwritten`, written **before** the fix and confirmed failing for the user's reason — `save_to` returned `Ok(())` and the file had been replaced. It covers both failure shapes and asserts the bytes on disk are identical afterwards.

Each half of the fix was reverted individually to confirm it is load-bearing: reverting the `save_to` guard reddens 4 tests, reverting the `merged_document` guard reddens its own.

Controls, so the red means what it should: `a_readable_document_still_saves_and_still_merges` (a good document still saves and still keeps a section this build does not own — PRD #803's container promise, which the refusal must not have cost) and `an_absent_or_empty_document_is_not_a_problem`.

**Three existing claims were narrowed rather than left standing** (CLAUDE.md rule 17), and each is a case where the old assertion was true *because of the bug*:

| claim as written | what is true now |
| --- | --- |
| #827's sweep: *"the value does not survive on disk either"* | the scrub was a side effect of publishing defaults over a document nobody understood. A value that still lets the document load is overwritten as before; one that makes it unreadable stays in the user's own `0o600` file — the same answer already given for a key the schema does not own. Both arms are exercised and the test fails if either stops being reached. |
| `a_row_this_build_cannot_read_survives_the_next_save` | it proved the row survived by asserting the theme change landed beside it — which *is* the data loss. Renamed and rewritten to assert the file is byte-identical. |
| `no_settings_write_error_carries_a_value_from_the_document`: *"every `SettingsWriteError` … is constructed from a path and an `io::Error`"* | the refusal is built from `parse_locator`, which reads the document only to count newlines. The narrower claim is spelled out and the refusal is added to the swept set. |

The user-facing sentence is pinned whole on the Rust side, because the two frontend test files hard-code the same string — a rewording should be a deliberate diff on both sides rather than a silent drift on one.

## Gates run

| gate | result |
| --- | --- |
| `cargo test-fast` | **3125 passed, 0 skipped** (see the flake note below) |
| `cargo nextest run -p dot-agent-deck-desktop settings::` | 79 passed |
| `cargo nextest run -p xtask-linkage-check desktop_settings_secrets` | 11 passed, including the `DesktopSettingsSnapshotDto` pin |
| `pnpm vitest run` (desktop) | 518 passed, 19 files |
| `pnpm exec tsc --noEmit` and `pnpm test:browser:types` | clean |
| `pnpm test:browser` (Playwright, Chromium + WebKit) | 112 passed |

**Two unrelated reds met on the way, both phantom flakes from parallel contention.** Run 1 failed `idle_worker_008_closed_orchestrator_pane_id_reuse_receives_nothing`, run 2 failed a *different* test (`delegate_022_delegate_during_an_in_flight_close_brings_the_role_back`) and passed the first, run 3 was fully green. Both pass in isolation. That two-different-tests-per-run signature is exactly what #701 and #351 record, `delegate_022` has its own open issue (#954) about a missable precondition, and `idle_worker_008`'s own comment cites #709 for contention scaling. Neither can be caused by this branch: they are root-crate integration tests, the root crate does not depend on the desktop crate, and nothing in this diff is linked into those binaries. Reported rather than passed over, per rule 6.

**No e2e/L2 test.** The desktop app's PTY-free tiers are vitest and Playwright and both are covered. A Playwright test specifically for this cannot be honest: the browser preview's `FixtureDeckBridge` keeps settings in `localStorage` and structurally cannot reach a file, so there is no unreadable document for it to have. The user-visible assertion lives at the vitest level instead (`App.test.tsx`: open Settings against a snapshot carrying a problem, see the line and column, see no invented "saving failed" claim, and see the footer name the file).

## The two sibling issues

**Neither falls out of this, and neither is taken.**

**#825** (comments and inline-array formatting lost on save) needs `toml_edit` — a second TOML library in the desktop crate — which is a decision worth making on its own rather than as a footnote here. `an_unknown_section_keeps_its_data_but_not_its_formatting` still pins the limitation, unchanged. Worth knowing for whoever takes it: `merged_document` is the function #825 replaces, and the parse guard added in front of it here is a property that has to survive that rewrite, not a line to delete along the way. One adjacent thing did improve by accident — comments in a document this build *cannot* read are now preserved byte for byte, because the save is refused rather than re-rendered.

**#828** (two processes racing on the same known field) is untouched. `save_to` re-reads the document at save time and now also validates it, which is where a compare-and-retry transaction would go — but the merge still makes this process's struct authoritative over every field it owns, and that is the whole of #828. Its own trigger condition is unchanged: revisit before #741/#802 ships a section with more than one field a user edits.

## Scope

Stayed out of `AgentTile.tsx`, `App.tsx` and `AgentOverview.tsx` per the dispatch boundary. `desktop/src/lib/bridge.ts` was the one boundary file touched, on explicit instruction after asking: the frontend's only ingress for the snapshot is `normalizeDesktopSettingsSnapshot`, which constructs its result rather than copying its input (a #827 guard that is not going away), so a field that does not pass through it cannot reach the UI at all. The change there is 19 additive lines in the settings DTO region.


prageethw (approved):
Reviewed the full diff (Rust settings.rs + TS hook/panels). This is a real data-loss fix: a document this build can't read used to load as defaults and the next save silently overwrote it. The fix re-parses at save time against the actual DesktopSettings schema (not just TOML::Table) to catch both syntax and schema-rejection failures, refuses to write rather than merging into an empty table, and threads the reason through to the settings UI without ever including document bytes or a filesystem path in the public half — consistent with the existing #827 redaction discipline. Extensive test coverage on both the Rust and TS sides, including the credential-sink sweep updated to cover both the drop and refuse arms. No prior review threads open.

Issue #1109, opened 2026-09-15
A stray SIGTERM stops every agent, but 'daemon stop' refuses to: the signal path bypasses #770's guard


vfarcic:
## The asymmetry

`daemon stop` **refuses** while the daemon holds orchestration roles whose panes still have a live agent (#770). That refusal exists for a good reason, recorded in `CLAUDE.md` rule 15: the role maps (`pane_role_map`, `pane_orchestration_map`, `orchestrator_pane_ids`, `pane_cwd_map`) live in the daemon's memory and nowhere else, so stopping it deletes them. An agent that survives the restart keeps posting hook events, keeps looking healthy on its card, and **can never delegate again**. The refusal names the panes and roles at stake so the choice is informed, and `--force` is documented as abandoning a run.

A raw `SIGTERM` gets none of that. It produces a compliant graceful shutdown that stops every managed agent, with no refusal, no prompt, and no mention of what was running:

```
daemon received termination signal; initiating graceful shutdown
  (every managed agent will be stopped) signal="SIGTERM"
```

So the sanctioned path is guarded and the signal path is not, for the same destructive outcome.

## Why this is being filed now

Observed for real on 2026-09-15 — see [#428's occurrence #5](https://github.com/vfarcic/dot-agent-deck/issues/428#issuecomment-5688814518). An agent tearing down a cross-version test sandbox ran `pkill -f "daemon serve"`, which matched the production daemon's command line. Nine panes across three dispatched units were stopped 1.4 s later. `pkill` sends `SIGTERM`, so #770's guard was bypassed without anything having to defeat it.

That specific cause is best addressed by an instruction (a `CLAUDE.md` rule 12 note about sandbox teardown), and that is being handled separately. This issue is the narrower structural question it exposed.

## The question, which is genuinely open

**Should the signal path carry any of the same protection, and if so what?**

This is not obviously "yes". `SIGTERM` is how a service manager, a container runtime, a session logout or a system shutdown asks a daemon to stop. A daemon that argues back has real failure modes of its own:

- systemd escalates to `SIGKILL` after `TimeoutStopSec`, which is strictly worse — no graceful shutdown at all, and the agents are orphaned rather than reaped.
- A container runtime does the same on `docker stop`.
- Refusing to die is a classic way to make a machine hard to shut down.

So the plausible shapes are narrower than "ignore SIGTERM":

[...]

Issue #770, opened 2026-08-30
A daemon restart silently orphans running orchestrations: delegate dies, everything else keeps working


vfarcic:
Observed live on 2026-08-30 while running PRD #740, and separately by the maintainer in a second, unrelated orchestration. Not a regression — the mechanism has been there all along.

## Symptom

An orchestrator that has been delegating happily suddenly cannot:

```
delegate from pane sched-dispatch-issue-318-319-17-r0 failed: the daemon holds
no orchestration role for pane sched-dispatch-issue-318-319-17-r0, so this
delegate was routed nowhere. Only a pane spawned as part of an orchestration
can delegate.
```

Nothing else looks wrong. The pane keeps running, the agent keeps working, the card stays live.

## Mechanism

`State::register_orchestration_role` (`src/state.rs:5799`) populates `pane_role_map`, `pane_orchestration_map`, `orchestrator_pane_ids` and `pane_cwd_map` **in memory**. There is no persistence path — a search for serialization or restore of those maps finds nothing, and `~/.local/state/dot-agent-deck/` holds only logs and lock files.

Agent processes can outlive the daemon. A detached one (`ppid=1`) is not attached to the PTY it was born under, so it survives a daemon restart untouched. When the new daemon comes up its role maps are empty, and that pane is orphaned permanently — there is no recovery short of re-dispatching.

## What makes it nasty: the asymmetry

The daemon **rejects delegates** from an unregistered pane:

```rust
// src/state.rs:6026
if !self.pane_role_map.contains_key(&signal.pane_id) {
    warn!(pane_id = %signal.pane_id, "delegate from unknown pane");
```

but **accepts hook events** from one. Measured: the daemon logged `ToolStart` / `ToolEnd` / `Idle` events from the orphaned pane for hours, with the same `pane_id` it refused to route a delegate for.

So an orphaned pane looks completely healthy — status updates flow, work continues — and the failure surfaces only at the moment someone delegates, which for an orchestrator can be hours into a run.

## Why it is reachable by design, not by accident

CLAUDE.md rule 15 instructs maintainers to run `dot-agent-deck daemon stop` routinely while working on the desktop GUI, and nothing anywhere warns that doing so kills every running orchestration. Every daemon restart in the log for the observed incident was a deliberate `KIND_SHUTDOWN received — sending ack and beginning graceful daemon shutdown`, not a crash and not an idle timeout.

## Suggested slice, in priority order

[...]

Issue #428, opened 2026-08-08
Daemon killed by stray SIGTERM, taking all agents with it (4th occurrence; command evidence lost mid-tool)


vfarcic:
## Summary

A running daemon is periodically killed by a stray `SIGTERM`, taking every managed agent with it. It has now happened four times, always overnight or during unattended work, and each investigation has stalled at the same place: we can see *that* an agent was running a command when the daemon died, but not *which* command.

The most recent occurrence (2026-08-08 03:05:24 UTC) killed three working agents.

## Occurrences

| When (UTC) | Notes |
|---|---|
| 2026-07-28 23:05 | recorded by `dad-watchdog` |
| 2026-07-29 01:54 | daemon PID replaced (A→B) rather than going to none |
| 2026-07-29 02:09 | same shape |
| 2026-08-08 03:05:24 | 3 agents lost; best-documented, detailed below |

## 2026-08-08 evidence

Daemon PID 2097121, hosting exactly 3 agent panes:

```
2026-08-08T03:05:24.476666Z  WARN dot_agent_deck::daemon: daemon received termination
  signal; initiating graceful shutdown (every managed agent will be stopped) signal="SIGTERM"
2026-08-08T03:05:24.542281Z  INFO dot_agent_deck::daemon: Daemon hook loop exiting on shutdown signal
```

`dad-watchdog` death signature: `orphaned-on-PPID1=0` — agents were reaped through the shutdown code path, i.e. a well-formed SIGTERM, not a crash or external SIGKILL.

**Ruled out**, each verified rather than assumed:

- **OOM** — no kill events after 02:10; 14.4 GB free at the moment of death
- **Reboot** — uptime 11 days, boot unchanged
- **Network** — the host's Wi-Fi roam was at 03:05:47, **23 s after** the SIGTERM
- **An agent running a broad `pkill` / `daemon stop`** — see below

### What the three agents were doing

From their Claude Code transcripts:

| Session | Activity | Last transcript entry |
|---|---|---|
| `e23505a3` | `cd ../dot-agent-deck-pr-232 && cargo nextest run --features e2e -j1 opencode_auto_submits_daemon_injected_prompt delegate_015_…` (started 03:04:07.880, ToolEnd 03:05:11) | 03:05:12.006 |
| `e832ef89` | cross-version contract test — did issue kills, but **correctly scoped** | 03:05:08.231 |
[...]

PR #1111, opened 2026-09-15
docs(claude): warn against an unscoped pkill in sandbox teardown (#428)

vfarcic:
## What

Two edits to `CLAUDE.md`, nothing else. No `src/`, no changelog fragment (precedent: `5993cbaf`, which added rule 17 and touched `CLAUDE.md` alone — a contributor instruction has no user-visible release note).

**Rule 12** gains a teardown bullet. That rule already enumerates three ways the cross-version manual test silently goes wrong ("a second mouth", "a third mouth"); this is a **fourth**, placed on the teardown step because that is where an agent is reading when it needs the warning. It is categorically different from the other three, and the bullet says so: those cost you the run (two of them silently), this one destroys work outside the test.

The substance: tear a sandbox daemon down by its specific pid; never `pkill -f` a pattern the production daemon's command line also matches — and `daemon serve` matches it. `pkill` sends SIGTERM, so this goes *around* #770's `daemon stop` refusal rather than having to defeat it.

**Rule 15** gains a one-sentence cross-reference beside the #770 refusal it already explains. That rule already says reaching for `--force` abandons a run; the adjacent fact is that an unscoped `pkill` never reaches the refusal at all.

## Evidence

[#428 occurrence #5](https://github.com/vfarcic/dot-agent-deck/issues/428#issuecomment-5688814518), 2026-09-15. An agent building a rule-12 sandbox at `~/dev/xver-876` ran `pkill -f dot-agent-deck-linux-amd64` and then `pkill -f "daemon serve"`. The second matched the production daemon (`…/dot-agent-deck daemon serve`), which took the SIGTERM **1.42 s** later and gracefully stopped every agent it managed — **nine panes across three dispatched units**. Every number in the two edits is from that comment; none is rounded or inferred.

Two claims were verified in the tree rather than asserted, since rule 17 makes implementation absolutes the dangerous kind:

- *"never reaches that refusal"* — `src/daemon.rs:92-108` selects on `sigterm`/`sigint` and calls `shutdown_all_graceful` directly. `stop_refusal` is in `src/daemon_stop.rs:295`, reached only from the CLI stop path at `:252`.
- *"without naming one of them"* — the signal handler's `warn!` is a fixed string, `"(every managed agent will be stopped)"`, with no pane, role or agent id.
- The production daemon's command line was confirmed with `ps -eo pid,args` on this box: `/home/vfarcic/.local/bin/dot-agent-deck daemon serve`.

## What this deliberately does NOT claim

The note closes the **deliberate** hole — an agent running a broad kill. It does not close the accidental one, and the bullet says so in its own last sentence. #428's occurrence #4 had an agent whose kills **were** correctly scoped (`v0358|pr-390/target/release`, `xver-390`) and the production daemon died anyway; `daemon stop` resolves its target via `peer_pid()` on the **attach** socket, which is #428's own "Still open" item 1 and remains open. Nothing here says the daemon can no longer be killed by a test, or that this prevents a recurrence of #428.

## Scope

`CLAUDE.md` only. The daemon-side structural question is #1109 and the harness question is #428 item 1; both are deliberately untouched.

## Rule 12 self-check

This PR does not touch the daemon, the TUI↔daemon protocol, orchestration or hooks, so `PROTOCOL_VERSION` is unchanged and the cross-version manual test does not apply.

## Gates

- `cargo fmt --check` — clean
- `cargo clippy --workspace --all-targets --features e2e,e2e-live -- -D warnings` — exit 0
- `cargo test-fast` — **3118 passed, 0 skipped** (48.4s)

No test reads `CLAUDE.md` as data (checked across `xtask/`), so there are no change-scoped tests beyond the standard gates.


prageethw (approved):
CLAUDE.md-only documentation addition: a real incident (production daemon killed by an unscoped pkill -f pattern matching its own command line) turned into a clear, actionable warning for rule 12's cross-version test procedure and rule 16's daemon-stop discussion. No code risk.

